PR #1, opened 2019-03-21
Scaffold flask application

lcarva:


lcarva:
@mprahl PTAL

It's just a very simple layout for the application.

.gitignore
```diff
@@ -103,5 +103,8 @@ venv.bak/
 # mypy
 .mypy_cache/
 
+# Flask
+settings.cfg
```
mprahl: Why is it `.cfg` and not `.py`?

cachito/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+
+app = Flask(__name__)
+app.config.from_object('cachito.default_settings')
+app.config.from_envvar('CACHITO_SETTINGS')
+
+if not app.debug:
+    import logging
+    from logging.handlers import TimedRotatingFileHandler
+    # https://docs.python.org/3.6/library/logging.handlers.html#timedrotatingfilehandler
+    file_handler = TimedRotatingFileHandler(os.path.join(app.config['LOG_DIR'], 'cachito.log'),
```
mprahl: I don't think we want file logs. I think we want to stream them to stdout for OpenShift captures them.

cachito/__init__.py
```diff
@@ -0,0 +1,19 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+import os
+from flask import Flask
+
+app = Flask(__name__)
```
mprahl: Let's refactor this to use an application factory. That can be in a separate PR.

cachito/views.py
```diff
@@ -0,0 +1,10 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+from flask import render_template
+
+from cachito import app
+
+
+@app.route('/')
+def index():
+    app.logger.warning('sample message')
```
mprahl: This can be removed

tests/test_cachito.py
```diff
@@ -0,0 +1,18 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+import unittest
+
+import cachito
+
+
+class CachitoTestCase(unittest.TestCase):
```
mprahl: I'd rather use pytest tests and pytest fixtures for the test client like we do in Greenwave and Estuary

tox.ini
```diff
@@ -0,0 +1,27 @@
...
+[testenv:lint]
+deps =
+    flake8
+commands =
+    python -m flake8 {posargs}
+
+[flake8]
+show-source = True
+max-line-length = 100
+exclude = .git,.tox,dist,*egg,*fedmsg.d
```
mprahl: Oh no, not fedmsg!

tox.ini
```diff
@@ -0,0 +1,27 @@
...
+deps =
+    -rdev-requirements.txt
+whitelist_externals =
+    make
+    touch
+setenv =
+	CACHITO_SETTINGS={toxinidir}/settings.cfg
+commands =
+    touch {toxinidir}/settings.cfg
+    py.test tests/ {posargs}
```
mprahl: It's nice to use `-vvv` here so we can get more information about failures

mprahl:
I'd like to have the stuff I mentioned in the comments addressed but we can handle those in other PRs.

:+1:

cachito/__init__.py
lcarva: Removed it. Part of the cookiecutter stuff. Dropped it for now.

cachito/__init__.py
lcarva: Awesome, thanks for suggesting that. This is done now.

tox.ini
lcarva: Honestly, I don't know how that got in there. Removed.

.gitignore
lcarva: Changed it to .py.

lcarva: Silly cookiecutter...

lcarva:
@mprahl, thanks for the review! This should be ready for another round.

.gitignore
```diff
@@ -97,11 +97,17 @@ venv.bak/
 # Rope project settings
 .ropeproject
 
+# VS Code settings
+.vscode
```
mprahl: 👍

PR #2, opened 2019-03-26
Reorganize cachito to use the web and workers subpackages

mprahl:
Reorganize cachito to use the web and workers subpackages and a couple small unrelated commits


mprahl:
I'm fixing the tests now

cachito/web/splash.py
```diff
@@ -2,7 +2,7 @@
 from flask import render_template, Blueprint
 
 
-splash = Blueprint('splash', __name__, static_folder='static', template_folder='template')
```
lcarva: Oh interesting! Not needed because it defaults to those values anyways?

requirements-workers.txt
```diff
@@ -0,0 +1 @@
+celery
```
lcarva: Don't you need celery also in the main `requirements.txt` ? I thought that was needed to actually create celery tasks.

cachito/web/splash.py
mprahl: Yup, it works just fine.

requirements-workers.txt
mprahl: You're right. I'll fix it.

mprahl: It's actually not needed yet, so I'll just remove it from this PR and add it back later

mprahl:
@lcarva this should be all set now.

PR #3, opened 2019-03-27
Add the boilerplate for the database integration

mprahl:
This also adds the ability to configure Cachito.

### DIFF
--- a/cachito/web/config.py
+++ b/cachito/web/config.py
@@ -1,0 +1,21 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+
+class Config(object):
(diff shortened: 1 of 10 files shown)


mprahl:
@lcarva note that anything under `cachito/web/migrations` is auto-generated boilerplate from Flask-Migrate.

cachito/web/config.py
```diff
@@ -0,0 +1,21 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+
+class Config(object):
+    """The base Cachito Flask configuration."""
+    pass
```
lcarva: nitpick: `pass` is not needed if there's a docstring.

PR #4, opened 2019-03-28
WIP: Add a way to develop Cachito using docker-compose

mprahl:
This mostly works except for a couple issues:
* A permission denied error occurs in the Flask API container when accessing `setup.py` when the Flask API container and the Celery container both mount the source directory
* We need to add a way for Celery to automatically reload when the code changes like Flask does. StackOverflow suggests `watchdog`. This is not a blocker, but it'd be nice to have.

I also need to add unit tests, especially for the `configure_celery` function.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,53 @@
+version: '3'
+services:
+  db:
+    image: postgres:9.6
+    restart: always
+    environment:
+      POSTGRES_USER: cachito
+      POSTGRES_PASSWORD: cachito
+      POSTGRES_DB: cachito
+      POSTGRES_INITDB_ARGS: "--auth='ident' --auth='trust'"
+
+  rabbitmq:
+    image: rabbitmq:3.7-management
+    environment:
+      RABBITMQ_DEFAULT_USER: cachito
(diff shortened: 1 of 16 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,52 @@
...
+      pip3 uninstall -y cachito &&
+      python3 setup.py develop --no-deps &&
+      cachito wait-for-db &&
+      cachito db upgrade &&
+      flask-3 run --reload --host 0.0.0.0 --port 8080
+    environment:
+      FLASK_ENV: development
+      FLASK_APP: cachito/web/wsgi.py
+    volumes:
+      - ./:/src:Z
```
lcarva: You want lower case z here. From `man docker run`:
> To change a label in the container context, you can add either of two suffixes :z or :Z to the volume mount.
These suffixes tell Docker to relabel file objects on the shared volumes. The z option tells Docker that two
containers share the volume content. As a result, Docker labels the content with a shared content label.
Shared volume labels allow all containers to read/write content.  The Z option tells Docker to label the
content with a private unshared label.  Only the current container can use a private volume.

PR #6, opened 2019-04-05
Add GET /requests/<id> and POST /requests APIs

lcarva:

### DIFF
--- a/cachito/web/config.py
+++ b/cachito/web/config.py
@@ -1,5 +1,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
+TEST_DB_FILE = '/tmp/cachito.db'
+
 
 class Config(object):
     """The base Cachito Flask configuration."""
@@ -19,4 +21,10 @@ class DevelopmentConfig(Config):
 
 class TestingConfig(DevelopmentConfig):
     """The testing Cachito Flask configuration."""
-    pass
+    # IMPORTANT: don't use in-memory sqlite. Alembic migrations will create a new
+    # connection producing a new instance of the database which is deleted immediately
+    # after the migration completes...
+    #   https://github.com/miguelgrinberg/Flask-Migrate/issues/153
+    SQLALCHEMY_DATABASE_URI = f'sqlite:///{TEST_DB_FILE}'
+    DEBUG = True
+    TESTING = True
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -1,11 +1,79 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
+from copy import deepcopy
 
+from cachito.errors import ValidationError
 from cachito.web import db
+
+
+request_pkg_manager_table = db.Table(
+    'request_pkg_manager',
+    db.Column('request_id', db.Integer, db.ForeignKey('request.id')),
+    db.Column('pkg_manager_id', db.Integer, db.ForeignKey('package_manager.id'))
+)
 
 
 class Request(db.Model):
     """A Cachito user request."""
     id = db.Column(db.Integer, primary_key=True)
+    git_repo = db.Column(db.String, nullable=False)
+    git_ref = db.Column(db.String, nullable=False)
+    pkg_managers = db.relationship('PackageManager', secondary=request_pkg_manager_table,
+                                   backref='requests')
 
     def __repr__(self):
         return '<Request {0!r}>'.format(self.id)
+
+    def to_json(self):
+        pkg_managers = [pkg_manager.to_json() for pkg_manager in self.pkg_managers]
+        return {
+            'id': self.id,
+            'git_repo': self.git_repo,
+            'git_ref': self.git_ref,
+            'pkg_managers': pkg_managers,
+        }
+
+    @classmethod
+    def from_json(cls, kwargs):
+        # Validate all required parameters are present
+        required_params = {'git_repo', 'git_ref', 'pkg_managers'}
+        missing_params = required_params - set(kwargs.keys())
+        if missing_params:
+            raise ValidationError('Missing required parameter(s): {}'
+                                  .format(', '.join(missing_params)))
+        kwargs = deepcopy(kwargs)
+
+        # Validate package managers are correctly provided
+        pkg_managers_names = kwargs.pop('pkg_managers', None)
+        if not pkg_managers_names:
+            raise ValidationError('At least one package manager is required')
+
+        pkg_managers_names = set(pkg_managers_names)
+        found_pkg_managers = (PackageManager.query
+                              .filter(PackageManager.name.in_(pkg_managers_names))
+                              .all())
+        if len(pkg_managers_names) != len(found_pkg_managers):
+            found_pkg_managers_names = set(pkg_manager.name for pkg_manager in found_pkg_managers)
+            invalid_pkg_managers = pkg_managers_names - found_pkg_managers_names
+            raise ValidationError('Invalid package manager(s): {}'
+                                  .format(', '.join(invalid_pkg_managers)))
+
+        kwargs['pkg_managers'] = found_pkg_managers
+
+        try:
+            request = cls(**kwargs)
+        except TypeError as e:
+            # Handle extraneous parameters.
+            raise ValidationError(str(e))
+        return request
+
+
+class PackageManager(db.Model):
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String, nullable=False)
+
+    def to_json(self):
+        return self.name
+
+    @classmethod
+    def from_json(cls, name):
+        return cls(name=name)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,11 +1,36 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
+import os
+
 import pytest
+import flask_migrate
 
-
+from cachito.web import db as _db
+from cachito.web.config import TEST_DB_FILE
 from cachito.web.wsgi import create_app
 
 
 @pytest.fixture(scope='session')
-def client():
+def app():
+    """Return Flask application for the pytest session."""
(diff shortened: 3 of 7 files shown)


hanzz:
+1

cachito/web/migrations/versions/app_source_migration.py
```diff
@@ -0,0 +1,49 @@
+"""empty message
```
mprahl: Can you add a message here?

Also, this should replace `initial_migration.py` since that was there as a placeholder just to get a working development environment.

cachito/web/models.py
```diff
@@ -1,11 +1,69 @@
...
+    'request_pkg_manager',
+    db.Model.metadata,
+    db.Column('request_id', db.Integer, db.ForeignKey('request.id')),
+    db.Column('pkg_manager_id', db.Integer, db.ForeignKey('pkg_manager.id'))
+)
+
+
 class Request(db.Model):
     """A Cachito user request."""
+    __tablename__ = 'request'
```
mprahl: Optional: This isn't necessary since Flask-SqlAlchemy will automatically generate the table name from the class name in snake case form.

lcarva: This was needed for the association table below. Do you know of a way around it?

cachito/web/models.py
```diff
@@ -1,11 +1,69 @@
...
     id = db.Column(db.Integer, primary_key=True)
+    git_repo = db.Column(db.String, nullable=False)
+    git_ref = db.Column(db.String, nullable=False)
+    pkg_managers = db.relationship('PackageManager', secondary=request_pkg_manager_table)
```
mprahl: It might be nice to use `backref` here so that from a package manager, we can get all the requests.

cachito/web/models.py
```diff
@@ -1,11 +1,69 @@
...
+        pkg_managers = [pkg_manager.to_json() for pkg_manager in self.pkg_managers]
+        return {
+            'id': self.id,
+            'git_repo': self.git_repo,
+            'git_ref': self.git_ref,
+            'pkg_managers': pkg_managers,
+        }
+
+    @classmethod
+    def from_json(klass, kwargs):
```
mprahl: I've never seen this convention of `klass` before. I've always seen `cls`. Is there a PEP rule that prefers `klass`?

cachito/web/models.py
```diff
@@ -1,11 +1,69 @@
...
+        # Validate all required parameters are present
+        required_params = {'git_repo', 'git_ref', 'pkg_managers'}
+        missing_params = required_params - set(kwargs.keys())
+        if missing_params:
+            raise ValueError('Missing required parameter(s): {}'.format(', '.join(missing_params)))
```
mprahl: It'd be nice to create a custom exception class, and add an error handler in Flask so that it automatically returns a JSON error to the user like Estuary does:
https://github.com/release-engineering/estuary-api/blob/master/estuary/app.py#L115

cachito/web/models.py
```diff
@@ -1,11 +1,69 @@
...
+        kwargs = deepcopy(kwargs)
+
+        # Validate package managers are correctly provided
+        pkg_managers = kwargs.pop('pkg_managers', None)
+        if not pkg_managers:
+            raise ValueError('At least one package manager is required')
```
mprahl: Same as above

cachito/web/models.py
```diff
@@ -1,11 +1,69 @@
...
+        for name in set(pkg_managers):
+            pkg_manager = PackageManager.query.filter_by(name=name).first()
```
mprahl: You might be able to replace this for loop with a single query using `_in` and then compare the length of `set(pkg_managers)` and the results.

cachito/web/models.py
```diff
@@ -1,11 +1,69 @@
...
+            if not pkg_manager:
+                raise ValueError('Invalid package manager: {}'.format(name))
+
+            kwargs.setdefault('pkg_managers', [])
+            kwargs['pkg_managers'].append(pkg_manager)
```
mprahl: Why `pop` the `pkg_managers` only to re-add it? Is this just to filter out duplicates?

tests/conftest.py
```diff
@@ -2,10 +2,33 @@
...
-    return create_app('cachito.web.config.TestingConfig').test_client()
+    return app.test_client()
+
+
+@pytest.fixture()
+def db(app):
+    """Yields a DB with required app tables but with no records."""
+    with app.app_context():
+        _db.create_all()
+        # TODO: The default pkg managers are defined in the migration scripts. How
```
mprahl: Instead of `_db.create.all()`, you may be able to run `flask_migrate.migrate()`.

tests/conftest.py
```diff
@@ -2,10 +2,33 @@
...
+        # to call those from here?
+        pkg_manager = PackageManager(name='gomod')
+        _db.session.add(pkg_manager)
+        _db.session.commit()
+        yield _db
+        _db.session.remove()
+        _db.drop_all()
```
mprahl: Optional: WaiverDB rolls back the transaction to avoid dropping and recreating the database for every test. This would increase performance of the unit tests:
https://pagure.io/waiverdb/blob/master/f/tests/conftest.py#_46

tests/test_api_v1.py
```diff
@@ -1,7 +1,80 @@
...
+def test_missing_request(client, db):
+    rv = client.get('/api/v1/requests/1')
+    assert rv.status_code == 404
+
+
+def test_malformed_request_id(client, db):
+    rv = client.get('/api/v1/requests/spam')
+    assert rv.status_code == 400
+    error_msg = json.loads(rv.data.decode('utf-8'))['error']
+    assert 'invalid literal' in error_msg
```
mprahl: This seems like a scary message that should be improved

cachito/web/models.py
mprahl: The Flask-SQLAlchemy docs recommend the following:
```python
request_pkg_manager_table = db.Table(
    'request_pkg_manager',
    db.Column('request_id', db.Integer, db.ForeignKey('request.id')),
    db.Column('pkg_manager_id', db.Integer, db.ForeignKey('pkg_manager.id'))
)
```

You could give that a try. I'm not sure what the `db.Model.metadata` argument provides, but it's not in the docs.

http://flask-sqlalchemy.pocoo.org/2.3/models/#many-to-many-relationships

mprahl:
I'm done with the first round. It looks good, but I left a few comments.

cachito/web/models.py
lcarva: Looks like the `db.Model.metadata` is a parameter for plain SQLAlchemy. It's apparently undocumented for Flask-SQLAlchemy but it doesn't seem to cause any errors. Anyways, thanks for spotting this. Removed it.

cachito/web/migrations/versions/app_source_migration.py
lcarva: Folded these into the existing migration script.

cachito/web/models.py
lcarva: Good call! Done.

cachito/web/models.py
lcarva: I've always used `klass`. Oh well, changed it to `cls` since that seems to be more common in our team.

cachito/web/models.py
lcarva: Created shared `ValidationError` exception. I placed this directly under `cachito` module because it's generic enough to apply to either `web` or `worker`.

cachito/web/models.py
lcarva: Done! (Although, I don't think the code is much cleaner with that approach)

cachito/web/models.py
lcarva: It's actually needed to convert the pkg_manager **names** into pkg_manager **objects**.

tests/conftest.py
lcarva: done!

tests/test_api_v1.py
lcarva: Changed it to "X is not a valid request ID"

lcarva:
@hanzz, @mprahl, thanks for the reviews! This is ready for another round.

cachito/web/config.py
```diff
@@ -19,4 +19,10 @@ class DevelopmentConfig(Config):
 
 class TestingConfig(DevelopmentConfig):
     """The testing Cachito Flask configuration."""
-    pass
+    # IMPORTANT: don't use in-memory sqlite. Alembic migrations will create a new
+    # connection producing a new instance of the database which is deleted immediately
+    # after migration completes...
```
mprahl: `after migration` => `after the migration`

lcarva: Done!

tests/conftest.py
```diff
@@ -1,11 +1,34 @@
...
+
+@pytest.fixture()
+def db(app, tmpdir):
+    """Yields a DB with required app tables but with no records."""
+    with app.app_context():
+        flask_migrate.upgrade()
+        connection = _db.engine.connect()
+        # Clear each table to ensure tests are idempotent. The exception is
+        # the package_manager table which contains required data to run app.
+        for table in _db.metadata.tables.keys():
```
mprahl: It might be easier to just delete the SQLite database file before running `flask_migrate.upgrade()`. Also, does `_db.drop_all()` not work when run before `flask_migrate.upgrade()`?

lcarva: `_db.drop_all()` doesn't seem to re-trigger the upgrade process again. I had it deleting the file initially, but thought cleaning up the data would give us more flexibility. I don't have a strong preference either way, which makes me lean towards leaving it as is :)

lcarva: Although... calling out a table here creates a special case which is not ideal. I'll change it.

lcarva: Done!

PR #7, opened 2019-04-12
Add the ability to fetch the source code from the requests API endpoint

mprahl:
This still needs documentation and tests written, but I'd like to get your opinion on this.

### DIFF
--- a/cachito/workers/scm.py
+++ b/cachito/workers/scm.py
@@ -1,0 +1,208 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+import logging
+import os
+from abc import ABC, abstractmethod
+import urllib.parse
+import shutil
+import tarfile
+import tempfile
+import subprocess
+
+import requests
+
+from cachito.errors import CachitoError
+
+
+log = logging.getLogger(__name__)
+
+
+class SCM(ABC):
+    """The base class for interacting with source control."""
+
+    def __init__(self, url, ref):
+        """
+        Initialize the SCM class.
+
+        :param str url: the source control URL to the repository to fetch
+        :param str ref: the source control reference
+        """
+        super().__init__()
+        self.url = url
+        self.ref = ref
+        self._archives_dir = None
+        self._archive_path = None
+        self._repo_name = None
+
+    @property
+    def archive_path(self):
+        """
+        Get the path to where the archive for a particular SCM reference should be.
+
+        :return: the path to the archive
+        :rtype: str
+        """
+        if not self._archive_path:
+            directory = os.path.join(self.archives_dir, *self.repo_name.split('/'))
+            # Create the directories if they don't exist
+            os.makedirs(directory, exist_ok=True)
+            self._archive_path = os.path.join(directory, f'{self.ref}.tar.gz')
+
+        return self._archive_path
+
+    @property
+    def archives_dir(self):
+        """
+        Get the absolute path of the archives directory from the Celery configuration.
+
+        :returns: the absolute path of the archives directory
+        :rtype: str
+        """
+        if not self._archives_dir:
+            # Import this here to avoid a circular import
+            import cachito.workers.tasks
+
+            self._archives_dir = os.path.abspath(
+                cachito.workers.tasks.app.conf.cachito_archives_dir
+            )
+            log.debug('Using "%s" as the archives directory', self._archives_dir)
+
+        return self._archives_dir
+
+    def download_source_archive(self, url):
+        """
+        Download the compressed archive of the source and place it in long-term storage.
+
+        This is useful for services like GitHub, where they already provide a mechanism to
+        download the compressed archive directly.
+
+        :param str url: the URL to download the compressed source archive from
+        :raises CachitoError: if the download fails
+        """
+        with tempfile.TemporaryDirectory(prefix='cachito-') as temp_dir:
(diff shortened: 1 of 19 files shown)


cachito/workers/scm.py
```diff
@@ -0,0 +1,201 @@
...
+    @property
+    def archive_path(self):
+        """
+        Get the path to where the archive for a particular SCM reference should be.
+
+        :return: the path to the archive
+        :rtype: str
+        """
+        if not self._archive_path:
+            directory = os.path.join(self.archives_dir, *self.repo_name.split('/'))
```
lcarva: Is splitting necessary? Won't this produce the same thing? 
```python
os.path.join(self.archives_dir, self.repo_name)
```

cachito/workers/scm.py
```diff
@@ -0,0 +1,201 @@
...
+        :param str ref: the source control reference
+        """
+        super().__init__()
+        self.url = url
+        self.ref = ref
+        self._archives_dir = None
+        self._archive_path = None
+        self._repo_name = None
+        # Validate that the base directory can be retrieved
+        self.archives_dir
```
lcarva: It sounds like you don't want `self.archives_dir` to be a property :)
Maybe just have a helper method that is called here to do the work?

cachito/workers/scm.py
```diff
@@ -0,0 +1,201 @@
...
+
+        :returns: the absolute path of the archives directory
+        :rtype: str
+        :raise ConfigError: if "cachito_archives_dir" is not configured in Celery
+        """
+        if not self._archives_dir:
+            # Import this here to avoid a circular import
+            from cachito.workers.tasks import app as celery_app
+
+            if not celery_app.conf.get('cachito_archives_dir'):
```
lcarva: Can we make this a required attribute in the config?  I'd rather see an error during app start time then something that happens later on. IOW, let's move this to `validate_celery_config`.

cachito/workers/scm.py
```diff
@@ -0,0 +1,201 @@
...
+            cmd = ['git', 'clone', '-q', '--no-checkout', self.url, clone_path]
+            log.debug('Cloning the repo with "%s"', ', '.join(cmd))
+            git_clone = subprocess.Popen(cmd, stderr=subprocess.PIPE)
+            _, error_output = git_clone.communicate()
+            if git_clone.returncode != 0:
+                error_output = error_output.decode('utf-8')
+                log.error(
+                    'Cloning the git repo with "%s" failed with: %s', ', '.join(cmd),
+                    error_output)
+                raise CachitoError(error)
```
lcarva: Maybe set `error` to indicate it failed to clone?

cachito/workers/tasks.py
```diff
@@ -15,3 +18,22 @@
...
+    """
+    Fetch the application source code that was requested and put it in long-term storage.
+
+    :param str url: the source control URL to pull the source from
+    :param str ref: the source control reference
+    """
+    log.info('Fetching the source from "%s" at reference "%s"', url, ref)
+    try:
+        # Default to Git for now
+        scm = Git.create_from_url(url, ref)
```
lcarva: `Git(url, ref)` ?

cachito/workers/scm.py
mprahl: It would on systems that use forward slashes in their paths. I tend to not assume that when writing code, but it's probably safe to assume we won't be supporting Windows :stuck_out_tongue:

cachito/workers/scm.py
mprahl: Is it safe to assume that all of the celery workers will need access to the archives directory? If so, I can make it mandatory.

cachito/workers/tasks.py
mprahl: Oops, I forgot to update this after I changed it.

cachito/workers/scm.py
mprahl: I'll keep it as a property and move the validation to `validate_celery_config` so this line can be removed.

mprahl:
@lcarva okay, all the feedback has been addressed and I've added unit tests. I've also started validating the provided refs, and renamed the database columns to be generic and not specific to git. I figure it's easier to change this now rather than later.

cachito/workers/scm.py
lcarva: Fair point :)

cachito/workers/tasks.py
```diff
@@ -1,17 +1,48 @@
...
 
-from cachito.workers.config import configure_celery
 
 log = logging.getLogger(__name__)
 logging.basicConfig()
 app = Celery()
 configure_celery(app)
+# Only validate the configuration when Celery is directly running this module. This allows
+# importers of this module (e.g. the Flask app) to not have Celery fully configured.
+if os.path.basename(sys.argv[0]).startswith('celery'):
```
lcarva: This is pretty hackish :)
Can we use [on_init](https://docs.celeryproject.org/en/latest/reference/celery.html#celery.Celery.on_init) ?

mprahl: I think `on_init` wouldn't solve the issue, but it looks like [worker_init signal](https://docs.celeryproject.org/en/master/userguide/signals.html?highlight=signals#worker-init) might work. I'll try that.

mprahl:
@lcarva this should be all set now. Thanks for the review!

PR #8, opened 2019-04-26
Add gomod package manager support

lcarva:


lcarva:
Unit tests are in the works!

cachito/workers/pkg_manager.py
```diff
@@ -0,0 +1,58 @@
...
+            'GOPATH': temp_dir,
+            'GO111MODULE': 'on',
+            'GOCACHE': temp_dir,
+            'GOPROXY': get_worker_config().athens_url,
+            'PATH': os.environ.get('PATH', ''),
+        }
+        cmd = ('go', 'list', '-m', 'all')
+
+        go_list = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
+                                   text=True, encoding='utf-8', env=env, cwd=source_dir)
```
lcarva: Bummer, the `test` parameter was only introduced in python 3.7. It seems that the existing `universal_newlines` will do the same. I'll change it.

https://docs.python.org/3/library/subprocess.html#subprocess.Popen

cachito/workers/pkg_manager.py
```diff
@@ -0,0 +1,58 @@
...
+            )
+            raise CachitoError('Fetching gomod dependencies failed')
+
+        deps = []
+        for line in output.splitlines():
+            parts = line.split(' ')
+            if len(parts) == 1:
+                # This is the application itself, not a dependency
+                continue
+            if len(parts) > 2:
```
mprahl: How about `if len(parts) != 2:`?

cachito/workers/pkg_manager.py
```diff
@@ -0,0 +1,58 @@
...
+        env = {
+            'GOPATH': temp_dir,
+            'GO111MODULE': 'on',
+            'GOCACHE': temp_dir,
+            'GOPROXY': get_worker_config().athens_url,
+            'PATH': os.environ.get('PATH', ''),
+        }
+        cmd = ('go', 'list', '-m', 'all')
+
+        go_list = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
```
mikebonnet: Could consider using `subprocess.run()` here instead.

cachito/workers/pkg_manager.py
```diff
@@ -0,0 +1,58 @@
...
+            if len(parts) > 2:
+                log.warning('Unexpected go module output: %s', line)
+                continue
+            if len(parts) == 2:
+                deps.append({'type': 'gomod', 'name': parts[0], 'version': parts[1]})
+
+        return deps
+
+
+def _mount_app_src(archive_path, parent_dir):
```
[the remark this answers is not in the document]
yashvardhannanavati: or `_get_app_source`?

cachito/workers/pkg_manager.py
lcarva: Actually, I like to separate the two use cases. In the case where there's only one part, we can just ignore, but we really want to warn if it unexpectedly has more than 2 parts.

cachito/workers/config.py
```diff
@@ -22,6 +22,7 @@ class DevelopmentConfig(Config):
     """The development Cachito Celery configuration."""
 
     broker_url = 'amqp://cachito:cachito@rabbitmq:5672//'
+    athens_url = 'http://athens:3000'
```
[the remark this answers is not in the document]
lcarva: Probably. Although I'm not sure if this should be considered validation or part of the startup process. Should we enhance the command 'wait-for-db' to also wait for athens? (With a different name)

I'm tempted to skip the validation at this point.

cachito/workers/pkg_manager.py
lcarva: Fixed!

cachito/workers/pkg_manager.py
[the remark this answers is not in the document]
lcarva: `_extract_app_src` is what I had stashed locally. Pushed now.

cachito/workers/pkg_manager.py
lcarva: Yeah. I was trying to keep consistency with the `scm.py` module. I'll make a separate commit to change them both over.

lcarva: Done!

lcarva:
Hey all, I think I've addressed all your comments either by making code changes, or by replying to your message. Please have another looks when you get a chance. Removing WIP from title now.

cachito/workers/pkg_manager.py
```diff
@@ -0,0 +1,69 @@
...
+        }
+        cmd = ('go', 'list', '-m', 'all')
+
+        go_list = subprocess.run(cmd, capture_output=True, universal_newlines=True,
+                                 encoding='utf-8', env=env, cwd=source_dir)
+        if go_list.returncode != 0:
+            log.error(
+                'Fetching gomod dependencies with "%s" failed with: %s',
+                ' '.join(cmd),
+                go_list.stdout,
```
mprahl: The error is outputted to `stdout` and not `stderr`?

cachito/workers/scm.py
```diff
@@ -136,14 +134,13 @@ def clone_and_archive(self):
 
             cmd = ['git', 'clone', '-q', '--no-checkout', self.url, clone_path]
             log.debug('Cloning the repo with "%s"', ' '.join(cmd))
-            git_clone = subprocess.Popen(cmd, stderr=subprocess.PIPE)
-            _, error_output = git_clone.communicate()
+            git_clone = subprocess.run(cmd, capture_output=True, universal_newlines=True,
+                                       encoding='utf-8')
```
mprahl: Optional: I'd prefer the following formatting:
```python
git_clone = subprocess.run(
    cmd, capture_output=True, universal_newlines=True, encoding='utf-8')
```

cachito/workers/scm.py
```diff
@@ -158,16 +155,15 @@ def clone_and_archive(self):
                 self.ref,
             ]
             log.debug('Creating the archive with "%s"', ' '.join(cmd))
-            git_archive = subprocess.Popen(cmd, stderr=subprocess.PIPE)
-            _, error_output = git_archive.communicate()
+            git_archive = subprocess.run(cmd, capture_output=True, universal_newlines=True,
+                                         encoding='utf-8')
```
mprahl: Optional: Same as above

tests/test_workers/test_pkg_manager.py
```diff
@@ -0,0 +1,106 @@
...
+    github.com/op/go-logging v0.0.0-20160315200505-970db520ece7
+    github.com/pkg/errors v0.8.1
+    golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2
+    golang.org/x/net v0.0.0-20190311183353-d8887717615a
+    golang.org/x/sys v0.0.0-20190215142949-d0b11bdaac8a
+    golang.org/x/text v0.3.0
+    golang.org/x/tools v0.0.0-20190325161752-5a8dccf5b48a
+    gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127
+    gopkg.in/yaml.v2 v2.2.2
+    """)
```
mprahl: Optional I'd prefer if there wasn't an indentation here.

cachito/workers/scm.py
mikebonnet: As counterpoint, I tend to use Lui's formatting. :)

PR #9, opened 2019-05-01
Download archive

lcarva:


cachito/workers/config.py
```diff
@@ -75,11 +78,12 @@ def validate_celery_config(conf, **kwargs):
...
     :raises ConfigError: if the configuration is invalid
     """
-    archives_dir = conf.get('cachito_archives_dir')
-    if archives_dir is None or not os.path.isdir(archives_dir):
-        raise ConfigError(
-            'The configuration "cachito_archives_dir" must be set to an existing directory'
-        )
+    for required_dir_conf in ('cachito_archives_dir', 'cachito_shared_dir'):
+        required_dir = conf.get(required_dir_conf)
+        if required_dir is None or not os.path.isdir(required_dir):
```
yashvardhannanavati: optional: you can use `if not required_dir or not os.path.isdir(required)`

cachito/workers/pkg_manager.py
```diff
@@ -65,6 +65,6 @@ def _extract_app_src(archive_path, parent_dir):
     :returns: the absolute path of the extracted application source code
     :rtype: str
     """
-    with tarfile.open(archive_path, 'r:gz') as archive:
+    with tarfile.open(archive_path, 'r:*') as archive:
```
mprahl: Cool trick!

cachito/workers/config.py
```diff
@@ -26,6 +26,9 @@ class DevelopmentConfig(Config):
     cachito_archives_dir = os.path.join(tempfile.gettempdir(), 'cachito-archives')
     cachito_log_level = 'DEBUG'
 
+    result_backend = 'rpc'
```
mprahl: Cool, if you're curious how this works, this is a good resource:
https://www.rabbitmq.com/tutorials/tutorial-six-python.html

cachito/workers/config.py
```diff
@@ -24,6 +24,8 @@ class DevelopmentConfig(Config):
     broker_url = 'amqp://cachito:cachito@rabbitmq:5672//'
     athens_url = 'http://athens:3000'
     cachito_archives_dir = os.path.join(tempfile.gettempdir(), 'cachito-archives')
+    # NOTE: cachito_shared_dir must also be available at the same location in frontend
```
mprahl: `frontend` => `the API`

docker-compose.yml
```diff
@@ -47,6 +47,12 @@ services:
       CACHITO_DEV: 'true'
     volumes:
       - ./:/src:z
+      - /tmp/cachito-archives:/tmp/cachito-archives:z
+      - /tmp/cachito-shared:/tmp/cachito-shared:z
+    # This is needed in order to allow cleaning up of temporary tarball
```
mprahl: `of temporary tarball` = > `a temporary tarball`

cachito/web/api_v1.py
```diff
@@ -77,3 +111,20 @@ def create_request():
...
+    Retrieve request object for given ID
+
+    :param int request_id: the value of the request ID
+    :return: a Request
+    :rtype: cachito.models.Request
+    :raise NotFound: if the request is not found
+    :raise ValidationError: if the request ID is of invalid format
+    """
+    try:
+        request_id = int(request_id)
```
mprahl: Instead, you could just set the route to:
```python
@api_v1.route('/requests/<int:request_id>', methods=['GET'])
```

Reference: http://flask.pocoo.org/docs/1.0/quickstart/#variable-rules

cachito/web/api_v1.py
```diff
@@ -35,12 +40,41 @@ def get_request(request_id):
...
-    try:
-        request_id = int(request_id)
-    except ValueError:
-        raise ValidationError(f'{request_id} is not a valid request ID')
-    request = Request.query.get_or_404(request_id)
-    return flask.jsonify(request.to_json())
+    return flask.jsonify(_get_request_obj(request_id).to_json())
+
+
+@api_v1.route('/requests/<request_id>/download', methods=['GET'])
```
mprahl: `<request_id>` => `<int:request_id>`

cachito/web/api_v1.py
```diff
@@ -35,12 +40,41 @@ def get_request(request_id):
...
+    :param int request_id: the value of the request ID
+    :return: a Flask send_file response
+    :rtype: flask.Response
+    :raise NotFound: if the request is not found
+    :raise ValidationError: if the request ID is of invalid format
+    """
+    request = _get_request_obj(request_id)
+    # TODO: Verify request has already been processed.
+
+    app_archive_path = Git(request.repo, request.ref).archive_path
```
mprahl: This will require the API to have `cachito_archives_dir` configured in its Celery config, which I was hoping would be an implementation detail for the worker. Couldn't `fetch_gomod_source` just do this?

cachito/web/api_v1.py
```diff
@@ -35,12 +40,41 @@ def get_request(request_id):
...
+
+    cachito_shared_dir = get_worker_config().cachito_shared_dir
```
mprahl: This makes configuring the API split between the Flask configuration and the Celery configuration. Ideally, the Celery configuration on the API should simply contain the connection details to send requests to the workers. Can this just be a Flask configuration called the same thing?

cachito/workers/config.py
```diff
@@ -24,8 +24,13 @@ class DevelopmentConfig(Config):
...
+    cachito_shared_dir = os.path.join(tempfile.gettempdir(), 'cachito-shared')
```
mprahl: Can this be documented in the read me?

cachito/web/api_v1.py
```diff
@@ -35,12 +40,41 @@ def get_request(request_id):
...
+    with tempfile.TemporaryDirectory(prefix='cachito-', dir=cachito_shared_dir) as temp_dir:
+        deps_path = os.path.join(temp_dir, 'deps')
+        result = tasks.fetch_gomod_source.delay(app_archive_path, copy_cache=deps_path)
+        result.wait(timeout=None, interval=0.5)
```
mprahl: What happens if the workers are down and the HTTP request times out? This should probably be handled gracefully.

cachito/workers/pkg_manager.py
```diff
@@ -68,3 +71,29 @@ def _extract_app_src(archive_path, parent_dir):
     with tarfile.open(archive_path, 'r:*') as archive:
         archive.extractall(parent_dir)
     return os.path.join(parent_dir, 'app')
+
+
+def _run_cmd(cmd, params):
```
mprahl: This is nice

cachito/workers/tasks.py
```diff
@@ -47,15 +47,16 @@ def fetch_app_source(url, ref):
 
 
 @app.task
-def fetch_gomod_source(archive_path):
+def fetch_gomod_source(archive_path, copy_cache=None):
     """
     Resolve and fetch gomod dependencies for given app source archive.
 
     :param str archive_path: the full path to the application source code
+    :param str copy_cache: path to copy artifacts from gomod cache
```
mprahl: `copy_cache` and `archive_path` should ideally be relative paths instead of absolute paths. With this current implementation, the API has to know where the storage mount is in all the workers. Where the storage mount should be an implementation detail, and the API doesn't actually need to know about `archive_path`, it just needs to know the worker has it stored somewhere.

cachito/web/api_v1.py
lcarva: To get the right type? Yeah, cool hint. I'll try it out!

cachito/web/api_v1.py
lcarva: You're right. I got too into the weeds and forgot about this nasty detail. 

I'll play with it in the morning. Maybe fetch app source also handles the copy cache parameter. Or maybe we introduce a task just for generating the final archive tarball. I like the second one.

cachito/workers/tasks.py
lcarva: Yup. I'll try to make that work.

lcarva: Thanks for the suggestion. Made the change to use relative paths, much better.

cachito/web/api_v1.py
lcarva: I added a configurable timeout with a default of 120 seconds. I think that's as graceful as we can get here.

cachito/workers/config.py
lcarva: Done!

cachito/web/api_v1.py
lcarva: Removed dependency.

cachito/web/api_v1.py
lcarva: Done! This is now done on the `fetch_app_source` task itself.

lcarva:
This is ready for another review, thanks!

cachito/web/api_v1.py
```diff
@@ -33,14 +36,54 @@ def get_request(request_id):
...
+    :param int request_id: the value of the request ID
+    :return: a Flask send_file response
+    :rtype: flask.Response
+    :raise NotFound: if the request is not found
+    """
     request = Request.query.get_or_404(request_id)
-    return flask.jsonify(request.to_json())
+    # TODO: Verify request has already been processed.
+
+    cachito_shared_dir = flask.current_app.config.get('CACHITO_SHARED_DIR')
```
mprahl: Some sort of validation should occur so that `cachito_shared_dir` is not `None`. I'm also fine if a KeyError is raised since it'd only happen due to a misconfiguration.

cachito/web/api_v1.py
```diff
@@ -33,14 +36,54 @@ def get_request(request_id):
...
+    wait_timeout = flask.current_app.config.get('CACHITO_WAIT_TIMEOUT')
```
mprahl: Optional: Since the base configuration has a default value, you could directly access the value instead of using `.get` to show that it is expected that that value will always be defined. It's just a minor suggestion to improve readability.

cachito/workers/pkg_manager.py
```diff
@@ -12,37 +13,33 @@
 log = logging.getLogger(__name__)
 
 
-def resolve_gomod_deps(archive_path):
+def resolve_gomod_deps(archive_path, copy_cache=None):
```
mprahl: Optional: I prefer the keyword argument of `copy_cache_to` so it's clear a path is expected. In my opinion, `copy_cache` sounds like a bool is expected.

cachito/web/api_v1.py
```diff
@@ -33,14 +36,54 @@ def get_request(request_id):
...
+        chain_result = chain(
+            tasks.fetch_app_source.s(request.repo, request.ref,
+                                     copy_cache=relative_app_archive_path),
+            tasks.fetch_gomod_source.s(copy_cache=relative_deps_path)
+        ).delay()
+        chain_result.wait(timeout=wait_timeout)
+
+        # Generate a tarball containing the application and dependencies source code
+        response_archive_path = os.path.join(temp_dir, 'temp.tar.gz')
+        with tarfile.open(response_archive_path, mode='w:gz') as response_archive:
```
mprahl: It's not a requirement, but I'd prefer if instead of chaining two tasks together and then generating the final tar file on the API, you just had one synchronous task on the worker that provides you with the final tar file to stream to the user. It just feels a bit off to me that part of the work happens on the worker and part of it happens on the API.

tests/test_api_v1.py
```diff
@@ -102,3 +106,66 @@ def test_validate_extraneous_params(client, db):
...
+@mock.patch('tempfile.TemporaryDirectory')
+@mock.patch('cachito.web.api_v1.Request')
+@mock.patch('cachito.web.api_v1.chain')
+def test_download_archive(
+    mock_chain, mock_request, mock_temp_dir, client, db, app, tmpdir
+):
+    shared_volume = tmpdir.mkdir('shared')
+    shared_workdir = shared_volume.mkdir('ephemeral')
+
+    app.config.update(CACHITO_SHARED_DIR=str(shared_volume))
```
mprahl: You should use `patch.dict` since the `app` fixture is scoped to the session. We don't want this influencing other tests.

cachito/web/api_v1.py
lcarva: I did consider creating a new task in the worker to handle this. The problem is that this would be a task that calls other tasks synchronously. This is not recommended as it may cause deadlocks.

cachito/web/api_v1.py
lcarva: Sticking to `KeyError` for now. 

NOTE: I think we should make the configuration a bit more user friendly. Since the main deployment target for cachito is a containerized environment, we could optimize for that. I have some thoughts on this, but not yet ready to share. I'm mentioning it here to give some reasoning why I think we should stick to `KeyError` for now :)

tests/test_api_v1.py
lcarva: Cool! Didn't know about this. Thanks.

lcarva:
@mprahl, comments addressed.

cachito/web/api_v1.py
mprahl: @lcarva the task doesn't have to call other tasks. The current tasks are merely wrappers around a couple functions. The new task would just call those two functions.

mprahl:
@lcarva I left a reply to one of the comments you left. The other changes look good though!

lcarva:
@mprahl, one more time, please.

lcarva:
Thanks a lot!

PR #10, opened 2019-05-13
Add API endpoint for listing all requests

lcarva:

### DIFF
--- a/tests/test_api_v1.py
+++ b/tests/test_api_v1.py
@@ -44,6 +44,40 @@ def test_create_and_fetch_request(mock_chain, client, db):
     assert created_request == fetched_request
 
 
+@mock.patch('cachito.web.api_v1.chain')
+def test_fetch_paginated_requests(mock_chain, client, db):
+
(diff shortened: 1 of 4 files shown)


tests/test_api_v1.py
```diff
@@ -44,6 +44,40 @@ def test_create_and_fetch_request(mock_chain, client, db):
...
+        assert request['repo'] == repo_template.format(repo_number)
+
+    # per_page and page parameters are honored
+    rv = client.get('/api/v1/requests?page=2&per_page=10')
+    assert rv.status_code == 200
+    response = json.loads(rv.data.decode('utf-8'))
+    fetched_requests = response['items']
+    assert len(fetched_requests) == 10
+    for repo_number, request in enumerate(fetched_requests, 1):
+        # Add 10 because each page contains 10 items and we're processing the second page
```
mprahl: Could you just set `enumerate` to start at `11` instead?

lcarva: Yes, of course! Also, I realize that I ended up introducing the 1-index based sequence while populating test data. Changed it to 0-index based and it's a bit simpler.

PR #11, opened 2019-05-14
Set the request state after the async tasks complete

mprahl:
This also:
* Handles all default Flask HTTP exceptions and returns them as JSON
* Doesn't allow the columns in the request_pkg_manager association table to be null

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -120,9 +120,12 @@ def create_request():
         raise ValidationError('The "ref" parameter must be a 40 character hex string')
 
     # Chain tasks
+    error_callback = tasks.failed_request_callback.s(request.id)
     chain(
-        tasks.fetch_app_source.s(request.repo, request.ref),
-        tasks.fetch_gomod_source.s()
+        tasks.fetch_app_source.s(
+            request.repo, request.ref, request_id_to_update=request.id).on_error(error_callback),
+        tasks.fetch_gomod_source.s(request_id_to_update=request.id).on_error(error_callback),
+        tasks.set_request_state.si(request.id, 'complete', 'Completed successfully'),
     ).delay()
 
     return flask.jsonify(request.to_json()), 201
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -126,8 +126,3 @@ def create_request():
     ).delay()
 
     return flask.jsonify(request.to_json()), 201
-
-
-@api_v1.errorhandler(ValidationError)
-def handle_validation_error(e):
-    return flask.jsonify(error=str(e)), 400
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -126,3 +126,47 @@ def create_request():
     ).delay()
 
     return flask.jsonify(request.to_json()), 201
+
+
+@api_v1.route('/requests/<int:request_id>', methods=['PATCH'])
+def patch_request(request_id):
+    """
+    Modify the given request.
+
+    :param int request_id: the request ID from the URL
+    :return: a Flask JSON response
+    :rtype: flask.Response
+    :raise NotFound: if the request is not found
+    :raise ValidationError: if the JSON is invalid
+    """
+    # TODO: Setup admin authentication
+    payload = flask.request.get_json()
+    if not isinstance(payload, dict):
+        raise ValidationError('The input data must be a JSON object')
+
+    if not payload:
+        raise ValidationError('At least one key must be specified to update the request')
+
+    valid_keys = {'state', 'state_reason'}
+    invalid_keys = set(payload.keys()) - valid_keys
+    if invalid_keys:
+        raise ValidationError(
+            'The following keys are not allowed: {}'.format(', '.join(invalid_keys)))
+
+    for key, value in payload.items():
+        if not isinstance(value, str):
+            raise ValidationError(
+                'The value for "{}" must be a string. It was the type {}.'
+                .format(key, type(value).__name__)
+            )
+
+    if 'state' in payload and 'state_reason' not in payload:
+        raise ValidationError('The "state_reason" key is required when "state" is supplied')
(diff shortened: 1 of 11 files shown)


cachito/workers/tasks.py
```diff
@@ -90,3 +88,35 @@ def assemble_source_code_archive(app_archive_path, deps_path, bundle_archive_pat
...
+    :raise CachitoError: if the request to the Cachito API fails
+    """
+    config = get_worker_config()
+    request_url = f'{config.cachito_api_url.rstrip("/")}/requests/{request_id}'
+
+    log.info('Setting the status of request %d to "%s"', request_id, state)
+    try:
+        rv = requests.patch(request_url, json={'state': state}, timeout=30)
+    except requests.RequestException:
+        msg = f'The connection failed when setting the state to "{state}"" on request {request_id}'
```
lcarva: Additional double quote here?

cachito/web/errors.py
```diff
@@ -0,0 +1,31 @@
...
+            msg = error.description
+        response = jsonify({'error': msg})
+        response.status_code = error.code
+    else:
+        status_code = 500
+        message = None
+        if isinstance(error, ValidationError):
+            status_code = 400
+
+        response = jsonify({'error': message or str(error)})
```
lcarva: `message` is always `None` here. Did you mean to set a default value a few lines back?

cachito/web/api_v1.py
```diff
@@ -121,13 +121,49 @@ def create_request():
...
+    new_state_obj = State.query.filter_by(name=new_state).first()
+    if not new_state_obj:
+        all_state_names = sorted([state.name for state in State.query.all()])
+        raise ValidationError(
+            'The state "{}" is invalid. Please select from: {}.'
+            .format(new_state, ', '.join(all_state_names))
+        )
+
+    request = Request.query.get_or_404(request_id)
+    request.state = new_state_obj
```
lcarva: During design we talked about doing two things:
1. Add some metadata when updating the state - timestamp for now.
2. Track the history of the state. For example, when a state is updated, it's also added to `state_history`.

You could combine both items by having the `state` field simply being a property that maps to the latest state in `state_history`.

I'm fine with doing those things as a follow up PR, but I do think it's important to add this capability in the early stage of Cachito.

cachito/web/errors.py
mprahl: It was a copy-paste error from Estuary. I'll fix it.

mprahl:
@lcarva this is ready for review.

cachito/web/models.py
```diff
@@ -12,25 +15,62 @@
...
+        # redundant information about the request itself
+        states = [
+            {
+                'state': RequestStateMapping(state.state).name,
+                'state_reason': state.state_reason,
+                'updated': state.updated.isoformat(),
+            }
+            for state in self.states
+        ]
+        latest_state = states[-1]
```
lcarva: Not most recent state first?

cachito/web/models.py
```diff
@@ -47,6 +47,33 @@ class Request(db.Model):
     def __repr__(self):
         return '<Request {0!r}>'.format(self.id)
 
+    @property
+    def state(self):
```
lcarva: Should `state` be `latest_state_name` and `latest_state` be `latest_state_object` ? Both of them return information about the latest state.

cachito/web/models.py
```diff
@@ -47,6 +47,33 @@ class Request(db.Model):
...
+    @property
+    def latest_state(self):
+        """
+        Get the latest RequestState associated with this request.
+
+        :return: the latest RequestState associated with this request
+        :rtype: RequestState
+        """
+        return (
+            RequestState.query
```
lcarva: Shouldn't this be accessible through the backreference states_history?

cachito/workers/tasks.py
```diff
@@ -90,3 +88,39 @@ def assemble_source_code_archive(app_archive_path, deps_path, bundle_archive_pat
...
+    """
+    config = get_worker_config()
+    request_url = f'{config.cachito_api_url.rstrip("/")}/requests/{request_id}'
+
+    log.info('Setting the status of request %d to "%s"', request_id, state)
+    payload = {'state': state}
+    if state_reason:
+        payload['state_reason'] = state_reason
+    try:
+        rv = requests.patch(request_url, json=payload, timeout=30)
```
lcarva: Let's add a config for timeout here.

cachito/web/models.py
mprahl: Yes, but calling that would retrieve all the states associated with the request. I figured it'd be more efficient to grab just the one RequestState. Regardless, after refactoring this PR, this is no longer needed so I'll just remove it.

cachito/web/models.py
mprahl: It will be after refactoring the PR.

cachito/workers/tasks.py
mprahl: The request for downloading the source archive is set to 120. That seems reasonable, but this shouldn't be that high. So that means we'd have to have separate config options for both.

I like the idea of making this configurable, but let's make it a separate PR.

mprahl:
@lcarva this is ready for review again. The main differences are:
* Failed async tasks in `create_request` now set the state reason based on the exception that caused the failure.
* The request state is updated when fetching the application source and when fetching the golang dependencies.
* The state reason is no longer optional/nullable. At first I thought it'd be nice to not have to provide a state reason for certain use-cases, but I think it's best to be consistent, and being verbose is better than too little information in my opinion.
* The status code for the patch route returns `200` on success instead of `201`. I had originally taken a different approach and had forgotten to update the status code after refactoring.

Here is the diff from the last review, but some of the changes were cleaning up the commit history in the PR:
https://github.com/release-engineering/cachito/compare/75786fe0b5c5551df9753ff72cb2be0cca51cdc1..62af154af8174a95d2db089aecc92db24d6d42bc

PR #12, opened 2019-06-07
Add support for user authentication

mprahl:
This implementation relies on a web server that supplies the `REMOTE_USER` environment variable when the user is properly authenticated. A common deployment option is using httpd (Apache web server) with the `mod_auth_gssapi` module.

This also includes a couple commits that are logically separate but related.

### DIFF
--- a/cachito/web/auth.py
+++ b/cachito/web/auth.py
@@ -1,0 +1,64 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+import logging
+
+from flask import current_app
+
+from cachito.web import db
+from cachito.web.models import User
+
+
+log = logging.getLogger(__name__)
+
+
+def user_loader(username):
+    """
+    Get the user by their username from the database.
+
+    This is used by the Flask-Login library.
+
+    :param str username: the username of the user
+    :return: the User object associated with the username or None
+    :rtype: cachito.web.models.User
+    """
+    return User.query.filter_by(username=username).first()
+
+
+def load_user_from_request(request):
+    """
+    Load the user that authenticated from the current request.
+
+    When authentication is turned on (default in production), this relies on the "REMOTE_USER"
+    environment variable being set. This is usually set by the mod_auth_gssapi Apache authentication
+    module.
+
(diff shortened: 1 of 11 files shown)


cachito/web/config.py
```diff
@@ -21,14 +22,17 @@ class DevelopmentConfig(Config):
...
 class TestingConfig(DevelopmentConfig):
     """The testing Cachito Flask configuration."""
+    CACHITO_WORKER_USERNAMES = ['worker']
     # IMPORTANT: don't use in-memory sqlite. Alembic migrations will create a new
     # connection producing a new instance of the database which is deleted immediately
     # after the migration completes...
     #   https://github.com/miguelgrinberg/Flask-Migrate/issues/153
     SQLALCHEMY_DATABASE_URI = f'sqlite:///{TEST_DB_FILE}'
     DEBUG = True
+    LOGIN_DISABLED = False
```
lcarva: Consider defining this in the `Config` class so it's always defined.

mprahl: This is just here because the `TestingConfig` class inherits from the `DevelopmentConfig`. It defaults to `False`.

cachito/web/auth.py
```diff
@@ -0,0 +1,65 @@
...
+    if not remote_user:
+        if current_app.config.get('LOGIN_DISABLED', False) is True:
+            log.info(
+                'The REMOTE_USER environment variable wasn\'t set on the request, but the '
+                'LOGIN_DISABLED configuration is set to True.'
+            )
+        return
+
+    log.info(f'The user "{remote_user}" was authenticated successfully by httpd')
+    if '@' in remote_user:
```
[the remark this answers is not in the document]
mprahl: I'm not sure. I suppose it was just because we have one Kerberos realm, and it seemed cleaner to just store the username, but you're right. We might as well keep it. I'll change it to all lower-case though in case capitalization comes out differently. Freshmaker currently does this as well, so I don't know if having it consistent matters to you.

mprahl:
@lcarva thanks for the review! I addressed your comments. Could you please take another look?

PR #21, opened 2019-07-24
Don't add a request state if it matches the previous state exactly

mprahl:
This is to protect against a Celery task getting executed twice and setting the state each time.


mprahl:
@lcarva do you think the API endpoint should be doing this check or should it be in the `set_request_state` Celery task? I can't really think of a reason why you'd want to add a duplicate state, so I added it to the API endpoint, but let me know if you disagree.

lcarva:
It looks like this was merged as part of https://github.com/release-engineering/cachito/pull/23

PR #25, opened 2019-07-26
Override some of the Celery defaults for production use

mprahl:
This relies on PR #24.

### DIFF
--- a/cachito/workers/config.py
+++ b/cachito/workers/config.py
@@ -3,18 +3,43 @@ import os
 import tempfile
 import logging
 
+import kombu
+
 from cachito.errors import ConfigError
 
 
 class Config(object):
     """The base Cachito Celery configuration."""
 
-    # Don't use the default 'celery' queue
-    task_default_queue = 'cachito'
     cachito_auth_type = None
     cachito_log_level = 'INFO'
     result_backend = 'rpc'
     result_persistent = True
+    # The task messages will be acknowledged after the task has been executed,
+    # instead of just before
+    task_acks_late = True
+    # Don't use the default 'celery' queue and routing key
+    task_default_queue = 'cachito'
+    task_default_routing_key = 'cachito'
+    # By default, have the worker process general and golang tasks
+    task_queues = (
+        kombu.Queue('cachito'),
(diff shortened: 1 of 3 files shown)


cachito/workers/config.py
```diff
@@ -3,16 +3,41 @@
 import tempfile
 import logging
 
+import kombu
```
mprahl: @lcarva this dependency comes from Celery, but since I'm directly importing it, should it be part of the requirements.txt file?

lcarva: Yes. I think so since it's now a direct dependency.

cachito/workers/config.py
```diff
@@ -3,16 +3,41 @@
...
 class Config(object):
     """The base Cachito Celery configuration."""
 
-    # Don't use the default 'celery' queue
-    task_default_queue = 'cachito'
     cachito_auth_type = None
     cachito_log_level = 'INFO'
+    # The task messages will be acknowledged after the task has been executed,
+    # instead of just before
+    task_acks_late = True
```
lcarva: Some tasks may take a while to complete. If we have more than worker for a given queue, won't this cause messages to be resent while actually being processed? 

Even if we set `worker_concurrency` to 1, as soon as we have more than one worker pod, we may hit this situation.

mprahl: @lcarva why do you think it will be resent if it takes too long? The heartbeat of the Celery worker will be kept alive with RabbitMQ during this time.

mprahl:
@lcarva I rebased the PR by the way.

cachito/workers/config.py
lcarva: Something to watch out for. I'm not sure if the heartbeat will be maintained. I was seeing some odd behavior when using Cachito to fetch the sources for kubernetes.

mprahl:
@lcarva is this good to merge?

cachito/workers/config.py
mprahl: Was it using this configuration or the configuration before this PR?

lcarva: Before this PR.

PR #32, opened 2019-08-09
Don't retry more than 5 times when publishing a message

mprahl:


lcarva:
Something to keep in mind is that if when we do a deployment of RabbitMQ, which will restart a new pod, it takes about a minute for it to be fully functional.

mprahl:
@lcarva okay, based on my reading of the kombu code, it seems like by default it sleeps 2 seconds and adds 2 seconds each time per failed attempt. So I think in this case it should be 30 seconds of sleeping. I can bump this up to 10 retries so it's over a minute of sleeping.

PR #35, opened 2019-08-09
Fix logging in the API

mprahl:
This also adds a couple extra log statements to let us know what's going on in the API.

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -1,5 +1,4 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
-import logging
 import tempfile
 import os
 import re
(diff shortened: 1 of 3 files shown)


cachito/web/api_v1.py
```diff
@@ -196,5 +200,10 @@ def patch_request(request_id):
                 request.dependencies.append(dep_obj)
 
     db.session.commit()
+    if current_user.is_authenticated:
+        flask.current_app.logger.info(
+            'The user %s patched request %d', current_user.username, request.id)
```
lcarva: What's `current_user.username` for anonymous users?

mprahl: It raises an `AttributeError`.

lcarva: Bummer... Ideally, this would be set to something reasonable such as `None`. I don't necessarily like the conditional logging. It's easy to forget to do it in all places. But I don't see a way around it.

PR #42, opened 2019-08-30
Create the bundle in the archives dir

mprahl:
This makes it so that the download API endpoint doesn't wait for a worker to assemble the bundle. Instead, once the request is complete, the bundle will be present in the archives directory. This will prevent timeouts on large projects and it will address the unstable behavior seen on NFS volumes.

In another PR, I'll work on being able to mark a request as stale through the API, which will cause it to be deleted.

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -1,12 +1,11 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
-import tempfile
 import os
 import re
 
 from celery import chain
 import flask
 from flask_login import current_user, login_required
-from werkzeug.exceptions import Unauthorized
+from werkzeug.exceptions import Unauthorized, InternalServerError
 
 from cachito.errors import ValidationError
 from cachito.web import db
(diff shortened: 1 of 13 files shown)


cachito/web/api_v1.py
```diff
@@ -65,29 +64,21 @@ def download_archive(request_id):
...
-        return flask.send_file(absolute_bundle_archive_path, mimetype='application/gzip')
+    cachito_archives_dir = flask.current_app.config['CACHITO_ARCHIVES_DIR']
+    bundle_archive_path = os.path.join(
+        cachito_archives_dir, 'cachito_bundles', f'{request_id}.tar.gz')
+    if not os.path.exists(bundle_archive_path):
+        flask.current_app.logger.error(
+            'The bundle archive at %s for request %d doesn\'t exist',
+            bundle_archive_path, request_id,
+        )
+        raise InternalServerError()
```
hanzz: Would `Not Found` make more sense here? I see your logic with InternalServerError in case the bundle should always be there for complete request.

cachito/web/api_v1.py
```diff
@@ -65,29 +64,21 @@ def download_archive(request_id):
...
-        # workers and the API, they may be mounted at different locations. Use relative
-        # paths to agree on data location within the shared volume.
-        relative_temp_dir = os.path.basename(temp_dir)
-        relative_deps_path = os.path.join(relative_temp_dir, 'deps')
-        relative_bundle_archive_path = os.path.join(relative_temp_dir, 'bundle.tar.gz')
-        absolute_bundle_archive_path = os.path.join(
-            cachito_shared_dir, relative_bundle_archive_path)
-
-        # Chain tasks
-        chain_result = chain(
```
hanzz: I think I don't know the code well enough to understand how these three tasks are replaced by single `add_deps_to_bundle_archive` call run on backend. I only presume that those tasks move somehow files from backend to frontend, so the frontend code can repackage them into bundle.

hanzz:
This looks OK to me. Instead of generating the bundle on frontend, you generate it as part of the task itself and store it into CACHITO_ARCHIVES_DIR so it is available on frontend and user can download it directly.

+1.

cachito/web/api_v1.py
mprahl: Yes, it should always exist. If it doesn't, it's an issue such as the volume not being mounted or it got accidentally deleted.

cachito/web/api_v1.py
mprahl: The previous code to get the download bundle was basically duplicating a lot of tasks that happen as part of the request handling. This is no longer needed since the bundle is created as part of the request.

mprahl:
Thanks for the review @hanzz. I added another commit which deletes the bundle archive when the request is transitioned to the new `stale` state. This is to save space since some of these bundles are quite large.

Eventually, we'll have a scheduled task that expires old requests.

cachito/workers/pkg_manager.py
```diff
@@ -74,13 +75,12 @@ def resolve_gomod_deps(archive_path, copy_cache_to=None):
...
-            # Copy gomod cache to requested location
+        # Add the gomod cache to the bundle the user will later download
+        if request_id is not None:
             cache_path = os.path.join('pkg', 'mod', 'cache', 'download')
             src_cache_path = os.path.join(temp_dir, cache_path)
-            dest_cache_path = os.path.join(
-                worker_config.cachito_shared_dir, copy_cache_to, 'gomod', cache_path)
-            shutil.copytree(src_cache_path, dest_cache_path)
+            dest_cache_path = os.path.join('gomod', cache_path)
+            add_deps_to_bundle_archive(request_id, archive_path, src_cache_path, dest_cache_path)
```
lcarva: It seems a bit odd for `resolve_gomod_deps` to actually create the final bundle. This will be problematic when we add support for other package managers and we want to parallelize dependency resolution.

What if we introduced a new task to finalize the bundle? In that way, each task essentially dumps content in a known location, and the finalize task creates the downloadable archive which simply includes everything from that location.

_This is not a must because currently only gomod is supported._

cachito/web/models.py
```diff
@@ -208,6 +212,8 @@ def add_state(self, state, state_reason):
                 .format(state, ', '.join(RequestStateMapping.get_state_names()))
             )
 
+        if state == 'stale' and os.path.exists(self.bundle_archive):
+            os.remove(self.bundle_archive)
```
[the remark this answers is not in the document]
mprahl: Good point. I'll move the code to after the database session is commited.

cachito/workers/pkg_manager.py
mprahl: As discussed in IRC. I'll refactor this in a separate PR.

PR #45, opened 2019-09-04
Assemble the bundle archive after all the package managers complete

mprahl:
Assemble the bundle archive after all the package managers to allow running the package managers in parallel.

This works by having each package manager add their dependencies to "bundles/temp/<request ID>/deps". Once all the package managers have been completed, a final Celery task assembles the bundle archive.

### DIFF
--- a/cachito/workers/tasks/general.py
+++ b/cachito/workers/tasks/general.py
@@ -1,5 +1,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import logging
+import os
+import tarfile
 
 import requests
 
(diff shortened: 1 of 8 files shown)


cachito/workers/tasks/general.py
```diff
@@ -95,3 +98,38 @@ def failed_request_callback(context, exc, traceback, request_id):
         msg = 'An unknown error occurred'
 
     set_request_state(request_id, 'failed', msg)
+
+
+@app.task
+def create_bundle_archive(app_archive_path, request_id):
```
lcarva: Don't we need to change `resolve_gomod_deps` to return the `app_archive_path` for this to work? It currently returns the list of deps.

mprahl: The Celery task is `fetch_gomod_source` which returns `app_archive_path`.  `fetch_gomod_source` wraps `resolve_gomod_deps`.

PR #49, opened 2019-09-06
Include gomod environment variables in request

lcarva:
When gomod package manager is used, certain environment variables should
be set to ensure the cached content is used. This change enhances
request objects to include which environment variables, along with their
corresponding values, should be set.

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -1,4 +1,5 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
+from collections import OrderedDict
 from copy import deepcopy
 from enum import Enum
 import os
@@ -23,6 +24,13 @@ request_dependency_table = db.Table(
     db.Column('request_id', db.Integer, db.ForeignKey('request.id'), nullable=False),
     db.Column('dependency_id', db.Integer, db.ForeignKey('dependency.id'), nullable=False),
     db.UniqueConstraint('request_id', 'dependency_id'),
+)
+
+request_environment_variable_table = db.Table(
+    'request_environment_variable',
+    db.Column('request_id', db.Integer, db.ForeignKey('request.id'), nullable=False),
+    db.Column('env_var_id', db.Integer, db.ForeignKey('environment_variable.id'), nullable=False),
+    db.UniqueConstraint('request_id', 'env_var_id'),
 )
 
 
@@ -103,6 +111,9 @@ class Request(db.Model):
                                    backref='requests')
     states = db.relationship(
         'RequestState', back_populates='request', order_by='RequestState.updated')
+    environment_variables = db.relationship(
+        'EnvironmentVariable', secondary=request_environment_variable_table, backref='requests',
+        order_by='EnvironmentVariable.name')
     user = db.relationship('User', back_populates='requests')
 
     def __repr__(self):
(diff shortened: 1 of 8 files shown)


tests/test_api_v1.py
```diff
@@ -371,7 +375,7 @@ def test_set_deps(app, client, db, worker_auth_env, sample_deps):
     db.session.add(request)
     db.session.commit()
 
-    payload = {'dependencies': sample_deps}
+    payload = {'dependencies': sample_deps, 'environment_variables': env_vars}
```
mprahl: Could you please verify the environment variables are actually recorded?

cachito/web/models.py
```diff
@@ -273,6 +283,27 @@ def __repr__(self):
...
+    name = db.Column(db.String, nullable=False)
+    value = db.Column(db.String, nullable=False)
+
+    __table_args__ = (
+        db.UniqueConstraint('name', 'value'),
+    )
+
+    @classmethod
+    def validate_json(cls, name, value):
+        if not isinstance(value, str):
```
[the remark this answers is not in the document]
lcarva: Let's skip on this for now. We can add this validation if there's a need in the future.

cachito/web/models.py
```diff
@@ -103,6 +110,8 @@ class Request(db.Model):
                                    backref='requests')
     states = db.relationship(
         'RequestState', back_populates='request', order_by='RequestState.updated')
+    environment_variables = db.relationship(
+        'EnvironmentVariable', secondary=request_environment_variable_table, backref='requests')
```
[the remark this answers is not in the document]
lcarva: That won't do much since this gets mapped into a python dict.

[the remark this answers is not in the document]
lcarva: Ok, added this and changed it to use OrderedDict.

cachito/web/models.py
```diff
@@ -158,6 +167,7 @@ def to_json(self):
             'pkg_managers': pkg_managers,
             'state_history': states,
             'user': user,
+            'environment_variables': {var.name: var.value for var in self.environment_variables}
```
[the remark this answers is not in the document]
lcarva: Yes. The challenge here is that we're pulling apart pieces of the EnvironmentVariable object. I've added the `to_json` method which returns a tuple of two items which represent the key and the value in the dict.

tests/test_api_v1.py
lcarva: Good catch! I also added a  verification to ensure the environment variables in the request json actually maps to the expected value.

lcarva:
@mprahl, this is ready for another review! Thanks!

cachito/web/models.py
```diff
@@ -273,6 +286,30 @@ def __repr__(self):
...
+            raise ValidationError(
+                'The value of environment variables must be a string')
+
+    @classmethod
+    def from_json(cls, name, value):
+        cls.validate_json(name, value)
+        return cls(name=name, value=value)
+
+    def to_json(self):
+        return self.name, self.value
```
mprahl: I see your point from your previous comment. I liked it better without the `to_json` method for this use case.

lcarva: I don't mind it this way. The benefit is that this gives us the opportunity to normalize the key or value in the future. Of course, we don't have a need for that now.

Since a tuple can be serialized as JSON (Array), the name of the method is accurate. 

If you feel strongly about this, I can change it back.

mprahl: That works for me.

PR #50, opened 2019-09-17
Add support for flags in requests

yashvardhannanavati:
@lcarva @mprahl can you PTAL?

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -33,6 +33,13 @@ request_environment_variable_table = db.Table(
     db.UniqueConstraint('request_id', 'env_var_id'),
 )
 
+request_flag_table = db.Table(
+    'request_flag',
+    db.Column('request_id', db.Integer, db.ForeignKey('request.id'), nullable=False),
+    db.Column('flag_id', db.Integer, db.ForeignKey('flag.id'), nullable=False),
+    db.UniqueConstraint('request_id', 'flag_id'),
+)
+
 
 class RequestStateMapping(Enum):
     """
@@ -115,6 +122,8 @@ class Request(db.Model):
         'EnvironmentVariable', secondary=request_environment_variable_table, backref='requests',
         order_by='EnvironmentVariable.name')
     user = db.relationship('User', back_populates='requests')
+    flags = db.relationship(
+        'Flag', secondary=request_flag_table, backref='requests', order_by='Flag.name')
 
     def __repr__(self):
         return '<Request {0!r}>'.format(self.id)
@@ -171,6 +180,7 @@ class Request(db.Model):
             'state_history': states,
             'user': user,
             'environment_variables': env_vars_json,
+            'flags': [flag.to_json() for flag in self.flags if flag.active],
         }
         # Show the latest state information in the first level of the JSON
         rv.update(latest_state)
@@ -180,13 +190,14 @@ class Request(db.Model):
     def from_json(cls, kwargs):
         # Validate all required parameters are present
         required_params = {'repo', 'ref', 'pkg_managers'}
-        missing_params = required_params - set(kwargs.keys())
+        optional_params = {'flags'}
+        missing_params = required_params - set(kwargs.keys()) - optional_params
         if missing_params:
             raise ValidationError('Missing required parameter(s): {}'
                                   .format(', '.join(missing_params)))
 
         # Don't allow the user to set arbitrary columns or relationships
-        invalid_params = kwargs.keys() - required_params
+        invalid_params = set(kwargs.keys()) - required_params - optional_params
         if invalid_params:
             raise ValidationError(
                 'The following parameters are invalid: {}'.format(', '.join(invalid_params)))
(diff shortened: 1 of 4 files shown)


cachito/web/models.py
```diff
@@ -115,6 +122,7 @@ class Request(db.Model):
         'EnvironmentVariable', secondary=request_environment_variable_table, backref='requests',
         order_by='EnvironmentVariable.name')
     user = db.relationship('User', back_populates='requests')
+    flags = db.relationship('Flag', secondary=request_flag_table, backref='requests')
```
lcarva: Maybe order by `Flag.name` so the response is consistent?

cachito/web/models.py
```diff
@@ -180,13 +189,14 @@ def to_json(self):
...
-        missing_params = required_params - set(kwargs.keys())
+        optional_params = {'flags'}
+        missing_params = required_params - set(kwargs.keys()) - optional_params
         if missing_params:
             raise ValidationError('Missing required parameter(s): {}'
                                   .format(', '.join(missing_params)))
 
         # Don't allow the user to set arbitrary columns or relationships
-        invalid_params = kwargs.keys() - required_params
+        invalid_params = kwargs.keys() - required_params - optional_params
```
lcarva: To compute `missing_params`, `kwargs.keys()` is converted into a `set`. Why not do it here also?

cachito/web/models.py
```diff
@@ -209,10 +219,26 @@ def from_json(cls, kwargs):
                                   .format(', '.join(invalid_pkg_managers)))
 
         request_kwargs['pkg_managers'] = found_pkg_managers
+
+        flags_kwargs = request_kwargs.pop('flags', None)
```
lcarva: `flag_names` ?

yashvardhannanavati:
@lcarva made the changes. Can you PTAL?

cachito/web/models.py
```diff
@@ -115,6 +122,8 @@ class Request(db.Model):
         'EnvironmentVariable', secondary=request_environment_variable_table, backref='requests',
         order_by='EnvironmentVariable.name')
     user = db.relationship('User', back_populates='requests')
+    flags = db.relationship('Flag', secondary=request_flag_table, backref='requests',
```
lcarva: This is nitpicking, but we prefer this form of indentation:
```python
flags = db.relationship(
    'Flag', secondary=request_flag_table, backref='requests', order_by='Flag.name')
```

tests/test_api_v1.py
```diff
@@ -153,6 +154,20 @@ def test_malformed_request_id(client, db):
     assert data == {'error': 'The requested resource was not found'}
 
 
+def test_create_request_invalid_flag(auth_env, client, db):
```
lcarva: It would be nice to add a valid test as well. (Will need to add test flags to the test db)

yashvardhannanavati: Added a new test which submits a request when the flag is active and fetches the request when the flag is set to inactive. :)

yashvardhannanavati:
@lcarva made the changes. PTAL?

cachito/web/models.py
```diff
@@ -171,6 +180,7 @@ def to_json(self):
             'state_history': states,
             'user': user,
             'environment_variables': env_vars_json,
+            'flags': [flag.to_json() for flag in self.flags if flag.active],
```
mprahl: This doesn't look right to me. Don't we want to show the flags that were used historically even if they were disabled?

cachito/web/models.py
```diff
@@ -209,10 +220,26 @@ def from_json(cls, kwargs):
                                   .format(', '.join(invalid_pkg_managers)))
 
         request_kwargs['pkg_managers'] = found_pkg_managers
+
+        flag_names = request_kwargs.pop('flags', None)
+        if flag_names:
+            flag_names = set(flag_names)
+            found_flags = (Flag.query
```
mprahl: Nitpick: This indentation doesn't match the rest of the codebase

cachito/web/models.py
lcarva: I'm struggling to find the right balance here. If we display them, we have a transparent way of showing historical results. However, users can't "replay" the same request. Knowledge of what's active is required.

For auditing purposes, we could still inspect the database for all the flags used on any request, but this requires DB access.

We could represent the active state of a flag in the response but that seems a bit verbose, and potentially unnecessary, to me.

mprahl: @lcarva, I think we should show flags as they were historically. If the user replays a request and the flags that are disabled are omitted, the old request isn't actually reproduced. This is misleading to the user in my opinion. I think we should show an error message in this case stating that the following flags in replayed request are not valid.

lcarva: Consider the case where we want to roll out a feature gradually. This will be done by having some users setting a flag. Eventually, that feature is main stream and all the requests will behave as if the flag is set. At this point, we should mark the flag as inactive.

It seems odd to me to make those previous requests not "replayable". However, I do see the point that hiding those flags may make things more complicated.

What if we always show the used flags regardless if they're no longer active, as you suggested, but change how we handle processing them for new requests.

Going back to the example I mentioned earlier, it would be a pretty bad user experience if we decide to deem a flag no longer necessary and mark it as inactive, only to break requests for those early adopters using the flag. In the POST processing, we could filter out inactive flags from the request, and continue to reject invalid flags. This also addresses the replaying scenario. To be clear, I don't think the inactive flags should be added to the request, but could be silently ignored.

mprahl: @lcarva, I had misunderstood the point of the "active" column on the flag then. I had assumed that a flag would be disabled if we didn't want to allow the use of the flag anymore. For instance, we tried a new feature and decided it wasn't a good idea and wanted to disable it.

Your approach works for me. @yashvardhannanavati, could you make the change so that the flags are not filtered? Since we don't allow a request to be "replayable" yet, there's nothing to do for that part of the feature that was described.

yashvardhannanavati: +1 I'll make that change and submit a separate PR.

PR #52, opened 2019-09-19
Add pagination metadata for the requests endpoint

yashvardhannanavati:
@lcarva  @mprahl can you PTAL?

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -11,6 +11,7 @@ from werkzeug.exceptions import Unauthorized, InternalServerError
 from cachito.errors import ValidationError
 from cachito.web import db
 from cachito.web.models import Request, Dependency, EnvironmentVariable
+from cachito.web.utils import pagination_metadata
 from cachito.workers import tasks
 
 
@@ -30,9 +31,11 @@ def get_requests():
     max_per_page = flask.current_app.config['CACHITO_MAX_PER_PAGE']
     # The call to `paginate` will inspect the current HTTP request for the
     # pagination parameters `page` and `per_page`.
-    requests = Request.query.paginate(max_per_page=max_per_page).items
+    pagination_query = Request.query.paginate(max_per_page=max_per_page)
+    requests = pagination_query.items
     response = {
         'items': [request.to_json() for request in requests],
+        'meta': pagination_metadata(pagination_query),
     }
     return flask.jsonify(response)
 
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -11,7 +11,7 @@ from werkzeug.exceptions import Unauthorized, InternalServerError
 from cachito.errors import ValidationError
 from cachito.web import db
 from cachito.web.models import Request, Dependency, EnvironmentVariable
-from cachito.web.utils import pagination_metadata
+from cachito.web.utils import pagination_metadata, str_to_bool
 from cachito.workers import tasks
 
 
@@ -28,14 +28,16 @@ def get_requests():
         value exceeds configuration's CACHITO_MAX_PER_PAGE.
     :rtype: flask.Response
     """
+    # Default versbose flag to False
+    verbose = str_to_bool(flask.request.args.get('verbose', False))
     max_per_page = flask.current_app.config['CACHITO_MAX_PER_PAGE']
     # The call to `paginate` will inspect the current HTTP request for the
     # pagination parameters `page` and `per_page`.
     pagination_query = Request.query.paginate(max_per_page=max_per_page)
     requests = pagination_query.items
     response = {
-        'items': [request.to_json() for request in requests],
-        'meta': pagination_metadata(pagination_query),
+        'items': [request.to_json(verbose=verbose) for request in requests],
+        'meta': pagination_metadata(pagination_query, verbose=verbose),
     }
     return flask.jsonify(response)
 
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -150,7 +150,19 @@ class Request(db.Model):
         cachito_bundles_dir = flask.current_app.config['CACHITO_BUNDLES_DIR']
         return os.path.join(cachito_bundles_dir, 'temp', str(self.id))
 
-    def to_json(self):
+    @property
+    def number_of_deps(self):
+        """
+        Get the total number of dependencies for a request.
+
+        :return: count of the number of dependencies
+        :rtype: int
+        """
+        req_dep_cols = request_dependency_table.c
+        return db.session.query(sqlalchemy.func.count(req_dep_cols.dependency_id)).filter(
+            req_dep_cols.request_id == self.id).scalar()
+
+    def to_json(self, verbose=True):
         pkg_managers = [pkg_manager.to_json() for pkg_manager in self.pkg_managers]
         # Use this list comprehension instead of a RequestState.to_json method to avoid including
         # redundant information about the request itself
@@ -172,18 +184,21 @@ class Request(db.Model):
 
         env_vars_json = OrderedDict(env_var.to_json() for env_var in self.environment_variables)
         rv = {
-            'dependencies': [dep.to_json() for dep in self.dependencies],
             'id': self.id,
             'repo': self.repo,
             'ref': self.ref,
             'pkg_managers': pkg_managers,
-            'state_history': states,
             'user': user,
             'environment_variables': env_vars_json,
             'flags': [flag.to_json() for flag in self.flags],
         }
         # Show the latest state information in the first level of the JSON
         rv.update(latest_state)
+        if verbose:
+            rv.update({'state_history': states})
+            rv.update({'dependencies': [dep.to_json() for dep in self.dependencies]})
+        else:
+            rv.update({'dependencies': self.number_of_deps})
         return rv
 
     @classmethod
--- a/cachito/web/utils.py
+++ b/cachito/web/utils.py
@@ -1,0 +1,37 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+from flask import request, url_for
+
+
+def pagination_metadata(pagination_query):
+    """
+    Return a dictionary containing metadata about the paginated query.
+    This must be run as part of a Flask request.
+
+    :param pagination_query: flask_sqlalchemy.Pagination object
+    :return: a dictionary containing metadata about the paginated query
+    """
+
+    pagination_data = {
+        'first': url_for(
+            request.endpoint, page=1, per_page=pagination_query.per_page, _external=True),
+        'last': url_for(
+            request.endpoint, page=pagination_query.pages, per_page=pagination_query.per_page,
+            _external=True),
+        'next': None,
+        'page': pagination_query.page,
+        'pages': pagination_query.pages,
+        'per_page': pagination_query.per_page,
+        'previous': None,
+        'total': pagination_query.total,
+    }
+
+    if pagination_query.has_prev:
+        pagination_data['previous'] = url_for(
+            request.endpoint, page=pagination_query.prev_num,
+            per_page=pagination_query.per_page, _external=True)
+    if pagination_query.has_next:
+        pagination_data['next'] = url_for(
+            request.endpoint, page=pagination_query.next_num,
+            per_page=pagination_query.per_page, _external=True)
+
+    return pagination_data
--- a/cachito/web/utils.py
+++ b/cachito/web/utils.py
@@ -2,7 +2,7 @@
 from flask import request, url_for
 
 
-def pagination_metadata(pagination_query):
+def pagination_metadata(pagination_query, verbose):
     """
     Return a dictionary containing metadata about the paginated query.
     This must be run as part of a Flask request.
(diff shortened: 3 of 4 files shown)


cachito/web/api_v1.py
```diff
@@ -12,6 +12,7 @@
 from cachito.web import db
 from cachito.web.models import Request, Dependency, EnvironmentVariable
 from cachito.workers import tasks
+from cachito.web.utils import pagination_metadata
```
mprahl: Nitpick: This is not alphabetized :smile:

cachito/web/utils.py
```diff
@@ -0,0 +1,32 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+from flask import request, url_for
+
+
+def pagination_metadata(pagination_query):
+    """
+    Returns a dictionary containing metadata about the paginated query.
```
mprahl: Nitpick: A new line is needed after this

The first word of the summary should also be singular.

Perhaps, I should add flake8-docstrings to Cachito to enforce this.

cachito/web/utils.py
```diff
@@ -0,0 +1,32 @@
...
+        'first': url_for(
+            request.endpoint, page=1, per_page=pagination_query.per_page, _external=True),
+        'last': url_for(
+            request.endpoint, page=pagination_query.pages, per_page=pagination_query.per_page,
+            _external=True)
+    }
+
+    if pagination_query.has_prev:
+        pagination_data['prev'] = url_for(request.endpoint, page=pagination_query.prev_num,
+                                          per_page=pagination_query.per_page, _external=True)
```
mprahl: Nitpick: Could you fix the indentation style so it matches the rest of the code?

cachito/web/utils.py
```diff
@@ -0,0 +1,32 @@
...
+    if pagination_query.has_next:
+        pagination_data['next'] = url_for(request.endpoint, page=pagination_query.next_num,
```
mprahl: Nitpick: Could you fix the indentation style so it matches the rest of the code?

cachito/web/utils.py
```diff
@@ -0,0 +1,32 @@
...
+
+def pagination_metadata(pagination_query):
+    """
+    Returns a dictionary containing metadata about the paginated query.
+    This must be run as part of a Flask request.
+    :param pagination_query: flask_sqlalchemy.Pagination object
+    :return: a dictionary containing metadata about the paginated query
+    """
+
+    pagination_data = {
```
mprahl: Can you set default values to `next` and `prev` so that they are just null when they aren't applicable.

cachito/web/utils.py
```diff
@@ -0,0 +1,32 @@
...
+        'pages': pagination_query.pages,
+        'first': url_for(
+            request.endpoint, page=1, per_page=pagination_query.per_page, _external=True),
+        'last': url_for(
+            request.endpoint, page=pagination_query.pages, per_page=pagination_query.per_page,
+            _external=True)
+    }
+
+    if pagination_query.has_prev:
+        pagination_data['prev'] = url_for(request.endpoint, page=pagination_query.prev_num,
```
mprahl: I'd prefer using `previous` instead of `prev`.

tests/test_api_v1.py
```diff
@@ -85,6 +85,12 @@ def test_fetch_paginated_requests(mock_chain, app, auth_env, client, db):
     # Start at 10 because each page contains 10 items and we're processing the second page
     for repo_number, request in enumerate(fetched_requests, 10):
         assert request['repo'] == repo_template.format(repo_number)
+    pagination_metadata = response['meta']
+    assert pagination_metadata['next'].endswith('?page=3&per_page=10')
+    assert pagination_metadata['last'].endswith('?page=5&per_page=10')
+    assert pagination_metadata['first'].endswith('?page=1&per_page=10')
+    assert pagination_metadata['prev'].endswith('?page=1&per_page=10')
```
mprahl: Is the order of the query parameters always guaranteed? I'd be worried that it's not depending on the system and it could cause these tests to fail.

yashvardhannanavati: I don't think the order is always guaranteed but the order shouldn't matter in the tests since they access individual keys and don't check for the order.

yashvardhannanavati:
@mprahl updated the PR :)

yashvardhannanavati:
@lcarva @mprahl can you PTAL? I added one more commit to it.

cachito/web/api_v1.py
```diff
@@ -27,12 +28,18 @@ def get_requests():
         value exceeds configuration's CACHITO_MAX_PER_PAGE.
     :rtype: flask.Response
     """
+    # Default versbose flag to False
+    verbose = False
```
hanzz: It should be possible to rewrite that as:

```
verbose = str_to_bool(flask.request.args.get('verbose', 'false'))
```

But just an idea, you can keep the current way.

cachito/web/models.py
```diff
@@ -172,18 +172,21 @@ def to_json(self):
...
             'ref': self.ref,
             'pkg_managers': pkg_managers,
-            'state_history': states,
             'user': user,
             'environment_variables': env_vars_json,
             'flags': [flag.to_json() for flag in self.flags if flag.active],
         }
         # Show the latest state information in the first level of the JSON
         rv.update(latest_state)
+        if verbose:
```
hanzz: Are there some users of existing API already? If so, are we sure the (by default) removed/changed `dependencies` and `state_history` won't break them?

mprahl: The API is not used yet which is why we're making this change now  :smile:

cachito/web/models.py
```diff
@@ -172,18 +172,21 @@ def to_json(self):
...
+            rv.update({'state_history': states})
+            rv.update({'dependencies': [dep.to_json() for dep in self.dependencies]})
+        else:
+            rv.update({'dependencies': len(self.dependencies)})
```
mprahl: This will pull down all the dependencies from the database and then provide the length of the list. This is inefficient and will make the REST API slower. How about using SQL to get the count?

For instance:
```python
import sqlalchemy

req_dep_cols = request_dependency_table.c
num_deps = db.session.query(sqlalchemy.func.count(req_dep_cols.dependency_id)).filter(req_dep_cols.request_id == self.id).scalar()
```

You could put the code snippet in it's own property method called `number_of_deps` for readability.

cachito/web/utils.py
```diff
@@ -0,0 +1,51 @@
...
+    :param pagination_query: flask_sqlalchemy.Pagination object
+    :return: a dictionary containing metadata about the paginated query
+    """
+
+    pagination_data = {
+        'page': pagination_query.page,
+        'per_page': pagination_query.per_page,
+        'total': pagination_query.total,
+        'pages': pagination_query.pages,
+        'first': url_for(
```
mprahl: Optional: For readability, it'd be nice if the keys were alphabetized

cachito/web/utils.py
```diff
@@ -0,0 +1,51 @@
...
+            request.endpoint, page=1, per_page=pagination_query.per_page, _external=True),
```
mprahl: If the user had a query parameter, it will not be reflected in this generated URL. That needs to be addressed. Here are the docs:
https://flask.palletsprojects.com/en/1.1.x/api/#flask.url_for

tests/test_api_v1.py
```diff
@@ -103,7 +103,8 @@ def test_create_and_fetch_request_with_flag(mock_chain, app, auth_env, client, d
 
 
 @mock.patch('cachito.web.api_v1.chain')
-def test_fetch_paginated_requests(mock_chain, app, auth_env, client, db):
+def test_fetch_paginated_requests(mock_chain, app, auth_env, client, db, sample_deps,
```
mprahl: Optional: We don't have a style guide (we should), but it's preferred to do the following:
```python
def test_fetch_paginated_requests(
    mock_chain, app, auth_env, client, db, sample_deps, worker_auth_env,
):
```

tests/test_api_v1.py
```diff
@@ -118,6 +119,10 @@ def test_fetch_paginated_requests(mock_chain, app, auth_env, client, db):
             db.session.add(request)
     db.session.commit()
 
+    payload = {'dependencies': sample_deps}
+    client.patch('/api/v1/requests/1', json=payload, environ_base=worker_auth_env)
+    client.patch('/api/v1/requests/11', json=payload, environ_base=worker_auth_env)
```
mprahl: What's the purpose of this?

tests/test_api_v1.py
```diff
@@ -126,16 +131,26 @@ def test_fetch_paginated_requests(mock_chain, app, auth_env, client, db):
...
+    rv = client.get('/api/v1/requests?page=2&per_page=10&verbose=True')
     assert rv.status_code == 200
     response = json.loads(rv.data.decode('utf-8'))
     fetched_requests = response['items']
     assert len(fetched_requests) == 10
     # Start at 10 because each page contains 10 items and we're processing the second page
     for repo_number, request in enumerate(fetched_requests, 10):
         assert request['repo'] == repo_template.format(repo_number)
+    pagination_metadata = response['meta']
+    assert pagination_metadata['next'].endswith('?page=3&per_page=10')
```
mprahl: I'll try to clarify my previous comment. My concern is that `?page=3&per_page=10` will not always be in the same order, so the tests will pass on some systems and not on others. To be clear, I think the URL could end up being `?page=3&per_page=10` or `?per_page=10&page=3`.

tests/test_api_v1.py
yashvardhannanavati: So this is to test the `verbose` flag. Without this, there will always be 0 dependencies in all the requests. I added dependencies to the first request of the first page and the second page. And I add the `verbose` flag while requesting the second page

yashvardhannanavati:
@mprahl made the changes. can you PTAL?

cachito/web/utils.py
yashvardhannanavati: Done!

cachito/web/models.py
```diff
@@ -150,7 +150,20 @@ def bundle_temp_files(self):
...
+    @property
+    def number_of_deps(self):
+        """
+        Get the total number of dependencies for a request.
+
+        :return: count of the number of dependencies
+        :rtype: int
+        """
+        req_dep_cols = request_dependency_table.c
+        num_deps = db.session.query(sqlalchemy.func.count(req_dep_cols.dependency_id)).filter(
```
mprahl: A variable isn't necessary. You can just return it directly.

tests/test_api_v1.py
```diff
@@ -126,16 +130,25 @@ def test_fetch_paginated_requests(mock_chain, app, auth_env, client, db):
...
     assert rv.status_code == 200
     response = json.loads(rv.data.decode('utf-8'))
     fetched_requests = response['items']
     assert len(fetched_requests) == 10
     # Start at 10 because each page contains 10 items and we're processing the second page
     for repo_number, request in enumerate(fetched_requests, 10):
         assert request['repo'] == repo_template.format(repo_number)
+    pagination_metadata = response['meta']
+    for page, page_num in [('next', 3), ('last', 5), ('previous', 1), ('first', 1)]:
+        assert 'page={}'.format(page_num) in pagination_metadata[page]
```
mprahl: Optional: Since we're using Python 3, you can use:
```python
f'page={page}'
```

tests/test_api_v1.py
```diff
@@ -126,16 +130,25 @@ def test_fetch_paginated_requests(mock_chain, app, auth_env, client, db):
...
+        assert 'per_page=10' and 'verbose=True' in pagination_metadata[page]
```
mprahl: This assert statement isn't checking if `'per_page=10'` is in `pagination_metadata[page]`. It's just checking if that's truthy. How about the following?
```python
assert 'per_page=10' in pagination_metadata[page]
assert 'verbose=True' in pagination_metadata[page]
```

cachito/web/utils.py
```diff
@@ -0,0 +1,52 @@
...
+            verbose=verbose, _external=True),
+        'last': url_for(
+            request.endpoint, page=pagination_query.pages, per_page=pagination_query.per_page,
+            verbose=verbose, _external=True),
+        'next': None,
+        'page': pagination_query.page,
+        'pages': pagination_query.pages,
+        'per_page': pagination_query.per_page,
+        'previous': None,
+        'total': pagination_query.total
```
mprahl: Optional: It's nice to use a trailing comma

cachito/web/models.py
yashvardhannanavati: It was looking ugly that way so I decided to put it in a variable. But I can do that too.

cachito/web/api_v1.py
```diff
@@ -27,12 +28,16 @@ def get_requests():
         value exceeds configuration's CACHITO_MAX_PER_PAGE.
     :rtype: flask.Response
     """
+    # Default versbose flag to False
```
mprahl: Can you fix the typo in this comment please?

yashvardhannanavati:
Thanks guys!

PR #53, opened 2019-09-26
Show all active/inactive flags in a request

yashvardhannanavati:
Currently, if a flag is active when a request is made and then it is made inacitve, it will not show up in the response when the user requests to see his/her past request. This change will show that flag in the response even though it is inactive now.


yashvardhannanavati:
@lcarva @mprahl can you PTAL? Is this in-line with the discussion we had in #50 ?

yashvardhannanavati:
Thanks Matt!

PR #57, opened 2019-10-02
Document the API using the OpenAPI spec

mprahl:
Once this is merged, I'll enable GitHub pages.


README.md
```diff
@@ -117,3 +117,25 @@ If you are planning to deploy Cachito with authentication enabled, you'll need t
...
+OpenAPI 3.0 format.
+
+To update the documentation deployed on GitHub pages, install `redoc-cli`:
+```bash
+$ npm install redoc-cli
+```
+
+Generate the documentation from the `master` branch:
+```
+$ node_modules/.bin/redoc-cli bundle ./spec/api_v1.yaml --output docs/index.html
```
lcarva: Can we wrap this up in the `Makefile`?

docs/index.html
```diff
@@ -0,0 +1,338 @@
+<!DOCTYPE html>
```
lcarva: Can we serve this page in flask? It would be really useful to have `https://cachito.example.com/docs` serving this file to provide the accurate documentation for a given deployment.

spec/api_v1.yaml
```diff
@@ -0,0 +1,368 @@
...
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  error:
+                    type: string
+                    example: The requested resource was not found
+        "500":
+          description: The input is invalid
```
lcarva: Are we really returning 500 on user provided input? We should fix that.

spec/api_v1.yaml
```diff
@@ -0,0 +1,368 @@
...
+  license:
+    name: "GPL-3.0"
+  title: Cachito API
+  version: "1.0"
+servers:
+  - url: https://cachito.domain.local/api/v1
+paths:
+  /requests:
+    get:
+      description: Returns all the Cachito requests
```
lcarva: The wording `Returns all the Cachito requests` is in a different verb tense than `Create a new request`. FWIW, I think "Return all the Cachito requests" would look better in the UI context.

spec/api_v1.yaml
```diff
@@ -0,0 +1,368 @@
+openapi: "3.0.2"
```
lcarva: Are there any recommendations/suggestions on how to keep this file up to date with the code?

spec/api_v1.yaml
mprahl: That was a typo on my part, and then copied to the rest of the document. It's actually a 400 status code. I'll fix it.

spec/api_v1.yaml
mprahl: I'm not sure. I was planning to make this a part of the code review process.

If we want to make this automated, maybe we can create a pytest plugin/hook that records all of API calls in the unit tests and verifies there is a corresponding entry in the OpenAPI spec for that route, method, and status code. It could be something to explore if we want to add this to the rest of our projects.

PR #58, opened 2019-10-04
Add cleanup job to mark requests as stale after a configurable amount…

yashvardhannanavati:
… of time

Refers to FACTORY-5132

### DIFF
--- a/cachito/workers/cleanup_job.py
+++ b/cachito/workers/cleanup_job.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python3
+import logging
+from datetime import datetime, timedelta
+
+import requests
+
+from cachito.workers.requests import get_requests_session
+from cachito.workers.config import get_worker_config
+from cachito.errors import CachitoError
+
+log = logging.getLogger(__name__)
+
+auth_session = get_requests_session(auth=True)
+session = get_requests_session()
+config = get_worker_config()
+
+# state, state_reason and payload values for marking a request in Cachito API
+state = 'stale'
+state_reason = 'The request has expired'
+payload = {'state': state, 'state_reason': state_reason}
+
+
+def main():
+    """
+    Get all completed requests from the Cachito API, identify the stale ones,
+    and mark them as stale using the API.
+    """
+    url = config.cachito_api_url.rstrip('/') + '/requests'
+    while True:
+        json_response = get_completed_requests(url)
+        identify_and_mark_stale_requests(json_response['items'])
+        if json_response['meta']['next']:
+            url = json_response['meta']['next']
+        else:
+            break
+
+
+def get_completed_requests(url):
+    """
+    Get one page of completed requests from the Cachito API
+
+    :param str url: the URL to fetch the Cachito requests from
+    :raise CachitoError: if the request to the Cachito API fails
+    :rtype: dict
+    """
+    response = session.get(url)
+    if not response.ok:
+        raise CachitoError('Could not reach Cachito API to get all completed requests')
+    return response.json()
+
+
+def identify_and_mark_stale_requests(requests_json):
+    """
+    Identify completed Cachito requests which have reached end of life and mark them as stale.
+
+    :param dict requests_json: the JSON representation of a Cachito API response page
+    """
+    current_time = datetime.utcnow()
+    for request in requests_json:
+        if request['state'] != 'complete':
+            continue
+        date_time_obj = datetime.strptime(request['updated'], '%Y-%m-%dT%H:%M:%S.%f')
+        if current_time - date_time_obj > timedelta(config.cachito_request_lifetime):
+            mark_as_stale(request['id'])
+
+
+def mark_as_stale(request_id):
+    """
+    Mark the identified stale request ID as `stale` in Cachito.
+
+    :param int request_id: request ID identified as stale
+    :raise CachitoError: if the request to the Cachito API fails
+    """
+    try:
+        log.info('Setting state of request %d to `stale`', request_id)
+        request_rv = auth_session.patch(
+            f'{config.cachito_api_url.rstrip("/")}/requests/{request_id}',
+            json=payload, timeout=config.cachito_api_timeout)
+    except requests.RequestException:
+        msg = f'The connection failed when setting the `stale` state on request {request_id}'
+        log.exception(msg)
+        raise CachitoError(msg)
+
+    if not request_rv.ok:
+        log.error(
+            'Failed to set the `stale` state on request %d. The status was %d. '
+            'The text was:\n%s',
+            request_id, request_rv.status_code, request_rv.text,
+        )
+
+
+if __name__ == '__main__':
+    main()
--- a/cachito/workers/config.py
+++ b/cachito/workers/config.py
@@ -23,6 +23,8 @@ class Config(object):
     cachito_download_timeout = 120
     # The timeout when making a Cachito API request
     cachito_api_timeout = 60
+    # Configurable number of days before which a request becomes stale
+    cachito_request_lifetime = 1
     # The task messages will be acknowledged after the task has been executed,
     # instead of just before
     task_acks_late = True
--- a/tests/test_workers/test_cleanup_job.py
+++ b/tests/test_workers/test_cleanup_job.py
@@ -1,0 +1,71 @@
+from unittest import mock
+from datetime import datetime
+
+from cachito.workers.cleanup_job import main
+
+
+expected = {
+    'items': [
+        {
+            'dependencies': 309,
+            'environment_variables': {},
+            'flags': [],
+            'id': 50,
+            'pkg_managers': [
+                'gomod'
+            ],
+            'ref': 'a7ac8d4c0b7fe90d51fb911511cbf6939655c877',
+            'repo': 'https://github.com/kubernetes/kubernetes.git',
+            'state': 'complete',
+            'state_reason': 'Completed successfully',
+            'updated': '2019-09-05T18:24:50.857861',
+            'user': 'mprahl@redhat.com'
+        }
+    ],
+    'meta': {
+        'first': 'https://cachito.stage.engineering.redhat.com/api/v1/requests'
+                 '?page=1&per_page=20&verbose=False',
+        'last': 'https://cachito.stage.engineering.redhat.com/api/v1/requests'
+                '?page=1&per_page=20&verbose=False',
+        'next': None,
+        'page': 1,
+        'pages': 1,
+        'per_page': 20,
+        'previous': None,
+        'total': 13
+    }
+}
+
+
+@mock.patch('cachito.workers.config.Config.cachito_request_lifetime', 1)
+@mock.patch('cachito.workers.cleanup_job.datetime')
+@mock.patch('cachito.workers.cleanup_job.auth_session.patch')
+@mock.patch('cachito.workers.cleanup_job.session.get')
+def test_cleanup_job_success(mock_requests, mock_auth_requests, mock_dt):
+    mock_dt.utcnow = mock.Mock(return_value=datetime(2019, 9, 7))
+    mock_dt.strptime = mock.Mock(return_value=datetime(2019, 9, 5))
+    mock_requests.return_value.json.return_value = expected
+    mock_auth_requests.return_value.ok = True
+    main()
+    calls = [
+        mock.call(
+            'http://cachito.domain.local/api/v1/requests/50',
+            json={'state': 'stale', 'state_reason': 'The request has expired'},
+            timeout=60),
+    ]
+    assert mock_requests.call_count == 1
+    assert mock_auth_requests.call_count == 1
+    mock_auth_requests.assert_has_calls(calls)
+
+
+@mock.patch('cachito.workers.config.Config.cachito_request_lifetime', 1)
+@mock.patch('cachito.workers.cleanup_job.datetime')
+@mock.patch('cachito.workers.cleanup_job.mark_as_stale')
+@mock.patch('cachito.workers.cleanup_job.session.get')
+def test_cleanup_job_request_not_stale(mock_requests, mock_mark_as_stale, mock_dt):
+    mock_dt.utcnow = mock.Mock(return_value=datetime(2019, 9, 5))
+    mock_dt.strptime = mock.Mock(return_value=datetime(2019, 9, 5))
+    mock_requests.return_value.json.return_value = expected
+    main()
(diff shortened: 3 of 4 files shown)


yashvardhannanavati:
@mprahl can you PTAL?

cachito/workers/config.py
```diff
@@ -50,6 +50,8 @@ class Config(object):
     worker_prefetch_multiplier = 1
     # Configurable batch size of the json payload so that the cachito API request doesn't time out
     cachito_deps_patch_batch_size = 50
+    # Configurable number of days before which a request becomes stale
+    cachito_request_lifetime = 1
```
mprahl: Optional: It'd be nice if these were alphabetical like the rest of the settings. It just makes it easier to find things.

cleanup_job.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env python3
+import logging
+import requests
```
mprahl: We separate imports as follows:
```
# stdlib

# third-party

# local
```

cleanup_job.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env python3
```
mprahl: Can you add this file somewhere under the `cachito/workers` directory? I imagine we want this bundled as part of the Python package.

Also, could you add an entry in `console_scripts` in `setup.py`:
https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point

That way you can just call it with `cachito-cleanup` or something like that.

cleanup_job.py
```diff
@@ -0,0 +1,77 @@
...
+    response = session.get(config.cachito_api_url + 'requests?state=complete')
+    if not response.ok:
+        raise CachitoError('Could not reach cachito API to get all completed requests')
+    json_response = response.json()
+    identify_stale_requests(json_response)
+
+
+def identify_stale_requests(requests_json):
+    """
+    Check if a completed Cachito request has reached end of life.
```
mprahl: This docstring doesn't mention that it actually marks the requests as stale that are passed in

cleanup_job.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    # Get all completed requests
+    response = session.get(config.cachito_api_url + 'requests?state=complete')
+    if not response.ok:
+        raise CachitoError('Could not reach cachito API to get all completed requests')
+    json_response = response.json()
+    identify_stale_requests(json_response)
+
+
+def identify_stale_requests(requests_json):
```
mprahl: This function name is misleading because the function actually modifies the requests, but you wouldn't expect that by the function name.

cleanup_job.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    Check if a completed Cachito request has reached end of life.
+
+    :param dict requests_json: json representation of all complete requests
+    """
+    stale_request_ids = []
+    current_time = datetime.now()
```
mprahl: Could you use UTC time for this?

cleanup_job.py
```diff
@@ -0,0 +1,77 @@
...
+        date_time_obj = datetime.strptime(request['updated'], '%Y-%m-%dT%H:%M:%S.%f')
+        if current_time - date_time_obj > timedelta(config.cachito_request_lifetime):
+            stale_request_ids.append(request['id'])
+
+    mark_as_stale(stale_request_ids)
+
+
+def mark_as_stale(stale_request_ids):
+    """
+    Mark the identified stale request ids as `stale` in cachito.
```
mprahl: `ids` => `IDs`

mprahl: `cachito` => `Cachito`

cleanup_job.py
```diff
@@ -0,0 +1,77 @@
...
+
+    :param list stale_request_ids: list of request ids identified as stale
```
mprahl: `ids` => `IDs`

cleanup_job.py
```diff
@@ -0,0 +1,77 @@
...
+                f'{config.cachito_api_url.rstrip("/")}/requests/{request_id}',
+                json=payload, timeout=config.cachito_api_timeout)
+        except requests.RequestException:
+            msg = f'The connection failed when setting the `stale` state on request {request_id}'
+            log.exception(msg)
+            raise CachitoError(msg)
+
+        if not request_rv.ok:
+            log.error(
+                'The worker failed to set the `stale` state on request %d. The status was %d. '
```
mprahl: This won't be run by the `worker`, so perhaps this log message could remove the `The worker` section.

yashvardhannanavati:
@mprahl can you PTAL? Made the changes.

cachito/workers/cleanup_job.py
```diff
@@ -0,0 +1,78 @@
...
+log = logging.getLogger(__name__)
+
+auth_session = get_requests_session(auth=True)
+session = get_requests_session()
+config = get_worker_config()
+
+
+def main():
+    """
+    Get all completed requests from the Cachito API.
```
mprahl: Could you update the docstring please?

cachito/workers/cleanup_job.py
```diff
@@ -0,0 +1,78 @@
...
+
+    :raise CachitoError: if the request to the Cachito API fails
+    """
+    # Get all completed requests
+    response = session.get(config.cachito_api_url + 'requests?state=complete')
```
mprahl: What if there's more than one page of completed requests that need to be marked as stale? This only grabs a single page.

cachito/workers/cleanup_job.py
```diff
@@ -0,0 +1,78 @@
...
+
+    :param list stale_request_ids: list of request IDs identified as stale
+    :raise CachitoError: if the request to the Cachito API fails
+    """
+    state = 'stale'
+    state_reason = 'The request has expired'
+    payload = {'state': state, 'state_reason': state_reason}
+    for request_id in stale_request_ids:
+        try:
+            log.info(f'Setting state of request {request_id} to `stale`')
```
mprahl: Optional:
```python
log.info('Setting state of request %d to `stale`', request_id)
```

cachito/workers/cleanup_job.py
```diff
@@ -0,0 +1,78 @@
...
+    Identify completed Cachito requests which have reached end of life and mark them as stale.
+
+    :param dict requests_json: json representation of all complete requests
+    """
+    stale_request_ids = []
+    current_time = datetime.utcnow()
+    for request in requests_json['items']:
+        date_time_obj = datetime.strptime(request['updated'], '%Y-%m-%dT%H:%M:%S.%f')
+        if current_time - date_time_obj > timedelta(config.cachito_request_lifetime):
+            stale_request_ids.append(request['id'])
```
mprahl: Optional: It'd be more efficient if you marked the requests as stale as you identify them instead of finding them all and then marking them. Basically O(n) instead of O(2n).

cachito/workers/config.py
```diff
@@ -50,6 +52,8 @@ class Config(object):
...
```
mprahl: This is a duplicate

cachito/workers/cleanup_job.py
yashvardhannanavati: But the complexity in this case remains the same which is O(kn) == O(n). Doing it separately just makes it more readable. :)

mprahl: @yashvardhannanavati how so? You're currently looping through the requests to find the ones to mark and then you loop through those requests to mark them as stale. My suggestion is as you loop through the requests the first time, just mark them as stale as you identify them. That way you loop through the requests once, not twice.

cachito/workers/cleanup_job.py
yashvardhannanavati: Done

cachito/workers/cleanup_job.py
yashvardhannanavati: Nice catch! Done :)

cachito/workers/cleanup_job.py
yashvardhannanavati: Changed it but this is what I meant [Why is O(kn) == O(n)](https://stackoverflow.com/questions/13161397/is-okn-linear-complexity-or-quadratic-complexity-or-does-it-depend-on-the-k)

In the grand scheme of things, unless k is O(n) (nested for loops), a constant value of k doesn't increase the complexity of an algorithm. Ofcourse, O(2n) will run a bit slower than O(n) but that is a very minimal difference. :)

cachito/workers/cleanup_job.py
```diff
@@ -0,0 +1,93 @@
...
+log = logging.getLogger(__name__)
+
+auth_session = get_requests_session(auth=True)
+session = get_requests_session()
+config = get_worker_config()
+
+# state, state_reason and payload values for marking a request in Cachito API
+state = 'stale'
+state_reason = 'The request has expired'
+payload = {'state': state, 'state_reason': state_reason}
```
mprahl: This is only used in the `mark_as_stale` function. I'd prefer if this wasn't global to the `cleanup_job` module.

cachito/workers/cleanup_job.py
```diff
@@ -0,0 +1,93 @@
...
+    and mark them as stale using the API.
+    """
+    # Process the first page of requests
+    json_response = get_completed_requests(config.cachito_api_url + 'requests')
+    identify_and_mark_stale_requests(json_response['items'])
+
+    # Process the remaining pages
+    while(json_response['meta']['next']):
+        json_response = get_completed_requests(json_response['meta']['next'])
+        identify_and_mark_stale_requests(json_response['items'])
```
mprahl: Unfortunately, Python doesn't support a "do while" loop, but the way around this to avoid duplicating code is as follows:

```python
url = config.cachito_api_url + 'requests'
while True:
    json_response = get_completed_requests(url)
    identify_and_mark_stale_requests(json_response['items'])
    if json_response['meta']['next']:
        url = json_response['meta']['next']
    else:
        break
```

cachito/workers/cleanup_job.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def get_completed_requests(url):
+    """
+    Get one page of completed requests from the Cachito API
+
+    :param str url: URL to fetch the cachito requests from
```
mprahl:
```suggestion
    :param str url: the URL to fetch the Cachito requests from
```

cachito/workers/cleanup_job.py
```diff
@@ -0,0 +1,93 @@
...
+    if not response.ok:
+        raise CachitoError('Could not reach Cachito API to get all completed requests')
+    return response.json()
+
+
+def identify_and_mark_stale_requests(requests_json):
+    """
+    Identify completed Cachito requests which have reached end of life and mark them as stale.
+
+    :param dict requests_json: json representation of all complete requests
```
mprahl:
```suggestion
    :param dict requests_json: the JSON representation of a Cachito page
```

cachito/workers/cleanup_job.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    Mark the identified stale request ID as `stale` in Cachito.
+
+    :param int request_id: request ID identified as stale
+    :raise CachitoError: if the request to the Cachito API fails
+    """
+    try:
+        log.info('Setting state of request %d to `stale`', request_id)
+        request_rv = auth_session.patch(
+            f'{config.cachito_api_url.rstrip("/")}/requests/{request_id}',
```
mprahl: You use `rstrip` here but not in `main`. Could you make them consistent?

cachito/workers/cleanup_job.py
yashvardhannanavati: I made it global so that it doesn't get initialized on every call to that function (which will happen for every stale request id).

mprahl: Alright. I don't have a strong preference then.  :)

cachito/workers/cleanup_job.py
```diff
@@ -0,0 +1,93 @@
...
+state_reason = 'The request has expired'
+payload = {'state': state, 'state_reason': state_reason}
+
+
+def main():
+    """
+    Get all completed requests from the Cachito API, identify the stale ones,
+    and mark them as stale using the API.
+    """
+    url = config.cachito_api_url.rstrip("/") + '/requests'
```
mprahl: Could you use single quotes here to be consistent?

yashvardhannanavati:
@mprahl yes, now I'll add the tests for it :)

cachito/workers/cleanup_job.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    Mark the identified stale request ID as `stale` in Cachito.
+
+    :param int request_id: request ID identified as stale
+    :raise CachitoError: if the request to the Cachito API fails
+    """
+    try:
+        log.info('Setting state of request %d to `stale`', request_id)
+        request_rv = auth_session.patch(
+            f'{config.cachito_api_url.rstrip('/')}/requests/{request_id}',
```
mprahl: I believe double quotes will be needed inside the single quotes

tests/test_workers/test_cleanup_job.py
```diff
@@ -0,0 +1,71 @@
+from unittest import mock
+from datetime import datetime
+
+from cachito.workers.cleanup_job import main
+
+
+expected = {
+    "items": [
```
mprahl: Could you use single quotes to be consistent with the rest of the project? It's annoying to go back and forth.

tests/test_workers/test_cleanup_job.py
```diff
@@ -0,0 +1,71 @@
...
+            "ref": "a7ac8d4c0b7fe90d51fb911511cbf6939655c877",
+            "repo": "https://github.com/kubernetes/kubernetes.git",
+            "state": "complete",
+            "state_reason": "Completed successfully",
+            "updated": "2019-09-05T18:24:50.857861",
+            "user": "mprahl@redhat.com"
+        }
+    ],
+    "meta": {
+        "first": ("https://cachito.stage.engineering.redhat.com/api/v1/requests"
```
mprahl: Optional: The parentheses aren't necessary

tests/test_workers/test_cleanup_job.py
yashvardhannanavati: doh! I had it in mind while writing the tests and then it slipped. I apologize. Changing it now...

yashvardhannanavati:
Rebased the PR :)

PR #60, opened 2019-10-09
Add support for replacing dependencies

mprahl:
To accomplish this, fetch_app_source now extracts the application
source code archive to the `<bundle_dir>/temp/<request_id>/app`
directory. This can then be used by the package manager tasks to
modify the source for dependency overrides. For Golang, this ends
up being the go.mod and go.sum files.

The create_bundle_archive task then creates an archive from the
contents from `<bundle_dir>/temp/<request_id>/app` and
`<bundle_dir>/temp/<request_id>/deps`.

The dependencies in a request's JSON now has a new key called
"replaces". If this is set, a JSON object representing the
dependency that was replaced is shown.

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -126,12 +126,11 @@ def create_request():
     ]
     if 'gomod' in pkg_manager_names or auto_detect:
         chain_tasks.append(
-            tasks.fetch_gomod_source.s(request_id_to_update=request.id, auto_detect=auto_detect)
-                 .on_error(error_callback)
+            tasks.fetch_gomod_source.s(request.id, auto_detect=auto_detect).on_error(error_callback)
         )
 
     chain_tasks.extend([
-        tasks.create_bundle_archive.s(request_id=request.id).on_error(error_callback),
+        tasks.create_bundle_archive.si(request.id).on_error(error_callback),
         tasks.set_request_state.si(request.id, 'complete', 'Completed successfully'),
     ])
 
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -1,4 +1,5 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
+import copy
 import os
 import re
 import shutil
@@ -125,8 +126,15 @@ def create_request():
         tasks.fetch_app_source.s(request.repo, request.ref, request.id).on_error(error_callback),
     ]
     if 'gomod' in pkg_manager_names or auto_detect:
+        gomod_dependency_replacements = [
+            dependency_replacement
+            for dependency_replacement in payload.get('dependency_replacements', [])
+            if dependency_replacement['type'] == 'gomod'
+        ]
         chain_tasks.append(
-            tasks.fetch_gomod_source.s(request.id, auto_detect=auto_detect).on_error(error_callback)
+            tasks.fetch_gomod_source.si(
+                request.id, auto_detect, gomod_dependency_replacements,
+            ).on_error(error_callback)
         )
 
     chain_tasks.extend([
@@ -177,7 +185,7 @@ def patch_request(request_id):
 
         if key == 'dependencies':
             for dep in value:
-                Dependency.validate_json(dep)
+                Dependency.validate_json(dep, for_update=True)
         elif key == 'pkg_managers':
             for pkg_manager in value:
                 if not isinstance(pkg_manager, str):
@@ -213,14 +221,15 @@ def patch_request(request_id):
             request.add_state(new_state, new_state_reason)
 
     if 'dependencies' in payload:
-        for dep in payload['dependencies']:
-            dep_obj = Dependency.query.filter_by(**dep).first()
-            if not dep_obj:
-                dep_obj = Dependency.from_json(dep)
-                db.session.add(dep_obj)
+        for dep_and_replaces in payload['dependencies']:
+            dep = copy.deepcopy(dep_and_replaces)
+            replaces = dep.pop('replaces', None)
 
-            if dep_obj not in request.dependencies:
-                request.dependencies.append(dep_obj)
+            dep_object = Dependency.get_or_create(dep)
+            replaces_object = None
+            if replaces:
+                replaces_object = Dependency.get_or_create(replaces)
+            request.add_dependency(dep_object, replaces_object)
 
     if 'pkg_managers' in payload:
         pkg_managers = PackageManager.get_pkg_managers(payload['pkg_managers'])
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -122,8 +122,7 @@ def create_request():
     # Chain tasks
     error_callback = tasks.failed_request_callback.s(request.id)
     chain_tasks = [
-        tasks.fetch_app_source.s(
-            request.repo, request.ref, request_id_to_update=request.id).on_error(error_callback),
+        tasks.fetch_app_source.s(request.repo, request.ref, request.id).on_error(error_callback),
     ]
     if 'gomod' in pkg_manager_names or auto_detect:
         chain_tasks.append(
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -19,13 +19,6 @@ request_pkg_manager_table = db.Table(
     db.UniqueConstraint('request_id', 'pkg_manager_id'),
 )
 
-request_dependency_table = db.Table(
-    'request_dependency',
-    db.Column('request_id', db.Integer, db.ForeignKey('request.id'), nullable=False),
-    db.Column('dependency_id', db.Integer, db.ForeignKey('dependency.id'), nullable=False),
-    db.UniqueConstraint('request_id', 'dependency_id'),
-)
-
 request_environment_variable_table = db.Table(
     'request_environment_variable',
     db.Column('request_id', db.Integer, db.ForeignKey('request.id'), nullable=False),
@@ -106,6 +99,28 @@ class Dependency(db.Model):
         }
 
 
+class RequestDependency(db.Model):
+    """An association table between requests and dependencies."""
+    # A primary key is required by SQLAlchemy when using declaritive style tables, so a composite
+    # primary key is used on the two required columns
+    request_id = db.Column(
+        db.Integer,
+        db.ForeignKey('request.id'),
+        autoincrement=False,
+        primary_key=True,
+    )
+    dependency_id = db.Column(
+        db.Integer,
+        db.ForeignKey('dependency.id'),
+        autoincrement=False,
+        primary_key=True,
+    )
+
+    __table_args__ = (
+        db.UniqueConstraint('request_id', 'dependency_id'),
+    )
+
+
 class Request(db.Model):
     """A Cachito user request."""
     id = db.Column(db.Integer, primary_key=True)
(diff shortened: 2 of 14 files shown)


cachito/web/models.py
```diff
@@ -162,6 +175,18 @@ def number_of_deps(self):
         return db.session.query(sqlalchemy.func.count(req_dep_cols.dependency_id)).filter(
             req_dep_cols.request_id == self.id).scalar()
 
+    @property
+    def number_of_dep_overrides(self):
```
yashvardhannanavati: How about renaming this to dependency_overrides_count

cachito/web/models.py
```diff
@@ -242,6 +271,22 @@ def from_json(cls, kwargs):
 
             request_kwargs['flags'] = found_flags
 
+        dependency_overrides = request_kwargs.get('dependency_overrides', [])
+        if not isinstance(dependency_overrides, list):
+            raise ValidationError('"dependency_overrides" must be an array')
+
+        dependency_override_objects = []
+        for dep in dependency_overrides:
```
yashvardhannanavati: I prefer having the entire name `dependency` instead of `dep`.

I blame luiz's book for that. :p

yashvardhannanavati: But I'll leave it up to you :)

cachito/workers/pkg_manager.py
```diff
@@ -70,6 +79,11 @@ def resolve_gomod_deps(archive_path, request_id):
             if len(parts) == 1:
                 # This is the application itself, not a dependency
                 continue
+            elif len(parts) == 3:
```
yashvardhannanavati: What are your thoughts on converting this piece into a switch statement instead? I think it would look cleaner.

cachito/workers/pkg_manager.py
```diff
@@ -148,31 +162,17 @@ def add_deps_to_bundle(src_deps_path, dest_cache_path, request_id):
         content of src_deps_path to
     :param int request_id: the request the bundle is for
     """
-    config = get_worker_config()
-    deps_path = os.path.join(config.cachito_bundles_dir, 'temp', str(request_id), 'deps')
+    bundle_dir = get_request_bundle_dir(request_id)
```
yashvardhannanavati: Optional: You can skip creating this variable here and directly call the method in the next statement.

cachito/web/models.py
mprahl: @yashvardhannanavati I'm just using that name to be consistent with `number_of_deps`. What's funny is you named it that way in #52.  :laughing:

If I rename this one, I'll also rename the other. Which do you prefer?

cachito/web/models.py
mprahl: Sure, I'll use `dependency_override` to be more descriptive.

cachito/workers/pkg_manager.py
mprahl: @yashvardhannanavati Python doesn't support switch statements.

cachito/workers/pkg_manager.py
mprahl: I did it the other way for readability, but I'll change it.

cachito/workers/pkg_manager.py
yashvardhannanavati: Ah, that's right. My bad. Java is playing with my mind again.

cachito/web/models.py
yashvardhannanavati: @mprahl lol, Sorry I didn't realize it while writing #52. Can I please request you to rename both if possible?

mprahl:
@yashvardhannanavati I addressed your comments. Could you take another look?

cachito/workers/pkg_manager.py
```diff
@@ -61,19 +61,27 @@ def resolve_gomod_deps(archive_path, request_id=None):
...
+            # If there is no "replace" directive used on the dependency, then the last column will
+            # be "<nil>"
+            parts = [part for part in line.split(' ') if part not in ('', '<nil>')]
             if len(parts) == 1:
                 # This is the application itself, not a dependency
                 continue
-            if len(parts) > 2:
-                log.warning('Unexpected go module output: %s', line)
-                continue
+            elif len(parts) == 4:
```
lcarva: Consider using `if` instead of `elif` here. This seems to create some sort of relationship with the previous `if` statement that I don't think is accurate.

tests/test_workers/test_pkg_manager.py
```diff
@@ -22,7 +22,7 @@
     github.com/kr/pty v1.1.1
     github.com/kr/text v0.1.0
     github.com/op/go-logging v0.0.0-20160315200505-970db520ece7
-    github.com/pkg/errors v0.8.1
+    github.com/pkg/errors v0.8.0 github.com/pkg/errors v0.8.1
```
lcarva: It would be nice to have a test case that verified v0.8.1 is used. Also, to ensure that the second module name is used, maybe change the line to something like:
```
github.com/pkg/ignored v0.8.0 github.com/pkg/errors v0.8.1
```

I don't think we need to do it for every dep here, but a few keys one would be useful.

tests/test_workers/test_tasks.py
```diff
@@ -34,34 +34,29 @@ def test_fetch_app_source_request_timed_out(mock_git, mock_set_request_state):
...
 
 
-@pytest.mark.parametrize('request_id_to_update, auto_detect, contains_go_mod', (
-    (None, False, False),
-    (1, False, False),
-    (None, True, True),
-    (None, True, False),
+@pytest.mark.parametrize('auto_detect, contains_go_mod', (
+    (False, False),
+    (False, False),
```
lcarva: Duplicated test case.

tests/test_workers/test_tasks.py
```diff
@@ -34,34 +34,29 @@ def test_fetch_app_source_request_timed_out(mock_git, mock_set_request_state):
...
 ):
     mock_archive_contains_path.return_value = contains_go_mod
     app_archive_path = 'path/to/archive.tar.gz'
     mock_resolve_gomod_deps.return_value = sample_deps
-    tasks.fetch_gomod_source(
-        app_archive_path, request_id_to_update=request_id_to_update, auto_detect=auto_detect,
-    )
-    if request_id_to_update:
+    tasks.fetch_gomod_source(app_archive_path, 1, auto_detect=auto_detect)
+    if not auto_detect or contains_go_mod:
```
lcarva: Consider adding a new test param `expect_state_update` instead.

cachito/web/api_v1.py
```diff
@@ -125,12 +125,21 @@ def create_request():
...
     ]
     if 'gomod' in pkg_manager_names or auto_detect:
+        gomod_dep_overrides = [
+            dep_override.to_json()
+            for dep_override in request.dependency_overrides
+            if dep_override.type == 'gomod'
+        ]
         chain_tasks.append(
-            tasks.fetch_gomod_source.s(request.id, auto_detect=auto_detect).on_error(error_callback)
+            tasks.fetch_gomod_source.si(
```
lcarva: +1 to using immutable signatures. I think this makes the code easier to understand.

cachito/web/models.py
```diff
@@ -162,6 +175,18 @@ def number_of_deps(self):
...
+    def dependency_overrides_count(self):
+        """
+        Get the total number of dependency overrides for a request.
+
+        :return: the number of dependency overrides
+        :rtype: int
+        """
+        req_dep_cols = request_dependency_override_table.c
+        return db.session.query(sqlalchemy.func.count(req_dep_cols.dependency_id)).filter(
+            req_dep_cols.request_id == self.id).scalar()
```
lcarva: Didn't know about [scalar](https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.scalar). We should use it more often.

cachito/web/models.py
```diff
@@ -197,15 +222,19 @@ def to_json(self, verbose=True):
         if verbose:
             rv.update({'state_history': states})
             rv.update({'dependencies': [dep.to_json() for dep in self.dependencies]})
+            rv.update(
+                {'dependency_overrides': [dep.to_json() for dep in self.dependency_overrides]}
```
lcarva: Hmm... does this mean the `dependencies` attribute is not an accurate representation of the dependencies being used?

lcarva: It would be more useful if the existing items in the `dependencies` list had additional information to determine whether or not they are overrides. That way clients only have to consume a single list, but still have the overrides information if needed.

cachito/workers/pkg_manager.py
```diff
@@ -34,20 +34,23 @@ def __exit__(self, exc, value, tb):
...
 
-def resolve_gomod_deps(archive_path, request_id):
+def resolve_gomod_deps(app_source_path, request_id, dep_overrides=None):
     """
     Resolve and fetch gomod dependencies for given app source archive.
 
-    :param str archive_path: the full path to the application source code
+    :param str app_source_path: the full path to the application source code
     :param int request_id: the request ID of the bundle to add the gomod deps to
+    :param list dep_overrides: dependency overrides with the keys "name" and "version; this
```
lcarva: Missing double quote after `version`.

cachito/workers/pkg_manager.py
```diff
@@ -56,7 +59,13 @@ def resolve_gomod_deps(archive_path, request_id):
...
         }
 
-        run_params = {'env': env, 'cwd': source_dir}
+        run_params = {'env': env, 'cwd': app_source_path}
+
+        for dep_override in dep_overrides:
+            name = dep_override['name']
+            version = dep_override['version']
+            log.info('Applying the gomod override %s@%s', name, version)
+            _run_cmd(('go', 'mod', 'edit', '-replace', f'{name}={name}@{version}'), run_params)
```
lcarva: I think we want to have the ability to override across different names. This would be needed for handling replacing a dependency with a fork done in another host, for example. `github.com/foo/bar` gets replaced with `git.internal.com/foo/bar`.

tests/test_workers/test_pkg_manager.py
mprahl: This actually does test that `v0.8.1` is used. I can add a test to for the different name though.

cachito/web/models.py
mprahl: The `dependencies` attribute is accurate. It's the dependencies that golang will use during a build, which includes the replace directives in the `go.mod` file.

I'll reach out on IRC on how we want to approach the comment on the JSON representation.

mprahl:
@lcarva and @yashvardhannanavati this is ready for review again.

cachito/web/models.py
```diff
@@ -143,6 +236,36 @@ class Request(db.Model):
...
+        :param Dependency replaced_dependency: an optional Dependency object to mark as being
+            replaced by the input dependency for this request
+        """
+        # If the ID is not set, then the dependency was just created and is not part of the
+        # database's transaction buffer.
+        if not dependency.id or (replaced_dependency and not replaced_dependency.id):
+            # Send the changes queued up in SQLAlchemy to the database's transaction buffer
+            db.session.flush()
+
+        mapping = RequestDependency.query.filter_by(
```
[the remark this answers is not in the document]
mprahl: An exception is raised if this occurs now. I also added a test for this.

mprahl:
@lcarva I addressed your comments. Here is the diff for your convenience:
https://github.com/release-engineering/cachito/compare/2afeb0f4829b179129337ac21dab41af90b07d09..a8822bcd6512f194112fc800ed4aaa668d8a39a0

PR #61, opened 2019-10-09
Add indexes to commonly used columns

mprahl:
The database migration relies on #60.

The approach taken here is to add indexes to all the foreign keys on the join tables, and any commonly used columns that are currently filtered on that will grow significantly in production.


mprahl:
@lcarva and @yashvardhannanavati could you please review?

PR #62, opened 2019-10-16
Set the logging level in the cleanup job

yashvardhannanavati:


cachito/workers/cleanup_job.py
```diff
@@ -91,3 +91,4 @@ def mark_as_stale(request_id):
 
 if __name__ == '__main__':
     main()
+    log.setLevel(logging.INFO)
```
mprahl:
```suggestion
    log.setLevel(logging.INFO)

```

mprahl: This needs to be called before `main()`. Also, could you add a new line at the end of the file please?

PR #63, opened 2019-10-16
Make the Flask logger configurable

mprahl:


cachito/web/config.py
```diff
@@ -7,6 +7,11 @@
 
 class Config(object):
     """The base Cachito Flask configuration."""
+    # Additional loggers to set to the level defined in CACHITO_LOG_LEVEL
+    CACHITO_ADDITIONAL_LOGGERS = []
+    # This sets the level of the "flask.app" logger, which is accessed from current_app.logger
+    CACHITO_LOG_LEVEL = 'INFO'
```
lcarva: Can we set this to `DEBUG` for `DevelopmentConfig` and `TestingConfig` ?

cachito/web/app.py
```diff
@@ -68,6 +70,17 @@ def create_app(config_obj=None):
     else:
         load_config(app)
 
+    # Configure logging
+    default_handler.setFormatter(
+        logging.Formatter(fmt=app.config['CACHITO_LOG_FORMAT'], datefmt='%Y-%m-%d %H:%M:%S')
+    )
+    app.logger.setLevel(app.config['CACHITO_LOG_LEVEL'])
+    for logger_name in app.config['CACHITO_ADDITIONAL_LOGGERS']:
+        logger = logging.getLogger(logger_name)
```
lcarva: Would this be for setting the logging level of third party libraries?
I think that's pretty neat.

mprahl: Yes, exactly!

cachito/web/config.py
mprahl: Good point! I just added that.

mprahl:

PR #64, opened 2019-10-21
Allow privileged users to submit a request on behalf of another user

mprahl:
Addresses FACTORY-4595

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -7,6 +7,7 @@ import os
 import flask
 from flask_login import UserMixin, current_user
 import sqlalchemy
+from werkzeug.exceptions import Forbidden
 
 from cachito.errors import ValidationError
 from cachito.web import db
@@ -223,6 +224,7 @@ class Request(db.Model):
     repo = db.Column(db.String, nullable=False)
     ref = db.Column(db.String, nullable=False)
     user_id = db.Column(db.Integer, db.ForeignKey('user.id'))
+    submitted_by_id = db.Column(db.Integer, db.ForeignKey('user.id'))
     dependencies = db.relationship(
         'Dependency',
         backref='requests',
@@ -248,7 +250,8 @@ class Request(db.Model):
     environment_variables = db.relationship(
         'EnvironmentVariable', secondary=request_environment_variable_table, backref='requests',
         order_by='EnvironmentVariable.name')
-    user = db.relationship('User', back_populates='requests')
+    submitted_by = db.relationship('User', foreign_keys=[submitted_by_id])
+    user = db.relationship('User', foreign_keys=[user_id], back_populates='requests')
     flags = db.relationship(
         'Flag', secondary=request_flag_table, backref='requests', order_by='Flag.name')
 
@@ -370,8 +373,13 @@ class Request(db.Model):
             'environment_variables': env_vars_json,
             'flags': [flag.to_json() for flag in self.flags],
         }
+        if self.submitted_by:
+            rv['submitted_by'] = self.submitted_by.username
+        else:
+            rv['submitted_by'] = None
         # Show the latest state information in the first level of the JSON
         rv.update(latest_state)
+
         if verbose:
             rv['state_history'] = states
             replacement_id_to_replacement = {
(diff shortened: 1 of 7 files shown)


cachito/web/models.py
```diff
@@ -440,7 +447,29 @@ def from_json(cls, kwargs):
 
         # current_user.is_authenticated is only ever False when auth is disabled
         if current_user.is_authenticated:
-            request_kwargs['user_id'] = current_user.id
+            user_override_username = request_kwargs.pop('user', None)
```
yashvardhannanavati: What are your thoughts on using `proxy_username` instead of `user_override_username`. I find the latter a bit difficult to comprehend.

cachito/web/models.py
```diff
@@ -440,7 +447,29 @@ def from_json(cls, kwargs):
...
+                # lower-case for consistency
+                allowed_users = {
+                    user.lower() for user in flask.current_app.config['CACHITO_PROXY_USERNAMES']
+                }
+                if current_user.username not in allowed_users:
+                    raise Forbidden(
+                        'You are not authorized to create a request on behalf of another user'
+                    )
+
+                user_override = User.query.filter_by(username=user_override_username).first()
```
yashvardhannanavati: similarly..how about `proxied_user` instead  of `user_override`?

cachito/web/models.py
mprahl: Good point. I used `submitted_for_username`. How does that sound?

cachito/web/models.py
```diff
@@ -382,19 +389,19 @@ def to_json(self, verbose=True):
...
-            rv.update({'dependencies': self.dependencies_count})
+            rv['dependencies'] = self.dependencies_count
         return rv
 
     @classmethod
     def from_json(cls, kwargs):
         # Validate all required parameters are present
         required_params = {'repo', 'ref'}
-        optional_params = {'dependency_replacements', 'flags', 'pkg_managers'}
+        optional_params = {'dependency_replacements', 'flags', 'pkg_managers', 'user'}
```
yashvardhannanavati: Keeping the param name as `user` might be confusing. Should we make it more specific like `proxy_username`?

tests/test_api_v1.py
```diff
@@ -238,13 +248,27 @@ def test_create_request_invalid_parameter(auth_env, client, db):
...
+
+
+def test_create_request_cannot_set_user(client, db):
+    data = {
+        'repo': 'https://github.com/release-engineering/retrodep.git',
+        'ref': 'c50b93a32df1c9d700e3e80996845bc2e13be848',
+        'user': 'tom_hanks@domain.local',
+    }
+
+    auth_env = {'REMOTE_USER': 'homer_simpson@domain.local'}
```
yashvardhannanavati: lol, nice use of names!

cachito/web/models.py
mprahl: I chose `user` because that is the key used on the JSON output. I thought it'd be good to be consistent in that regard.

I'm not opposed to changing it but would like @lcarva's opinion on this.

cachito/web/models.py
```diff
@@ -373,6 +376,10 @@ def to_json(self, verbose=True):
         # Show the latest state information in the first level of the JSON
         rv.update(latest_state)
         if verbose:
+            if self.submitted_by:
```
lcarva: Can we always display this, instead of just when verbose mode?
Not doing so may cause confusion.

cachito/web/models.py
```diff
@@ -440,7 +447,30 @@ def from_json(cls, kwargs):
...
+                allowed_users = {
+                    user.lower()
+                    for user in flask.current_app.config['CACHITO_USER_REPRESENTATIVES']
+                }
+                if current_user.username not in allowed_users:
+                    raise Forbidden(
+                        'You are not authorized to create a request on behalf of another user'
+                    )
+
+                submitted_for = User.query.filter_by(username=submitted_for_username).first()
```
lcarva: It would be nice if the get or create logic was wrapped in a helper method.

cachito/web/models.py
```diff
@@ -373,7 +373,7 @@ def to_json(self, verbose=True):
         # Show the latest state information in the first level of the JSON
         rv.update(latest_state)
         if verbose:
-            rv.update({'state_history': states})
```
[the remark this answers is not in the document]
mprahl: It was just a style issue I found while looking at the code. It was in a separate commit since it wasn't related to the PR.

cachito/web/models.py
mprahl: Sure. I see it as more of an implementation detail which is why I thought it'd be hidden to just verbose output, but I'll change that.

PR #67, opened 2019-10-23
Remove cachito-cli.py since it's not used and outdated

mprahl:


lcarva:
Yes! I've been meaning to do this for some time. Thanks!

PR #68, opened 2019-11-04
Add filtering based on request state

sarah256:


sarah256:
This fails because it says the name 'json' is not defined, yet it is used elsewhere in this file and is imported at the top. I'm completely stumped, any ideas? (The tests also pass locally)

hluk:
The failed test ran for a commit that is missing the `import json` (https://github.com/release-engineering/cachito/blob/8b9a4d844564949b191bf217f7b2bf7b4c529afb/tests/test_api_v1.py).

I believe this happens because it makes a **merge automatically** before running tests.

cachito/web/api_v1.py
```diff
@@ -28,15 +30,23 @@ def get_requests():
         value exceeds configuration's CACHITO_MAX_PER_PAGE.
     :rtype: flask.Response
     """
+    # Check if user is filtering requests by state
+    state = flask.request.args.get('state', default='')
```
mprahl: The more "Pythonic" approach would be to let `state` be `None` if it wasn't in the request.

cachito/web/api_v1.py
```diff
@@ -28,15 +30,23 @@ def get_requests():
...
     # Default versbose flag to False
     verbose = str_to_bool(flask.request.args.get('verbose', False))
     max_per_page = flask.current_app.config['CACHITO_MAX_PER_PAGE']
     # The call to `paginate` will inspect the current HTTP request for the
     # pagination parameters `page` and `per_page`.
     pagination_query = Request.query.paginate(max_per_page=max_per_page)
     requests = pagination_query.items
+    if state != '':
+        state_int = RequestStateMapping.__members__[state].value
+        items = [request.to_json(verbose=verbose) for request in requests
```
mprahl: You are filtering at the Python level and after pagination occurs. Instead, you need to add a SQLAlchemy filter to `Request.query` before `.paginate(max_per_page=max_per_page)` is called on it. This will filter at the SQL level which is much more efficient and will cause the pagination to be done on the filtered results.

mprahl:
> This fails because it says the name 'json' is not defined, yet it is used elsewhere in this file and is imported at the top. I'm completely stumped, any ideas? (The tests also pass locally)

The commit https://github.com/release-engineering/cachito/commit/78e50e99091696e147bbc39bd7eeb19dc7da6c7b is what's causing your tests to fail.

We moved away from `json.loads` and just use the `json` property on return values from the Flask client.

cachito/web/api_v1.py
```diff
@@ -29,12 +33,26 @@ def get_requests():
...
+    if state:
+        state_int = RequestStateMapping.__members__[state].value
+        query = Request.query
+        state_subquery = (
+            db.session.query(RequestState.request_id,
+                             func.max(RequestState.updated).label('updated'))
+            .group_by(RequestState.request_id)
+            .filter_by(state=state_int)
+        ).subquery()
+        filtered_query = query.join(state_subquery, Request.id == state_subquery.c.request_id)
```
sarah256: @mprahl Do you think it is necessary to explain any parts of the queries with comments?

cachito/web/api_v1.py
```diff
@@ -29,12 +33,26 @@ def get_requests():
         value exceeds configuration's CACHITO_MAX_PER_PAGE.
     :rtype: flask.Response
     """
-    # Default versbose flag to False
+    # Check if user is filtering requests by state
+    state = flask.request.args.get('state', default=None)
```
mprahl: You should be able to remove `default=None`.

cachito/web/api_v1.py
```diff
@@ -29,12 +33,26 @@ def get_requests():
...
+    # Default verbose flag to False
     verbose = str_to_bool(flask.request.args.get('verbose', False))
     max_per_page = flask.current_app.config['CACHITO_MAX_PER_PAGE']
     # The call to `paginate` will inspect the current HTTP request for the
     # pagination parameters `page` and `per_page`.
-    pagination_query = Request.query.paginate(max_per_page=max_per_page)
+    if state:
+        state_int = RequestStateMapping.__members__[state].value
```
mprahl: Could you add code that validates the `state` that was provided? This currently would produce a `KeyError` and provide a 500 error with no explanation to the user if the user put in an invalid state.

One of the special exceptions we use is `ValidationError`, which will return a 400 error to the user in JSON format.

cachito/web/api_v1.py
```diff
@@ -29,12 +33,26 @@ def get_requests():
...
+        query = Request.query
+        state_subquery = (
+            db.session.query(RequestState.request_id,
```
mprahl: To make the styling consistent, could you use the following?
```python
db.session.query(
    RequestState.request_id, func.max(RequestState.updated).label('updated'),
)
```

cachito/web/api_v1.py
```diff
@@ -29,12 +33,26 @@ def get_requests():
...
+                             func.max(RequestState.updated).label('updated'))
+            .group_by(RequestState.request_id)
+            .filter_by(state=state_int)
+        ).subquery()
+        filtered_query = query.join(state_subquery, Request.id == state_subquery.c.request_id)
+        pagination_query = filtered_query.paginate(max_per_page=max_per_page)
```
mprahl: If you assign `query = Request.query` outside of the if block and instead of using `filtered_query`, you just overwrite the value of `query`, then you can just do the pagination once outside of the if block as `pagination_query = query.paginate(max_per_page=max_per_page)`.

This develops a pattern where the pagination is performed the same way regardless of any filters added by the user.

tests/test_api_v1.py
```diff
@@ -189,6 +189,40 @@ def test_fetch_paginated_requests(
...
+                'repo': repo_template.format(i),
+                'ref': 'e1be527f39ec31323f0454f7d1422c6260b00580',
+                'pkg_managers': ['gomod'],
+            }
+            request_complete = Request.from_json(data_complete)
+            request_complete.add_state('complete', 'Completed successfully')
+            db.session.add(request_complete)
+    db.session.commit()
+
+    rv = client.get('/api/v1/requests?state=complete')
```
mprahl: All your requests are in the `complete` state, so this test doesn't actually verify if the filter was applied. Could you please change your test environment to have at least one request in a different state?

tests/test_api_v1.py
```diff
@@ -189,6 +189,40 @@ def test_fetch_paginated_requests(
     assert type(fetched_requests[0]['dependencies']) == list
 
 
+def test_create_request_filter_state(app, auth_env, client, db):
+    """Test that requests can be filtered by state."""
+    repo_template = 'https://github.com/release-engineering/retrodep{}.git'
+    # flask_login.current_user is used in Request.from_json, which requires a request context
+    with app.test_request_context(environ_base=auth_env):
+        for i in range(10):
```
mprahl: Only two entries are necessary to test filtering. Each entry will just have a different state. By just having two, it makes the test a little faster since it reduces the amount of disk IO necessary.

cachito/web/api_v1.py
mprahl: I think a comment explaining `state_subquery` would be nice, but I think it's pretty clear as is.

cachito/web/api_v1.py
mprahl: Could you please document this query parameter in the OpenAPI document at `cachito/web/static/api_v1.yaml`? It's how we generate our API documentation at https://release-engineering.github.io/cachito.

Here is the documentation: https://swagger.io/docs/specification/describing-parameters/#query-parameters

cachito/web/api_v1.py
sarah256: @mprahl Should we just go to the else clause if the state is invalid instead of throwing an error?

mprahl: @sarah256 I would prefer an error message over silently failing. It makes it easier to debug issues from the consumer perspective.

tests/test_api_v1.py
sarah256: The requests I tested are half in progress and half complete. They start out as in progress and then for the ones using the `request_complete` variable, they get a `complete` state added to them on line 213. (Unless I am using them wrong, which may be so!) What I intended to do was make 2 requests for every loop, for a total of 20, and then half of them would be those 10 fetched requests I assert later on.

mprahl: I'm sorry. I misread your code. This test is accurate, but I'd still like to see the other comment about making it more efficient addressed.

cachito/web/api_v1.py
```diff
@@ -29,12 +33,27 @@ def get_requests():
...
+    state = flask.request.args.get('state')
+    # Default verbose flag to False
     verbose = str_to_bool(flask.request.args.get('verbose', False))
     max_per_page = flask.current_app.config['CACHITO_MAX_PER_PAGE']
     # The call to `paginate` will inspect the current HTTP request for the
     # pagination parameters `page` and `per_page`.
-    pagination_query = Request.query.paginate(max_per_page=max_per_page)
+    query = Request.query
+    if state:
+        if state not in ['in_progress', 'complete', 'failed', 'stale']:
```
mprahl: If we ever add a state, we'll need to update this list. My suggestion is to use `cachito.web.models.RequestStateMapping.get_state_names()`.

An alternative is to do:
```python
try:
    state_int = RequestStateMapping.__members__[state].value
except KeyError:
    raise ValidationError('error goes here')
```

cachito/web/api_v1.py
```diff
@@ -29,12 +33,27 @@ def get_requests():
...
+            raise ValidationError('{} is not a valid request state.'.format(state))
```
mprahl: Optional: We try to use f-strings in the code base. In this case, you would do:
```python
raise ValidationError(f'{state} is not a valid request state.')
```

mprahl: Could you add the states which are valid in the error message? This just improves the UX.

cachito/web/static/api_v1.yaml
```diff
@@ -34,6 +34,13 @@ paths:
             type: boolean
             example: true
             default: false
+        - name: state
```
mprahl: Could you put this above `verbose`? It's nice to have the documentation in alphabetical order for readability.

cachito/web/static/api_v1.yaml
```diff
@@ -34,6 +34,13 @@ paths:
...
+          in: query
+          description: The state to filter requests by
+          schema:
+            type: string
+            example: in_progress
+            default: None
```
mprahl: This should be `null` instead of `None` since it's YAML.

cachito/web/api_v1.py
```diff
@@ -29,12 +33,28 @@ def get_requests():
         value exceeds configuration's CACHITO_MAX_PER_PAGE.
     :rtype: flask.Response
     """
-    # Default versbose flag to False
+    # Check if user is filtering requests by state
```
mprahl: Optional: `if user` => `if the user`

cachito/web/api_v1.py
```diff
@@ -29,12 +33,28 @@ def get_requests():
...
+    # Default verbose flag to False
     verbose = str_to_bool(flask.request.args.get('verbose', False))
     max_per_page = flask.current_app.config['CACHITO_MAX_PER_PAGE']
     # The call to `paginate` will inspect the current HTTP request for the
     # pagination parameters `page` and `per_page`.
-    pagination_query = Request.query.paginate(max_per_page=max_per_page)
+    query = Request.query
+    if state:
+        if state not in RequestStateMapping.get_state_names():
+            raise ValidationError(f'{state} is not a valid request state. Valid states are: '
```
mprahl: Could you please add a unit test for this failure case?

mprahl: To match the code style in the project, could you do:
```python
raise ValidationError(
    f'{state} is not a valid request state. Valid states are: '
    'in_progress, complete, failed, or stale.'
)
```

cachito/web/api_v1.py
```diff
@@ -29,12 +33,28 @@ def get_requests():
...
+                                  'in_progress, complete, failed, or stale.')
```
mprahl: `'in_progress, complete, failed, or stale.'` => `'{":".join(RequestStateMapping.get_state_names())}'`

cachito/web/api_v1.py
```diff
@@ -29,12 +33,30 @@ def get_requests():
...
     max_per_page = flask.current_app.config['CACHITO_MAX_PER_PAGE']
     # The call to `paginate` will inspect the current HTTP request for the
     # pagination parameters `page` and `per_page`.
-    pagination_query = Request.query.paginate(max_per_page=max_per_page)
+    query = Request.query
+    if state:
+        if state not in RequestStateMapping.get_state_names():
+            raise ValidationError(
+                f'{state} is not a valid request state. Valid states are: '
+                f'{":".join(RequestStateMapping.get_state_names())}'
```
mprahl:
```suggestion
                ':'.join(RequestStateMapping.get_state_names())
```

mprahl:
@sarah256 one small nitpick to address and then this is good to merge!

cachito/web/api_v1.py
sarah256: I changed the quotes, but also put it into a separate variable because I think it makes it more readable.

PR #69, opened 2019-11-08
Split pkg_manager.py into its own subpackage

mprahl:
This will make it cleaner when future code is added to the Golang package manager.


mprahl:
I split `test_pkg_manager.py` to match the new layout in the latest force-push.

PR #70, opened 2019-11-11
Validate Cachito Flask configuration

yashvardhannanavati:
Refers to FACTORY-5085

### DIFF
--- a/tests/test_cachito_config.py
+++ b/tests/test_cachito_config.py
@@ -1,0 +1,25 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+from unittest.mock import patch
+
+import pytest
+
+from cachito.web.config import validate_cachito_config
+from cachito.errors import ConfigError
+
+
+@patch('os.path.isdir', return_value=True)
+def test_validate_cachito_config_success(mock_isdir, app):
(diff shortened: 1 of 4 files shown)


tests/test_cachito_config.py
```diff
@@ -0,0 +1,27 @@
...
+    mock_isdir.assert_any_call('/tmp/cachito-archives/bundles')
+
+
+@patch('os.path.isdir', return_value=False)
+@pytest.mark.parametrize('variable_name', (
+    'CACHITO_BUNDLES_DIR', 'CACHITO_LOG_LEVEL', 'CACHITO_MAX_PER_PAGE', 'CACHITO_LOG_FORMAT'))
+def test_validate_cachito_config_failure(mock_isdir, app, variable_name):
+    expected = f'The configuration "{variable_name}" must be set'
+    temp = app.config[variable_name]
+    app.config[variable_name] = None
```
mprahl: Could you use https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch.dict? This is a safer way of doing this.

yashvardhannanavati: really cool!

PR #71, opened 2019-11-11
Rename the dependency table to package

mprahl:
This will allow a future change to reuse this table to map the top level packages associated with the request.

Relates to FACTORY-5064

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -67,8 +67,11 @@ class RequestStateMapping(Enum):
         return sorted([state.name for state in cls])
 
 
-class Dependency(db.Model):
-    """A dependency (e.g. gomod dependency) associated with the request."""
+class Package(db.Model):
+    """A package associated with the request."""
+    # Statically set the table name so that the inherited classes uses this value instead of one
+    # derived from the class name
+    __tablename__ = 'package'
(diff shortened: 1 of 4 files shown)


mprahl:
@lcarva I refactored it so that the API is better. Could you please take a look again?

yashvardhannanavati:
:+1:

cachito/web/models.py
```diff
@@ -67,8 +67,9 @@ def get_state_names(cls):
         return sorted([state.name for state in cls])
 
 
-class Dependency(db.Model):
-    """A dependency (e.g. gomod dependency) associated with the request."""
+class Package(db.Model):
+    """A package associated with the request."""
+    __tablename__ = 'package'
```
lcarva: Can we add a comment so we don't accidentally remove this in the future?

cachito/web/models.py
```diff
@@ -77,10 +78,19 @@ class Dependency(db.Model):
         db.UniqueConstraint('name', 'type', 'version'),
     )
 
+    def __repr__(self):
+        return f'<Package id={self.id}, name={self.name} type={self.type} version={self.version}>'
```
lcarva: Can we just use `self.__class__.__name__` so we don't have to override it on every sub class?

PR #72, opened 2019-11-14
Show the top level packages associated with a request

mprahl:


cachito/workers/pkg_managers/golang.py
```diff
@@ -84,6 +91,7 @@ def resolve_gomod_deps(app_source_path, request_id, dep_replacements=None):
             parts = [part for part in line.split(' ') if part not in ('', '<nil>')]
             if len(parts) == 1:
                 # This is the application itself, not a dependency
+                module_name = parts[0]
```
lcarva: Can we add a sanity check if `module_name` is not `None`? We should raise an error as that is unsupported/unexpected behavior.

cachito/workers/pkg_managers/golang.py
```diff
@@ -123,10 +131,150 @@ def resolve_gomod_deps(app_source_path, request_id, dep_replacements=None):
...
         dest_cache_path = os.path.join('gomod', cache_path)
-        add_deps_to_bundle(src_cache_path, dest_cache_path, request_id)
+        add_deps_to_bundle(src_cache_path, dest_cache_path, request['id'])
+
+        return module, deps
+
+
+def _get_golang_pseudo_version(commit, semantic_version=None, module_major_version=None):
+    """
+    Get the Go module's pseudo version when a non-version commit is used.
```
lcarva: Let's add a link here to where this algorithm is defined in golang docs.

cachito/workers/pkg_managers/golang.py
```diff
@@ -123,10 +131,150 @@ def resolve_gomod_deps(app_source_path, request_id, dep_replacements=None):
...
+    # the end of the module path. If the module is version v0 or v1, the major version is omitted
+    # from the module path.
+    module_major_version = None
+    match = re.match(r'(?:.+/v)(?P<major_version>\d+)$', module_name)
+    if match:
+        module_major_version = int(match.groupdict()['major_version'])
+
+    repo = git.Repo(git_path)
+    if update_tags:
+        repo.remote().fetch(force=True, tags=True)
```
lcarva: Something to think about in the future: we may want to treat failure to update the tags due to network issues as a warning, and allow the rest to proceed. Otherwise, we lose the offline capability for things we have already processed. (I do understand that we should always _try_ to update the tags.)

cachito/workers/pkg_managers/golang.py
```diff
@@ -123,10 +131,150 @@ def resolve_gomod_deps(app_source_path, request_id, dep_replacements=None):
...
+            'parsed_version': parsed_version,
+            'tag_name': tag.name,
+        })
+
+    target_commit = None
+    for commit in repo.iter_commits(target_ref):
+        if not target_commit:
+            target_commit = commit
+
+        for tag_info in commit_to_tags_info.get(commit.hexsha, []):
```
lcarva: It took me a while to realize that as soon as we hit a commit that is tagged with a semantic version, we stop processing additional commits. Can we do something like this to make it more obvious?

```python
tags_info = commit_to_tags_info.get(commit.hexsha)

if not tags_info:
  # Skip commits without a semantic version tag
  continue

for tag_info in commit_to_tags_info.get(commit.hexsha, []):
  ...
```

lcarva: What if there is a commit tagged with multiple semantic versions? I think it may just arbitrarily pick one of them? Should we consider this an error, pick the highest one, or something else?

cachito/workers/pkg_managers/golang.py
mprahl: Good point. I'll wrap this in a try/except.

mprahl: Good point. I'll add a try/except.

cachito/workers/pkg_managers/golang.py
mprahl: I've been trying to figure out what algorithm Go uses to pick the right tag when there are multiple tags on the same commit. I'm getting results that differ from what I've read in the code. I made a post and will try again on Monday.
https://forum.golangbridge.org/t/how-is-a-version-determined-on-a-commit-with-multiple-tags/16265

mprahl:
@lcarva I refactored the algorithm after inspecting the `go mod` command source code. Please take a look when you can.

cachito/workers/pkg_managers/golang.py
```diff
@@ -123,10 +138,208 @@ def resolve_gomod_deps(app_source_path, request_id, dep_replacements=None):
...
+                '%(refname:lstrip=-1)',
+                'refs/tags',
+                '--merged',
+                target_commit.hexsha,
+            ]
+        else:
+            # Get the tags that point to this commit
+            cmd = ['git', 'tag', '--points-at', target_commit.hexsha]
+
+        tag_names = g.execute(cmd).split('\n')
```
lcarva: `splitlines()`

PR #73, opened 2019-11-19
Make run_cmd more generic

mprahl:
run_cmd was moved to the pkg_managers.general module in a previous commit but it still had a Go specific error message. This makes it more generic for other package managers to use.


cachito/workers/pkg_managers/golang.py
```diff
@@ -10,6 +11,7 @@
 __all__ = ['resolve_gomod_deps']
 
 log = logging.getLogger(__name__)
+run_gomod_cmd = functools.partial(run_cmd, exc_msg='Processing gomod dependencies failed')
```
lcarva: Good call using `partial` here.

PR #76, opened 2019-12-10
Handle user set when auth is disabled

lcarva:
With this change, clients are given a meaningful error message when a
request is made with a specified user.

### DIFF
--- a/cachito/web/config.py
+++ b/cachito/web/config.py
@@ -48,6 +48,16 @@ class TestingConfig(DevelopmentConfig):
     TESTING = True
 
 
+class TestingConfigNoAuth(TestingConfig):
+    """The testing Cachito Flask configuration without authentication."""
+    # This is needed because Flask seems to read the LOGIN_DISABLED setting
+    # and configure the relevant extensions at app creation time. Changing this
+    # during a test run still leaves login enabled. This behavior also applies
+    # to ENV and DEBUG config values:
+    #   https://flask.palletsprojects.com/en/1.1.x/config/#environment-and-debug-features
+    LOGIN_DISABLED = True
+
+
 def validate_cachito_config(config, cli=False):
     """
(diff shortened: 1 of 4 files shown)


mprahl:
@lcarva could you please add a unit test for this? Otherwise, it looks good.

lcarva:
> @lcarva could you please add a unit test for this? Otherwise, it looks good.

I can't figure out how to add a unit test. To trigger this behavior, we need an instance of the app with login disabled. However, simply setting `LOGIN_DISABLED` to `True` is not sufficient. Flask must be configuring the login handlers at creation time so changing this value does nothing.

The behavior described [here](https://flask.palletsprojects.com/en/1.1.x/config/#environment-and-debug-features) seems to match what I observed.

I went as far as creating a new Config with `LOGIN_DISABLE = True` and create new `app` and `client` fixtures to create a new app based on that. However, this caused random "popped invalid context" errors in unrelated tests. 

I'm open to ideas.

lcarva:
> > @lcarva could you please add a unit test for this? Otherwise, it looks good.
> 
> I can't figure out how to add a unit test. To trigger this behavior, we need an instance of the app with login disabled. However, simply setting `LOGIN_DISABLED` to `True` is not sufficient. Flask must be configuring the login handlers at creation time so changing this value does nothing.
> 
> The behavior described [here](https://flask.palletsprojects.com/en/1.1.x/config/#environment-and-debug-features) seems to match what I observed.
> 
> I went as far as creating a new Config with `LOGIN_DISABLE = True` and create new `app` and `client` fixtures to create a new app based on that. However, this caused random "popped invalid context" errors in unrelated tests.
> 
> I'm open to ideas.

@mprahl, thanks for helping out. I removed the `db` fixture and changed the scope of the `app*` and `client*` fixtures to be function based. It seems to be working now.

PR #77, opened 2019-12-11
Fix a bug that caused duplicate files to be included in the bundle archive

mprahl:


mprahl:
> LGTM!
> 
> Can you just explain to me why this fixes the issue?

Originally, the code listed all the files in the directory. For each file, it added `source_path` to the bundle instead of the actual file in `source_path`. This means that you'd get all of `source_path` added for every file in the `source_path` directory.

This now adds each file in the `source_path` directory except for `.git` to the `app` directory in the bundle.

PR #79, opened 2019-12-11
Add support for client SSL certificate authentication

mprahl:
Addresses FACTORY-5679

### DIFF
--- a/cachito/web/auth.py
+++ b/cachito/web/auth.py
@@ -18,13 +18,44 @@ def user_loader(username):
     return User.query.filter_by(username=username).first()
 
 
+def _get_kerberos_principal(request):
+    """
+    Get the Kerberos principal from the current request.
+
+    This relies on the "REMOTE_USER" environment variable being set. This is usually set by the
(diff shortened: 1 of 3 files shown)


cachito/web/auth.py
```diff
@@ -18,14 +18,45 @@ def user_loader(username):
...
+    Get the client certificate's subject's distinguished name.
+
+    This relies on the "SSL_CLIENT_S_DN" environment variable being set. This is set by the mod_ssl
+    Apache module. If Apache is unable to verify the client certificate, no user will be returned.
+
+    :param flask.Request request: the Flask request
+    :return: the client certificate's subject's distinguished name or None
+    :rtype: str
+    """
+    ssl_client_verify = request.environ.get('SSL_CLIENT_VERIFY')
```
lcarva: Does this ensure the client cert has been issued by a trusted CA? Can we specify which CA should be used?

lcarva: This is answered by the fact that you're specifying the CA in the playbooks.

PR #82, opened 2020-01-07
Add the new request state to the database session before flushing

mprahl:
This will prevent the request_state_id column from being null after a new state is added. This only seems to happen with Postgresql since the SQLite tests pass.


mprahl:
@lcarva and @yashvardhannanavati, since you're both on PTO, I'm going to merge this one-line change to fix production.

lcarva:
:+1:

PR #85, opened 2020-01-15
Log when a user unsuccessfully acts as a representative

mprahl:


mprahl:
@lcarva alternatively, we could log all exceptions in the `json_error` function.

PR #89, opened 2020-02-11
Add the state query parameter in the pagination when used

mprahl:


mprahl:
@lcarva I came across this while writing IIB. I figured I'd fix it here before adding the code to IIB.

PR #94, opened 2020-02-17
added tox to dev dependencies

### LINKED ISSUE #93: make: tox: No such file or directory
tox dependency is missing, so running `make test` as a first-time user I get an ugly failure with: 

```
make: tox: No such file or directory
```

gsaslis:
- Added missing tox dependency to dev env
- Makefile switched to tox version inside venv

Fixes #93

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,5 @@ python:
   - "3.7"
 install:
   - pip install -r requirements-dev.txt
-  - pip install tox
 script:
   - tox
--- a/Makefile
+++ b/Makefile
@@ -11,4 +11,4 @@ run:
 	docker-compose up
 
 test:
-	tox
+	venv/bin/tox
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -3,3 +3,4 @@
 
 flake8
 pytest
+tox


gsaslis:
(sidenote: I'm a total beginner in the python world - please treat gently :D )

mprahl:
@gsaslis thank you for the contribution but I will not accept this change. The main reason is that tox should be installed outside of the test environment via a system package or if one does not exist, then by using pip with the `--user` flag or in a separate virtualenv. By putting it in the `requirements-dev.txt` file, it insinuates that you must run `pip install -r requirements-dev.txt` before running the tests, which is not the case since tox handles virtualenv creation for the tests. `requirements-dev.txt` is more or less an implementation detail of the dependencies required by tox outside of the direct dependencies used by Cachito.

gsaslis:
Thanks for taking the time to explain @mprahl !  This makes sense. 

I didn't have tox installed, so I was wondering. 

Does it perhaps make sense to have this added to the Makefile then (under `venv` target) ?

mprahl:
@gsaslis yes, that is a good idea. Feel free to file a PR with that change.

PR #99, opened 2020-02-20
Requeue the message if the worker abruptly exits or is signaled

mprahl:


cachito/workers/config.py
```diff
@@ -36,6 +36,8 @@ class Config(object):
         kombu.Queue('cachito'),
         kombu.Queue('cachito_golang', routing_key='cachito.golang'),
     )
+    # Requeue the message if the worker abruptly exits or is signaled
+    task_reject_on_worker_lost = True
```
lcarva: Will this cause duplicated images if the worker takes too long to process a message?

mprahl: What do you mean by duplicated images?

lcarva: If the worker is busy processing a task, it may lose communication with the broker. The broker will then resend the message, no?

mprahl: I think you are getting confused with the `task_acks_late` configuration which we've had on since Cachito was first deployed. With that on, if the worker is taken offline abruptly, the message should be requeued for another worker to pick up. This is desirable since we need that level of reliability and the Celery tasks are idempotent.

The `task_reject_on_worker_lost` is used for when the Celery worker is taken offline such as by a signal. In this case, as part of the shutdown, the Celery worker should reject the message so that RabbitMQ requeues the message for another work to take it.

lcarva: Aha! That makes sense. Thanks for the explanation.

PR #103, opened 2020-02-26
Install mercurial on workers

lcarva:
Some golang modules are hosted on bitbucket.org which is a Mercurial
server.

### DIFF
--- a/docker/Dockerfile-workers
+++ b/docker/Dockerfile-workers
@@ -8,6 +8,7 @@ RUN dnf -y install \
     --setopt=tsflags=nodocs \
     golang \
     git-core \
+    mercurial \
     python3-celery \
     python3-GitPython \
     python3-pip \


twaugh:
That's probably something that ought to be fixed in packages which are using those dependencies:
https://bitbucket.org/blog/sunsetting-mercurial-support-in-bitbucket

lcarva:
> That's probably something that ought to be fixed in packages which are using those dependencies:
> https://bitbucket.org/blog/sunsetting-mercurial-support-in-bitbucket

Oh nice! So maybe we can revert this when that happens. In the mean time, there's a need to use Cachito for some images by the end of March which rely on some go modules hosted in mercurial. Howe does that sound?

twaugh:
Yes, fine. Just maybe those projects want to make sure they aren't depending on hg on bitbucket.org any more upstream.

PR #104, opened 2020-03-03
add tox to dependencies

### LINKED ISSUE #93: make: tox: No such file or directory

gsaslis:
Fixes #93

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,4 +11,4 @@ run:
 	docker-compose up
 
 test:
-	venv/bin/tox
+	PATH="$PWD/venv/bin:$PATH" tox
--- a/Makefile
+++ b/Makefile
@@ -5,10 +5,10 @@ clean:
 	docker-compose down
 
 venv:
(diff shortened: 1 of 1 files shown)


gsaslis:
@mprahl as per discussion in #94

Makefile
```diff
@@ -5,10 +5,10 @@ clean:
...
 venv:
-	virtualenv --python=python3 venv && venv/bin/python setup.py develop && venv/bin/pip install -r requirements-dev.txt
+	virtualenv --python=python3 venv && venv/bin/python setup.py develop && venv/bin/pip install -r requirements-dev.txt && venv/bin/pip install tox
 
 run:
 	docker-compose up
 
 test:
-	tox
+	venv/bin/tox
```
mprahl: How about the following?
```suggestion
	PATH="$PWD/venv/bin:$PATH" tox
```

This would allow the user to fallback to the system installed version of tox if they haven't created the virtualenv.

PR #106, opened 2020-03-12
WIP: Add ruby / bundler support

gsaslis:
This PR teaches cachito how to handle projects with ruby gems as dependencies. 

It relies on an existing `Gemfile.lock` in the root of the folder and eventually downloads all packages with `bundle package --no-install --all `, which it then adds to the archive made available for download.

Some notes: 

* `replacements` are not necessary in ruby, so logic is significantly simpler than golang
* Uses git repo URL as "app name", since there's no standard app naming conventions for ruby projects. As such, bundler does not offer a way to retrieve a name for the app it is installing dependencies for, but we can rely on project URL as a unique name.
* Uses Sonatype Nexus 3 as rubygems.org proxy (we are already relying on Nexus for storing ruby gems during productisation at 3scale)

TODO: 
- [x] Add test cases for ruby / bundler 
- [ ] Document requirements for ruby projects
- [ ] Explain usage of download archive with example Dockerfile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,5 +10,9 @@ venv:
 run:
 	docker-compose up
 
+proxy_setup:
+	curl -u admin:admin123 -X POST --header 'Content-Type: application/json' http://localhost:8881/service/rest/v1/script -d @nexus.json
+	curl -u admin:admin123 -X POST --header 'Content-Type: text/plain' --header "accept: application/json" http://localhost:8881/service/rest/v1/script/proxy/run
+
 test:
 	tox
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -151,6 +151,13 @@ def create_request():
             ).on_error(error_callback)
         )
 
+    if 'bundler' in pkg_manager_names or auto_detect:
+        chain_tasks.append(
+            tasks.fetch_bundler_source.si(
+                request.id, auto_detect,
+            ).on_error(error_callback)
+        )
+
     chain_tasks.extend([
         tasks.create_bundle_archive.si(request.id).on_error(error_callback),
         tasks.set_request_state.si(request.id, 'complete', 'Completed successfully'),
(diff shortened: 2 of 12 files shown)


gsaslis:
Now that there are some tests I'm removing WIP, so we can get some 👀 here. 

@mprahl / @lcarva I could use some help / feedback on how to improve this one. 

Thanks in advance!

mprahl:
Hi @gsaslis, PR #130 added support for the npm package manager using Nexus as described in the design document you've read. PR #138 added support for non-registry dependencies. If the Ruby use-case doesn't allow non-registry dependencies, then it should be a lot simpler. If it does and it's not needed, be sure that the code fails any Cachito request that encounters a non-registry dependency.

Assuming non-registry dependencies are not needed, you'll need to modify [configure-nexus.groovy](https://github.com/release-engineering/cachito/blob/master/docker/configure-nexus.groovy) to create a proxy repository to rubygems.org which will be the catch all for all ruby dependencies used by a container first build. This is simpler than the JavaScript use-case since a repository group is needed that points to a hosted repository that contains the non-registry dependencies and the proxy to the npm registry. This specific script is only used for the docker-compose environment, but it may also be used directly or partially for the production deployment.

You'll also need to modify the [Nexus scripts](https://github.com/release-engineering/cachito/tree/master/cachito/workers/nexus_scripts) to be more generic and not assume npm. These scripts are executed using the following functions for JavaScript. The Ruby use-case will need its equivalent functions and it'd be nice to reuse some of that logic where possible.
* https://github.com/release-engineering/cachito/blob/74b37e831ad042447afd188724002eabd7ffc5d4/cachito/workers/pkg_managers/general_js.py#L294-L315
* https://github.com/release-engineering/cachito/blob/74b37e831ad042447afd188724002eabd7ffc5d4/cachito/workers/pkg_managers/general_js.py#L99-L126

If you need more detail or information, please feel free to comment here, send me an email, or set up a meeting.

mprahl:
I'm going to close this PR due to inactivity. Please reopen it if you end up pursuing this again.

I'm available for any help or advice if you need it.

PR #107, opened 2020-03-13
Do not copy downloaded dependencies to bundle dir

tkdchen:
Resolves: FACTORY-5982

### DIFF
--- a/cachito/workers/pkg_managers/golang.py
+++ b/cachito/workers/pkg_managers/golang.py
@@ -82,6 +82,7 @@ def resolve_gomod(app_source_path, request, dep_replacements=None):
 
         log.info('Downloading the gomod dependencies')
         run_gomod_cmd(('go', 'mod', 'download'), run_params)
+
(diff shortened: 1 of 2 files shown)


mprahl:
@tkdchen on second thought, I think the directory structure for the Go cache of `deps/gomod/pkg/mod/cache/download` should still exist in the bundle so that applications that use Cachito can rely on the path existing.

tkdchen:
@mprahl Could you please review #109 firstly? If that PR is accepted, it would be easy to fix this pr according to your comment. Thanks.

lcarva:
> @tkdchen on second thought, I think the directory structure for the Go cache of `deps/gomod/pkg/mod/cache/download` should still exist in the bundle so that applications that use Cachito can rely on the path existing.

Alternatively, as long as we don't set the `environment_variables`, clients should not assume the `deps/gomod` directory exists.

lcarva:
We should close this because of #111

PR #110, opened 2020-03-23
Add integration test test_http_post

ejegrova:
Implement the integration test for the HTTP POST to API scenario.

### DIFF
--- a/tests/integration/README.md
+++ b/tests/integration/README.md
@@ -1,0 +1,27 @@
+# Integration tests for Cachito
+
+This directory stores the integration tests for Cachito.
+
+## Configuration
+
+Input data for tests should be configured in the file `test_env_vars.yaml`. The file should be placed in top-level 
+directory of this repository. The path can be changed by setting `CACHITO_TEST_CONFIG` to a different path.
+
+See `tests/integration/example_test_env_vars.yaml` for a complete list of configuration options and examples.
+
+## Running the tests
+
+Tests can be triggered from the top-level directory of this repository with:
+
+    tox -e integration
+
+The integration environment is not part of the default `tox` envlist.
+
+`REQUESTS_CA_BUNDLE` can be passed in `tox.ini` for the `integration`
+environment in order to enable running the tests against Cachito instances which
+have certificates issued by a custom root certificate authority. Example usage:
+
+    REQUESTS_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.crt tox -e integration
+
+`KRB5CCNAME` can be passed in `tox.ini` for the `integration`
+environment when using Kerberos authentication in requests.
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,0 +1,20 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import os
+
+import pytest
+import yaml
+
+
+@pytest.fixture(scope="session")
+def test_env():
+    """
+    Load the test environment configuration.
+
+    :return: Test environment configuration.
+    :rtype:  dict
+    """
+    config_file = os.getenv("CACHITO_TEST_CONFIG", "test_env_vars.yaml")
+    with open(config_file) as f:
+        env = yaml.safe_load(f)
+    return env
--- a/tests/integration/test_creating_new_request.py
+++ b/tests/integration/test_creating_new_request.py
@@ -1,0 +1,37 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import utils
+
+
+def test_creating_new_request(test_env):
+    """
+    Send a new request to the Cachito API.
+
+    Checks:
+    * Check that response code is 201
+    * Check that response contains id number, same ref and repo as in request,
+        state_reason is: The request was initiated
+    """
+    client = utils.Client(test_env["api_url"], test_env["api_auth_type"])
+    response_created_req = client.create_new_request(
+        payload={
+            "repo": test_env["post"]["repo"],
+            "ref": test_env["post"]["ref"],
+            "pkg_managers": test_env["post"]["pkg_managers"],
+        },
+    )
+
+    assert response_created_req.status == 201
+
+    assert "id" in response_created_req.data
+    assert response_created_req.id > 0
+    response_specific_req = client.fetch_request(response_created_req.id)
+    assert response_created_req.id == response_specific_req.id
+
+    assert test_env["post"]["pkg_managers"] == response_created_req.data["pkg_managers"]
+    assert test_env["post"]["ref"] == response_created_req.data["ref"]
+    assert test_env["post"]["repo"] == response_created_req.data["repo"]
+    assert test_env["post"]["ref"] == response_specific_req.data["ref"]
+    assert test_env["post"]["repo"] == response_specific_req.data["repo"]
+
+    assert response_created_req.data["state_reason"] == "The request was initiated"
--- a/tests/integration/utils.py
+++ b/tests/integration/utils.py
@@ -1,0 +1,55 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from collections import namedtuple
+
+import requests
+from requests_kerberos import HTTPKerberosAuth
+
+
+Response = namedtuple("Response", "data id status")
+
+
+class Client:
+    """Simplifies access to the Cachito API."""
+
+    def __init__(self, cachito_api_url, cachito_api_auth_type):
+        """
+        Initialize the Client class.
+
+        :attribute str _cachito_api_url: URL of the Cachito API
+        :attribute _cachito_api_auth_type: kind of authentication used
+        """
+        self._cachito_api_url = cachito_api_url
+        self._cachito_api_auth_type = cachito_api_auth_type
+
+    def fetch_request(self, request_id):
+        """
+        Fetch a request from the Cachito API.
+
+        :param int request_id: ID of the request in Cachito
+        :return: Object that contains response from the Cachito API
+        :rtype: Response
+        :raises requests.exceptions.HTTPError: if the request to the Cachito API fails
+        """
+        resp = requests.get(f"{self._cachito_api_url}/requests/{request_id}")
+        resp.raise_for_status()
+        return Response(resp.json(), resp.json()["id"], resp.status_code)
+
+    def create_new_request(self, payload):
+        """
+        Create a new request in Cachito.
+
+        :param dict payload: Payload for request
+        :return: Object that contains response from the Cachito API
+        :rtype: Response
+        :raises requests.exceptions.HTTPError: if the request to the Cachito API fails
+        """
+        authentication_mapping = {"kerberos": HTTPKerberosAuth()}
+        resp = requests.post(
+            f"{self._cachito_api_url}/requests",
+            auth=authentication_mapping.get(self._cachito_api_auth_type),
+            headers={"Content-Type": "application/json"},
+            json=payload,
+        )
+        resp.raise_for_status()
+        return Response(resp.json(), resp.json()["id"], resp.status_code)
--- a/tox.ini
+++ b/tox.ini
@@ -6,12 +6,15 @@ deps =
     -rrequirements.txt
     -rrequirements-web.txt
     pytest
+    pyyaml
 whitelist_externals =
     make
 setenv =
 	CACHITO_TESTING=true
 commands =
-    py.test -vvv {posargs}
+    py.test -vvv \
+        --ignore tests/integration \
+        {posargs}
 
 [testenv:black]
 description = black checks [Mandatory]
@@ -40,3 +43,11 @@ per-file-ignores =
     # Ignore missing docstrings in the tests and migrations
     tests/*:D103
     cachito/web/migrations/*:D103
+
+[testenv:integration]
+basepython = python3
+skipsdist = true
+commands =
+    pytest -rA -vvv \
+        --confcutdir=tests/integration \
+        {posargs:tests/integration}
--- a/tests/integration/example_test_env_vars.yaml
+++ b/tests/integration/example_test_env_vars.yaml
@@ -1,0 +1,13 @@
+---
+# Cachito API endpoint
+api_url: https://cachito.example.com/api/v1
+# Authentication type, user either null for no authentication or kerberos
+api_auth_type: kerberos
+# Test data for the Creating new request test
+post:
+  # The URL for the upstream git repository
+  repo: https://github.com/release-engineering/retrodep.git
+  # A git reference at the given git repository
+  ref: e1be527f39ec31323f0454f7d1422c6260b00580
+  # Package managers
+  pkg_managers: ["gomod"]


ejegrova:
@lcarva @mprahl ptal :-)

requirements.txt
```diff
@@ -4,3 +4,4 @@ kombu # A celery dependency but it's directly imported
 requests_kerberos
 requests
 semver
+pyyaml
```
lcarva: This implies that we need `pyyaml` to run Cachito itself. Can we just add this to `requirements-dev.txt` instead? Or maybe a new file, `requirements-integration-test.txt` ?

lcarva: Seems like reusing `requirements-dev.txt` might be a good idea since it already includes tox?

tests/integration/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Running tests
+
+Tests can be triggered from the top-level directory of this repository with:
+
+    tox -e integration
+
+The integration environment is not part of the default `tox` envlist.
+
+`REQUESTS_CA_BUNDLE` is passed in `tox.ini` for the `integration`
```
lcarva: I think we want to say "can be passed" here because we don't necessarily have to do this. It depends on the Cachito instance you're running against.

BTW, I always set this variable in my `.bashrc` to prevent headaches. Is that not sufficient here?

tests/integration/README.md
```diff
@@ -0,0 +1,23 @@
...
+environment in order to enable running the tests against Cachito instances which
+have self-signed certificates. Example usage:
```
lcarva: Not necessarily self-signed certs, but issue by a custom root certificate authority.

tests/integration/conftest.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+@pytest.fixture(scope="session")
+def test_env():
+    """Load test environment configuration
+
+    :return: Test environment configuration.
+    :rtype:  dict
+    """
+    config_file = os.getenv("Cachito_TEST_CONFIG", "test_env_vars.yaml")
```
lcarva: Can we use all upper case on the environment variable name, `CACHITO_TEST_CONFIG`?

You may also want to mention this env var in the `README.md`.

tests/integration/test_http_post.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+# SPDX-License-Identifier: MIT
+
+import utils
+
+
+def test_http_post(test_env):
```
lcarva: Can we name this after the use case?

tests/integration/test_http_post.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
```
lcarva: I don't think we need this.

tests/integration/test_http_post.py
```diff
@@ -0,0 +1,31 @@
...
+# SPDX-License-Identifier: MIT
```
lcarva: The Cachito source code is labeled as: `GPL-3.0-or-later`
Do we want to deviate the integration tests from it?

tests/integration/test_http_post.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    client = utils.Client(test_env["api_url"])
+    response_created_req = client.create_new_request(
+        repo=test_env["post"]["repo"],
+        ref=test_env["post"]["ref"],
+        pkg_managers=test_env["post"]["pkg_managers"])
+
+    assert response_created_req.status == 201
+
+    assert "id" in response_created_req.data.keys()
```
lcarva: No need for `.keys()`

lcarva: It would be great to also validate that in the `response_created_req`, `ref`, `repo`, and `pkg_managers` are correctly set.

We should also verify the `id` is an integer greater than 0.

tests/integration/utils.py
```diff
@@ -0,0 +1,59 @@
...
+        return Response(resp.json(), None, resp.status_code)
+
+    def fetch_request(self, request_id):
+        """Fetch a request from Cachito API
+
+        :param request_id: ID of the request in Cachito
+        :return: Object that contains response from Cachito API
+        :rtype: Response
+        """
+        resp = requests.get(f"{self._cachito_api_url}requests/{request_id}")
```
lcarva: Should we also `raise_for_status` here?

tests/integration/utils.py
```diff
@@ -0,0 +1,59 @@
...
+        return Response(resp.json(), resp.json()["id"], resp.status_code)
+
+    def create_new_request(self, **kwargs):
```
lcarva: Suggestion: use `payload` instead of `**kwargs`. That way if we do need to pass in another parameter to this method, it doesn't end up in the request payload.

tests/integration/utils.py
```diff
@@ -0,0 +1,59 @@
+# -*- coding: utf-8 -*-
+# SPDX-License-Identifier: MIT
+
+import requests
+from requests_kerberos import HTTPKerberosAuth
+
+
+class Response:
```
lcarva: Consider using a [namedtuple](https://docs.python.org/2/library/collections.html#collections.namedtuple), unless you plan on adding more functionality to this class later.

tests/integration/utils.py
```diff
@@ -0,0 +1,59 @@
...
+    def __init__(self, cachito_api_url):
+        self._cachito_api_url = cachito_api_url
+
+    def fetch_all_requests(self):
+        """Fetch a list of requests from the Cachito API
+
+        :return: Object that contains response from Cachito API
+        :rtype: Response
+        """
+        resp = requests.get(f"{self._cachito_api_url}requests")
```
lcarva: It might be easier to read if there's a slash before `requests`: `f"{self._cachito_api_url}/requests"`
Consider changing how the Cachito API URL is defined to adjust for this.

tests/integration/utils.py
```diff
@@ -0,0 +1,59 @@
...
+        return Response(resp.json(), resp.json()["id"], resp.status_code)
+
+    def create_new_request(self, **kwargs):
+        """Create new request in Cachito
+
+        :param dict kwargs: Payload for request
+        :return: Object that contains response from Cachito API
+        :rtype: Response
+        """
+        resp = requests.post(f"{self._cachito_api_url}requests", auth=HTTPKerberosAuth(),
```
lcarva: I'd like if `auth` was configurable to either use Kerberos, SSL Certs, or no auth. This would make the tests more portable between the various environment. The local dev environment, `docker-compose up`, does not use auth, for example.

tox.ini
```diff
@@ -23,3 +25,12 @@ max-line-length = 100
...
 ignore = W504
+
+[testenv:integration]
+basepython = python3
+skipsdist = true
+commands =
+    pytest -rA -vvv \
+        --confcutdir=tests/integration \
+        {posargs:tests/integration}
+passenv = KRB5CCNAME
```
lcarva: What's this for?

tests/integration/README.md
ejegrova: You can set it in .bashrc too and it will work but I think it's good to mention this here.

ejegrova: Changed that to "can be passed"

requirements.txt
ejegrova: I moved it to `requirements-dev.txt`

tests/integration/test_http_post.py
ejegrova: deleted

tests/integration/test_http_post.py
ejegrova: deleted .keys()

tests/integration/test_http_post.py
ejegrova: yes, we want that (it's copy-paste mistake)

tests/integration/README.md
ejegrova: changed

tests/integration/conftest.py
ejegrova: done

tests/integration/conftest.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
lcarva: Remove this too?

tests/integration/test_http_post.py
ejegrova: I renamed it to `test_creating_new_request`, is it ok?

tests/integration/utils.py
ejegrova: changed

tests/integration/utils.py
ejegrova: done, namedtuples are cool  :)

tests/integration/utils.py
ejegrova: changed

tests/integration/utils.py
ejegrova: done

tests/integration/conftest.py
ejegrova: done

tox.ini
ejegrova: When using Kerberos authentication, tox needs this variable set (credential cache location) otherwise it will fail with error: kerberos.GSSError: ('No Kerberos credentials available: No KCM server found', -1765328243). But since there does not have to be Kerberos authentication used, I'll remove the variable from here and mention it in Readme, ok?

lcarva: +1 sounds like a good idea.

tests/integration/utils.py
lcarva: They sure are! :)

tests/integration/utils.py
lcarva: I think this would need a bit more changes to make it truly configurable. For instance, it would be ideal if alongside `api_url` in the config yaml, we had `api_auth_type` which could be set to either 'kerberos', or `null` (we can add `ssl` later). Then, in `create_new_request`, we'd map the string `'kerberos'` to 'HTTPKerberosAuth()', for example.

This can be done later in a follow up PR.

tests/integration/conftest.py
```diff
@@ -0,0 +1,18 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import os
+import yaml
+import pytest
+
+
+@pytest.fixture(scope="session")
```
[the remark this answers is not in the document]
mprahl: PR #113 actually changed this. So this is no longer an issue.

mprahl:
@ejegrova could you please rebase? There is additional linting now that will help resolve some style issues in the doc strings.

tests/integration/utils.py
ejegrova: I did the changes, can you check that? I am not sure if there's a better way to do the mapping.

tests/integration/utils.py
```diff
@@ -0,0 +1,50 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import requests
+from collections import namedtuple
+
+
+Response = namedtuple('Response', 'data id status')
```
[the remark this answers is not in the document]
ejegrova: I think it could be useful to have it there because ID will be used often. But it's included in data attribute so it's not necessary.

tests/integration/utils.py
```diff
@@ -0,0 +1,50 @@
...
+    def __init__(self, cachito_api_url):
+        self._cachito_api_url = cachito_api_url
+
+    def fetch_all_requests(self):
+        """Fetch a list of requests from the Cachito API
+
+        :return: Object that contains response from Cachito API
+        :rtype: Response
+        """
+        resp = requests.get(f"{self._cachito_api_url}/requests")
```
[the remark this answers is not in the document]
lcarva: Good catch. Maybe we can just omit this method for this PR since it's not really used.

tests/integration/utils.py
lcarva: In python, `None` is a hashable value. So you could have:
```python
authentication_mapping = {
  None: None,
  "kerberos": HTTPKerberosAuth(),
}
```
This is better than having to force a string of value `'null``.

But there's no point to map a value to itself. Consider this:
```python
authentication_mapping = {
  "kerberos": HTTPKerberosAuth()
}
```

Then use `get` with a default value:
```python
auth=authentication_mapping.get(auth, auth),
```

ejegrova: Thanks for the explanation, it's changed now.

tests/integration/conftest.py
```diff
@@ -0,0 +1,18 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import os
+import yaml
+import pytest
```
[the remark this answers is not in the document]
ejegrova: changed

tests/integration/test_creating_new_request.py
```diff
@@ -0,0 +1,38 @@
...
+    * Check that response contains id number, same ref and repo as in request,
+        state_reason is: The request was initiated
+    """
+    client = utils.Client(test_env["api_url"])
+    response_created_req = client.create_new_request(
+        auth=HTTPKerberosAuth(),
+        payload={"repo": test_env["post"]["repo"],
+                 "ref": test_env["post"]["ref"],
+                 "pkg_managers": test_env["post"]["pkg_managers"]
+                 }
```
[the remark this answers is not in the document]
ejegrova: chenged

tests/integration/utils.py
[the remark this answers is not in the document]
ejegrova: removing the method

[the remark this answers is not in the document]
lcarva: I think this resolves this conversation, but I'll let @mprahl confirm and close it out.

tests/integration/test_creating_new_request.py
```diff
@@ -0,0 +1,38 @@
...
+    response_specific_req = client.fetch_request(response_created_req.id)
+    assert response_created_req.id == response_specific_req.id
+
+    assert test_env["post"]["pkg_managers"] == response_created_req.data["pkg_managers"]
+    assert test_env["post"]["ref"] == response_created_req.data["ref"]
+    assert test_env["post"]["repo"] == response_created_req.data["repo"]
+    assert test_env["post"]["ref"] == response_specific_req.data["ref"]
+    assert test_env["post"]["repo"] == response_specific_req.data["repo"]
+
+    assert response_created_req.data["state_reason"] == "The request was initiated"
```
mprahl, line 38: Isn't this susceptible to a race condition since Cachito could have changed the status quickly? I think checking that the state is `in_progress` or just not `failed` would be safer.

tests/integration/utils.py
```diff
@@ -0,0 +1,52 @@
...
+        self._cachito_api_url = cachito_api_url
+
+    def fetch_request(self, request_id):
+        """
+        Fetch a request from the Cachito API.
+
+        :param int request_id: ID of the request in Cachito
+        :return: Object that contains response from the Cachito API
+        :rtype: Response
+        """
```
mprahl, line 30: It'd nice to mention that this method can raise an exception with `:raises exception_type_here:  if the request to the Cachito API fails`.

tests/integration/utils.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def create_new_request(self, auth, payload):
+        """
+        Create a new request in Cachito.
+
+        :param auth: kind of authentication used
+        :param dict payload: Payload for request
+        :return: Object that contains response from the Cachito API
+        :rtype: Response
+        """
```
mprahl, line 43: It'd nice to mention that this method can raise an exception with `:raises exception_type_here:  if the request to the Cachito API fails`.

tests/integration/utils.py
```diff
@@ -0,0 +1,52 @@
...
+
+        :param int request_id: ID of the request in Cachito
+        :return: Object that contains response from the Cachito API
+        :rtype: Response
+        """
+        resp = requests.get(f"{self._cachito_api_url}/requests/{request_id}")
+        resp.raise_for_status()
+        return Response(resp.json(), resp.json()["id"], resp.status_code)
+
+    def create_new_request(self, auth, payload):
```
mprahl, line 35: Why did you decide to not have authentication automatically configured using the configuration rather than having to pass it in all the time?

It's okay that you did but I'm just wondering if there is a reason that this needs to be explicit.

tests/integration/test_creating_new_request.py
ejegrova: The variable response_created_req that checks the state_reason is the response that Cachito sends in the moment when it gets the new request. Based on what I tried with API it sends this state_reason or the state is failed immediately.
The status is changing quickly when you fetch request by its ID, that is stored in variable response_specific_req.
Is this correct or am I missing something?

tests/integration/utils.py
ejegrova: There's no reason, I didn't realise I could do it like that. It's changed now.

tests/integration/test_creating_new_request.py
mprahl: It's my mistake. I just misread the code.

tests/integration/utils.py
```diff
@@ -0,0 +1,57 @@
...
+        :param dict payload: Payload for request
+        :return: Object that contains response from the Cachito API
+        :rtype: Response
+        :raises requests.exceptions.HTTPError: if the request to the Cachito API fails
+        """
+        authentication_mapping = {"kerberos": HTTPKerberosAuth()}
+        resp = requests.post(
+            f"{self._cachito_api_url}/requests",
+            auth=authentication_mapping.get(
+                self._cachito_api_auth_type, self._cachito_api_auth_type
```
mprahl, line 51: If the authentication type is not found, you should be passing `auth=None` since it should default to no authentication.
```suggestion
                self._cachito_api_auth_type
```

PR #112, opened 2020-03-24
Add Nexus Repository Manager to the docker-compose environment

mprahl:
This is a prerequisite for adding JavaScript support.


mprahl:
@jdcasey could you please review the Groovy script that is added as part of this PR?

docker-compose.yml
```diff
@@ -27,6 +27,18 @@ services:
     ports:
       - 3000:3000
 
+  nexus:
+    # Run as the root user so that the container can write to the volume
+    user: "0:0"
```
lcarva: How come that's not needed for the other containers?

lcarva: There's a "Persistent Data" section that might be helpful: https://hub.docker.com/r/sonatype/nexus3/

Also, they mention a Red Hat certified image, have you considered using that?

cachito/nexus_scripts/configure_nexus.groovy
```diff
@@ -0,0 +1,139 @@
+// This script configures Nexus to be ready for Cachito and is idempotent
```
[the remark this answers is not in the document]
mprahl: I was thinking it would expand to cover other languages. In the future, we could split up in the following way:
* General configuration
* JavaScript repositories
* Other language repositories

cachito/nexus_scripts/configure_nexus.groovy
```diff
@@ -0,0 +1,139 @@
...
+import com.google.common.collect.Sets
```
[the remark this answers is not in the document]
mprahl: I can do that. I did intentionally create a lot of variables so that it was clear what was being input in the methods. I could just add comments instead.

cachito/nexus_scripts/configure_nexus.groovy
```diff
@@ -0,0 +1,139 @@
...
+proxyRepoConfig.attributes('storage').set('strictContentTypeValidation', true)
+
+def proxy = proxyRepoConfig.attributes('proxy')
+proxy.set('remoteUrl', 'https://registry.npmjs.org')
+// Don't expire the cache
+proxy.set('contentMaxAge', -1)
+proxy.set('metadataMaxAge', -1)
+
+def httpclient = proxyRepoConfig.attributes('httpclient')
+httpclient.set('autoBlock', true)
```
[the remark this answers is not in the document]
mprahl: Which attributes? All attributes for the repo or just the `httpclient` attributes?

cachito/nexus_scripts/configure_nexus.groovy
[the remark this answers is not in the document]
lcarva: Having a lot of variables is not a bad thing. However, they all have the same scope. I could accidentally use reuse a variable later in the script. Reduce the scope of each variable, not the amount of variables.

cachito/nexus_scripts/configure_nexus.groovy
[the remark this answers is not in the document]
lcarva: All attributes would be great. If there's a page that lists them all, just link to that.

docker-compose.yml
mprahl: I'll change it to use a named volume.

The image being used is based on UBI 8 so I assumed that was good enough. Did you want one specifically running on RHEL?

lcarva: I think the certified Red Hat image is the same image, just from the Red Hat registry.

mprahl: @lcarva don't you need to login to the Red Hat container registry before using it? If so, I'll just stuck with the one from Docker Hub since this is just for development.

lcarva: You're right. For the sake of usability, let's use the current one.

mprahl:

cachito/nexus_scripts/configure_nexus.groovy
```diff
@@ -0,0 +1,206 @@
...
+import org.sonatype.nexus.security.authz.AuthorizationManager
+import org.sonatype.nexus.security.role.NoSuchRoleException
+import org.sonatype.nexus.security.role.Role
+import static org.sonatype.nexus.security.user.UserManager.DEFAULT_SOURCE
+import org.sonatype.nexus.security.user.UserNotFoundException
+
+
+def createBlobStore(String blobStoreName) {
+    if (!blobStore.blobStoreManager.get(blobStoreName)) {
+        String blobStorePath = blobStoreName
```
lcarva: This looks odd. Why do we need to make a reference copy to `blobStoreName`? Are both variables really supposed to have the same value? If so, let's comment why.

mprahl: This is intentional. I will add a comment.

mprahl: @lcarva the extra variable is not necessary, but it makes it clear what the input is to the method.

mprahl:
@lcarva I addressed your comment. Could you please review it again?

mprahl:
@lcarva I added a bit more to the script. There is now an additional user that is created that has read access on the cachito-js repo. This is so that the per request proxy repos can authenticate with the cachito-js repo.

PR #113, opened 2020-03-30
Use Python black to format the code

mprahl:
Additionally, start linting the docstrings with flake8-docstrings

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = lint,py37
+envlist = black,flake8,py37
 
 [testenv]
 deps =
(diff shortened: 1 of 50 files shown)


tox.ini
```diff
@@ -11,15 +11,28 @@ setenv =
 commands =
     py.test -vvv {posargs}
 
-[testenv:lint]
+[testenv:black]
+description = black checks [Mandatory]
+skip_install = true
 deps =
-    flake8
+    black==19.10b0
```
lcarva: What's the reason for pinning the deps here and in `testenv:flake8` ?

mprahl: I added a comment explaining this.

PR #123, opened 2020-03-31
Add the scripts to configure Nexus for Cachito JavaScript support

mprahl:
This PR also fixes the Nexus development configuration and a typo in a docstring.

Please see each commit for more details.

### DIFF
--- a/cachito/workers/nexus_scripts/js_after_content_staged.groovy
+++ b/cachito/workers/nexus_scripts/js_after_content_staged.groovy
@@ -1,0 +1,98 @@
+/*
+This script configures Nexus so that the NPM proxy repository for the Cachito request is blocked from getting additional
+content.
+
+In addition, a temporary user for the request is created and is given permission to access this NPM proxy repository.
+It'd be preferable to give access to the Nexus anonymous user instead, but there is no way to add a role to a user.
+You can only set the entire set of roles at once. This is an issue since if more than one Cachito request is in progress
+and modifies the set of roles at the same time, one of the additions will be lost.
+ */
+import com.google.common.collect.Sets
+import groovy.json.JsonSlurper
+import groovy.transform.Field
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import org.sonatype.nexus.security.role.NoSuchRoleException
+import org.sonatype.nexus.security.user.UserStatus
+import org.sonatype.nexus.repository.config.Configuration
+import org.sonatype.nexus.security.authz.AuthorizationManager
+import org.sonatype.nexus.security.role.Role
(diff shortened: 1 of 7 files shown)


cachito/workers/nexus_scripts/js_after_content_staged.groovy
```diff
@@ -0,0 +1,97 @@
+/*
+This script configures Nexus so that the NPM proxy repository for the Cachito request is blocked from getting additional
+content.
+
+In addition, a temporary user for the request is created and is given permission to access this NPM proxy repository.
+It'd be preferable to give access to the Nexus anonymous user instead, but there is no way to add a role to a user.
```
lcarva, line 6: Is there a way to use a single role to grant access only to all the request specific repos? Not sure if we can name these repos with a common prefix and use something like `"nx-repository-view-npm-public*-read"` as the role added to the anonymous user.

mprahl: I will investigate to see if I can create a [privilege](https://help.sonatype.com/repomanager3/security/privileges#Privileges-Example:CreatingaRepositoryViewPrivilege) that does that.

I've been going with the privileges that Nexus creates for us.

docker/configure-nexus.py
```diff
@@ -32,13 +32,36 @@
...
+    json=realms,
+    timeout=15,
+)
+
+if not rv_set_realms.ok:
+    print(
+        f'Setting the realms to {realms} failed with the status code: {rv_set_realms.status_code}\n'
+        f'The output was: {rv_set_realms.text}',
+        file=sys.stderr,
+    )
```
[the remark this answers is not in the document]
mprahl: Good catch! I'll fix that.

cachito/workers/nexus_scripts/js_after_content_staged.groovy
mprahl: So for the privilege `nx-repository-view-npm-cachito-js-1-read`, the permission behind the privilege is `nexus:repository-view:npm:cachito-js-1:read`. It seems like between the colons, you can use the `*` wildcard to denote things such as any repo or any action on the repo, but I haven't figured out how to say read access on all repos with the pattern `cachito-js-\d+`.

There's like no documentation on this, so I'll contact Sonatype support to see if they have any suggestions.

mprahl: @lcarva according to Sonatype support, they say this is not possible.

lcarva: :sob:

Issue #86, opened 2020-01-30
Set Content-Disposition: attachment with filename in /requests/{id}/download


lcarva:
Set [Content-Disposition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) with a meaningful filename.

This would improve the user experience when users plug in the download url in the browser. The name should include the request ID and have `tar.gz` suffix.

lcarva:
It looks like [flask.send_file](https://flask.palletsprojects.com/en/1.1.x/api/#flask.send_file) supports this.

tkdchen:
This has been fixed in #124 and can be closed.

PR #124, opened 2020-04-01
Set header Content-Disposition for request bundle archive download

tkdchen:

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -104,7 +104,12 @@ def download_archive(request_id):
     flask.current_app.logger.debug(
         "Sending the bundle at %s for request %d", bundle_dir.bundle_archive_file, request_id
     )
-    return flask.send_file(str(bundle_dir.bundle_archive_file), mimetype="application/gzip")
+    return flask.send_file(
+        str(bundle_dir.bundle_archive_file),
+        mimetype="application/gzip",
+        as_attachment=True,
+        attachment_filename=f"cachito-{request_id}.tar.gz",
+    )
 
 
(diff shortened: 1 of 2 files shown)


cachito/web/api_v1.py
```diff
@@ -104,7 +104,9 @@ def download_archive(request_id):
     flask.current_app.logger.debug(
         "Sending the bundle at %s for request %d", bundle_dir.bundle_archive_file, request_id
     )
-    return flask.send_file(str(bundle_dir.bundle_archive_file), mimetype="application/gzip")
+    return flask.send_file(
+        str(bundle_dir.bundle_archive_file), mimetype="application/gzip", as_attachment=True,
```
mprahl, line 108: Could you also set `attachment_filename` as requested in #86?

tkdchen: @mprahl I tested the `send_file`, with `as_attachment=True`, the `Content-Disposition` header already has `filename` set with the same filename as the bundle archive file.

The doc tells:
> attachment_filename – the filename for the attachment if it **differs from** the file’s filename.

tkdchen: @mprahl Are you thinking of to set a different filename rather than just the `{request.id}.tar.gz`?

mprahl: Got it. I didn't know that it was automatically set.

Optional: I'm thinking it might be nice to name it as `f"cachito-{request.id}.tar.gz"`.

tkdchen: PR is modified to set this filename. Meanwhile, I also refactored the test without mocking flask.send_file, so that we can verify the expected result from response explicitly.

PR #127, opened 2020-04-06
Add the configuration API endpoint

mprahl:
This allows Cachito to specify configuration files that must be set by the consumer of the bundle for it to be usable.

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import copy
+import functools
 
 from celery import chain
 import flask
@@ -9,6 +10,7 @@ from werkzeug.exceptions import Unauthorized, InternalServerError
 from cachito.errors import ValidationError
 from cachito.web import db
 from cachito.web.models import (
+    ConfigFileBase64,
     Dependency,
     EnvironmentVariable,
     Package,
@@ -73,6 +75,22 @@ def get_request(request_id):
     :raise NotFound: if the request is not found
     """
     return flask.jsonify(Request.query.get_or_404(request_id).to_json())
+
+
+@api_v1.route("/requests/<int:request_id>/configuration-files", methods=["GET"])
+def get_request_config_files(request_id):
+    """
+    Retrieve the configuration files associated with the given request.
+
+    :param int request_id: the value of the request ID
+    :return: a Flask JSON response
+    :rtype: flask.Response
+    :raise NotFound: if the request is not found
+    """
+    config_files = Request.query.get_or_404(request_id).config_files_base64
+    config_files_json = [config_file.to_json() for config_file in config_files]
+    config_files_json = sorted(config_files_json, key=lambda c: c["path"])
+    return flask.jsonify(config_files_json)
 
 
 @api_v1.route("/requests/<int:request_id>/download", methods=["GET"])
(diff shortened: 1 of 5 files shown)


cachito/web/api_v1.py
```diff
@@ -75,6 +76,22 @@ def get_request(request_id):
...
+
+    :param int request_id: the value of the request ID
+    :return: a Flask JSON response
+    :rtype: flask.Response
+    :raise NotFound: if the request is not found
+    """
+    config_files = Request.query.get_or_404(request_id).config_files_base64
+    config_files_json = [config_file.to_json() for config_file in config_files]
+    config_files_json = sorted(config_files_json, key=lambda c: c["path"])
+    return flask.jsonify(config_files_json)
```
lcarva, line 92: This should just return an empty list if there are no config files, right? Can we add a unit test for it?

cachito/web/api_v1.py
```diff
@@ -188,10 +217,7 @@ def patch_request(request_id):
     :raise NotFound: if the request is not found
     :raise ValidationError: if the JSON is invalid
     """
-    allowed_users = flask.current_app.config["CACHITO_WORKER_USERNAMES"]
-    # current_user.is_authenticated is only ever False when auth is disabled
-    if current_user.is_authenticated and current_user.username not in allowed_users:
-        raise Unauthorized("This API endpoint is restricted to Cachito workers")
+    assert_user_is_worker()
```
lcarva, line 220: Consider using a decorator for this.

cachito/web/models.py
```diff
@@ -743,3 +760,80 @@ def to_json(self):
...
+    @classmethod
+    def validate_json(cls, payload):
+        """
+        Validate the input configuration file.
+
+        :param list payload: the list of dictionaries of configuration files
+        :raises ValidationError: if the configuration file is invalid
+        """
+        if not isinstance(payload, dict):
+            raise ValidationError("The base64 configuration file must be a JSON object")
```
lcarva, line 815: If type is a parameter, it doesn't right to hard code `base64` in these log messages.

cachito/web/static/api_v1.yaml
```diff
@@ -185,6 +185,78 @@ paths:
               $ref: '#/components/schemas/RequestUpdate'
       security:
       - "Kerberos Authentication": []
+  "/requests/{id}/configuration-files":
```
lcarva, line 188: Can we add a link to the configuration-files from the request object in the API response? This will make it simpler for clients to realize there's additional information they need to fetch. I'd always include the link, even if there are no configuration files.

cachito/web/models.py
mprahl: @lcarva after thinking about this, I'll move most of this method to the base class and require each child class have the `type_name` class property.

lcarva: IMO base class is unnecessary when there's only a single type being introduced, but I'll leave that up to you :)

cachito/web/static/api_v1.yaml
lcarva: Did this make into api_v1.yaml?

lcarva: @mprahl ping

mprahl: @lcarva just added it. Thanks!

mprahl:
@lcarva thanks for the review! This is ready for another round.

PR #128, opened 2020-04-07
Add integration test test_valid_data_in_request

ejegrova:
Implement the integration test for validating data in the request.

### DIFF
--- a/tests/integration/example_test_env_vars.yaml
+++ b/tests/integration/example_test_env_vars.yaml
@@ -11,3 +11,28 @@ post:
   ref: e1be527f39ec31323f0454f7d1422c6260b00580
   # Package managers
   pkg_managers: ["gomod"]
+# Test data for the Valid data in request test
+get:
+  repo: https://github.com/release-engineering/retrodep.git
+  ref: e1be527f39ec31323f0454f7d1422c6260b00580
+  pkg_managers: ["gomod"]
+  # The list of fetched packages
+  # Every package is defined by name, type and version.
+  packages:
+    - ["github.com/release-engineering/retrodep/v2", "gomod", "v2.1.1"]
+  # The list of fetched dependencies
+  # Every dependency is defined by name, type and version.
+  dependencies:
+  - ["github.com/Masterminds/semver", "gomod", "v1.4.2"]
+  - ["github.com/kr/pretty", "gomod", "v0.1.0"]
+  - ["github.com/kr/pty", "gomod", "v1.1.1"]
+  - ["github.com/kr/text", "gomod", "v0.1.0"]
+  - ["github.com/op/go-logging", "gomod", "v0.0.0-20160315200505-970db520ece7"]
+  - ["github.com/pkg/errors", "gomod", "v0.8.1"]
+  - ["golang.org/x/crypto", "gomod", "v0.0.0-20190308221718-c2843e01d9a2"]
+  - ["golang.org/x/net", "gomod", "v0.0.0-20190311183353-d8887717615a"]
+  - ["golang.org/x/sys", "gomod", "v0.0.0-20190215142949-d0b11bdaac8a"]
+  - ["golang.org/x/text", "gomod", "v0.3.0"]
+  - ["golang.org/x/tools", "gomod", "v0.0.0-20190325161752-5a8dccf5b48a"]
+  - ["gopkg.in/check.v1", "gomod", "v1.0.0-20180628173108-788fd7840127"]
+  - ["gopkg.in/yaml.v2", "gomod", "v2.2.2"]
--- a/tests/integration/test_valid_data_in_request.py
+++ b/tests/integration/test_valid_data_in_request.py
@@ -1,0 +1,61 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import operator
+import time
+
+import utils
+
+
+def test_valid_data_in_request(test_env):
+    """
+    Validate data in the request.
+
+    Process:
+    Send new request to the Cachito API
+    Send request to check status of existing request
+
+    Checks:
+    * Check that response code is 200
+    * Check in the response that state is complete
+    * Check that "packages" and "dependencies" keys have appropriate values
+    """
+    client = utils.Client(test_env["api_url"], test_env["api_auth_type"])
+    response_created_req = client.create_new_request(
+        payload={
+            "repo": test_env["get"]["repo"],
+            "ref": test_env["get"]["ref"],
+            "pkg_managers": test_env["get"]["pkg_managers"],
+        },
+    )
+
+    response = client.fetch_request(response_created_req.id)
+    assert response.status == 200
+
+    start_time = time.time()
+    while response.data["state"] == "in_progress":
+        # 300 seconds
+        if time.time() - start_time >= 300:
+            raise TimeoutError("The Cachito request did not complete within 5 minutes")
+        time.sleep(5)
+        response = client.fetch_request(response_created_req.id)
+    assert response.data["state"] == "complete"
+
+    response_dependencies = list_of_dict_to_list_of_name_type_version(response.data["dependencies"])
+    expected_dependencies = test_env["get"]["dependencies"]
+    assert response_dependencies == expected_dependencies
+
+    response_packages = list_of_dict_to_list_of_name_type_version(response.data["packages"])
+    expected_packages = test_env["get"]["packages"]
+    assert response_packages == expected_packages
+
+
+def list_of_dict_to_list_of_name_type_version(data):
+    """
+    Convert the list of dictionaries to a list of lists from the keys name, type, and version.
+
+    :param data: list of dictionaries containing keys name, type and version
+    :return: list of lists with values name, type and version in this order
+    """
+    sorted_packages = sorted(data, key=operator.itemgetter("name"))
+
+    return [[i["name"], i["type"], i["version"]] for i in sorted_packages]


tests/integration/example_test_env_vars.yaml
```diff
@@ -11,3 +11,15 @@ post:
...
   pkg_managers: ["gomod"]
+# Test data for the Valid data in request test
+get:
+  repo: https://github.com/release-engineering/retrodep.git
+  ref: e1be527f39ec31323f0454f7d1422c6260b00580
+  pkg_managers: ["gomod"]
+  # The name of the package
+  packages_names: "github.com/release-engineering/retrodep/v2"
+  # The names of fetched dependencies
+  dependencies_names: ["gopkg.in/yaml.v2", "gopkg.in/check.v1", "golang.org/x/tools", "golang.org/x/text",
```
lcarva, line 22: Consider using the syntax:
```
dependencies_names:
- gopkg.in/yaml.v2
- gopkg.in/check.v1
...
```
Order them alphabetically, and it should much easier to find items in it :)

tests/integration/example_test_env_vars.yaml
```diff
@@ -11,3 +11,15 @@ post:
   ref: e1be527f39ec31323f0454f7d1422c6260b00580
   # Package managers
   pkg_managers: ["gomod"]
+# Test data for the Valid data in request test
+get:
+  repo: https://github.com/release-engineering/retrodep.git
+  ref: e1be527f39ec31323f0454f7d1422c6260b00580
+  pkg_managers: ["gomod"]
+  # The name of the package
+  packages_names: "github.com/release-engineering/retrodep/v2"
```
lcarva, line 20: If a single value is expected, let's rename it to `package_name`. Otherwise, turn it into a list.

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,42 @@
...
+            "repo": test_env["get"]["repo"],
+            "ref": test_env["get"]["ref"],
+            "pkg_managers": test_env["get"]["pkg_managers"],
+        },
+    )
+
+    response = client.fetch_request(response_created_req.id)
+    assert response.status == 200
+
+    while True:
```
lcarva, line 30: Let's build some sort of timeout so this doesn't run forever if there's a bug in Cachito.

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,42 @@
...
+        if response.data["state"] == ("complete" or "failed"):
```
lcarva, line 31: I think this is only matching when the state goes into "complete". 
The expression `("complete" or "failed")` is evaluated immediately and since the first value is truthy, that's what it evaluates to :
```python
>>> "complete" or "failed"
'complete'
```
You probably want:
```python
if response.data["state"] in ("complete", "failed"):
```

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,42 @@
...
+            break
+        time.sleep(5)
+        response = client.fetch_request(response_created_req.id)
+    assert response.data["state"] == "complete"
+
+    assert response.data["packages"][0]["name"] == test_env["get"]["packages_names"]
```
lcarva, line 37: Also assert there's a single item in `packages`.

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,42 @@
...
+
+    response_dependencies_names = set()
+    for data in response.data["dependencies"]:
+        response_dependencies_names = response_dependencies_names.union({data["name"]})
```
lcarva, line 41: No need to create a new set for each dependency. Just call `response_dependencies_names.add(data["name"])`.

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,42 @@
...
+    assert response_dependencies_names == set(test_env["get"]["dependencies_names"])
```
lcarva, line 42: Note that we can also reliably expect the version of each dependency to not change. Maybe we should check for that too?

lcarva:
Travis CI is failing due to flake8 errors.

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
+# SPDX-License-Identifier: MIT
+
+import utils
+import time
```
mprahl, line 5: The imports should be in sections of stdlib, third-party, and local imports.
```suggestion
import time

import utils
```

tests/integration/example_test_env_vars.yaml
```diff
@@ -11,3 +11,15 @@ post:
   ref: e1be527f39ec31323f0454f7d1422c6260b00580
   # Package managers
   pkg_managers: ["gomod"]
+# Test data for the Valid data in request test
+get:
```
mprahl, line 15: Why did you decided to make this test data configurable? It seems to me that these are static values and could be set in the tests themselves.

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,45 @@
...
+            "repo": test_env["get"]["repo"],
+            "ref": test_env["get"]["ref"],
+            "pkg_managers": test_env["get"]["pkg_managers"],
+        },
+    )
+
+    response = client.fetch_request(response_created_req.id)
+    assert response.status == 200
+
+    while True:
```
mprahl, line 33: Why not set the while loop condition to be `while response.data["state"] == "in_progress":`? Then you could remove your if branch below that breaks out of the loop.

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,45 @@
...
+        if response.data["state"] == ("complete" or "failed"):
+            break
+        time.sleep(5)
+        response = client.fetch_request(response_created_req.id)
+    assert response.data["state"] == "complete"
+
+    assert response.data["packages"][0]["name"] == test_env["get"]["packages_names"]
```
mprahl, line 40: Could you also verify the package version?

tests/integration/example_test_env_vars.yaml
lcarva: The idea is that we can use different test data without having to update the code. It seems pretty powerful to me.

tests/integration/example_test_env_vars.yaml
ejegrova: Since data["packages"] is a list, I'll turn it into a list.

tests/integration/test_valid_data_in_request.py
ejegrova: I'm not entirely sure what to do with this - @lcarva mentioned:
> Let's build some sort of timeout so this doesn't run forever if there's a bug in Cachito.

I like the approach with the suggested loop condition but is there a possibility that the state of request will stay in `in_progress` forever? In that case I could add a condition with timeout inside the loop. What do you think?

lcarva: You could do both:

```python
state = 'in_progress'
while state == 'in_progress':
  if timeout_expired:
    raise error
  response = client.fetch_request(response_created_req.id)
  state = response.data["state"]
```

It's a little awkward to initialize the state but I guess that's ok. Once we switch to python 3.8 we could use the [walrus operator](https://www.python.org/dev/peps/pep-0572/) and remove the state initialization.

I just want a timeout for now :)

lcarva: @ejegrova pointed out to me that a few lines above, we already fetch the response and have the status. So it should be possible to just use the response directly in the `while` expression: `while response.data["state"] == 'in_progress':`

ejegrova:
@mprahl @lcarva I made some changes, ptal :)

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,65 @@
...
+    assert len(response_dependencies) == len(expected_dependencies)
+    assert response_dependencies == expected_dependencies
+
+    response_packages = list_of_dict_to_list_of_tuples(response.data["packages"])
+    expected_packages = [tuple(l) for l in test_env["get"]["packages"]]
+    assert len(response_packages) == len(expected_packages)
+    assert response_packages == expected_packages
+
+
+def list_of_dict_to_list_of_tuples(data):
```
lcarva, line 53: This is neat :)

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,65 @@
...
+    Validate data in the request.
+
+    Process:
+    Send new request to the Cachito API
+    Send request to check status of existing request
+
+    Checks:
+    * Check that response code is 200
+    * Check in the response that state is complete
+    * Check that “packages” and “dependencies” keys have appropriate values
```
mprahl, line 20: For some reason, the quotes look off.
```suggestion
    * Check that "packages" and "dependencies" keys have appropriate values
```

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def list_of_dict_to_list_of_tuples(data):
+    """
+    Convert the list of dictionaries to the list of tuples.
+
+    :param data: list of dictionaries containing keys name, type and version
+    :return: list of tuples with values name, type and version in this order
+    """
+    sorted_packages = sorted(data, key=lambda x: x["name"])
```
mprahl, line 60: Optional: It's preferred to use [operator.itemgetter](https://docs.python.org/3/library/operator.html#operator.itemgetter) instead of created an anonymous function.

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,65 @@
...
+            "ref": test_env["get"]["ref"],
+            "pkg_managers": test_env["get"]["pkg_managers"],
+        },
+    )
+
+    response = client.fetch_request(response_created_req.id)
+    assert response.status == 200
+
+    start_time = time.time()
+    while response.data["state"] == "in_progress":
```
mprahl, line 35: Optional: It'd be nice to have this loop in a separate function so it can be reused for future tests.

tests/integration/test_valid_data_in_request.py
mprahl, line 53: This function name insinuates that it does this for any list of dictionaries but it's actually only for dictionaries with the keys name, type, and version. Could you please rename this to make it clear?

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,65 @@
...
+
+def list_of_dict_to_list_of_tuples(data):
+    """
+    Convert the list of dictionaries to the list of tuples.
+
+    :param data: list of dictionaries containing keys name, type and version
+    :return: list of tuples with values name, type and version in this order
+    """
+    sorted_packages = sorted(data, key=lambda x: x["name"])
+    response_names = [item["name"] for item in sorted_packages]
```
mprahl, line 61: This function is clever but not efficient. You are iterating through the the original list of dictionaries three times and then the zip iterates through the three new lists once.

Why not just do the following?
```python
rv = []
for package in sorted(data, key=lambda x: x["name"]):
    rv.append((package["name"], package["type"], package["version"]))

return rv
```

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,65 @@
...
+            "pkg_managers": test_env["get"]["pkg_managers"],
+        },
+    )
+
+    response = client.fetch_request(response_created_req.id)
+    assert response.status == 200
+
+    start_time = time.time()
+    while response.data["state"] == "in_progress":
+        if time.time() - start_time >= 300:
```
mprahl, line 36: Please add a comment saying this is 300 seconds.

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,65 @@
+# -*- coding: utf-8 -*-
+# SPDX-License-Identifier: MIT
```
mprahl, line 2: The license we use in Cachito is GPL 3+

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,65 @@
+# -*- coding: utf-8 -*-
```
mprahl, line 1: Please remove this line since we don't use this elsewhere.

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,65 @@
...
+    start_time = time.time()
+    while response.data["state"] == "in_progress":
+        if time.time() - start_time >= 300:
+            raise TimeoutError
+        time.sleep(5)
+        response = client.fetch_request(response_created_req.id)
+    assert response.data["state"] == "complete"
+
+    response_dependencies = list_of_dict_to_list_of_tuples(response.data["dependencies"])
+    expected_dependencies = [tuple(l) for l in test_env["get"]["dependencies"]]
```
mprahl, line 43: Why not make `response_dependencies` be a list of lists rather than a list of tuples to avoid having to convert `test_env["get"]["dependencies"]`?

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,65 @@
...
+    assert len(response_dependencies) == len(expected_dependencies)
```
mprahl, line 44: What is the purpose of this assertion since you are checking the actual values below?

tests/integration/test_valid_data_in_request.py
ejegrova: I changed that in a different way, what do you think about that? Is it ok?

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,63 @@
...
+
+    response_packages = list_of_dict_to_list_of_name_type_version(response.data["packages"])
+    expected_packages = test_env["get"]["packages"]
+    assert len(response_packages) == len(expected_packages)
+    assert response_packages == expected_packages
+
+
+def list_of_dict_to_list_of_name_type_version(data):
+    """
+    Convert the list of dictionaries to the list of tuples.
```
mprahl, line 56:
```suggestion
    Convert the list of dictionaries to a list of lists from the keys name, type, and version.
```

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,63 @@
...
+
+    :param data: list of dictionaries containing keys name, type and version
+    :return: list of tuples with values name, type and version in this order
```
mprahl, line 59:
```suggestion
    :return: list of lists with values name, type and version in this order
```

tests/integration/test_valid_data_in_request.py
mprahl: @ejegrova could you please address this comment? Either explaining why you have this check or removing this check would be sufficient to resolve this comment.

ejegrova: Oh, forgot this one. It's not needed now, I'll remove it.

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+
+    response = client.fetch_request(response_created_req.id)
+    assert response.status == 200
+
+    start_time = time.time()
+    while response.data["state"] == "in_progress":
+        # 300 seconds
+        if time.time() - start_time >= 300:
+            raise TimeoutError
```
mprahl, line 38:
```suggestion
            raise TimeoutError("The Cachito request did not complete within 5 minutes")
```

tests/integration/test_valid_data_in_request.py
```diff
@@ -0,0 +1,63 @@
...
+    assert response.data["state"] == "complete"
+
+    response_dependencies = list_of_dict_to_list_of_name_type_version(response.data["dependencies"])
+    expected_dependencies = test_env["get"]["dependencies"]
+    assert len(response_dependencies) == len(expected_dependencies)
+    assert response_dependencies == expected_dependencies
+
+    response_packages = list_of_dict_to_list_of_name_type_version(response.data["packages"])
+    expected_packages = test_env["get"]["packages"]
+    assert len(response_packages) == len(expected_packages)
```
mprahl, line 50: Same as above

Issue #126, opened 2020-04-02
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/cachito-archives/bundles/1.tar.gz'


lcarva:
On a clean local environment, submitting the example request from `README.md` fails.

```
🐚  git rev-parse HEAD
8672b8b04af92cf761d4d256559c50c3bf81738f

🐚  docker-compose down -v
Removing network cachito_default
WARNING: Network cachito_default not found.
Removing volume cachito_nexus-volume
WARNING: Volume cachito_nexus-volume not found.
lucarval@localhost ~/src/exd/osbs-guild/cachito 

🐚  docker-compose up
...

🐚  curl -X POST -H "Content-Type: application/json" http://localhost:8080/api/v1/requests -d \
>         '{
>            "repo": "https://github.com/release-engineering/retrodep.git",
>            "ref": "e1be527f39ec31323f0454f7d1422c6260b00580",
>            "pkg_managers": ["gomod"]
>          }'
{
  "dependencies": [], 
  "environment_variables": {}, 
  "flags": [], 
  "id": 1, 
  "packages": [], 
  "pkg_managers": [
    "gomod"
  ], 
  "ref": "e1be527f39ec31323f0454f7d1422c6260b00580", 
  "repo": "https://github.com/release-engineering/retrodep.git", 
  "state": "in_progress", 
  "state_history": [
    {
      "state": "in_progress", 
      "state_reason": "The request was initiated", 
      "updated": "2020-04-02T17:18:03.671757"
    }
  ], 
  "state_reason": "The request was initiated", 
  "submitted_by": null, 
  "updated": "2020-04-02T17:18:03.671757", 
  "user": null
}
```
The request is marked as failed. The cachito-worker logs contain the following traceback:
```
cachito-worker_1  | [2020-04-02 17:18:03,739: INFO/ForkPoolWorker-1] Fetching the source from "https://github.com/release-engineering/retrodep.git" at reference "e1be527f39ec31323f0454f7d1422c6260b00580"
cachito-worker_1  | [2020-04-02 17:18:03,792: INFO/ForkPoolWorker-1] Setting the state of request 1 to "in_progress" with the reason "Fetching the application source"
cachito-worker_1  | [2020-04-02 17:18:03,826: DEBUG/ForkPoolWorker-1] Parsed the repository name "release-engineering/retrodep" from https://github.com/release-engineering/retrodep.git
[...]

PR #129, opened 2020-04-09
Fix incorrect archive path when extract fetched source to request bun…

tkdchen:
…dle dir


tkdchen:
Fixes #126

PR #130, opened 2020-04-09
Add support for the npm package manager

mprahl:
For now, only dependencies from the npm registry are supported. Any dependencies
in the package lock file that is directly from a URL, GitHub, Git repo, or
etc. is not supported.

Please note that this depends on #127 and thus, the first four commits in this PR are from #127.


cachito/workers/pkg_managers/general_js.py
```diff
@@ -0,0 +1,272 @@
...
+    Get the username that has read access on the NPM proxy repository for the request.
+
+    :param int request_id: the ID of the request this repository is for
+    :return: the username
+    :rtype: str
+    """
+    return f"cachito-js-{request_id}"
+
+
+def get_npm_privileged_token(request_id):
```
tkdchen, line 197: What's the intention to make this simple function that just wraps `get_npm_token`?

mprahl: @tkdchen it was just for readability. I can remove it if you'd like.

tkdchen: Not necessary. :) It makes sense.

cachito/workers/pkg_managers/npm.py
```diff
@@ -0,0 +1,120 @@
...
+    :param dict request: the Cachito request this is for
+    :return: a tuple of the npm package itself and the list of dictionaries representing the
+        dependencies
+    :rtype: (dict, list)
+    :raises CachitoError: if fetching the dependencies fails
+    """
+    # npm-shrinkwrap.json and package-lock.json share the same format but serve slightly
+    # different purposes. See the following documentation for more information:
+    # https://docs.npmjs.com/files/package-lock.json.
+    for lock_file in ("npm-shrinkwrap.json", "package-lock.json"):
```
lcarva, line 102: If neither is found, we'd get an obscure `NameError`. Can we make the error nicer?

cachito/workers/pkg_managers/npm.py
```diff
@@ -0,0 +1,120 @@
...
+        package_lock_path = os.path.join(app_source_path, lock_file)
+        if os.path.exists(package_lock_path):
+            break
+
+    try:
+        package, deps = get_package_and_deps(package_lock_path)
+    except KeyError:
```
lcarva, line 109: Logging the actual error would help debug issues.

cachito/workers/tasks/npm.py
```diff
@@ -0,0 +1,97 @@
...
+@app.task
+def fetch_npm_source(request_id, auto_detect=False):
+    """
+    Resolve and fetch npm dependencies for a given request.
+
+    :param int request_id: the Cachito request ID this is for
+    :param bool auto_detect: automatically detect if the archive uses npm
+    :raise CachitoError: if the task fails
+    """
+    validate_nexus_config()
```
lcarva, line 50: Let's do this at startup time instead.

cachito/workers/pkg_managers/general_js.py
```diff
@@ -0,0 +1,273 @@
...
+    """
+    conf = get_worker_config()
+    return get_npm_token(request_id, conf.cachito_nexus_username, conf.cachito_nexus_password)
+
+
+def get_npm_token(request_id, username, password):
+    """
+    Get an npm token for the Nexus npm proxy repository for the request.
+
+    This is an alternative to calling `npm login`, which doesn't have the ability to accept the
```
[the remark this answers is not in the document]
mprahl: The token does not expire from what I can tell.

I'll look into trying this approach out instead to avoid this function:
https://help.sonatype.com/repomanager3/formats/npm-registry#npmRegistry-AuthenticationUsingBasicAuth

cachito/workers/pkg_managers/general_js.py
```diff
@@ -0,0 +1,273 @@
...
+    Prepare Nexus so that Cachito can stage JavaScript content.
+
+    :param int request_id: the ID of the request that Nexus should be configured for
+    :raise CachitoError: if the script execution fails
+    """
+    config = get_worker_config()
+    payload = {
+        "repository_name": get_js_proxy_repo_name(request_id),
+        "http_password": config.cachito_nexus_unprivileged_password,
+        "http_username": config.cachito_nexus_unprivileged_username,
```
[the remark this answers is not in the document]
mprahl: This specifies the user that the new NPM proxy should use to authenticate to the `cachito-js` Nexus repository group. So no, you don't need admin privileges for that.

cachito/workers/tasks/npm.py
mprahl: As discussed, there's no clean way to determine what queues the Celery worker is listening on since the configuration can be overridden with the `-Q` option.

cachito/workers/pkg_managers/npm.py
```diff
@@ -0,0 +1,120 @@
...
+    a list. If it were a list, there'd be a time complexity of O(n) every time a dependency is to be
+    inserted.
+
+    :param dict package_lock_deps: the value of a "dependencies" key in a package-lock.json file
+    :param dict name_to_deps: the current mapping of dependencies
+    :raise CachitoError: if the lock file contains a dependency not from the registry
+    """
+    for name, info in package_lock_deps.items():
+        # Note that a bundled dependencies will not have the "resolved" key, but those are supported
+        # since they are properly cached in the parent dependency in Nexus
```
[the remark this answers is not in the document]
mprahl: If a dependency has bundled dependencies, they will not have the "resolved" key set. This is okay since they are included in the dependency that bundled them.

I'll clarify the comment.

cachito/workers/pkg_managers/npm.py
mprahl: Sure. I was relying on the check in the calling function of `fetch_npm_source`, but you're right, it's best to be explicit.

cachito/workers/config.py
```diff
@@ -131,6 +150,26 @@ def validate_celery_config(conf, **kwargs):
         raise ConfigError('The configuration "cachito_api_url" must be set')
 
 
+def validate_nexus_config():
```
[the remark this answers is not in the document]
mprahl: As discussed, there's no clean way to determine what queues the Celery worker is listening on since the configuration can be overridden with the `-Q` option.

mprahl:

tests/test_api_v1.py
```diff
@@ -512,6 +541,20 @@ def test_create_request_connection_error(mock_chain, app, auth_env, client, db):
     assert rv.json == {"error": "Failed to connect to the broker to schedule a task"}
 
 
+def test_create_request_unsupported_pkg_manager(app, auth_env, client, db):
```
lcarva, line 544: "unsupported" is misleading. It's support, just not enabled. Will there be a different error message if an unsupported package manager is provided?

tests/test_api_v1.py
```diff
@@ -612,15 +655,28 @@ def test_set_pkg_managers(app, client, db, worker_auth_env):
     assert get_rv.json["pkg_managers"] == ["gomod"]
 
 
-@pytest.mark.parametrize("bundle_exists", (True, False))
+@pytest.mark.parametrize(
+    "pkg_managers, bundle_exists", ((["gomod"], True), (["gomod"], False), (["npm"], True))
```
lcarva, line 659: Do we need the following scenarios?
* `['npm'], False`
* `['golang', 'npm'], False`
* `['golang', 'npm'], True`

cachito/workers/pkg_managers/general_js.py
```diff
@@ -0,0 +1,222 @@
...
+        fetch-retries=5
+        fetch-retry-factor=2
+        """
+    )
+    nexus_ca_cert = nexus.get_ca_cert()
+    if nexus_ca_cert:
+        npm_rc = textwrap.dedent(
+            f"""\
+            {npm_rc}
+            ca="{nexus_ca_cert}"
```
mprahl, line 158: As discussed over chat with @lcarva, I'll file another PR after this is merged to split the CA contents to a separate configuration file entry to avoid data duplication.

tests/test_workers/test_pkg_managers/test_npm.py
```diff
@@ -0,0 +1,214 @@
...
+def test_resolve_npm(mock_dd, mock_gpad, mock_exists, shrink_wrap, package_lock, package_and_deps):
+    mock_exists.side_effect = [shrink_wrap, package_lock]
+    mock_gpad.return_value = package_and_deps
+
+    expected_package, expected_deps = package_and_deps
+
+    src_path = "/tmp/cachito-bundles/temp/1/app"
+    package, deps = npm.resolve_npm(src_path, {"id": 1})
+
+    # Note that the deps returned by the get_package_and_deps function were modified as part of the
```
lcarva, line 178: Do we need to make a deep copy of  `expected_package` and `expected_deps` to ensure we're comparing the right thing?

tests/test_api_v1.py
mprahl: Good point. A different error does occur for a package manager that is not supported. This occurs in `cachito.web.models.py::PackageManager.get_pkg_managers`.

I'll rename this test.

tests/test_api_v1.py
mprahl: I'll use two parameterize decorators to cover all scenarios.

mprahl:
Thanks for the review @lcarva. I addressed your comments. Could you please take another look? In the meantime, I'll review it one more time myself.

mprahl:
@lcarva I reviewed the code again (not the tests) and just modified a few comments and added some.

PR #131, opened 2020-04-10
Ensure go.sum update-to-date to include checksum of replaced dep's so…

tkdchen:
…urce

### DIFF
--- a/cachito/workers/pkg_managers/golang.py
+++ b/cachito/workers/pkg_managers/golang.py
@@ -83,6 +83,8 @@ def resolve_gomod(app_source_path, request, dep_replacements=None):
 
         log.info("Downloading the gomod dependencies")
         run_gomod_cmd(("go", "mod", "download"), run_params)
+        if dep_replacements:
+            run_gomod_cmd(("go", "mod", "tidy"), run_params)
         go_list_output = run_gomod_cmd(
             ("go", "list", "-m", "-f", "{{.Path}} {{.Version}} {{.Replace}}", "all"), run_params
         )
--- a/tests/test_workers/test_pkg_managers/test_golang.py
+++ b/tests/test_workers/test_pkg_managers/test_golang.py
@@ -84,13 +84,12 @@ def test_resolve_gomod(
     run_side_effects = []
     if dep_replacement:
         run_side_effects.append(mock.Mock(returncode=0, stdout=None))  # go mod edit -replace
-    run_side_effects.extend(
-        [
-            mock.Mock(returncode=0, stdout=None),  # go mod download
-            mock.Mock(returncode=0, stdout=mock_cmd_output),  # go list -m all
-        ]
-    )
+    run_side_effects.append(mock.Mock(returncode=0, stdout=None))  # go mod download
+    if dep_replacement:
+        run_side_effects.append(mock.Mock(returncode=0, stdout=None))  # go mod tidy
+    run_side_effects.append(mock.Mock(returncode=0, stdout=mock_cmd_output))  # go list -m all
     mock_run.side_effect = run_side_effects
+
     mock_golang_version.return_value = "v2.1.1"
 
     archive_path = "/this/is/path/to/archive.tar.gz"
@@ -113,6 +112,8 @@ def test_resolve_gomod(
             "-replace",
             expected_replace,
         )
+        if dep_replacement:
+            assert mock_run.call_args_list[2][0][0] == ("go", "mod", "tidy")
 
     assert module == sample_package
     assert resolved_deps == expected_deps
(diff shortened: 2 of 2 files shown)


tkdchen:
Resolves CLOUDBLD-184

tkdchen:
@athos-ribeiro `go.mod` is changed after `go mod edit -replace`. I'm a newbie of golang, I can't figure out what issue would happen if `go.mod` is changed. Could you please elaborate?

BTW, after `go mod edit -replace` and `go mod tidy`, both `go.mod` and `go.sum` are changed, I have no idea why don't need to save them and how the build could work as expected without saving those changes. Could you help to explain? Thanks.

athos-ribeiro:
`go mod tidy` checks if the `go.mod` file information matches the module source code. It adds missing dependencies to the `go.mod` file and removes the unused ones (if any). Then it fixes go.sum to match the (possibly new) state of go.mod.

If anything is changed in `go.mod` at this point, it means that additional unnecessary dependencies were fetched or that cachito failed to provide required dependencies due to bogus information in the `go.mod` file provided by the upstream project.

Thus, it may make sense to always run `go mod tidy`. We could either use a reactive approach, where we run it after downloading the dependencies and we get the cachito request to fail due to bogus upstream `go.mod` file; or a proactive approach, where we run it before downloading the dependencies and if the `go.mod` file is changed, we just throw a warning to let the user know the upstream project file is incorrect. If we go for the latter approach, we would have to run `go mod tidy` twice to ensure the `go.sum` file is also correct.

WDYT?

Cc @mprahl @lcarva

lcarva:
> `go mod tidy` checks if the `go.mod` file information matches the module source code. It adds missing dependencies to the `go.mod` file and removes the unused ones (if any). Then it fixes go.sum to match the (possibly new) state of go.mod.
> 
> If anything is changed in `go.mod` at this point, it means that additional unnecessary dependencies were fetched or that cachito failed to provide required dependencies due to bogus information in the `go.mod` file provided by the upstream project.

Good catch. If there are additional unnecessary  dependencies then there would be additional unexpected modifications in `go.mod`.  I don't think the second case is valid. If there are bogus references in `go.mod`, then `go mod download` should fail, no?

> 
> Thus, it may make sense to always run `go mod tidy`. We could either use a reactive approach, where we run it after downloading the dependencies and we get the cachito request to fail due to bogus upstream `go.mod` file; or a proactive approach, where we run it before downloading the dependencies and if the `go.mod` file is changed, we just throw a warning to let the user know the upstream project file is incorrect. If we go for the latter approach, we would have to run `go mod tidy` twice to ensure the `go.sum` file is also correct.

Having inaccurate information is a problem, so we should address this. Let's file a separate issue for always running `go mod tidy`. I'm not sure what's the best course of action to take if unnecessary dependencies are removed from `go.mod`. Failing the request seems straight forward. Otherwise, we'd also have to remove the unnecessary dependencies from the source bundle of the request. This shouldn't be part of this PR.

> 
> WDYT?

I think for this PR, we should simply run `go mod tidy` after replace.

1. Run `go mod edit -replace ...`
1. Run `go mod download`
1. (NEW) Run `go mod tidy`

Then in a follow up story, we can sort out what to do if there are unnecessary dependencies in `go.mod`. This may turn out to be extremely tricky in the case where there are unnecessary dependencies and `go mod edit -replace` was used.

> 
> Cc @mprahl @lcarva

mprahl:
It looks like the tests passed in TravisCI but for some reason, TravisCI didn't mark the result on the PR.

athos-ribeiro:
> Good catch. If there are additional unnecessary dependencies then there would be additional unexpected modifications in `go.mod`. I don't think the second case is valid. If there are bogus references in `go.mod`, then `go mod download` should fail, no?

Here I should have said that the file itself is bogus. I was thinking about the case where a dependency is missing in `go.mod` before running `go mod download`. The download command will succeed in this case. It seems that In both cases cachito will provide the sources. Either with extra dependencies (which will may never be noticed) or with missing dependencies (which should be noticed during build time).

> Having inaccurate information is a problem, so we should address this. Let's file a separate issue for always running `go mod tidy`. I'm not sure what's the best course of action to take if unnecessary dependencies are removed from `go.mod`. Failing the request seems straight forward. Otherwise, we'd also have to remove the unnecessary dependencies from the source bundle of the request. This shouldn't be part of this PR.

That's right. This PR does the job of fixing the `go.sum` in case of replacements.

tkdchen:
@athos-ribeiro What is the difference between `unused ones` and `unnecessary dependencies` mentioned in your comment?

athos-ribeiro:
I may have used them to refer to the same thing. An unused reference in `go.mod` would result in a an unnecessary dependency shipped in the final cachito bundle.

PR #132, opened 2020-04-10
Use shutil.unpack_archive to extract app source

tkdchen:

This proposes an alternative way to simplify the code to extract app source to request bundle directory.


cachito/workers/utils.py
```diff
@@ -1,17 +1 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
```
athos-ribeiro, line 1: Any reasons for not removing the entire module?

tkdchen: Just thought `utils.py` might be also used to include other utility code in the future.

lcarva: It's easy enough to add it back in. I'd prefer if we removed it.

tkdchen: Done. The empty files are removed.

PR #135, opened 2020-04-17
Add integration test test_check_all_requests

ejegrova:
Implement the integration test for verifying the filtering parameters in
requests.

### DIFF
--- a/tests/integration/test_check_all_requests.py
+++ b/tests/integration/test_check_all_requests.py
@@ -1,0 +1,34 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import utils
+
+
+def test_check_all_requests(test_env):
+    """
+    Verify that the filtering parameters are working properly.
+
+    Process:
+    * Send request to check all requests in the Cachito API
+    * Send requests to check all requests with filtering parameters
+
+    Checks:
+    * Check that response code is 200
+    * Check that filtering parameters are working properly (page, per_page, state and verbose)
+    """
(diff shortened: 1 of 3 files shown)


tests/integration/test_check_all_requests.py
```diff
@@ -0,0 +1,34 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import utils
+
+
+def test_check_all_requests(test_env):
```
lcarva, line 6: I think we need to verify that `first`, `last`, `next`, and `previous` are working as expected.

Consider something like this:

* Get all requests with `per_page` = 15
* Verify there are at most 15 items in the response
* While there's a `next` link
  * Use `next` link to get the next set of requests
  * Verify there are at most 15 items in the response
* Verify `last` link matches the last seen `next` link
* While there's a `previous` link
  * Use `previous` link to get the next set of requests
  * Verify there are at most 15 items in the response
* Verify `first` link matches the last seen `first` link

For every request, `meta.total`, `meta.first`, and `meta.last` should not change. 
If it does, it implies there's a bug or the expected state has changed.

We should do the same check with no `state` filter, and with some values of the `state` filter. let's break this up into different tests, e.g. `test_check_all_requests`, `test_check_all_completed_requests`, `test_check_all_failed_requests`, etc.

mprahl:
@ejegrova I will close this PR to keep the PR backlog clean. Feel free to reopen it if you end up pursuing this again.

PR #139, opened 2020-04-23
Response forbidden when requester is not a worker

### LINKED ISSUE #117: Raise Forbidden instead of Unauthorized in the PATCH API endpoint
The PATH API endpoint currently raises `Unauthorized` (401 status code) if the user is not a worker. This should be raising `Forbidden` (403 status code).

tkdchen:
This change affects two endpoints, 1) patch a request and 2) add
configuration files to a specific request.

Fixes #117

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -5,7 +5,7 @@ import functools
 from celery import chain
 import flask
 from flask_login import current_user, login_required
-from werkzeug.exceptions import Unauthorized, InternalServerError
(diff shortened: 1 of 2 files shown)


athos-ribeiro:
LGTM

mprahl:
> It would be nice to add these to the API yaml spec.

It is partially there but it could use a little love:
https://github.com/release-engineering/cachito/blob/9f22b1f14efd75524b0f9fcacf362568e2fc020c/cachito/web/static/api_v1.yaml#L106-L115

Issue #116, opened 2020-03-30
Use SQLAlchemy relationships instead of the foreign keys directly


mprahl:
In models.py, there are a few places where the foreign key is directly used instead of a relationship. We should use a relationship when possible.

For example:
https://github.com/release-engineering/cachito/blob/b3a7c5a256f450cd16f419dbf4e420690ed8f533/cachito/web/models.py#L540-L543

This is another instance:
https://github.com/release-engineering/cachito/blob/b3a7c5a256f450cd16f419dbf4e420690ed8f533/cachito/web/models.py#L574

Investigate if its worth using an [alternate join condition](https://docs.sqlalchemy.org/en/13/orm/join_conditions.html?highlight=boston#specifying-alternate-join-conditions) here instead of this method:
https://github.com/release-engineering/cachito/blob/b3a7c5a256f450cd16f419dbf4e420690ed8f533/cachito/web/models.py#L387-L391

This might be another potential area for improvement (needs investigation):
https://github.com/release-engineering/cachito/blob/b3a7c5a256f450cd16f419dbf4e420690ed8f533/cachito/web/models.py#L312-L349

This might be also another potential area for improvement (needs investigation):
https://github.com/release-engineering/cachito/blob/b3a7c5a256f450cd16f419dbf4e420690ed8f533/cachito/web/models.py#L351-L377

tkdchen:
@mprahl I makde some changes to the code for this issue, however I can't figure out an improvement to `dependencies_count` and `packages_count` except changing the filter criterion like `RequestDependency.request_id == self.id` based on a relationship property. Any idea? What is your thought on the potential area for improvement originally?

mprahl:
> @mprahl I makde some changes to the code for this issue, however I can't figure out an improvement to `dependencies_count` and `packages_count` except changing the filter criterion like `RequestDependency.request_id == self.id` based on a relationship property. Any idea? What is your thought on the potential area for improvement originally?

@tkdchen I think it'd be cleaner to use a relationship instead of the foreign keys directly in the filter as mentioned, but it's fine to keep it as is. I'll leave the decision up to you if you'd like to change it.

PR #142, opened 2020-04-26
Use relationships to assign and access associated objects

tkdchen:

Relative to #116

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -553,10 +553,6 @@ class Request(db.Model):
                     )
 
                 submitted_for = User.get_or_create(submitted_for_username)
-                if not submitted_for.id:
-                    # Send the changes queued up in SQLAlchemy to the database's transaction buffer.
-                    # This will generate an ID that can be used below.
-                    db.session.flush()
                 request_kwargs["user"] = submitted_for
                 request_kwargs["submitted_by"] = current_user
             else:
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -369,16 +369,7 @@ class Request(db.Model):
         :raises ValidationError: if the dependency is already associated with the request, but
             replaced_dependency is different than what is already associated
         """
-        # If the ID is not set, then the dependency was just created and is not part of the
-        # database's transaction buffer.
-        if not dependency.id or (replaced_dependency and not replaced_dependency.id):
-            # Send the changes queued up in SQLAlchemy to the database's transaction buffer. This
-            # will genereate an ID that can be used for the mapping below.
-            db.session.flush()
-
-        mapping = RequestDependency.query.filter_by(
-            request_id=self.id, dependency_id=dependency.id
-        ).first()
+        mapping = RequestDependency.query.filter_by(request=self, dependency=dependency).first()
 
         if mapping:
             if mapping.replaced_dependency_id != getattr(replaced_dependency, "id", None):
@@ -387,11 +378,11 @@ class Request(db.Model):
                 )
             return
 
-        mapping = RequestDependency(request_id=self.id, dependency_id=dependency.id)
-        if replaced_dependency:
-            mapping.replaced_dependency_id = replaced_dependency.id
-
-        db.session.add(mapping)
+        db.session.add(
+            RequestDependency(
+                request=self, dependency=dependency, replaced_dependency=replaced_dependency
+            )
+        )
 
     @property
     def dependencies_count(self):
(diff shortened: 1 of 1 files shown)


cachito/web/models.py
```diff
@@ -590,10 +581,10 @@ def from_json(cls, kwargs):
                     # Send the changes queued up in SQLAlchemy to the database's transaction buffer.
                     # This will generate an ID that can be used below.
                     db.session.flush()
```
mprahl, line 567: I think this `db.session.flush()` is no longer needed since the code is now using the relationship. Could you please verify this?

cachito/web/models.py
```diff
@@ -590,10 +581,10 @@ def from_json(cls, kwargs):
...
-                request_kwargs["user_id"] = submitted_for.id
-                request_kwargs["submitted_by_id"] = current_user.id
+                request_kwargs["user"] = submitted_for
+                request_kwargs["submitted_by"] = current_user
             else:
-                request_kwargs["user_id"] = current_user.id
+                request_kwargs["user"] = User.query.get(current_user.id)
```
mprahl, line 571: Could you please try the following instead to avoid an extra query?
```suggestion
                request_kwargs["user"] = current_user._get_current_object()
```

https://werkzeug.palletsprojects.com/en/1.0.x/local/#werkzeug.local.LocalProxy._get_current_object

cachito/web/models.py
```diff
@@ -301,10 +301,15 @@ class RequestDependency(db.Model):
         db.Integer, db.ForeignKey("request.id"), autoincrement=False, index=True, primary_key=True
     )
     dependency_id = db.Column(
-        db.Integer, db.ForeignKey("package.id"), autoincrement=False, index=True, primary_key=True
+        db.Integer, db.ForeignKey("package.id"), autoincrement=False, index=True, primary_key=True,
```
mprahl, line 304: Could you please remove the comma you added here since it's not relevant to the change?

cachito/web/models.py
```diff
@@ -325,7 +326,7 @@ class Request(db.Model):
         db.Integer, db.ForeignKey("request_state.id"), index=True, unique=True
     )
     state = db.relationship("RequestState", foreign_keys=[request_state_id])
-    request_dependencies = db.relationship("RequestDependency")
+    # request_dependencies = db.relationship("RequestDependency")
```
mprahl, line 329: Did you mean to do this?
```suggestion
    request_dependencies = db.relationship("RequestDependency", backref="request")
```

tkdchen:
@mprahl Hi, thanks for review. All comments should be addressed. PTAL.

cachito/web/models.py
mprahl: @tkdchen were you able to verify if this is still needed?

mprahl:
@tkdchen it seems like in the last push, you made some changes in the last commit that should have been in the commit before that. Could you please try to clean up the commits?

cachito/web/models.py
tkdchen: @mprahl This is removed in bef4f16.

tkdchen:
@mprahl Hi, the commits are cleaned up. Please review again.

cachito/web/models.py
mprahl: @tkdchen This is a different instance of `db.session.flush`. Look at the diff directly and see where the comment is and it'll make more sense.

tkdchen:
Hi @mprahl, I learned you've been busy recently. When you have time, could you please review this pr again? Thanks for your time.

mprahl:
> Hi @mprahl, I learned you've been busy recently. When you have time, could you please review this pr again? Thanks for your time.

@tkdchen sorry about the confusion but I'm waiting on you addressing this comment:
https://github.com/release-engineering/cachito/pull/142#discussion_r422218366

cachito/web/models.py
tkdchen: @mprahl I removed these piece of code. What I see is SQLAlchemy flushes the changes into database automatically when using relationships.

tkdchen:
@mprahl PTAL :)

cachito/web/models.py
mprahl: @tkdchen so can you remove the `db.session.flush` after all? It's still in the diff.

Same for the one below.

tkdchen: @mprahl Oh, I think I misunderstood. My comment is for the flush call in `Request.add_dependency`.

tkdchen: The flush call in `add_state` is still needed. Removing it will cause a CircularDependencyError. The other one is removed.

PR #143, opened 2020-04-28
Add integration test test_check_downloaded_output

ejegrova:
Implement the integration test for checking the bundle download.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,9 +1,14 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import os
+from collections import namedtuple
 
 import pytest
 import yaml
+
+import utils
+
+DefaultRequest = namedtuple("DefaultRequest", "initial_response complete_response")
 
 
 @pytest.fixture(scope="session")
@@ -18,3 +23,25 @@ def test_env():
     with open(config_file) as f:
         env = yaml.safe_load(f)
     return env
+
+
+@pytest.fixture(scope="session")
+def default_request(test_env):
+    """
+    Create a new request in Cachito.
+
+    :param test_env: Test environment configuration
+    :return: a tuple that contains initial and completed response from the Cachito API
+    :rtype: tuple
+    """
+    client = utils.Client(test_env["api_url"], test_env["api_auth_type"])
+    initial_response = client.create_new_request(
+        payload={
+            "repo": test_env["package"]["repo"],
+            "ref": test_env["package"]["ref"],
+            "pkg_managers": test_env["package"]["pkg_managers"],
+        },
+    )
+    completed_response = client.wait_for_complete_request(initial_response)
+
+    return DefaultRequest(initial_response, completed_response)
--- a/tests/integration/test_check_downloaded_output.py
+++ b/tests/integration/test_check_downloaded_output.py
@@ -7,7 +7,7 @@ from pathlib import Path
 import utils
 
 
-def test_check_downloaded_output(test_env, tmpdir):
+def test_check_downloaded_output(test_env, default_request, tmpdir):
     """
     Check that the bundle has all the necessities.
 
@@ -23,18 +23,9 @@ def test_check_downloaded_output(test_env, tmpdir):
     * Check that dir app/ contains application source code
     * Check that the same full path filename is not duplicated
     """
-
+    response = default_request.complete_response
+    assert response.status == 200
     client = utils.Client(test_env["api_url"], test_env["api_auth_type"])
-    response = client.create_new_request(
-        payload={
-            "repo": test_env["download_output"]["repo"],
-            "ref": test_env["download_output"]["ref"],
-            "pkg_managers": test_env["download_output"]["pkg_managers"],
-        },
-    )
-
-    assert response.status == 201
-    response = client.wait_for_complete_request(response)
     assert response.data["state"] == "complete"
 
     file_name = tmpdir.join(f"download_{str(response.id)}")
--- a/tests/integration/test_check_downloaded_output.py
+++ b/tests/integration/test_check_downloaded_output.py
@@ -1,0 +1,76 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from os import path
+import tarfile
+from pathlib import Path
+
+import utils
+
+
+def test_check_downloaded_output(test_env, tmpdir):
+    """
+    Check that the bundle has all the necessities.
+
+    Process:
+    * Send new request to Cachito API
+    * Send request to download appropriate bundle from Cachito
+
+    Checks:
+    * Check that response code is 200
+    * Check that state is "complete"
+    * Check the downloaded data are in gzip format and valid
+    * Check that dir deps/gomod/… contains cached dependencies
+    * Check that dir app/ contains application source code
+    * Check that the same full path filename is not duplicated
+    """
+
+    client = utils.Client(test_env["api_url"], test_env["api_auth_type"])
+    response = client.create_new_request(
+        payload={
+            "repo": test_env["download_output"]["repo"],
+            "ref": test_env["download_output"]["ref"],
+            "pkg_managers": test_env["download_output"]["pkg_managers"],
+        },
+    )
+
+    assert response.status == 201
+    response = client.wait_for_complete_request(response)
+    assert response.data["state"] == "complete"
+
+    file_name = tmpdir.join(f"download_{str(response.id)}")
+    file_name_tar = tmpdir.join(f"download_{str(response.id)}.tar.gz")
+
+    resp = client.download_bundle(response.id, file_name_tar)
+    assert resp.status == 200
+    assert tarfile.is_tarfile(file_name_tar)
+
+    with tarfile.open(file_name_tar, "r:gz") as tar:
+        tar.extractall(file_name)
+
+    dependencies_path = "deps/gomod/pkg/mod/cache/download/"
+    names = [i["name"] for i in response.data["dependencies"]]
+    for dependency in names:
+        package_name = utils.escape_path_go(dependency)
+        dependency_path = path.join(file_name, dependencies_path, package_name)
+        assert path.exists(dependency_path)
+
+    assert path.exists(f"{file_name}/app/go.mod")
+    with open(f"{file_name}/app/go.mod", "r") as file:
+        module_names = []
+        for line in file:
+            if line.startswith("module "):
+                module_names.append(line.split()[-1])
+                break
+        assert module_names == [i["name"] for i in response.data["packages"]]
+
+    list_go_files = []
+    for app_path in Path(f"{file_name}/app/").rglob("*.go"):
+        list_go_files.append(app_path)
+    assert len(list_go_files) > 0
+
+    with tarfile.open(file_name_tar, mode="r:gz") as tar:
+        members = tar.getmembers()
+        path_names = set()
+        for dependency in members:
+            assert dependency.name not in path_names
+            path_names.add(dependency.name)
--- a/tests/integration/utils.py
+++ b/tests/integration/utils.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 from collections import namedtuple
+import time
 
 import requests
 from requests_kerberos import HTTPKerberosAuth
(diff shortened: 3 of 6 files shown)


tests/integration/test_check_downloaded_output.py
```diff
@@ -0,0 +1,82 @@
...
+
+import subprocess
+from os import path
+import tarfile
+from pathlib import Path
+
+import utils
+
+
+def test_check_downloaded_output(test_env):
```
mprahl, line 11: Would it be a good idea to reuse the completed request from `test_valid_data_in_request` to avoid needing to create a new request for each new test? I think this pattern would speed up the tests significantly over time. You could even use a pytest fixture that creates the request and then waits for it to complete. Once complete, it returns the request response. Then subsequent tests that use that fixture would just get the request response immediately.

I'll defer to your testing expertise on this though.

tests/integration/test_check_downloaded_output.py
```diff
@@ -0,0 +1,82 @@
...
+    assert response.status == 201
+    response = client.wait_for_complete_request(response)
+    assert response.data["state"] == "complete"
+
+    resp = client.download_bundle(response.id)
+    assert resp.status == 200
+
+    file_name = f"download_{str(response.id)}"
+    file_name_tar = f"{file_name}.tar.gz"
+    with open(file_name_tar, "wb") as file:
```
mprahl, line 45: Could you please download this to a temporary directory that gets automatically cleaned up after the test? Feel free to use the `tmpdir` pytest fixture to do this.

tests/integration/utils.py
```diff
@@ -53,3 +54,56 @@ def create_new_request(self, payload):
...
+    def download_bundle(self, request_id):
+        """
+        Download a bundle archive.
+
+        :param int request_id:  ID of the request in Cachito
+        :return: Object that contains response from the Cachito API
+        :rtype: Response
+        :raises requests.exceptions.HTTPError: if the request to the Cachito API fails
+        """
+        resp = requests.get(f"{self._cachito_api_url}/requests/{request_id}/download")
```
mprahl, line 67: These bundles can be quite large and this approach loads the bundle in memory. Could you please consider using the streaming approach as described here?
https://requests.readthedocs.io/en/latest/user/advanced/#body-content-workflow
https://stackoverflow.com/a/16696317/9706416

tests/integration/test_check_downloaded_output.py
```diff
@@ -0,0 +1,82 @@
...
+        file.write(resp.data)
+
+    process = subprocess.run(["gzip", "-t", file_name_tar])
```
mprahl, line 48: Good idea!

I'm not sure if this does the same thing, but Python has an [is_tarfile](https://docs.python.org/3/library/tarfile.html#tarfile.is_tarfile) function. If it does the same thing, it might be nice to use that instead of calling out to a command that may or may not be present on the system.

tests/integration/utils.py
```diff
@@ -53,3 +54,56 @@ def create_new_request(self, payload):
...
+            # 300 seconds
+            if time.time() - start_time >= 300:
+                raise TimeoutError("The Cachito request did not complete within 5 minutes")
+            time.sleep(5)
+            response = self.fetch_request(response.id)
+
+        return response
+
+
+def escape_path_go(dependency):
```
mprahl, line 91: Could you please explain why this is needed? This looks very interesting.

tests/integration/test_check_downloaded_output.py
```diff
@@ -0,0 +1,82 @@
...
+    names = [i["name"] for i in response.data["dependencies"]]
+    for dependency in names:
+        package_name = utils.escape_path_go(dependency)
+        dependency_path = path.join(file_name, dependencies_path, package_name)
+        assert path.exists(dependency_path)
+
+    with open(f"{file_name}/app/go.mod", "r") as file:
+        module_name = list()
+        for line in file:
+            if "module" in line:
```
mprahl, line 64: How about the following to be a little more precise?
```suggestion
            if line.startswith("module "):
```

I'm just concerned that some other line would have the text `module` in it such as a comment.

tests/integration/test_check_downloaded_output.py
```diff
@@ -0,0 +1,82 @@
...
+        tar.extractall(file_name)
+
+    dependencies_path = "deps/gomod/pkg/mod/cache/download/"
+    names = [i["name"] for i in response.data["dependencies"]]
+    for dependency in names:
+        package_name = utils.escape_path_go(dependency)
+        dependency_path = path.join(file_name, dependencies_path, package_name)
+        assert path.exists(dependency_path)
+
+    with open(f"{file_name}/app/go.mod", "r") as file:
```
mprahl, line 61: Optional: It might be good to assert that the path exists before trying to open it.

tests/integration/test_check_downloaded_output.py
```diff
@@ -0,0 +1,82 @@
...
+                module_name.append(line.split()[-1])
+                break
+        assert module_name == [i["name"] for i in response.data["packages"]]
+
+    list_go_files = list()
+    for app_path in Path(f"./{file_name}/app/").rglob("*.go"):
+        list_go_files.append(app_path)
+    assert len(list_go_files) > 0
+
+    with tarfile.open(f"{file_name_tar}", mode="r:gz") as tar:
```
mprahl, line 74:
```suggestion
    with tarfile.open(file_name_tar, mode="r:gz") as tar:
```

tests/integration/test_check_downloaded_output.py
```diff
@@ -0,0 +1,82 @@
...
+        members = tar.getmembers()
+        path_names = set()
+        for dependency in members:
+            assert dependency.name not in path_names
+            path_names.add(dependency.name)
+
+    subprocess.run(["rm", "-rf", file_name])
```
mprahl, line 81: As mentioned above, it'd be better to use the `tmpdir` pytest fixture since if the test fails, these would never get called.

tests/integration/test_check_downloaded_output.py
```diff
@@ -0,0 +1,82 @@
...
+
+import utils
+
+
+def test_check_downloaded_output(test_env):
+    """
+    Check that the bundle has all the necessities.
+
+    Process:
+    * Send new request to cachito api
```
mprahl, line 16:
```suggestion
    * Send new request to Cachito API
```

tests/integration/test_check_downloaded_output.py
```diff
@@ -0,0 +1,82 @@
...
+    * Send request to download appropriate bundle from cachito
```
mprahl, line 17:
```suggestion
    * Send request to download appropriate bundle from Cachito
```

tests/integration/test_check_downloaded_output.py
ejegrova: I didn't realise we can do it like this. Some of the scenarios will still need to create the request on their own because of the different payload (e.g. dependency replacement) but the fixture will be used for the majority of tests.

tests/integration/test_check_downloaded_output.py
ejegrova: It seems that it's doing the same thing. Changed it to use is_tarfile function.

tests/integration/test_check_downloaded_output.py
ejegrova: that's better :)

tests/integration/utils.py
ejegrova: I noticed that in the bundle the paths for dependencies changes when there is an uppercase letter. Escaping of paths is described in https://github.com/golang/mod/blob/2addee1ccfb22349ab47953a3046338e461eb4d1/module/module.go#L46
I hope it's relevant for all go modules.

ejegrova:
@mprahl  please review that

tests/integration/test_check_downloaded_output.py
```diff
@@ -0,0 +1,68 @@
...
+    * Check that state is "complete"
+    * Check the downloaded data are in gzip format and valid
+    * Check that dir deps/gomod/… contains cached dependencies
+    * Check that dir app/ contains application source code
+    * Check that the same full path filename is not duplicated
+    """
+    response = create_new_request
+    assert response.status == 201
+    client = utils.Client(test_env["api_url"], test_env["api_auth_type"])
+    response = client.wait_for_complete_request(response)
```
mprahl, line 29: Optional: It might be nice to include this in the fixture so that all tests that use the fixture don't have to duplicate this.

In essence, the fixture would guarantee that it would supply a Cachito request that is either complete or failed.

tests/integration/test_check_downloaded_output.py
```diff
@@ -0,0 +1,68 @@
...
+    dependencies_path = "deps/gomod/pkg/mod/cache/download/"
+    names = [i["name"] for i in response.data["dependencies"]]
+    for dependency in names:
+        package_name = utils.escape_path_go(dependency)
+        dependency_path = path.join(file_name, dependencies_path, package_name)
+        assert path.exists(dependency_path)
+
+    assert path.exists(f"{file_name}/app/go.mod")
+    with open(f"{file_name}/app/go.mod", "r") as file:
+        module_name = list()
```
mprahl, line 51: When you create variables that are lists, could you make the names plural (e.g. `module_names`)? It makes it easier to read.

mprahl, line 51: Also, please initialize lists with `module_names = []` and not `module_names = list()`.

tests/integration/test_check_downloaded_output.py
```diff
@@ -0,0 +1,68 @@
...
+        for line in file:
+            if line.startswith("module "):
+                module_name.append(line.split()[-1])
+                break
+        assert module_name == [i["name"] for i in response.data["packages"]]
+
+    list_go_files = list()
```
mprahl, line 58:
```suggestion
    list_go_files = []
```

tests/integration/utils.py
mprahl: Could you add this link to the docstring? It was very helpful.

tests/integration/utils.py
```diff
@@ -53,3 +54,62 @@ def create_new_request(self, payload):
...
+    Escape uppercase characters in names of Golang packages.
+
+    Replacing every uppercase letter with an exclamation mark followed by the lowercase letter.
+
+    :param str dependency: Name of the dependency
+    :return: Escaped dependency name
+    :rtype: str
+    """
+    if not dependency.islower():
+        package_name = str()
```
mprahl, line 108:
```suggestion
        package_name = ''
```

ejegrova:
@mprahl please check again :)

tests/integration/conftest.py
```diff
@@ -18,3 +20,25 @@ def test_env():
...
+    response = client.create_new_request(
+        payload={
+            "repo": test_env["package"]["repo"],
+            "ref": test_env["package"]["ref"],
+            "pkg_managers": test_env["package"]["pkg_managers"],
+        },
+    )
+    response = client.wait_for_complete_request(response)
+
+    return response
```
mprahl, line 44: Optional: You could return a tuple where the first element is the initial response and second element is completed response. This way, you could use this fixture in `test_creating_new_request`.

tests/integration/conftest.py
```diff
@@ -18,3 +20,25 @@ def test_env():
     with open(config_file) as f:
         env = yaml.safe_load(f)
     return env
+
+
+@pytest.fixture(scope="session")
+def create_new_request(test_env):
```
mprahl, line 26: Could you rename this to something that describes what is being returned? For example you could name it `default_request` or something like that.

I'm sorry I missed this in my previous review.

mprahl:
Just a couple of comments. This is looking great!

tests/integration/conftest.py
```diff
@@ -18,3 +20,25 @@ def test_env():
...
+    initial_response = client.create_new_request(
+        payload={
+            "repo": test_env["package"]["repo"],
+            "ref": test_env["package"]["ref"],
+            "pkg_managers": test_env["package"]["pkg_managers"],
+        },
+    )
+    completed_response = client.wait_for_complete_request(initial_response)
+
+    return initial_response, completed_response
```
mprahl, line 44: Sorry again for not mentioning this earlier, but perhaps a named tuple would make this easier to use:
```python
from collections import namedtuple
DefaultRequest = namedtuple('DefaultRequest', 'initial_response complete_response')
```

Then in your tests, you can use it like `default_request.initial_response` and `default_request.complete_response`

tests/integration/conftest.py
```diff
@@ -18,3 +20,25 @@ def test_env():
...
     return env
+
+
+@pytest.fixture(scope="session")
+def default_request(test_env):
+    """
+    Create a new request in Cachito.
+
+    :param test_env: Test environment configuration
+    :return: Object that contains initial and completed response from the Cachito API
```
mprahl, line 31:
```suggestion
    :return: a tuple that contains initial and completed response from the Cachito API
```

mprahl:
@ejegrova after these two comments are addressed, this is good to merge. Great work!

ejegrova:
Done the changes, @mprahl please check

PR #145, opened 2020-05-08
Mark stuck in_progress requests as stale in the cachito-cleanup script

mprahl:
This also adds a timeout to the GET requests, additional logging, and error handling.


mprahl:
@lcarva nice suggestion! Could you please take another look?

PR #146, opened 2020-05-13
Add integration test test_dependency_replacement

ejegrova:
Implement the integration test for checking the replacement of dependency
version

### DIFF
--- a/tests/integration/example_test_env_vars.yaml
+++ b/tests/integration/example_test_env_vars.yaml
@@ -33,3 +33,13 @@ get:
   - ["golang.org/x/tools", "gomod", "v0.0.0-20190325161752-5a8dccf5b48a"]
   - ["gopkg.in/check.v1", "gomod", "v1.0.0-20180628173108-788fd7840127"]
   - ["gopkg.in/yaml.v2", "gomod", "v2.2.2"]
+# Test data for the Dependency replacement test
+dep_replacement:
+  # The list of dependencies that are replaced with specific versions
+  dependency_replacements:
+  - name: github.com/pkg/errors
+    type: gomod
+    version: v0.8.0
+  - name: github.com/Masterminds/semver
+    type: gomod
+    version: v1.4.0
--- a/tests/integration/test_dependency_replacement.py
+++ b/tests/integration/test_dependency_replacement.py
@@ -1,0 +1,80 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from os import path
+import tarfile
+
+import utils
+
+
+def test_dependency_replacement(test_env, tmpdir):
+    """
+    Check that proper versions of dependencies were used.
+
+    Process:
+    * Send new request to Cachito API to fetch retrodep with another version of dependency package
+    * Download a bundle archive
+
+    Checks:
+    * Check that the state of request is complete
+    * Check that in the response there is a key "replaces" with dict values which was replaced
+    * Check that dir deps/gomod/pkg/mod/cache/download/github.com/pkg/errors/@v/… contains
+        only the required version
+    * Check that app/go.mod file has replace directive for the specified package
+    """
+    dependency_replacements = test_env["dep_replacement"]["dependency_replacements"]
+    client = utils.Client(test_env["api_url"], test_env["api_auth_type"])
+    response_created_req = client.create_new_request(
+        payload={
+            "repo": test_env["package"]["repo"],
+            "ref": test_env["package"]["ref"],
+            "pkg_managers": test_env["package"]["pkg_managers"],
+            "dependency_replacements": dependency_replacements,
+        },
+    )
+    response = client.wait_for_complete_request(response_created_req)
+    assert response.data["state"] == "complete"
+
+    names_replaced_dependencies = {
+        i["replaces"]["name"] for i in response.data["dependencies"] if i["replaces"] is not None
+    }
+    supposed_replaced_dependencies = set(i["name"] for i in dependency_replacements)
+    assert names_replaced_dependencies == supposed_replaced_dependencies
+
+    bundle_dir_name = tmpdir.join(f"download_{str(response.id)}")
+    bundle_archive = tmpdir.join(f"download_{str(response.id)}.tar.gz")
+    resp = client.download_bundle(response.id, bundle_archive)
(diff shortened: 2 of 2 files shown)


ejegrova:
@mprahl @lcarva ptal :)

tests/integration/test_dependency_replacement.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    Check that proper versions of dependencies were used.
+
+    Process:
+    * Send new request to Cachito API to fetch retrodep with another version of dependency package
+    * Download a bundle archive
+
+    Checks:
+    * Check that the state of request is complete
+    * Check that in the response there is a key “replaces” with dict values which was replaced
```
mprahl, line 19:
```suggestion
    * Check that in the response there is a key "replaces" with dict values which was replaced
```

tests/integration/example_test_env_vars.yaml
```diff
@@ -33,3 +33,9 @@ get:
   - ["golang.org/x/tools", "gomod", "v0.0.0-20190325161752-5a8dccf5b48a"]
   - ["gopkg.in/check.v1", "gomod", "v1.0.0-20180628173108-788fd7840127"]
   - ["gopkg.in/yaml.v2", "gomod", "v2.2.2"]
+# Test data for the Dependency replacement test
+dep_replacement:
+  # The list of dependencies that are replaced with specific versions
+  dependency_replacements:
+  - ["github.com/pkg/errors", "gomod", "v0.8.0"]
+  - ['github.com/Masterminds/semver', 'gomod', 'v1.4.0']
```
mprahl, line 41: Rather than modify the current structure into what Cachtio expects in the payload, how about just setting it that way from the beginning? It's also easier to understand.
```suggestion
  - name: github.com/pkg/errors
    type: gomod
    version: v0.8.0
  - name: github.com/Masterminds/semver
    type: gomod
    version: v1.4.0
```

tests/integration/test_dependency_replacement.py
```diff
@@ -0,0 +1,74 @@
...
+        },
+    )
+    response = client.wait_for_complete_request(response_created_req)
+    assert response.data["state"] == "complete"
+
+    replaced_dependencies = [
+        i["replaces"] for i in response.data["dependencies"] if i["replaces"] is not None
+    ]
+    names_replaced_dependencies = sorted([i["name"] for i in replaced_dependencies])
+    supposed_replaced_dependencies = sorted([i["name"] for i in dependency_replacements])
```
mprahl, line 43: I think this is a good use-case for sets. Also, you can avoid an extra iteration over the list by not storing `replaced_dependencies`
```suggestion
    names_replaced_dependencies = {
        i["replaces"]["name"] for i in response.data["dependencies"] if i["replaces"] is not None
    }
    supposed_replaced_dependencies = set([i["name"] for i in dependency_replacements])
```

tests/integration/test_dependency_replacement.py
```diff
@@ -0,0 +1,74 @@
...
+    assert names_replaced_dependencies == supposed_replaced_dependencies
+
+    file_name = tmpdir.join(f"download_{str(response.id)}")
```
mprahl, line 46: Rather than `file_` for these variables, how about `bundle_`?

So instead of `file_name`, it could be bundle_dir_name` and instead of `file_name_tar`, it could be `bundle_archive`.

tests/integration/test_dependency_replacement.py
```diff
@@ -0,0 +1,74 @@
...
+    assert resp.status == 200
+    assert tarfile.is_tarfile(file_name_tar)
+
+    with tarfile.open(file_name_tar, "r:gz") as tar:
+        tar.extractall(file_name)
+
+    for dependency in dependency_replacements:
+        dep_name = utils.escape_path_go(dependency["name"])
+        dependency_version_file = (
+            f"{file_name}/deps/gomod/pkg/mod/cache/download/{dep_name}/@v/list"
```
mprahl, line 58: Could you construct paths using `os.path.join` to make the tests not OS specific?

I filed https://github.com/release-engineering/cachito/pull/148 to fix a previous test.

ejegrova:
@mprahl I made the changes, can you check it again?

tests/integration/test_dependency_replacement.py
```diff
@@ -0,0 +1,81 @@
...
+            "dependency_replacements": dependency_replacements,
+        },
+    )
+    response = client.wait_for_complete_request(response_created_req)
+    assert response.data["state"] == "complete"
+
+    names_replaced_dependencies = {
+        i["replaces"]["name"] for i in response.data["dependencies"] if i["replaces"] is not None
+    }
+    supposed_replaced_dependencies = set([i["name"] for i in dependency_replacements])
```
mprahl, line 40: This is more efficient since you don't have to first create a list and then convert it to a set.
```suggestion
    supposed_replaced_dependencies = set(i["name"] for i in dependency_replacements)
```

tests/integration/test_dependency_replacement.py
```diff
@@ -0,0 +1,81 @@
...
+            "mod",
+            "cache",
+            "download",
+            dep_name,
+            "@v",
+            "list",
+        )
+
+        assert path.exists(dependency_version_file)
+        with open(dependency_version_file, "r") as file:
```
mprahl, line 68: Could you move this `open` context manager to above the for loop and store the value of `lines`? Then you don't need to keep opening and reading the file for each replacement.

mprahl:
Two minor comments. Afterwards, this is good to merge. Nice job!

tests/integration/test_dependency_replacement.py
ejegrova: It's opening different file each time (checks the version in all replaced dependencies).
For example: `deps/gomod/pkg/mod/cache/download/github.com/pkg/errors/@v/list`

ejegrova: For some dependencies, there are more versions listed in the file:
```
deps/gomod/pkg/mod/cache/download/gopkg.in/check.v1/@v/list
v0.0.0-20161208181325-20d25e280405
v1.0.0-20180628173108-788fd7840127
```

PR #148, opened 2020-05-14
Use Python to construct paths in the integration tests

mprahl:
This makes the path construction not OS specific.


mprahl:
@ejegrova could you please review?

PR #149, opened 2020-05-15
Add support for npm dependencies from private repositories

mprahl:
This documents how authentication works to private repositories and it keeps `npm pack` from prompting to trust unknown hosts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,12 +201,16 @@ registry in the event that the application needs to be built without Cachito and
 it manages.
 
 Cachito can also handle dependencies that are not from the npm registry such as those directly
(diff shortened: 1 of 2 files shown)


mprahl:
@nirzari @athos-ribeiro @lcarva what are your thoughts here? Do you think it'd be better to populate `known_hosts` instead on the Cachito worker?

lcarva:
It would be great if we could pre-populate the known_hosts via ansible to avoid having to bypass the check.

mprahl:
@nirzari @athos-ribeiro @lcarva okay, the PR is adjusted. Could you please take another look?

nirzari:
:+1:

PR #151, opened 2020-05-21
Set render_as_batch=True when generating migrations

mprahl:
Since we support SQLite, this makes generating those migrations easier.


mprahl:
@nirzari FYI

PR #155, opened 2020-05-22
Handle task scheduling failures from the REST API

mprahl:
If RabbitMQ is unavailable and a new request is submitted, Cachito will now fail the request and return an error to the user.

If RabbitMQ is unavailable when a request with the npm package manager is set to the "stale" or "failed" state, the PATCH request will still complete but a log message will state that the admin must clean this up manually.

Resolves CLOUDBLD-995.

### DIFF
--- a/cachito/errors.py
+++ b/cachito/errors.py
@@ -1,12 +1,12 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 
-class ValidationError(ValueError):
-    """An error was encountered during validation."""
-
-
 class CachitoError(RuntimeError):
     """An error was encountered in Cachito."""
+
+
+class ValidationError(CachitoError, ValueError):
+    """An error was encountered during validation."""
 
 
 class ConfigError(CachitoError):
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -5,9 +5,10 @@ import functools
 from celery import chain
 import flask
 from flask_login import current_user, login_required
+import kombu.exceptions
 from werkzeug.exceptions import Forbidden, InternalServerError
 
-from cachito.errors import ValidationError
+from cachito.errors import CachitoError, ValidationError
 from cachito.web import db
 from cachito.web.models import (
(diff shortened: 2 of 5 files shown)


mprahl:
@nirzari could you please also review?

cachito/errors.py
```diff
@@ -1,13 +1,13 @@
...
 
-class ValidationError(ValueError):
-    """An error was encountered during validation."""
-
-
 class CachitoError(RuntimeError):
     """An error was encountered in Cachito."""
 
 
+class ValidationError(CachitoError, ValueError):
```
lcarva, line 8: Interesting approach. This will match either RuntimeError, ValueError, or CachitoError. I think that's ok. We just have to be sure to catch any sort of ValidationError before catching RuntimeError.

cachito/web/api_v1.py
```diff
@@ -210,7 +211,16 @@ def create_request():
...
     )
 
-    chain(chain_tasks).delay()
+    try:
+        chain(chain_tasks).delay()
+    except kombu.exceptions.OperationalError:
+        flask.current_app.logger.exception(
+            "Failed to schedule the task for request %d. Failing the request.", request.id
+        )
+        error = "Failed to schedule the task to the workers. Please try again."
```
lcarva, line 220: Could Cachito try again?

nirzari:
This looked bit difficult to even review. Syntax looks good to me. That's all I can contribute at this point :+1:

cachito/errors.py
mprahl: In hindsight, I would have preferred that all these just inherit from just `Exception` but I didn't want to break compatibility and wanted to have all the exceptions at least inherit from `CachitoError`.

cachito/web/api_v1.py
mprahl: Celery will automatically retry for us.
https://github.com/release-engineering/cachito/blob/e1d7846bf7e1a896baa338fbb1b45880bb0c5150/cachito/workers/config.py#L18

PR #160, opened 2020-06-01
Fix the `npm pack` command execution

mprahl:
This fixes some `npm pack` failures in certain environments.

### DIFF
--- a/cachito/workers/pkg_managers/general_js.py
+++ b/cachito/workers/pkg_managers/general_js.py
@@ -60,6 +60,11 @@ def download_dependencies(request_id, deps):
             npm_rc_file, request_id, conf.cachito_nexus_username, conf.cachito_nexus_password
         )
         env = {
+            # This is set since the home directory must be determined by the HOME environment
+            # variable or by looking at the /etc/passwd file. The latter does not always work
+            # since some deployments (e.g. OpenShift) don't have an entry for the running user
+            # in /etc/passwd.
+            "HOME": os.environ.get("HOME", ""),
             "NPM_CONFIG_CACHE": os.path.join(temp_dir, "cache"),
             "NPM_CONFIG_USERCONFIG": npm_rc_file,
             "PATH": os.environ.get("PATH", ""),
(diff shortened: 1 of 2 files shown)


cachito/workers/pkg_managers/general_js.py
```diff
@@ -55,9 +55,17 @@ def download_dependencies(request_id, deps):
     conf = get_worker_config()
     with tempfile.TemporaryDirectory(prefix="cachito-") as temp_dir:
         npm_rc_file = os.path.join(temp_dir, ".npmrc")
+        if conf.cachito_nexus_ca_cert and os.path.exists(conf.cachito_nexus_ca_cert):
```
lcarva, line 58: If it's set, but the path doesn't exist, should we at least log a warning? Maybe even raise an exception?

cachito/workers/pkg_managers/general_js.py
```diff
@@ -345,6 +355,8 @@ def upload_non_registry_dependency(dep_identifier, version_suffix):
     """
     with tempfile.TemporaryDirectory(prefix="cachito-") as temp_dir:
         env = {
+            # This is set since the home directory must be writable for npm
+            "HOME": temp_dir,
```
lcarva, line 359: Why does it need access to the HOME dir?

cachito/workers/pkg_managers/general_js.py
mprahl: It has a default value so it's not unusual if it's set but the path doesn't exist:
https://github.com/release-engineering/cachito/blob/cb4d01ec113f3e48dcbc8ab4efeca22744c24cc0/cachito/workers/config.py#L26

I was just being more defensive by making sure it wasn't unset by the user before checking if the path exists.

cachito/workers/pkg_managers/general_js.py
mprahl: I added a comment explaining this and also changed this to inherit the home directory instead.

PR #161, opened 2020-06-02
Cloudbld 758 cachito tests jenkins

nirzari:
The example_test_env_vars.yaml  has been modified as it is accessed by the jenkins jobs

### DIFF
--- a/test_env_vars.yaml
+++ b/test_env_vars.yaml
@@ -1,6 +1,6 @@
 ---
 # Cachito API endpoint
-api_url: https://cachito.example.com/api/v1
+api_url: http://localhost:8080/api/v1
 # Authentication type, user either null for no authentication or kerberos
 api_auth_type: kerberos
 # Package that will be used for testing
--- a/tox.ini
+++ b/tox.ini
@@ -51,3 +51,4 @@ commands =
     pytest -rA -vvv \
         --confcutdir=tests/integration \
         {posargs:tests/integration}
+passenv = KRB5CCNAME REQUESTS_CA_BUNDLE KRB5_CLIENT_KTNAME


tests/integration/example_test_env_vars.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 # Cachito API endpoint
-api_url: https://cachito.example.com/api/v1
+api_url: https://cachito.${CACHITO_ENV_SUBURL}.com/api/v1
```
mprahl, line 3: How about just making the whole thing an environment variable that is called `CACHITO_API_URL`?

nirzari: Sure, I could do it. That might be easier. I just didn't want to change too much and went with this but I like your suggestion more.

nirzari: @mprahl  I made necessary changes in the respective MR as well. Please re-review. Resolving this.

lcarva:
I don't think `example_test_env_vars.yaml ` is named appropriately anymore. Can we make remove `example_` from it?

nirzari:
@lcarva Good catch! Fixed

test_env_vars.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 # Cachito API endpoint
-api_url: https://cachito.example.com/api/v1
```
lcarva, line 3: How's this set to a specific environment?

nirzari: It was in response for Matt's comment _I like option 2 but change the URL to point to http://localhost:8080/api/v1/ since that is the URL for the Cachito REST API in the docker-compose environment. Then you can use sed or some YAML editor to change the URL in the Jenkins job._ in https://gitlab.cee.redhat.com/exd-sp-cloud-build/cachito-tests/merge_requests/1

test_env_vars.yaml
```diff
@@ -1,6 +1,6 @@
...
+api_url: http://localhost:8080/api/v1/
```
mprahl, line 3:
```suggestion
api_url: http://localhost:8080/api/v1
```

test_env_vars.yaml
lcarva: Alright. That's fine. Instead of using `sed` to replace things, we could change the integration testing code itself to change the `api_url` based on an environment variable. Just brainstorming, no need to change anything for this PR.

PR #162, opened 2020-06-02
Update athens to v0.9.0

mprahl:


lcarva:
Is there a good way to verify the migration doesn't blow away the currently cached content?

mprahl:
@lcarva I tested the upgrade in docker-compose and the existing change wasn't changed.

PR #163, opened 2020-06-03
Remove the auto-detect feature

mprahl:
This is done to prevent requests that previously worked to fail if a new package manager is enabled. If no package manager is supplied in a request, a configurable default value is used. This configuration defaults to `["gomod"]`.

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -509,10 +509,16 @@ class Request(db.Model):
 
         # Validate package managers are correctly provided
         pkg_managers_names = request_kwargs.pop("pkg_managers", None)
-        # If no package managers are specified, then Cachito will detect them automatically
-        if pkg_managers_names:
-            pkg_managers = PackageManager.get_pkg_managers(pkg_managers_names)
-            request_kwargs["pkg_managers"] = pkg_managers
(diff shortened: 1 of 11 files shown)


athos-ribeiro:
LGTM. We should also update OSBS documentation on providing `pkg_managers` after this is merged. https://osbs.readthedocs.io/en/latest/users.html?highlight=cachito#remote-source

cachito/web/models.py
```diff
@@ -509,10 +509,16 @@ def from_json(cls, kwargs):
 
         # Validate package managers are correctly provided
         pkg_managers_names = request_kwargs.pop("pkg_managers", None)
-        # If no package managers are specified, then Cachito will detect them automatically
-        if pkg_managers_names:
-            pkg_managers = PackageManager.get_pkg_managers(pkg_managers_names)
-            request_kwargs["pkg_managers"] = pkg_managers
+        # Default to the default package managers
+        if not pkg_managers_names:
```
lcarva, line 513: What if I don't want any package managers at all?

mprahl:
@lcarva thanks for the review. Could you please take another look?

PR #164, opened 2020-06-04
Add a configuration to set default environment variables

mprahl:
These can be set for each package manager. Some defaults were added to npm to prevent certain post install scripts from reaching out to the internet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,10 @@ Custom configuration for the Celery workers are listed below:
 * `cachito_bundles_dir` - the directory for storing bundle archives which include the source archive
   and dependencies. This configuration is required, and the directory must already exist and be
   writeable.
+* `cachito_default_environment_variables` - a dictionary where the keys are names of package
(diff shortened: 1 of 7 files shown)


athos-ribeiro:
LGTM. Needs some adjustments in CI for those config expectations.

I am not familiar with JS development. Would that approach mean cachito will have to keep an updated list of env vars for different 3rd party testing tools? I can not think of a better approach on this one though...

mprahl:
> LGTM. Needs some adjustments in CI for those config expectations.
> 
> I am not familiar with JS development. Would that approach mean cachito will have to keep an updated list of env vars for different 3rd party testing tools? I can not think of a better approach on this one though...

Thanks for the review!

The idea is that Cachito would provide default environment variables for common dependencies, but it would be up to users to add their own environment variables in their Dockerfile for ones not covered by Cachito.

I just added these since these were the ones that I encountered in the ACM repos, so I thought it'd be a good default set.

PR #165, opened 2020-06-05
Add integration test test_using_cached_dependencies

ejegrova:
Implement the integration test that checks that the request sent to the Cachito API is fetched from the database if it already stored there

### DIFF
--- a/test_env_vars.yaml
+++ b/test_env_vars.yaml
@@ -43,3 +43,16 @@ dep_replacement:
   - name: github.com/Masterminds/semver
     type: gomod
     version: v1.4.0
+# Test data for the Using cached dependencies test
+cached_dependenices:
+  # The upstream git repository
+  seed_repo:
+    https_url: "https://github.com/release-engineering/retrodep.git"
+  # Repository that will be used for the test
+  test_repo:
+    # The tracked repository (remote)
+    ssh_url: "git@github.com:user/cachito_test_repo.git"
+    # The URL of the repository
+    https_url: "https://github.com/user/cachito_test_repo.git"
+    # Package managers
+    pkg_managers: ["gomod"]
--- a/tests/integration/test_using_cached_dependencies.py
+++ b/tests/integration/test_using_cached_dependencies.py
@@ -1,0 +1,80 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import random
+import string
+
+import git
+
+import utils
+
+
+def test_using_cached_dependencies(test_env, tmpdir):
+    """
+    Check that the cached dependencies are used instead of downloading them from repo again.
+
+    Preconditions:
+    * On git instance prepare an empty repository
+
+    Process:
+    * Clone the package from the upstream repository
+    * Create empty commit on new test branch and push it to prepared repository
+    * Send new request to Cachito API which would fetch data from the prepared repository
+    * Delete branch with the corresponding commit
+    * Send the same request to Cachito API
+
+    Checks:
+    * Check that the state of the first request is complete
+    * Check that the commit is not available in the repository after the branch is deleted
+    * Check that the state of the second request is complete
+    """
+    generated_suffix = "".join(
+        random.choice(string.ascii_letters + string.digits) for x in range(10)
+    )
+    branch_name = f"test-{generated_suffix}"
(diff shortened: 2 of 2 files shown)


ejegrova:
@mprahl @lcarva ptal :)

tests/integration/test_using_cached_dependencies.py
```diff
@@ -0,0 +1,69 @@
...
+            "ref": commit,
+            "pkg_managers": test_env["cached_dependenices"]["package"]["pkg_managers"],
+        },
+    )
+    response = client.wait_for_complete_request(response_created_req)
+    assert response.data["state"] == "complete"
+
+    assert repo.git.branch("-a", "--contains", commit)
+
+    repo.git.push("--delete", remote.name, branch_name)
```
lcarva, line 56: We need to be sure to always run this command to avoid cluttering the test repo.

tests/integration/test_using_cached_dependencies.py
```diff
@@ -0,0 +1,69 @@
...
+
+    response_created_req = client.create_new_request(
+        payload={
+            "repo": test_env["cached_dependenices"]["package"]["repo"],
+            "ref": commit,
+            "pkg_managers": test_env["cached_dependenices"]["package"]["pkg_managers"],
+        },
+    )
+    response = client.wait_for_complete_request(response_created_req)
+    assert response.data["state"] == "complete"
```
lcarva, line 69: It would be nice here to compare the packages, dependencies, etc between the two requests.

tests/integration/test_using_cached_dependencies.py
```diff
@@ -0,0 +1,69 @@
...
+    generated_suffix = "".join(
+        random.choice(string.ascii_letters + string.digits) for x in range(10)
+    )
+    branch_name = f"test-{generated_suffix}"
+    repo = git.repo.Repo.clone_from(test_env["cached_dependenices"]["seed_repo"], tmpdir)
+    remote = repo.create_remote("test", url=test_env["cached_dependenices"]["git_remote"])
+    assert remote.exists()
+
+    repo.create_head(branch_name).checkout()
+    repo.git.commit("--allow-empty", m="new commit")
```
lcarva, line 39: Can we make the commit message more descriptive?

test_env_vars.yaml
```diff
@@ -43,3 +43,15 @@ dep_replacement:
   - name: github.com/Masterminds/semver
     type: gomod
     version: v1.4.0
+# Test data for the Using cached dependencies test
+cached_dependenices:
+  # The URL for the upstream git repository
+  seed_repo: "https://github.com/release-engineering/retrodep.git"
+  # Package that will be used for the test
+  package:
```
lcarva, line 51: Grouping `repo` and `pkg_managers` under `package` doesn't sound right to me. How about something like:
```yaml
test_repo:
  ssh_url: git@github.com:user/cachito_test_repo.git
  https_url: https://github.com/user/cachito_test_repo.git
  pkg_managers: ["gomod"]
```

For consistency, you could use the same structure for the `seed_repo`:
```yaml
seed_repo:
  https_url: https://github.com/release-engineering/retrodep.git
```

tests/integration/test_using_cached_dependencies.py
```diff
@@ -0,0 +1,69 @@
...
+        },
+    )
+    response = client.wait_for_complete_request(response_created_req)
+    assert response.data["state"] == "complete"
+
+    assert repo.git.branch("-a", "--contains", commit)
+
+    repo.git.push("--delete", remote.name, branch_name)
+    repo.heads.master.checkout()
+    repo.git.branch("-D", branch_name)
```
lcarva, line 58: I'm wondering if we do care about cleaning up the local branch.

mprahl:
@ejegrova this looks really good. I just +1'd a few of Luiz's comments. Once addressed and the linting errors are fixed, this is good to merge from my point of view.

tests/integration/test_using_cached_dependencies.py
ejegrova: I think that we don't need to care if we run tests in container but if someone runs tests locally it should be cleaned up. So I would rather leave it there, what do you think?

ejegrova:
@lcarva can you check it again please?

PR #166, opened 2020-06-08
Allow workers to authenticate with SSL certificate

athos-ribeiro:
This can be used as an alternative to completely disabling
authentication requirements for environments without kerberos support

* See https://requests.readthedocs.io/en/master/user/advanced/#client-side-certificates

### DIFF
--- a/cachito/workers/config.py
+++ b/cachito/workers/config.py
@@ -171,6 +171,9 @@ def validate_celery_config(conf, **kwargs):
             "the other must also be set"
         )
 
+    if conf.get("cachito_auth_type") == "cert" and conf.get("cachito_auth_cert") is None:
+        raise ConfigError('cachito_auth_cert configuration must be set for "cert" authentication')
+
     if not isinstance(conf.get("cachito_default_environment_variables"), dict):
         raise ConfigError(
             'The configuration "cachito_default_environment_variables" must be a dictionary'
--- a/cachito/workers/requests.py
+++ b/cachito/workers/requests.py
@@ -1,9 +1,13 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
+import logging
+
 import requests
 from requests.packages.urllib3.util.retry import Retry
 import requests_kerberos
 
 from cachito.workers.config import get_worker_config
+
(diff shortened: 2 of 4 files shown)


cachito/workers/requests.py
```diff
@@ -16,10 +21,17 @@ def get_requests_session(auth=False):
...
-        )
+    if auth:
+        if config.cachito_auth_type == "kerberos":
+            session.auth = requests_kerberos.HTTPKerberosAuth(
+                mutual_authentication=requests_kerberos.OPTIONAL
+            )
+        elif config.cachito_auth_type == "cert":
+            try:
+                session.cert = config.cachito_auth_cert
+            except AttributeError:
```
mprahl, line 32: This isn't needed since you validate that it's always set in `validate_celery_config` right?

cachito/workers/requests.py
```diff
@@ -1,10 +1,15 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
+
```
mprahl, line 2:
```suggestion
```

cachito/workers/config.py
```diff
@@ -158,6 +158,9 @@ def validate_celery_config(conf, **kwargs):
             "the other must also be set"
         )
 
+    if conf.get("cachito_auth_type") == "cert" and conf.get("cachito_auth_cert") is None:
+        raise ConfigError("cachito_auth_cert configuration must be set for 'cert' authentication")
```
mprahl, line 162:
```suggestion
        raise ConfigError('The configuration "cachito_auth_cert" must be set for "cert" authentication')
```

cachito/workers/requests.py
athos-ribeiro: Right. I added it there just to be safe, in case the validation changes or is skipped for some reason in the future. Removed!

mprahl:
@athos-ribeiro could you please rebase? We can merge this afterwards.

athos-ribeiro:
Rebased.

I tested this in osbs-box with https://github.com/containerbuildsystem/osbs-box/pull/123 and it works as expected.

PR #173, opened 2020-06-23
Generate content manifest files for Go packages

athos-ribeiro:

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -12,6 +12,7 @@ from cachito.errors import CachitoError, ValidationError
 from cachito.web import db
 from cachito.web.models import (
     ConfigFileBase64,
+    ContentManifest,
     Dependency,
     EnvironmentVariable,
     Package,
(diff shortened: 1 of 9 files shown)


athos-ribeiro:
@mprahl, @lcarva:

When will we start supporting multiple package requests? To make a relation between the content manifests and packages, the packages would need to be allowed to have dependencies (instead of having the dependencies linked to the request). I started working on having such dependency, but then I realized the reason the dependencies are related to the requests is due to the dependency replacement feature.

Moreover I did not want to add the "go-package" packages to the request so they will not be displayed in the main cachito request endpoint (which could generate confusion as they would seem to be duplicated entries for Golang packages).

Ideally, we would store the packages data in the database, with a relation referencing the content manifest and the content manifest to_json method would generate the file on the fly.

Thoughts?

mprahl:
> @mprahl, @lcarva:
> 
> When will we start supporting multiple package requests? To make a relation between the content manifests and packages, the packages would need to be allowed to have dependencies (instead of having the dependencies linked to the request). I started working on having such dependency, but then I realized the reason the dependencies are related to the requests is due to the dependency replacement feature.
> 
> Moreover I did not want to add the "go-package" packages to the request so they will not be displayed in the main cachito request endpoint (which could generate confusion as they would seem to be duplicated entries for Golang packages).
> 
> Ideally, we would store the packages data in the database, with a relation referencing the content manifest and the content manifest to_json method would generate the file on the fly.
> 
> Thoughts?

@athos-ribeiro I'm planning to start working on that today. How about that I make a separate PR just for linking dependencies to packages rather than requests and then you can rebase your PR on that?

athos-ribeiro:
Sounds good to me!

mprahl:
@athos-ribeiro FYI, #181 was merged.

athos-ribeiro:
Closing in favor of https://github.com/release-engineering/cachito/pull/187

PR #175, opened 2020-06-23
Add an allow list of npm file dependencies

mprahl:
One project that Cachito must support is a plugin for Kui, which must specify its configuration via an npm dependency. This project wants to use a file dependency for this to avoid a separate repository. Since this is a special case, this should be allowed.

Resolves CLOUDBLD-1533

### DIFF
--- a/cachito/workers/pkg_managers/npm.py
+++ b/cachito/workers/pkg_managers/npm.py
@@ -27,7 +27,7 @@ __all__ = [
 log = logging.getLogger(__name__)
 
 
-def _get_deps(package_lock_deps, _name_to_deps=None):
+def _get_deps(package_lock_deps, file_deps_allowlist, _name_to_deps=None):
     """
     Get a mapping of dependencies to all versions of the dependency.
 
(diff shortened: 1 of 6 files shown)


mprahl:
@nirzari could you please review?

nirzari:
@mprahl The code changes look good to me! Please fix pipeline before merging. :+1:

cachito/workers/pkg_managers/npm.py
```diff
@@ -276,10 +286,15 @@ def get_package_and_deps(package_json_path, package_lock_path):
         package_lock = json.load(f)
 
     package_lock_original = copy.deepcopy(package_lock)
-    name_to_deps, top_level_replacements = _get_deps(package_lock.get("dependencies", {}))
+    package = {"name": package_lock["name"], "type": "npm", "version": package_lock["version"]}
```
lcarva, line 289: We're using values from `package_lock` before `_get_deps` is called. I think that's ok because the only thing that may change withing `package_lock` is the `dependencies` which we do not use to construct the `package` dict.

PR #179, opened 2020-06-26
Add a configuration to ignore missing go.mod files

mprahl:
This is useful if the user erroneously specifies the "gomod"
package manager in a Cachito request when the repository
is not a Go module project.

This is also useful if the default package manager is set to
"gomod" and you want to allow users to make requests without
any package managers and would like them not to need to explicitly
set `"pkg_managers": []` in their request.


mprahl:
@athos-ribeiro could you please review?

PR #181, opened 2020-06-26
Associate request dependencies with the package they came from

mprahl:
This adds the "dependencies" key to each package in the "packages" array in the GET API endpoints.
    
This is necessary for future tasks such as generating content manifests and adding support for multiple packages of the same type in the same repository.

Relates to CLOUDBLD-1257

### DIFF
--- a/cachito/web/migrations/versions/1714d8e3002b_map_deps_to_pkg.py
+++ b/cachito/web/migrations/versions/1714d8e3002b_map_deps_to_pkg.py
@@ -1,0 +1,136 @@
+"""
+Map dependencies to their associated package.
+
+Revision ID: 1714d8e3002b
+Revises: b46cf36806d7
+Create Date: 2020-06-24 13:56:28.738839
+"""
+import logging
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "1714d8e3002b"
+down_revision = "b46cf36806d7"
+branch_labels = None
+depends_on = None
+
+log = logging.getLogger("alembic")
+
+request_dependency_table = sa.Table(
+    "request_dependency",
+    sa.MetaData(),
+    sa.Column("request_id", sa.Integer()),
+    sa.Column("dependency_id", sa.Integer()),
+    sa.Column("package_id", sa.Integer()),
+)
+
+request_package_table = sa.Table(
+    "request_package",
+    sa.MetaData(),
+    sa.Column("request_id", sa.Integer()),
+    sa.Column("package_id", sa.Integer()),
+)
+
+package_table = sa.Table(
+    "package",
+    sa.MetaData(),
+    sa.Column("id", sa.Integer(), primary_key=True),
+    sa.Column("name", sa.String()),
+    sa.Column("type", sa.String()),
+)
+
+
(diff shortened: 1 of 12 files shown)


mprahl:
@athos-ribeiro FYI, this is what you needed for #173

athos-ribeiro:
I understand the JSON representation for a request did not change to avoid bumping the API version, is that a correct assumption?

It would be nce to have the dependencies under the packages in the representations in the future, to make it easier/intuitive to navigate through packages and their dependencies.

Thanks for the patch! LGTM

mprahl:
@athos-ribeiro could you please review again?

cachito/web/models.py
```diff
@@ -139,14 +139,19 @@ def from_json(cls, package):
         cls.validate_json(package)
         return cls(**package)
 
-    def to_json(self):
+    def to_json(self, dependencies=None):
```
athos-ribeiro, line 142: At first I was wondering if it would be worth to receive the request instead of a dependency here and moving the logic to fetch the dependencies in here. But then I couldn't think of a nice way to remove code duplications this would generate. I thought some level of introspection could solve it, but then it would probably break encapsulation and would end up being a bad idea.

cachito/web/migrations/versions/1714d8e3002b_map_deps_to_pkg.py
```diff
@@ -0,0 +1,136 @@
...
+        ).fetchone()
+        package = connection.execute(
+            package_table.select()
+            .select_from(
+                package_table.join(
+                    request_package_table, package_table.c.id == request_package_table.c.package_id
+                )
+            )
+            .where(request_package_table.c.request_id == request_dep.request_id)
+            .where(package_table.c.type == dependency.type)
```
lcarva, line 94: Why is dependency type needed if we're fetching by ID?

mprahl: @lcarva a request could have multiple package managers associated with it, so we want to get the package that is of the same type as the dependency.

cachito/web/migrations/versions/c8b2a3a26191_initial_migration.py
```diff
@@ -84,7 +84,9 @@ def upgrade():
         sa.ForeignKeyConstraint(
             ["request_id"], ["request.id"], "request_dependency_dependency_id_fkey"
         ),
-        sa.UniqueConstraint("request_id", "dependency_id"),
+        sa.UniqueConstraint(
+            "request_id", "dependency_id", name="request_dependency_request_id_dependency_id_key"
```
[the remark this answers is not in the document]
mprahl: Yes, this is what was created in the `docker-compose` environment.

cachito/web/static/api_v1.yaml
```diff
@@ -393,6 +393,19 @@ components:
         version:
           type: string
           example: "v0.1.1"
+    PackageRetroDep:
```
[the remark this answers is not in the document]
mprahl: I removed it. I originally had this so that the examples looked more real rather than a package depending on a package that replaces itself.

mprahl:

PR #182, opened 2020-06-29
Provide Content Manifest API endpoint

athos-ribeiro:
Register a new “GET /requests/REQUEST_ID/content-manifest” API
endpoint. All requests should serve at least an empty, valid, content
manifest.

Assembling content manifests should be implemented in a per package
manager manner.

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -62,6 +62,29 @@ request_config_file_base64_table = db.Table(
 )
 
 
+class ContentManifest:
+    """A content manifest associated with the request."""
+
+    version = 1
+    json_schema_url = (
+        "https://raw.githubusercontent.com/containerbuildsystem/atomic-reactor/"
+        "f4abcfdaf8247a6b074f94fa84f3846f82d781c6/atomic_reactor/schemas/content_manifest.json"
+    )
+    unknown_layer_index = -1
+
+    empty_icm = {
+        "metadata": {
+            "icm_version": version,
+            "icm_spec": json_schema_url,
+            "image_layer_index": unknown_layer_index,
+        },
+    }
+
+    def to_json(self):
+        """Generate the JSON representation of the content manifest."""
+        return self.empty_icm
+
+
 class RequestStateMapping(Enum):
     """An Enum that represents the request states."""
 
@@ -350,6 +373,10 @@ class Request(db.Model):
     config_files_base64 = db.relationship(
         "ConfigFileBase64", secondary=request_config_file_base64_table, backref="requests"
     )
+
+    # This is an empty content manifest which should be returned for all requests whose package
+    # manager does not implement content manifest creation
+    content_manifest = ContentManifest()
 
     def __repr__(self):
         return "<Request {0!r}>".format(self.id)
--- a/cachito/web/static/api_v1.yaml
+++ b/cachito/web/static/api_v1.yaml
@@ -277,6 +277,69 @@ paths:
                 $ref: "#/components/schemas/RequestConfiguration"
       security:
       - "Kerberos Authentication": []
+  "/requests/{id}/content-manifest":
+    get:
+      description: Return the content manifest of a request
+      parameters:
+      - name: id
+        in: path
+        required: true
+        description: The ID of the Cachito request to retrieve the content manifest for
+        schema:
+          type: integer
+      responses:
+        "200":
+          description: The content manifest of the Cachito request
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  metadata:
+                    type: object
+                    properties:
+                      icm_version:
+                        type: integer
+                        example: 1
+                      icm_spec:
+                        type: string
+                        example: "https://raw.githubusercontent.com/containerbuildsystem/\
+                                  atomic-reactor/f4abcfdaf8247a6b074f94fa84f3846f82d781c6/\
+                                  atomic_reactor/schemas/content_manifest.json"
+                      image_layer_index:
+                        type: integer
+                        example: -1
+                  image_contents:
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        purl:
+                          $ref: "#/components/schemas/Purl"
+                        dependencies:
+                          type: array
+                          items:
+                            type: object
+                            properties:
+                              purl:
+                                $ref: "#/components/schemas/Purl"
+                        sources:
+                          type: array
+                          items:
+                            type: object
+                            properties:
+                              purl:
+                                $ref: "#/components/schemas/Purl"
+        "404":
+          description: The request wasn't found
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  error:
+                    type: string
+                    example: The requested resource was not found
   "/requests/{id}/download":
     get:
       description: Download a Cachito request bundle
@@ -387,6 +450,9 @@ components:
         total:
           type: integer
           example: 45
+    Purl:
+      type: string
+      example: "pkg:golang/github.com%2Frelease-engineering%2Fretrodep%2Fv2@v2.0.2"
     RequestBase:
       type: object
       properties:
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -93,6 +93,21 @@ def get_request_config_files(request_id):
     return flask.jsonify(config_files_json)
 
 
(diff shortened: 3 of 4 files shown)


cachito/web/static/api_v1.yaml
```diff
@@ -277,6 +277,33 @@ paths:
...
+        description: The ID of the Cachito request to retrieve the content manifest for
+        schema:
+          type: integer
+      responses:
+        "200":
+          description: The content manifest of the Cachito request
+          content:
+            application/json:
+              schema:
+                $ref: "https://raw.githubusercontent.com/containerbuildsystem/atomic-reactor/f4abcfdaf8247a6b074f94fa84f3846f82d781c6/atomic_reactor/schemas/content_manifest.json#"
```
athos-ribeiro, line 296: Not sure how to get this one to work. It seems it is not happy with `"type": ["object", "null"]` and it just shows null. The file is being loaded, since the schema title is being imported.

athos-ribeiro:
@mprahl, @lcarva, @ben-alkov, PTAL

athos-ribeiro:
Partially adapted https://github.com/release-engineering/cachito/pull/173 here. Let's keep the other one open as a draft until we finish pulling the idea in. I will close it then.

cachito/web/models.py
```diff
@@ -62,6 +62,29 @@
 )
 
 
+class ContentManifest:
+    """A content manifest associated with the request."""
+
+    version = 1
+    json_schema_url = (
+        "https://raw.githubusercontent.com/containerbuildsystem/atomic-reactor/"
+        "f4abcfdaf8247a6b074f94fa84f3846f82d781c6/atomic_reactor/schemas/content_manifest.json"
```
ben-alkov, line 71: Using a specific commit hash seems brittle...

mprahl: Could you use a Git tag instead of a commit hash?

cachito/web/models.py
```diff
@@ -62,6 +62,29 @@
 )
 
 
+class ContentManifest:
```
mprahl, line 65: Could you move this to a new file? `models.py` is already rather large and this change is not really related to a database schema.

cachito/web/static/api_v1.yaml
mprahl: I'm not sure JSON schema is supported as OpenAPI is an extended subset of JSON schema:
https://swagger.io/docs/specification/data-models/keywords/

cachito/web/models.py
athos-ribeiro: The reason I used the commit in the first place was due to the floating nature of tags. I can move to a tag, knowing we do control the tags in atomic-reactor and they will not change.

cachito/web/static/api_v1.yaml
athos-ribeiro: Any suggestions to avoid duplicating the schema here?

cachito/web/models.py
athos-ribeiro: Actually, we do not have an atomic-reactor release with it, so we would need to wait for the next release cycle to change this. Any other suggestions?

cachito/web/static/api_v1.yaml
athos-ribeiro: I ended up partially pulling in the schema

cachito/web/models.py
mprahl: @athos-ribeiro that's fine. Let's just update it once it's released or you can set it now knowing the URL is invalid until the release.

cachito/web/static/api_v1.yaml
```diff
@@ -277,6 +277,75 @@ paths:
...
+        schema:
+          type: integer
+      responses:
+        "200":
+          description: The content manifest of the Cachito request
+          content:
+            application/json:
+              schema:
+                type: object
+                nullable: true
```
mprahl, line 297: Is this actually true?

cachito/web/static/api_v1.yaml
```diff
@@ -277,6 +277,75 @@ paths:
...
+                properties:
+                  metadata:
+                    type: object
+                    properties:
+                      icm_version:
+                        type: integer
+                        example: 1
+                      icm_spec:
+                        type: string
+                        example: "https://json-schema.icm.example.com"
```
mprahl, line 307: A real example would be nice here so people can go to the schema directly from the Cachito API documentation.

cachito/web/static/api_v1.yaml
```diff
@@ -277,6 +277,75 @@ paths:
...
+                      image_layer_index:
+                        type: integer
+                        example: 1
```
mprahl, line 310: Will Cachito ever return an integer other than `-1` since it has no context of the image that the request is used for?

cachito/web/static/api_v1.yaml
```diff
@@ -277,6 +277,75 @@ paths:
...
+                        purl:
+                          $ref: "#/components/schemas/Purl"
+                        dependencies:
+                          type: array
+                          items:
+                            type: object
+                            properties:
+                              purl:
+                                $ref: "#/components/schemas/Purl"
+                        sources:
```
mprahl, line 325: Will Cachito ever set this? If not, I think this should be removed.

cachito/web/static/api_v1.yaml
```diff
@@ -277,6 +277,75 @@ paths:
...
+                          type: array
+                          items:
+                            type: object
+                            properties:
+                              purl:
+                                $ref: "#/components/schemas/Purl"
+                  content_sets:
```
mprahl, line 332: Will Cachito ever set this? If not, I think this should be removed.

mprahl:
@athos-ribeiro this looks good. I left a few comments. Once addressed, please squash the commits and then it should be good to merge.

cachito/web/static/api_v1.yaml
athos-ribeiro: According to the original schema, it is. We could remove it here since we are not returning null though, since this is just for documenting what cachito is returning.

cachito/web/static/api_v1.yaml
athos-ribeiro: Yes, the idea is that cachito itself will return this one filled for gomod with the module level dependencies, which were the sources used in build time

cachito/web/static/api_v1.yaml
athos-ribeiro: Not really, I will remove this one

cachito/web/static/api_v1.yaml
athos-ribeiro: Removed

cachito/web/static/api_v1.yaml
athos-ribeiro: Changed

cachito/web/static/api_v1.yaml
athos-ribeiro: Changed to `-1`

cachito/web/static/api_v1.yaml
athos-ribeiro: removed

athos-ribeiro:
Comments addressed; commits squashed

cachito/web/static/api_v1.yaml
mprahl: I reread the design document and now remember what sources is for.

cachito/web/static/api_v1.yaml
```diff
@@ -277,6 +277,69 @@ paths:
...
+                    type: object
+                    properties:
+                      icm_version:
+                        type: integer
+                        example: 1
+                      icm_spec:
+                        type: string
+                        example: "https://raw.githubusercontent.com/containerbuildsystem/\
+                                  atomic-reactor/f4abcfdaf8247a6b074f94fa84f3846f82d781c6/\
+                                  atomic_reactor/schemas/content_manifest.json"
```
mprahl, line 308: I think this might be the preferred syntax for long strings in YAML:
```suggestion
                        example: >-
                          https://raw.githubusercontent.com/containerbuildsystem/
                          atomic-reactor/f4abcfdaf8247a6b074f94fa84f3846f82d781c6/
                          atomic_reactor/schemas/content_manifest.json"
```

athos-ribeiro: Wouldn't this add spaces after the slashes though?

mprahl: You're right

PR #185, opened 2020-07-01
Add support for multiple packages for the npm package manager in a single repo

mprahl:
This is done by supplying the `packages` parameter. For example:
`{"packages": {"npm": [{"path": "client"}, {"path": "proxy"}]}}`

Resolves CLOUDBLD-1257

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -2,6 +2,7 @@
 from collections import OrderedDict
 from copy import deepcopy
 from enum import Enum
+import os
 import re
 
 import flask
@@ -346,6 +347,65 @@ class RequestDependency(db.Model):
     __table_args__ = (db.UniqueConstraint("request_id", "dependency_id", "package_id"),)
 
 
+def _validate_request_package_configs(request_kwargs, pkg_managers_names):
+    """
+    Validate the "packages" parameter in a new request.
+
+    :param dict request_kwargs: the JSON parameters of the new request
+    :param list pkg_managers_names: the list of valid package manager names for the request
+    :raises ValidationError: if the "packages" parameter is invalid
+    """
+    # Validate the custom packages configuration. For example:
+    # {"packages": {"npm": [{"path": "client"}]}}
+    packages_configs = request_kwargs.get("packages", {})
+    if not isinstance(packages_configs, dict):
+        raise ValidationError('The "packages" parameter must be an object')
+
+    invalid_package_managers = packages_configs.keys() - set(pkg_managers_names)
+    if invalid_package_managers:
+        raise ValidationError(
+            'The following package managers in the "packages" object do not apply: '
+            + ", ".join(invalid_package_managers)
+        )
+
+    supported_packages_configs = {"npm"}
+    unsupported_packages_managers = packages_configs.keys() - supported_packages_configs
+    if unsupported_packages_managers:
+        raise ValidationError(
+            'The following package managers in the "packages" object are unsupported: '
+            + ", ".join(unsupported_packages_managers)
+        )
+
+    # Validate the values for each package manager configuration (e.g. packages.npm)
+    valid_package_config_keys = {"path"}
+    for pkg_manager, packages_config in packages_configs.items():
+        invalid_format_error = (
+            f'The value of "packages.{pkg_manager}" must be an array of objects with the following '
+            f'keys: {", ".join(valid_package_config_keys)}'
+        )
+        if not isinstance(packages_config, list):
+            raise ValidationError(invalid_format_error)
+
+        for package_config in packages_config:
+            if not isinstance(package_config, dict) or not package_config:
+                raise ValidationError(invalid_format_error)
+
+            invalid_keys = package_config.keys() - valid_package_config_keys
+            if invalid_keys:
+                raise ValidationError(invalid_format_error)
+
+            if not (
+                isinstance(package_config["path"], str)
+                and package_config["path"]
+                and not os.path.isabs(package_config["path"])
+                and os.pardir not in os.path.normpath(package_config["path"])
+            ):
+                raise ValidationError(
+                    f'The "path" values in the "packages.{pkg_manager}" value must be to a '
+                    "relative path in the source repository"
+                )
+
+
 class Request(db.Model):
     """A Cachito user request."""
 
@@ -435,7 +495,9 @@ class Request(db.Model):
         :rtype: int
         """
         return (
-            db.session.query(sqlalchemy.func.count(RequestDependency.dependency_id))
+            db.session.query(
+                sqlalchemy.func.count(sqlalchemy.distinct(RequestDependency.dependency_id))
+            )
             .filter(RequestDependency.request_id == self.id)
             .scalar()
         )
@@ -508,13 +570,21 @@ class Request(db.Model):
             rv["state_history"] = states
 
             package_to_deps = {}
+            # This is used to quickly see if a dependency has been added to the root "dependencies"
+            # array in the returned JSON. This uses a set instead of iterating through the
+            # "dependencies" array to avoid an O(n) operation for every dependency.
+            seen_dependencies = set()
             rv["dependencies"] = []
             for req_dep in self.request_dependencies:
                 dep_json = req_dep.dependency.to_json(
                     req_dep.replaced_dependency.to_json() if req_dep.replaced_dependency else None,
                     force_replaces=True,
                 )
-                rv["dependencies"].append(dep_json)
+                # Avoid duplicate dependencies in the root "dependencies" array
+                if req_dep.dependency.id not in seen_dependencies:
+                    rv["dependencies"].append(dep_json)
+                    seen_dependencies.add(req_dep.dependency.id)
+
                 package_to_deps.setdefault(req_dep.package.id, []).append(dep_json)
 
             rv["packages"] = [
@@ -540,7 +610,7 @@ class Request(db.Model):
         """
         # Validate all required parameters are present
         required_params = {"repo", "ref"}
-        optional_params = {"dependency_replacements", "flags", "pkg_managers", "user"}
+        optional_params = {"dependency_replacements", "flags", "packages", "pkg_managers", "user"}
 
         missing_params = required_params - set(kwargs.keys()) - optional_params
         if missing_params:
@@ -572,6 +642,11 @@ class Request(db.Model):
 
         pkg_managers = PackageManager.get_pkg_managers(pkg_managers_names)
         request_kwargs["pkg_managers"] = pkg_managers
+
+        _validate_request_package_configs(request_kwargs, pkg_managers_names or [])
+        # Remove this from the request kwargs since it's not used as part of the creation of
+        # the request object
+        request_kwargs.pop("packages", None)
 
         flag_names = request_kwargs.pop("flags", None)
         if flag_names:
--- a/cachito/web/static/api_v1.yaml
+++ b/cachito/web/static/api_v1.yaml
@@ -586,6 +586,16 @@ components:
           items:
             type: string
             example: some_flag
+        packages:
+          type: object
+          items:
+            type: array
+            items:
+              type: object
+          example:
+            npm:
+            - path: client
+            - path: legacy/client
         pkg_managers:
           type: array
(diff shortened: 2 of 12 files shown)


mprahl:
@athos-ribeiro could you also please review if you have time?

cachito/workers/tasks/npm.py
```diff
@@ -32,6 +32,51 @@
 log = logging.getLogger(__name__)
 
 
+def _verify_npm_files(bundle_dir, subpaths):
+    """
+    Verify that the correct are present for the npm package manager to proceed.
```
nirzari, line 37: Did you mean `correct npm files are present`?

cachito/web/models.py
```diff
@@ -346,6 +347,65 @@ class RequestDependency(db.Model):
...
+
+            invalid_keys = package_config.keys() - valid_package_config_keys
+            if invalid_keys:
+                raise ValidationError(invalid_format_error)
+
+            if not (
+                isinstance(package_config["path"], str)
+                and package_config["path"]
+                and not os.path.isabs(package_config["path"])
+                and os.path.pardir not in package_config["path"]
```
lcarva, line 401: It's odd. I don't see either `os.path.pardir` nor `os.path.curdir` in the python docs: https://docs.python.org/3/library/os.path.html

However, they do exist in there.

I do see them listed in the docs for the `os` module: https://docs.python.org/3/library/os.html

Could we use that instead in order to avoid some sort of breakage in the future?

cachito/web/models.py
```diff
@@ -346,6 +347,65 @@ class RequestDependency(db.Model):
...
+            ):
+                raise ValidationError(
+                    f'The "path" values in the "packages.{pkg_manager}" value must be to a '
```
lcarva, line 404: This message may be confusing to users. A path with `..` in it could be resolved to a relative path withing the application repo, e.g. `foo/bar/../baz`. This would be unusual, but the error message wouldn't necessarily point the user in the right direction.

cachito/web/models.py
```diff
@@ -503,13 +565,24 @@ def _state_to_json(state):
             rv["state_history"] = states
 
             package_to_deps = {}
+            # This dictionary will have the keys as dependency names and the values as sets of
+            # dependency IDs. This is used to quickly see if a dependency has been added to the
+            # root "dependencies" array in the returned JSON. This uses a dictionary instead of
+            # a list to avoid an O(n) operation for every dependency.
+            dep_name_to_dep_ids = {}
```
lcarva, line 572: If dependency IDs are unique, won't a simple set work?
```python
already_seen_dependencies = set()
...
if req_dep.dependency.id not in already_seen_dependencies:
  ...
  already_seen_dependencies.add(req_dep.dependency.id)
```

cachito/web/static/api_v1.yaml
```diff
@@ -534,6 +534,16 @@ components:
...
             example: some_flag
+        packages:
+          type: object
+          items:
+            type: array
+            items:
+              type: object
+          example:
+            npm:
+            - path: client
```
lcarva, line 545: Maybe provide an example with multiple directories, e.g. `client/foo/bar` ?

cachito/workers/pkg_managers/npm.py
```diff
@@ -332,16 +332,20 @@ def get_package_and_deps(package_json_path, package_lock_path):
...
 
     :param str app_source_path: the full path to the application source code
     :param dict request: the Cachito request this is for
+    :param set skip_deps: a set of dependency identifiers to not download because they've already
+        been downloaded for this request
     :return: a dictionary that has the keys "deps" which is the list of dependencies,
-        "lock_file" which is the lock file if it was modified, "package" which is the
-        dictionary describing the main package, and "package.json" which is the package.json file if
-        it was modified.
+        "downloaded_deps" which is a set of the dependency identifiers of the dependencies that
```
lcarva, line 344: It would be nice if we could structure this text as an unordered list.

cachito/workers/pkg_managers/npm.py
```diff
@@ -332,16 +332,20 @@ def get_package_and_deps(package_json_path, package_lock_path):
...
+        were downloaded as part of this function's execution, "lock_file" which is the lock file if
+        it was modified, "lock_file_name" is the name of the lock file that was used, "package"
```
lcarva, line 346: Is it just the name of the lock file, or the full path to it?

cachito/workers/tasks/npm.py
lcarva: Maybe it should be `expected npm files` ?

cachito/workers/tasks/npm.py
```diff
@@ -32,6 +32,51 @@
...
+    for subpath in subpaths:
+        bundle_dir_subpath = bundle_dir.app_subpath(subpath)
+        for lock_file in (
+            bundle_dir_subpath.npm_shrinkwrap_file,
+            bundle_dir_subpath.npm_package_lock_file,
+        ):
+            if lock_file.exists():
+                break
+        else:
+            shrinkwrap_rel_path = bundle_dir_subpath.get_relpath(
```
lcarva, line 53: It would be great if we didn't have to duplicate these entries here. Could we do something like:
```python
lock_files = (bundle_dir_subpath.npm_shrinkwrap_file, bundle_dir_subpath.npm_package_lock_file)

...

lock_files_relpath = [bundle_dir_subpath.get_relpath(lock_file) for lock_file in lock_files]
```

cachito/workers/tasks/npm.py
```diff
@@ -43,102 +88,128 @@ def cleanup_npm_request(request_id):
...
-    for lock_file in (bundle_dir.npm_shrinkwrap_file, bundle_dir.npm_package_lock_file):
-        if lock_file.exists():
-            break
-    else:
-        raise CachitoError(
-            "The npm-shrinkwrap.json or package-lock.json file must be present for the npm "
-            "package manager"
-        )
+    subpaths = []
+    if package_configs:
```
lcarva, line 111: `package_configs` is set to an empty list if it's `None`. We could completely remove this if statement, and the explicit initialization of `subpaths` above..

cachito/workers/tasks/npm.py
```diff
@@ -43,102 +88,128 @@ def cleanup_npm_request(request_id):
...
+    npm_config_files = []
+    downloaded_deps = set()
+    for i, subpath in enumerate(subpaths):
+        log.info("Fetching the npm dependencies for request %d in subpath %s", request_id, subpath)
+        request = set_request_state(
+            request_id,
+            "in_progress",
+            f'Fetching the npm dependencies at the "{subpath}" directory"',
+        )
+        package_source_path = str(bundle_dir.source_dir.joinpath(subpath))
```
lcarva, line 134: Shouldn't this be `app_subpath` ?

tests/test_api_v1.py
```diff
@@ -122,6 +122,35 @@ def test_create_and_fetch_request(
...
+        "repo": "https://github.com/release-engineering/web-terminal.git",
+        "ref": "c50b93a32df1c9d700e3e80996845bc2e13be848",
+        "packages": package_value,
+        "pkg_managers": ["npm"],
+    }
+
+    rv = client.post("/api/v1/requests", json=data, environ_base=auth_env)
+    assert rv.status_code == 201
+
+    error_callback = failed_request_callback.s(1)
```
lcarva, line 140: Could use the request ID from `rv` to make it clearer what the number `1` is.

tests/test_workers/test_pkg_managers/test_general_js.py
```diff
@@ -140,6 +140,63 @@ def test_download_dependencies(
...
+@mock.patch("cachito.workers.pkg_managers.general_js.generate_and_write_npmrc_file")
+@mock.patch("cachito.workers.pkg_managers.general_js.run_cmd")
+@mock.patch("shutil.move")
+@mock.patch("cachito.workers.paths.get_worker_config")
+def test_download_dependencies_skip_deps(
+    mock_gwc, mock_move, mock_run_cmd, mock_gawnf, mock_exists, mock_td, tmpdir,
+):
+    bundles_dir = tmpdir.mkdir("bundles")
+    mock_gwc.return_value.cachito_bundles_dir = str(bundles_dir)
+    mock_td.return_value.__enter__.return_value = "/tmp/cachito-agfdsk"
```
lcarva, line 154: Can we please use `tmpdir` here?

cachito/web/models.py
```diff
@@ -346,6 +347,65 @@ class RequestDependency(db.Model):
...
+        raise ValidationError(
+            'The following package managers in the "packages" value are unsupported: '
+            + ", ".join(unsupported_packages_managers)
+        )
+
+    # Validate the values for each package manager configuration (e.g. packages.npm)
+    valid_package_config_keys = {"path"}
+    for pkg_manager, packages_config in packages_configs.items():
+        invalid_format_error = (
+            f'The "packages.{pkg_manager}" value must be an array of objects with the following '
```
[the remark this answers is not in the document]
mprahl: json.org uses the term "names" but the usage "keys" is common such as on w3schools:
https://www.w3schools.com/js/js_json_objects.asp

We use the term "keys" for all of our error messages in Cachito already, so I'd like to stick to that.

cachito/web/models.py
mprahl: Okay, I'll check it against `os.path.normpath(package_config["path"])`.

cachito/web/models.py
mprahl: You're right. The time complexity of an `in` operation on a set averages to O(n) like accessing a dictionary key. That makes sense. I'll do that.

cachito/web/models.py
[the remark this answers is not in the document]
lcarva: TIL: the JSON Schema definition seems to use neither "keys" nor "attributes": https://json-schema.org/draft-04/json-schema-core.html#rfc.section.3.1

cachito/workers/pkg_managers/npm.py
mprahl: Just the name.

cachito/web/models.py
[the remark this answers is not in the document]
nirzari: +1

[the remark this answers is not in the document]
mprahl: @lcarva and @nirzari so are we saying to change all the existing error messages to use the term member, property, or name or are we just keeping it as keys for consistency?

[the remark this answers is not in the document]
nirzari: IMO we could keep it as "keys" for consistency here and later file a separate commit/MR to change the rest of them if we really want to.

cachito/workers/pkg_managers/npm.py
mprahl: I changed the structure a little so that it's easier to read and still valid restructured text.

cachito/workers/tasks/npm.py
mprahl: It's less efficient but it makes the code more readable and less duplication, so I changed it to use `app_subpath`.

tests/test_workers/test_pkg_managers/test_general_js.py
mprahl: I made the change, though it is not as bad as it looks since nothing writes to that directory since everything is mocked out but yes, it would be safer to use tmpdir.

mprahl:
The tests are failing after rebasing to master. I'll fix them.

mprahl:
The test are now passing again and all the comments have been addressed. Thanks for the reviews!

cachito/web/models.py
```diff
@@ -346,6 +347,65 @@ class RequestDependency(db.Model):
...
+        raise ValidationError(
+            'The following package managers in the "packages" object do not apply: '
+            + ", ".join(invalid_package_managers)
+        )
+
+    supported_packages_configs = {"npm"}
+    unsupported_packages_managers = packages_configs.keys() - supported_packages_configs
+    if unsupported_packages_managers:
+        raise ValidationError(
+            'The following package managers in the "packages" value are unsupported: '
```
lcarva, line 375: in the "packages" object?

PR #191, opened 2020-07-10
Make the map_deps_to_pkg migration more efficient

mprahl:
This should increase the performance of the migration on large Cachito deployments.

### DIFF
--- a/cachito/web/migrations/versions/1714d8e3002b_map_deps_to_pkg.py
+++ b/cachito/web/migrations/versions/1714d8e3002b_map_deps_to_pkg.py
@@ -68,74 +68,95 @@ def upgrade():
         )
 
 
+def _update_request_deps(connection, request_id, pkg_id_to_dep_ids):
+    for pkg_id, dep_ids in pkg_id_to_dep_ids.items():
+        connection.execute(
+            request_dependency_table.update()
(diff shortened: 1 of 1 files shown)


cachito/web/migrations/versions/1714d8e3002b_map_deps_to_pkg.py
```diff
@@ -68,74 +68,94 @@ def upgrade():
...
     connection = op.get_bind()
     last_request_id = None
     dummy_package_name = "cachito-migration-placeholder"
     dummy_package_version = "0.0.0"
+    pkg_type_to_pkg = {}
+    pkg_id_to_dep_ids = {}
     for request_dep in connection.execute(
         request_dependency_table.select().order_by(request_dependency_table.c.request_id)
     ).fetchall():
         if last_request_id != request_dep.request_id:
```
lcarva, line 91: Should we only do this if `last_request_id is not None` ?

mprahl: It's not necessary since `_update_request_deps` would do nothing but I think it makes the cleaner if we don't execute it when `last_request_id` is `None`.

I fixed it in the last push.

PR #192, opened 2020-07-13
Cloudbld 1262 unique path for npm dependencies

nirzari:


cachito/workers/pkg_managers/general_js.py
```diff
@@ -133,8 +169,22 @@ def download_dependencies(request_id, deps, proxy_repo_url, skip_deps=None):
             for dep_pair in list(zip(output.split("\n"), dep_batch)):
                 tarball = dep_pair[0]  # e.g. angular-animations-8.2.0.tgz
                 dir_path = dep_pair[1].rsplit("@", 1)[0]  # e.g. @angular/animations
+                version = dep_pair[1].rsplit("@", 1)[1]
```
athos-ribeiro, line 172: We could split it only once here with `dir_path, version = dep_pair[1].rsplit("@", 1)`

cachito/workers/pkg_managers/general_js.py
```diff
@@ -133,8 +169,22 @@ def download_dependencies(request_id, deps, proxy_repo_url, skip_deps=None):
...
+
+                # In case of external dependencies, create additional intermediate
+                # parent e.g. github/<org>/<repo> or external-<repo>
+                extra_dir = ""
+                # Set up unique path for external dependencies
+                if ":" in version and version.rsplit(":", 1)[0] in prefixes:
```
athos-ribeiro, line 178: Is `rsplit` what we want here? What if the version includes `http://example.com:8080/foo...`

Also, would using regular expressions with the `re` module help parsing this?

something like

```
match = re.match(r'(?P<prefix>[^:]+):(?P<is_external>//)?(?P<suffix>[^/]+)', version)
# do some checking for the prefix
if match:
  if match.group('is_external'):
    dir_path = f"external-{dir_path}"
  else:
    extra_dir = os.path.join(match.group('prefix'),  match.group('suffix'),

```

nirzari: Excellent catch/suggestion Athos. Will make the change. Thanks!

cachito/workers/pkg_managers/general_js.py
nirzari: While it's a one-liner and a good suggestion, I'd prefer to keep it as it is just for the sake of readability and having an example there in-line can help make things easier to understand IMO. lmk what do you think.

cachito/workers/pkg_managers/general_js.py
```diff
@@ -34,6 +36,32 @@
 log = logging.getLogger(__name__)
 
 
+def get_git_prefixes():
```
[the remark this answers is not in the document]
nirzari: I considered doing that but we need 2 different set of git_prefixes and http_prefixes in npm.py.
I thought it might be better to define them at one place and have if else logic in the function since in the future if we're adding more to the list, we have to do it at a single place only. Wdyt?

cachito/workers/pkg_managers/general_js.py
```diff
@@ -133,8 +169,22 @@ def download_dependencies(request_id, deps, proxy_repo_url, skip_deps=None):
...
+                # In case of external dependencies, create additional intermediate
+                # parent e.g. github/<org>/<repo> or external-<repo>
+                extra_dir = ""
+                # Set up unique path for external dependencies
+                if ":" in version and version.rsplit(":", 1)[0] in prefixes:
+                    if "://" in version:
+                        dir_path = f"external-{dir_path}"
+                    else:
+                        extra_dir = os.path.join(
+                            version.rsplit(":", 1)[0], version.rsplit(":", 1)[1].rsplit("/", 1)[0]
```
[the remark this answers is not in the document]
nirzari: I'm not sure about this requirement as AC originally says `github/<org>/<repo>` and then has an additional question about `github.com`. Also the comments on the related spike suggests `github`. I have raised this question in Cloudbuild channel. Maybe we can discuss it there what was actually agreed upon.

cachito/workers/pkg_managers/general_js.py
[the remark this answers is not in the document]
nirzari: Hmmm this looks like a big misunderstanding on my part while understanding the distinction between `version` and `version_in_nexus`. I will make necessary changes to fix this.

cachito/workers/pkg_managers/general_js.py
```diff
@@ -118,23 +152,40 @@ def download_dependencies(request_id, deps, proxy_repo_url, skip_deps=None):
...
                 tarball = dep_pair[0]  # e.g. angular-animations-8.2.0.tgz
-                dir_path = dep_pair[1].rsplit("@", 1)[0]  # e.g. @angular/animations
+                dir_path = dep_pair[1][0].rsplit("@", 1)[0]  # e.g. @angular/animations
+                external_dep_version = dep_pair[1][1]
+
+                # In case of external dependencies, create additional intermediate
+                # parent e.g. github/<org>/<repo> or external-<repo>
+                extra_dir = ""
+                if external_dep_version is not None:
+                    match = re.match(r"(?P<prefix>[^:]+):?(?P<suffix>[^/]+)", external_dep_version)
```
athos-ribeiro, line 181: Why is the ':' between the prefix and suffix optional?

Also, is `git://$STRING` a valid value here? if so, the suffix captured would be just the ':' character, since since the suffix should be "any non-empty string after the prefix (or a ':') until a '/' is found.

Finally, note that if no match is found, re.match will return `None`.

nirzari: I added the `:` between prefix and suffix because for I/P `github:ReactiveX/rxjs#78032157f5c1655436829017bbda787565b48c30`, I need O/P as `ReactiveX/rxjs`. Without the added `:`, it's giving me `:ReactiveX/rxjs`.

`git://$STRING` is a valid value but in that case, it falls under the `other_prefixes` category and so we don't care about what `suffix holds`.

if `external_dep_version` is not None, there **must** be a match. IMO, we should just let it fail and the failure would be captured in https://github.com/release-engineering/cachito/pull/192/files#diff-c372cb74d95ccecd07e2b5b76a539cf3R157 even before reaching here if it's not a valid dependency. 
@mprahl Please correct me if I'm wrong!

cachito/workers/pkg_managers/general_js.py
```diff
@@ -34,6 +36,27 @@
 log = logging.getLogger(__name__)
 
 
+def get_git_prefixes():
+    """Get the git prefixes supported by Cachito.
```
mprahl, line 40:
```suggestion
    """
    Get the git prefixes supported by Cachito.
```

mprahl, line 40:
```suggestion
    """
    Get the git prefixes supported by Cachito.
```

cachito/workers/pkg_managers/general_js.py
```diff
@@ -34,6 +36,27 @@
...
+
+    :return: the set of Cachito supported git prefixes
```
mprahl, line 42: This is now returning a tuple. Could you please update this?

cachito/workers/pkg_managers/general_js.py
```diff
@@ -34,6 +36,27 @@
...
+    :rtype: set
+    """
+    known_git_prefixes = {"github:", "bitbucket:", "gitlab:"}
+    other_git_prefixes = {"git://", "git+http://", "git+https://", "git+ssh://"}
+    return known_git_prefixes, other_git_prefixes
+
+
+def get_http_prefixes():
+    """Get the http prefixes supported by Cachito.
```
mprahl, line 51:
```suggestion
    """
    Get the HTTP prefixes supported by Cachito.
```

cachito/workers/pkg_managers/general_js.py
```diff
@@ -34,6 +36,27 @@
...
+
+    :return: the set of Cachito supported http prefixes
```
mprahl, line 53:
```suggestion
    :return: the set of Cachito supported HTTP prefixes
```

cachito/workers/pkg_managers/general_js.py
```diff
@@ -34,6 +36,27 @@
 log = logging.getLogger(__name__)
 
 
+def get_git_prefixes():
```
mprahl, line 39: Optional: I know this sounds a little obsessive, but the whole Python module is alphabetized so it'd be nice to keep this consistent. :stuck_out_tongue:

cachito/workers/pkg_managers/general_js.py
```diff
@@ -92,13 +115,24 @@ def download_dependencies(request_id, deps, proxy_repo_url, skip_deps=None):
...
+        other_prefixes = {i.rsplit(":", 1)[0] for i in get_git_prefixes()[1]}.union(
+            i.rsplit(":", 1)[0] for i in get_http_prefixes()
+        )
+
         # This must be done in batches to prevent Nexus from erroring with "Header is too large"
         deps_batches = []
         counter = 0
         batch_size = get_worker_config().cachito_js_download_batch_size
         for dep in deps:
+            # Initialize external dependency version
```
mprahl, line 131: This comment doesn't add much value since it's clear this is happening in the following line. If you'd like to keep the comment, you can explain why the variable is used.
```suggestion
```

cachito/workers/pkg_managers/general_js.py
```diff
@@ -118,23 +152,40 @@ def download_dependencies(request_id, deps, proxy_repo_url, skip_deps=None):
...
+                # In case of external dependencies, create additional intermediate
+                # parent e.g. github/<org>/<repo> or external-<repo>
+                extra_dir = ""
+                if external_dep_version is not None:
+                    match = re.match(r"(?P<prefix>[^:]+):?(?P<suffix>[^/]+)", external_dep_version)
+                    if match.group("prefix") in known_git_prefixes:
+                        extra_dir = os.path.join(match.group("prefix"), match.group("suffix"))
+                    elif match.group("prefix") in other_prefixes:
+                        dir_path = f"external-{dir_path}"
+
```
mprahl, line 186: I'm thinking this might be simpler to read and less fragile and then you wouldn't need to consult `known_git_prefixes` or `other_prefixes`.
```suggestion
                # In case of external dependencies, create additional intermediate
                # parent e.g. github/<namespace>/<repo> or external-<repo>
                external_dep_version = dep_pair[1][1]
                if external_dep_version:
                    match = re.match(
                        r"^(?P<host>.+)(?::)(?P<namespace>[^/]+)(?:\/)(?P<repo>[^/#]+)(?:.+)$",
                        external_dep_version,
                    )
                    if match:
                        groups = match.groupdict()
                        dir_path = os.path.join(groups["host"], groups["namespace"], groups["repo"])
                    else:
                        dir_path = f"external-{dir_path}"
                else:
                    dir_path = dep_pair[1][0].rsplit("@", 1)[0]  # e.g. @angular/animations
```

tests/test_workers/test_pkg_managers/test_general_js.py
```diff
@@ -14,6 +14,19 @@
 from cachito.workers.paths import RequestBundleDir
 
 
+def test_get_git_prefixes():
```
mprahl, line 17: Optional: I know this sounds a little obsessive, but the whole Python module is alphabetized so it'd be nice to keep this consistent. :stuck_out_tongue:

tests/test_workers/test_pkg_managers/test_general_js.py
```diff
@@ -42,6 +55,8 @@ def test_download_dependencies(
         angular-devkit-architect-0.803.26.tgz
         angular-animations-8.2.14.tgz
         rxjs-6.5.5-external-gitcommit-78032157f5c1655436829017bbda787565b48c30.tgz
+        exsp-external-archive-2.10.2.tar.gz
+        jsplumb-2.10.2
```
mprahl, line 59: Is this line needed?

tests/test_workers/test_pkg_managers/test_general_js.py
```diff
@@ -80,6 +95,13 @@ def test_download_dependencies(
             "version": "file:../jsplumb-2.10.2.tgz",
             "version_in_nexus": None,
         },
+        {
+            "bundled": False,
+            "dev": False,
+            "name": "exsp",
+            "version": "https://github.com/exsp/exsp/archive/2.10.2.tar.gz",
+            "version_in_nexus": "2.10.2-external",
```
mprahl, line 103:
```suggestion
            "version_in_nexus": "2.10.2-external-sha512-abcdefg",
```

cachito/workers/pkg_managers/general_js.py
nirzari: No, I was skeptical as well and felt it was unnecessary. Removing.

cachito/workers/pkg_managers/general_js.py
nirzari: This is outdated. Resolving

cachito/workers/pkg_managers/general_js.py
nirzari: This is outdated. Resolving

cachito/workers/pkg_managers/general_js.py
```diff
@@ -92,13 +94,23 @@ def download_dependencies(request_id, deps, proxy_repo_url, skip_deps=None):
...
         log.info("Processing %d npm dependencies to stage in Nexus", len(deps))
         downloaded_deps = set()
+
+        # Get Cachito supported prefixes. Include the part only before ":" in order to
+        # match conveniently with the dependency version
+        known_git_prefixes = {i.rsplit(":", 1)[0] for i in get_git_prefixes()[0]}
+        other_prefixes = {i.rsplit(":", 1)[0] for i in get_git_prefixes()[1]}.union(
+            i.rsplit(":", 1)[0] for i in get_http_prefixes()
+        )
+
```
mprahl, line 104: This can be removed now since you are pursuing the regex approach.

cachito/workers/pkg_managers/general_js.py
```diff
@@ -118,21 +120,40 @@ def download_dependencies(request_id, deps, proxy_repo_url, skip_deps=None):
...
+            dep_batch_download = [i[0] for i in dep_batch]
+            log.debug(
+                f"Downloading the following npm dependencies: {', '.join(dep_batch_download)}"
+            )
+            npm_pack_args = ["npm", "pack"] + dep_batch_download
             output = run_cmd(npm_pack_args, run_params, "Failed to download the npm dependencies")
 
             # Move dependencies to their respective folders
             # Iterate through the tuples made of dependency tarball and dep_identifier
             # e.g. ('angular-animations-8.2.0.tgz', '@angular/animations@8.2.0')
```
mprahl, line 139: This comment needs to be updated since the 1 index value is now a tuple.

cachito/workers/pkg_managers/general_js.py
```diff
@@ -118,21 +120,40 @@ def download_dependencies(request_id, deps, proxy_repo_url, skip_deps=None):
...
             for dep_pair in list(zip(output.split("\n"), dep_batch)):
                 tarball = dep_pair[0]  # e.g. angular-animations-8.2.0.tgz
-                dir_path = dep_pair[1].rsplit("@", 1)[0]  # e.g. @angular/animations
+                dir_path = dep_pair[1][0].rsplit("@", 1)[0]  # e.g. @angular/animations
```
mprahl, line 142: This makes it a little easier to read.
```suggestion
                dep_indentifer = dep_pair[1][0]
                dir_path = dep_indentifer.rsplit("@", 1)[0]  # e.g. @angular/animations
```

cachito/workers/pkg_managers/general_js.py
```diff
@@ -118,21 +120,40 @@ def download_dependencies(request_id, deps, proxy_repo_url, skip_deps=None):
...
+                external_dep_version = dep_pair[1][1]
+
+                # In case of external dependencies, create additional intermediate
+                # parent e.g. github/<org>/<repo> or external-<repo>
+                if external_dep_version:
+                    match = re.match(
+                        r"^(?P<host>.+)(?::)(?!//)(?P<repo_path>.+)(?:#.+)$", external_dep_version
+                    )
+                    if match:
+                        groups = match.groupdict()
```
mprahl, line 152: This makes it a little clearer what the regex is checking:
```suggestion
                    known_git_host_match = re.match(
                        r"^(?P<host>.+)(?::)(?!//)(?P<repo_path>.+)(?:#.+)$", external_dep_version
                    )
                    if known_git_host_match:
                        groups = known_git_host_match.groupdict()
```

tests/test_workers/test_pkg_managers/test_general_js.py
```diff
@@ -42,6 +42,7 @@ def test_download_dependencies(
         angular-devkit-architect-0.803.26.tgz
         angular-animations-8.2.14.tgz
         rxjs-6.5.5-external-gitcommit-78032157f5c1655436829017bbda787565b48c30.tgz
+        exsp-external-archive-2.10.2.tar.gz
```
mprahl, line 45: This should be named like this right?
```suggestion
        exsp-2.10.2-external-sha512-abcdefg.tar.gz
```

PR #196, opened 2020-07-16
Add integration test test_run_app_from_bundle

ejegrova:
Adding a test that will check that the application from bundle can run properly.

### DIFF
--- a/tests/integration/test_run_app_from_bundle.py
+++ b/tests/integration/test_run_app_from_bundle.py
@@ -1,0 +1,57 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from os import path
+import shutil
+import subprocess
+import tarfile
+
+import pytest
+
+import utils
+
+
+@pytest.mark.skipif(not shutil.which("go"), reason="requires go to be installed")
+def test_run_app_from_bundle(test_env, default_request, tmpdir):
+    """
+    Check that downloaded bundle could be used to run the application.
+
+    Process:
+    * Send new request to Cachito API
+    * Download a bundle from the request
+    * Run go build
+    * Run the application
+
+    Checks:
+    * Check that the state of request is complete
+    * Check that the bundle is properly downloaded
+    * Check that the application runs successfully
+    """
+    response = default_request.complete_response
+    assert response.status == 200
+    assert response.data["state"] == "complete"
+
+    file_name_tar = tmpdir.join(f"download_{str(response.id)}.tar.gz")
+    bundle_dir = tmpdir.mkdir(f"download_{str(response.id)}")
+    client = utils.Client(test_env["api_url"], test_env["api_auth_type"])
+    resp = client.download_bundle(response.id, file_name_tar)
+    assert resp.status == 200
+    assert tarfile.is_tarfile(file_name_tar)
+
+    with tarfile.open(file_name_tar, "r:gz") as tar:
+        tar.extractall(bundle_dir)
+
+    app_name = test_env["run_app"]["app_name"]
+    app_binary_file = str(tmpdir.join(app_name))
+    subprocess.run(
+        ["go", "build", "-o", app_binary_file, str(bundle_dir.join("app", "main.go"))],
+        env={
+            "GOPATH": str(bundle_dir.join("deps", "gomod")),
+            "GOCACHE": str(bundle_dir.join("deps", "gomod")),
+        },
+        cwd=str(bundle_dir.join("app")),
+        check=True,
+    )
+
+    assert path.exists(app_binary_file)
+    sp = subprocess.run([app_binary_file, "--help"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    assert sp.returncode == 0
--- a/test_env_vars.yaml
+++ b/test_env_vars.yaml
@@ -71,3 +71,7 @@ http_get_all:
 downloaded_output:
   # Package managers of cached dependencies
   pkg_managers: ["gomod"]
+# Test data for the Run application from bundle test
+run_app:
+  # Name of the applicaiton that is downloaded via bundle
+  app_name: retrodep


ejegrova:
@mprahl @lcarva @akhmanova ptal :-)

tests/integration/test_run_app_from_bundle.py
```diff
@@ -0,0 +1,50 @@
...
+    * Check that the state of request is complete
+    * Check that the bundle is properly downloaded
+    * Check that the application runs successfully
+    """
+    response = default_request.complete_response
+    assert response.status == 200
+    assert response.data["state"] == "complete"
+
+    file_name_tar = tmpdir.join(f"download_{str(response.id)}.tar.gz")
+    file_name = tmpdir.join(f"download_{str(response.id)}")
```
mprahl, line 29: This would make it easier to read.
```suggestion
    bundle_dir = tmpdir.join(f"download_{str(response.id)}")
```

mprahl, line 29: It seems like it doesn't matter, but it might be clearer if you used mkdir:
```suggestion
    bundle_dir = tmpdir.mkdir(f"download_{str(response.id)}")
```

tests/integration/test_run_app_from_bundle.py
```diff
@@ -0,0 +1,50 @@
...
+    resp = client.download_bundle(response.id, file_name_tar)
+    assert resp.status == 200
+    assert tarfile.is_tarfile(file_name_tar)
+
+    with tarfile.open(file_name_tar, "r:gz") as tar:
+        tar.extractall(file_name)
+
+    app_name = test_env["run_app"]["app_name"]
+    subprocess.run(
+        ["go", "build", "-o", f"{file_name}/app/bin/{app_name}", f"{file_name}/app/main.go"],
```
mprahl, line 40: Would this work so that you don't assume a Linux/Unix style path?
```suggestion
        ["go", "build", "-o", str(file_name.join("app", "bin", app_name)), str(file_name.join("app", "main.go"))],
```

tests/integration/test_run_app_from_bundle.py
```diff
@@ -0,0 +1,50 @@
...
+        env={"GOPATH": f"{file_name}/deps/gomod", "GOCACHE": f"{file_name}/deps/gomod"},
```
mprahl, line 41: Would this work so that you don't assume a Linux/Unix style path?
```suggestion
        env={"GOPATH": str(file_name.join("deps", "gomod")), "GOCACHE": str(file_name.join("deps", "gomod"))},
```

tests/integration/test_run_app_from_bundle.py
```diff
@@ -0,0 +1,50 @@
...
+        cwd=f"{file_name}/app",
```
mprahl, line 42: Would this work so that you don't assume a Linux/Unix style path?
```suggestion
        cwd=str(file_name.join("app")),
```

tests/integration/test_run_app_from_bundle.py
```diff
@@ -0,0 +1,50 @@
...
+        check=True,
+    )
+
+    sp = subprocess.run(
+        f"{file_name}/app/bin/{app_name}", stdout=subprocess.PIPE, stderr=subprocess.PIPE
```
mprahl, line 47: Would this work so that you don't assume a Linux/Unix style path?
```suggestion
        str(file_name.join("app", "bin", app_name)), stdout=subprocess.PIPE, stderr=subprocess.PIPE
```

tests/integration/test_run_app_from_bundle.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+
+    assert f"{app_name}:" in str(sp.stderr)
```
mprahl, line 50: Could you please verify the exit code and if the binary exists?

tests/integration/test_run_app_from_bundle.py
```diff
@@ -0,0 +1,62 @@
...
+    with tarfile.open(file_name_tar, "r:gz") as tar:
+        tar.extractall(bundle_dir)
+
+    app_name = test_env["run_app"]["app_name"]
+    subprocess.run(
+        [
+            "go",
+            "build",
+            "-o",
+            str(bundle_dir.join("app", "bin", app_name)),
```
mprahl, line 45: Optional: I think it'd be better if this wasn't in the the bundle directory.
```suggestion
            str(tmpdir.join(app_name)),
```

mprahl, line 45: It'd be to store this in a variable since you reuse this path below to execute the output binary.

tests/integration/test_run_app_from_bundle.py
```diff
@@ -0,0 +1,62 @@
...
+    client = utils.Client(test_env["api_url"], test_env["api_auth_type"])
+    resp = client.download_bundle(response.id, file_name_tar)
+    assert resp.status == 200
+    assert tarfile.is_tarfile(file_name_tar)
+
+    with tarfile.open(file_name_tar, "r:gz") as tar:
+        tar.extractall(bundle_dir)
+
+    app_name = test_env["run_app"]["app_name"]
+    subprocess.run(
```
mprahl, line 40: Could you please check the exit code of the build step?

tests/integration/test_run_app_from_bundle.py
```diff
@@ -0,0 +1,62 @@
...
+        env={
+            "GOPATH": str(bundle_dir.join("deps", "gomod")),
+            "GOCACHE": str(bundle_dir.join("deps", "gomod")),
+        },
+        cwd=str(bundle_dir.join("app")),
+        check=True,
+    )
+
+    sp = subprocess.run(
+        str(bundle_dir.join("app", "bin", app_name)), stdout=subprocess.PIPE, stderr=subprocess.PIPE
```
mprahl, line 57: Could you run this with the `--help` argument so that the exit code is `0`? This will tell us that the command executed successfully.

tests/integration/test_run_app_from_bundle.py
ejegrova: @mprahl I thought that argument `check=True` is doing this.

ejegrova:
@mprahl please check again

tests/integration/test_run_app_from_bundle.py
```diff
@@ -0,0 +1,53 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from os import path
+import subprocess
+import tarfile
+
+import utils
+
+
+def test_run_app_from_bundle(test_env, default_request, tmpdir):
```
mprahl, line 10: It'd be nice to skip this test if `go` is not installed on the system:
```suggestion
@pytest.mark.skipif(not shutil.which("go"), reason="requires go to be installed")
def test_run_app_from_bundle(test_env, default_request, tmpdir):
```

PR #197, opened 2020-07-16
Add image_contents data to npm content manifests

athos-ribeiro:

### DIFF
--- a/cachito/web/content_manifest.py
+++ b/cachito/web/content_manifest.py
@@ -23,6 +23,8 @@ class ContentManifest:
         self._gopkg_data = {}
         # dict to store go module level purl dependencies. Module names are used as keys
         self._gomod_data = {}
+        # dict to store npm package data; uses the package purl as key to identify a package
+        self._npm_data = {}
 
     def process_gomod(self, package, dependency):
         """
@@ -42,8 +44,8 @@ class ContentManifest:
         :param Package package: the go-package package to process
         :param Dependency dependency: the go-package package dependency to process
         """
-        purl = package.to_purl()
         if dependency.type == "go-package":
+            purl = package.to_purl()
             icm_dependency = {"purl": dependency.to_purl()}
             self._gopkg_data[purl]["dependencies"].append(icm_dependency)
 
@@ -71,6 +73,20 @@ class ContentManifest:
                     flask.current_app.logger.warning("Could not find a Go module for %s", purl)
             pkg_data.pop("name")
 
+    def process_npm_package(self, package, dependency):
+        """
+        Process npm package.
+
+        :param Package package: the npm package to process
+        :param Dependency dependency: the npm package dependency to process
+        """
+        if dependency.type == "npm":
+            purl = package.to_purl()
+            icm_dependency = {"purl": dependency.to_purl()}
+            self._npm_data[purl]["sources"].append(icm_dependency)
+            if not dependency.dev:
+                self._npm_data[purl]["dependencies"].append(icm_dependency)
+
     def to_json(self):
         """
         Generate the JSON representation of the content manifest.
@@ -80,6 +96,7 @@ class ContentManifest:
         """
         self._gopkg_data = {}
         self._gomod_data = {}
+        self._npm_data = {}
 
         # Address the possibility of packages having no dependencies
         for package in self.request.packages:
@@ -90,6 +107,9 @@ class ContentManifest:
                 )
             elif package.type == "gomod":
                 self._gomod_data.setdefault(package.name, [])
+            elif package.type == "npm":
+                purl = package.to_purl()
+                self._npm_data.setdefault(purl, {"purl": purl, "dependencies": [], "sources": []})
             else:
                 flask.current_app.logger.debug(
                     "No ICM implementation for '%s' packages", package.type
@@ -98,13 +118,15 @@ class ContentManifest:
         for req_dep in self.request.request_dependencies:
             if req_dep.package.type == "go-package":
                 self.process_go_package(req_dep.package, req_dep.dependency)
-            if req_dep.package.type == "gomod":
+            elif req_dep.package.type == "gomod":
                 self.process_gomod(req_dep.package, req_dep.dependency)
+            elif req_dep.package.type == "npm":
+                self.process_npm_package(req_dep.package, req_dep.dependency)
 
         # Adjust source level dependencies for go packages
         self.set_go_package_sources()
 
-        top_level_packages = list(self._gopkg_data.values())
+        top_level_packages = list(self._gopkg_data.values()) + list(self._npm_data.values())
         return self.generate_icm(top_level_packages)
 
     def generate_icm(self, image_contents=None):
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -4,6 +4,7 @@ from copy import deepcopy
 from enum import Enum
 import os
 import re
+import urllib.parse
 
 import flask
 from flask_login import UserMixin, current_user
@@ -184,8 +185,37 @@ class Package(db.Model):
         if self.type in ("go-package", "gomod"):
             # Use only the PURL "name" field to avoid ambiguity for Go modules/packages
             # see https://github.com/package-url/purl-spec/issues/63 for further reference
-            purl_name = self.name.replace("/", "%2F")
+            purl_name = urllib.parse.quote(self.name, safe="")
             return f"pkg:golang/{purl_name}@{self.version}"
+
+        elif self.type == "npm":
+            purl_name = urllib.parse.quote(self.name)
+            match = re.match(
+                r"(?P<protocol>[^:]+):(?P<has_authority>//)?(?P<suffix>.+)", self.version
+            )
+            if not match:
+                return f"pkg:npm/{purl_name}@{self.version}"
+            protocol = match.group("protocol")
+            suffix = match.group("suffix")
+            has_authority = match.group("has_authority")
+            if not has_authority:
+                # github:namespace/name#ref or gitlab:ns1/ns2/name#ref
+                match_forge = re.match(
+                    r"(?P<namespace>.+)/(?P<name>[^#/]+)#(?P<version>.+)$", suffix
+                )
+                if not match_forge:
+                    raise ValueError(f"Could not convert version {self.version} to purl")
+                forge = match_forge.groupdict()
+                return f"pkg:{protocol}/{forge['namespace']}/{forge['name']}@{forge['version']}"
+            elif protocol in ("git", "git+http", "git+https", "git+ssh"):
+                qualifier = urllib.parse.quote(self.version, safe="")
+                return f"pkg:generic/{purl_name}?vcs_url={qualifier}"
+            elif protocol in ("http", "https"):
+                qualifier = urllib.parse.quote(self.version, safe="")
+                return f"pkg:generic/{purl_name}?download_url={qualifier}"
+            else:
+                raise ValueError(f"Unknown protocol in npm package version: {self.version}")
+
         else:
             raise ValueError(f"The PURL spec is not defined for {self.type} packages")
 
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -13,7 +13,7 @@ import sqlalchemy.sql
 from werkzeug.exceptions import Forbidden
 
 from cachito.web import content_manifest
-from cachito.errors import ValidationError
+from cachito.errors import ContentManifestError, ValidationError
 from cachito.web import db
 
 
@@ -180,7 +180,7 @@ class Package(db.Model):
 
         :return: the PURL string of the Package object
         :rtype: str
-        :raise ValueError: if the there is no implementation for the package type
+        :raise ContentManifestError: if the there is no implementation for the package type
         """
         if self.type in ("go-package", "gomod"):
             # Use only the PURL "name" field to avoid ambiguity for Go modules/packages
@@ -204,7 +204,7 @@ class Package(db.Model):
                     r"(?P<namespace>.+)/(?P<name>[^#/]+)#(?P<version>.+)$", suffix
                 )
                 if not match_forge:
-                    raise ValueError(f"Could not convert version {self.version} to purl")
+                    raise ContentManifestError(f"Could not convert version {self.version} to purl")
                 forge = match_forge.groupdict()
                 return f"pkg:{protocol}/{forge['namespace']}/{forge['name']}@{forge['version']}"
             elif protocol in ("git", "git+http", "git+https", "git+ssh"):
@@ -214,10 +214,12 @@ class Package(db.Model):
                 qualifier = urllib.parse.quote(self.version, safe="")
                 return f"pkg:generic/{purl_name}?download_url={qualifier}"
             else:
-                raise ValueError(f"Unknown protocol in npm package version: {self.version}")
+                raise ContentManifestError(
+                    f"Unknown protocol in npm package version: {self.version}"
+                )
 
         else:
-            raise ValueError(f"The PURL spec is not defined for {self.type} packages")
+            raise ContentManifestError(f"The PURL spec is not defined for {self.type} packages")
 
 
 class Dependency(Package):
--- a/tests/test_api_v1.py
+++ b/tests/test_api_v1.py
@@ -1420,7 +1420,7 @@ def test_request_fetch_request_content_manifest_invalid(client, worker_auth_env)
 
 
 @pytest.mark.parametrize("state", ["complete", "stale", "in_progress", "failed"])
-def test_fetch_request_content_manifest(
+def test_fetch_request_content_manifest_go(
     app,
     client,
     db,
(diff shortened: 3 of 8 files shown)


cachito/web/models.py
```diff
@@ -186,6 +186,30 @@ def to_purl(self):
...
+                match_forge = re.match(
+                    r"(?P<namespace>[^/]+)/(?P<name>[^#]+)#(?P<version>.+)$", suffix
+                )
+                if not match_forge:
+                    raise ValueError(f"Could not convert version {self.version} to purl")
+                forge = match_forge.groupdict()
+                return f"pkg:{protocol}/{forge['namespace']}/{forge['name']}@{forge['version']}"
+            elif protocol in ("git", "git+http", "git+https", "git+ssh"):
+                return f"pkg:npm/{purl_name}?vcs_url={self.version}"
+            elif protocol in ("http", "https"):
```
athos-ribeiro, line 207: @mprahl do you have any examples for this case? I am not sure what to assume here (e.g., git repos, external npm repos, tarballs, etc)

mprahl:
@athos-ribeiro could you please rebase?

cachito/web/models.py
mprahl: This would just be a normal URL. For example:
https://github.com/jsplumb/jsplumb/archive/2.10.2.tar.gz

cachito/web/models.py
```diff
@@ -186,6 +186,30 @@ def to_purl(self):
...
+            if not match:
+                return f"pkg:npm/{purl_name}@{self.version}"
+            protocol = match.group("protocol")
+            suffix = match.group("suffix")
+            if protocol in ("github", "gitlab", "bitbucket"):  # github:namespace/name#ref
+                match_forge = re.match(
+                    r"(?P<namespace>[^/]+)/(?P<name>[^#]+)#(?P<version>.+)$", suffix
+                )
+                if not match_forge:
+                    raise ValueError(f"Could not convert version {self.version} to purl")
```
mprahl, line 202: FYI, raising this in the API will cause a 500 Internal Server Error response. I think that's okay since this is a sanity check that should never occur.

cachito/web/models.py
```diff
@@ -186,6 +186,30 @@ def to_purl(self):
             # see https://github.com/package-url/purl-spec/issues/63 for further reference
             purl_name = self.name.replace("/", "%2F")
             return f"pkg:golang/{purl_name}@{self.version}"
+
+        elif self.type == "npm":
+            purl_name = self.name.replace("@", "%40")
+            match = re.match(r"(?P<protocol>[^:]+):(?://)?(?P<suffix>.+)", self.version)
```
mprahl, line 192:
```suggestion
            match = re.match(r"^(?P<protocol>[^:]+):(?://)?(?P<suffix>.+)$", self.version)
```

athos-ribeiro:
> @athos-ribeiro could you please rebase?

Rebased

cachito/web/models.py
athos-ribeiro: Should we use some naive heuristics (e.g., checking for file extensions) here to try to figure out what kind of resource is provided by the URL?

This would impact on what purl param we add to it. Namely, `repository_url` or `download_url`

cachito/web/models.py
athos-ribeiro: Agreed. In this case it would be better to halt the request execution and set it to a failed state.

cachito/web/models.py
mprahl: @athos-ribeiro if you used `download_url`, would you need to have the version as well in the PURL? Cachito is aware of the version that is in the package.json of the tarball but it is not stored in the database, so it would not be available when generating the content manifest. My concern about adding the version is that the version can be inaccurate if the dependency comes from an HTTP URL as it may have diverged significantly from upstream.

My vote is either of the following options for the package jsplumb downloaded from a tarball:
* `pkg:npm/jsplumb@https%3A%2F%2Fgithub.com%2Fjsplumb%2Fjsplumb%2Farchive%2F2.10.2.tar.gz` (assuming the URL should be URL encoded)
* `pkg:npm/jsplumb?download_url=https%3A%2F%2Fgithub.com%2Fjsplumb%2Fjsplumb%2Farchive%2F2.10.2.tar.gz`

cachito/web/models.py
```diff
@@ -186,6 +186,30 @@ def to_purl(self):
...
+                )
+                if not match_forge:
+                    raise ValueError(f"Could not convert version {self.version} to purl")
+                forge = match_forge.groupdict()
+                return f"pkg:{protocol}/{forge['namespace']}/{forge['name']}@{forge['version']}"
+            elif protocol in ("git", "git+http", "git+https", "git+ssh"):
+                return f"pkg:npm/{purl_name}?vcs_url={self.version}"
+            elif protocol in ("http", "https"):
+                # assume this is an npm repository
+                return f"pkg:npm/{purl_name}?repository_url={self.version}"
```
mprahl, line 209: I don't think `repository_url` is quite right in this case as I think it would only apply to if the dependency came from an alternate npm registry.

tests/test_api_v1.py
```diff
@@ -1488,7 +1488,90 @@ def test_fetch_request_content_manifest(
...
+    )
+    npm_pkg_purls = [Package.from_json(pkg).to_purl() for pkg in npm_pkgs]
+
+    image_contents = [
+        {"purl": npm_pkg_purls[0], "dependencies": [], "sources": []},
+        {"purl": npm_pkg_purls[1], "dependencies": [], "sources": []},
+    ]
+
+    for i, d in enumerate(npm_deps):
+        dep = 0
```
mprahl, line 1542: I think this would be easier to read with a more precise variable name.
```suggestion
        pkg_index = 0
```

tests/test_api_v1.py
```diff
@@ -1488,7 +1488,90 @@ def test_fetch_request_content_manifest(
...
+        "f4abcfdaf8247a6b074f94fa84f3846f82d781c6/atomic_reactor/schemas/content_manifest.json"
+    )
+    npm_pkg_purls = [Package.from_json(pkg).to_purl() for pkg in npm_pkgs]
+
+    image_contents = [
+        {"purl": npm_pkg_purls[0], "dependencies": [], "sources": []},
+        {"purl": npm_pkg_purls[1], "dependencies": [], "sources": []},
+    ]
+
+    for i, d in enumerate(npm_deps):
```
mprahl, line 1541: Optional: I think it'd be easier to read if you did something like this instead:
```suggestion
    for pkg, deps in ((npm_pkgs[0], npm_deps[:2]), (npm_pkgs[1], npm_deps[2:])):
```

tests/test_content_manifest.py
```diff
@@ -0,0 +1,80 @@
...
+    assert expected_purl in cm._npm_data
+    assert cm._npm_data == expected_contents
+
+
+def test_to_json():
+    assert True
+
+
+def test_generate_icm():
+    assert True
```
mprahl, line 80: I think I'll require better coverage than that 😄

mprahl:
@athos-ribeiro could you please verify that you have 100% coverage of the new code?

cachito/web/models.py
athos-ribeiro: We do not need the version here. I will change it to `download_url` then. Would this URL ever be an alternative registry though?

cachito/web/models.py
athos-ribeiro: `re.match` already starts the match from the beginning of the string. I am wondering if there are any benefits in saying you want to match the end of a string when searching for `.+`.

cachito/web/models.py
mprahl: @athos-ribeiro as of now, registry dependencies only come from the official npm registry, if that changes for some reason, we can reevaluate using the `download_url` field.

cachito/web/models.py
mprahl: Ah, you're right. I didn't know that `re.match` did that already.

cachito/web/models.py
```diff
@@ -186,6 +186,29 @@ def to_purl(self):
...
+                    r"(?P<namespace>[^/]+)/(?P<name>[^#]+)#(?P<version>.+)$", suffix
+                )
+                if not match_forge:
+                    raise ValueError(f"Could not convert version {self.version} to purl")
+                forge = match_forge.groupdict()
+                return f"pkg:{protocol}/{forge['namespace']}/{forge['name']}@{forge['version']}"
+            elif protocol in ("git", "git+http", "git+https", "git+ssh"):
+                return f"pkg:npm/{purl_name}?vcs_url={self.version}"
+            elif protocol in ("http", "https"):
+                return f"pkg:npm/{purl_name}?download_url={self.version}"
```
mprahl, line 208: Does `self.version` have to be URL encoded? Especially if the version is a URL.

cachito/web/models.py
```diff
@@ -186,6 +186,29 @@ def to_purl(self):
...
+            suffix = match.group("suffix")
+            if protocol in ("github", "gitlab", "bitbucket"):  # github:namespace/name#ref
+                match_forge = re.match(
+                    r"(?P<namespace>[^/]+)/(?P<name>[^#]+)#(?P<version>.+)$", suffix
+                )
+                if not match_forge:
+                    raise ValueError(f"Could not convert version {self.version} to purl")
+                forge = match_forge.groupdict()
+                return f"pkg:{protocol}/{forge['namespace']}/{forge['name']}@{forge['version']}"
+            elif protocol in ("git", "git+http", "git+https", "git+ssh"):
```
mprahl, line 205: Could you please add test coverage for this?

cachito/web/models.py
```diff
@@ -186,6 +186,29 @@ def to_purl(self):
...
+                return f"pkg:npm/{purl_name}?vcs_url={self.version}"
+            elif protocol in ("http", "https"):
```
mprahl, line 207: Could you please add test coverage for this?

cachito/web/models.py
```diff
@@ -186,6 +186,29 @@ def to_purl(self):
...
+            if not match:
+                return f"pkg:npm/{purl_name}@{self.version}"
+            protocol = match.group("protocol")
+            suffix = match.group("suffix")
+            if protocol in ("github", "gitlab", "bitbucket"):  # github:namespace/name#ref
+                match_forge = re.match(
+                    r"(?P<namespace>[^/]+)/(?P<name>[^#]+)#(?P<version>.+)$", suffix
+                )
+                if not match_forge:
+                    raise ValueError(f"Could not convert version {self.version} to purl")
```
mprahl, line 202: Could you please add test coverage for this?

athos-ribeiro: done

cachito/web/models.py
athos-ribeiro: Done

cachito/web/models.py
athos-ribeiro: Done

cachito/web/models.py
athos-ribeiro: You are right!

Done

athos-ribeiro:
Hi  @mprahl,

I updated the PR increasing coverage and addressing your comments.

The only thing left it to eduplicate some code similar to #192

cachito/web/models.py
```diff
@@ -186,6 +187,32 @@ def to_purl(self):
...
+                # github:namespace/name#ref or gitlab:ns1/ns2/name#ref
+                match_forge = re.match(
+                    r"(?P<namespace>.+)/(?P<name>[^#/]+)#(?P<version>.+)$", suffix
+                )
+                if not match_forge:
+                    raise ValueError(f"Could not convert version {self.version} to purl")
+                forge = match_forge.groupdict()
+                return f"pkg:{protocol}/{forge['namespace']}/{forge['name']}@{forge['version']}"
+            elif protocol in ("git", "git+http", "git+https", "git+ssh"):
+                qualifier = urllib.parse.quote(self.version, safe="")
```
mprahl, line 208: Interestingly, the purl-spec repo does not "quote" this:
https://github.com/package-url/purl-spec

cachito/web/models.py
```diff
@@ -186,6 +187,32 @@ def to_purl(self):
...
+                return f"pkg:generic/{purl_name}?vcs_url={qualifier}"
+            elif protocol in ("http", "https"):
+                qualifier = urllib.parse.quote(self.version, safe="")
```
mprahl, line 211: Interestingly, the purl-spec repo does not "quote" this:
https://github.com/package-url/purl-spec

cachito/web/models.py
```diff
@@ -186,6 +187,32 @@ def to_purl(self):
...
+                return f"pkg:npm/{purl_name}@{self.version}"
+            protocol = match.group("protocol")
+            suffix = match.group("suffix")
+            if protocol in ("github", "gitlab", "bitbucket"):
+                # github:namespace/name#ref or gitlab:ns1/ns2/name#ref
+                match_forge = re.match(
+                    r"(?P<namespace>.+)/(?P<name>[^#/]+)#(?P<version>.+)$", suffix
+                )
+                if not match_forge:
+                    raise ValueError(f"Could not convert version {self.version} to purl")
```
mprahl, line 204: @athos-ribeiro and @lcarva I've been thinking about this... do you think a message should be exposed to the user rather than just a generic "Internal Server Error"? This would then be recorded by OSBS and it would make it easier for us to debug later on.

We could create a new exception that is called "CachitoContentManifestInvalid" and register it with Flask here:
https://github.com/release-engineering/cachito/blob/5d5ce769021101a2e7308cc35de01d9775a28193/cachito/web/app.py#L99-L102

cachito/web/models.py
```diff
@@ -186,6 +187,32 @@ def to_purl(self):
...
+                forge = match_forge.groupdict()
+                return f"pkg:{protocol}/{forge['namespace']}/{forge['name']}@{forge['version']}"
+            elif protocol in ("git", "git+http", "git+https", "git+ssh"):
+                qualifier = urllib.parse.quote(self.version, safe="")
+                return f"pkg:generic/{purl_name}?vcs_url={qualifier}"
+            elif protocol in ("http", "https"):
+                qualifier = urllib.parse.quote(self.version, safe="")
+                return f"pkg:generic/{purl_name}?download_url={qualifier}"
+            else:
+                raise ValueError(f"Unknown protocol in npm package version: {self.version}")
```
mprahl, line 214: Same as above

tests/test_api_v1.py
```diff
@@ -1488,7 +1488,90 @@ def test_fetch_request_content_manifest(
...
+        d.pop("replaces")
+        is_dev = d.pop("dev")
+        p = Package.from_json(d).to_purl()
+        image_contents[pkg_index]["sources"].append({"purl": p})
+        if not is_dev:
+            image_contents[pkg_index]["dependencies"].append({"purl": p})
+
+    expected = {
+        "metadata": {"icm_version": 1, "icm_spec": json_schema_url, "image_layer_index": -1},
+        "image_contents": image_contents,
```
mprahl, line 1554: Instead of dynamically creating the `image_contents` value that is expected, could you just directly compare against a static dictionary? I know it's more lines, but it ensures that the functions such as `to_purl` that are used to generate this value are returning the correct value.

It also makes it easier to read.

cachito/web/models.py
athos-ribeiro: It does say that for qualifiers, "A value must be a percent-encoded string"

cachito/web/models.py
athos-ribeiro: It does say that for qualifiers, "A value must be a percent-encoded string"

tests/test_api_v1.py
athos-ribeiro: Done

cachito/web/models.py
athos-ribeiro: I like the idea. I did it in a separate commit.

cachito/web/models.py
```diff
@@ -186,6 +186,30 @@ def to_purl(self):
...
             return f"pkg:golang/{purl_name}@{self.version}"
+
+        elif self.type == "npm":
+            purl_name = self.name.replace("@", "%40")
+            match = re.match(r"(?P<protocol>[^:]+):(?://)?(?P<suffix>.+)", self.version)
+            if not match:
+                return f"pkg:npm/{purl_name}@{self.version}"
+            protocol = match.group("protocol")
+            suffix = match.group("suffix")
+            if protocol in ("github", "gitlab", "bitbucket"):  # github:namespace/name#ref
```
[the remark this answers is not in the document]
athos-ribeiro: I changed it to look if there is any authority declaration as well

cachito/web/models.py
```diff
@@ -186,6 +186,30 @@ def to_purl(self):
...
+                match_forge = re.match(
+                    r"(?P<namespace>[^/]+)/(?P<name>[^#]+)#(?P<version>.+)$", suffix
```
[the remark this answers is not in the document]
athos-ribeiro: fixed

cachito/web/content_manifest.py
```diff
@@ -71,6 +77,21 @@ def set_go_package_sources(self):
...
+
+        :param Package package: the npm package to process
+        :param Dependency dependency: the npm package dependency to process
+        """
+        if dependency.type == "npm":
+            purl = package.to_purl()
+            self._npm_data.setdefault(purl, {"purl": purl, "dependencies": [], "sources": []})
+            icm_dependency = {"purl": dependency.to_purl()}
+            self._npm_data[purl]["sources"].append(icm_dependency)
+            if not dependency.dev:
```
lcarva, line 92: Is there a test for this?

cachito/web/models.py
mprahl: It's interesting that the examples don't show that. It makes sense to me though.

cachito/web/models.py
```diff
@@ -179,15 +180,46 @@ def to_purl(self):
...
+        :raise ContentManifestError: if the there is no implementation for the package type
         """
         if self.type in ("go-package", "gomod"):
             # Use only the PURL "name" field to avoid ambiguity for Go modules/packages
             # see https://github.com/package-url/purl-spec/issues/63 for further reference
             purl_name = self.name.replace("/", "%2F")
             return f"pkg:golang/{purl_name}@{self.version}"
+
+        elif self.type == "npm":
+            purl_name = self.name.replace("@", "%40")
```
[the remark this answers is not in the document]
mprahl: So the [documentation](https://docs.npmjs.com/files/package.json#name) says "the name ends up being part of a URL, an argument on the command line, and a folder name. Therefore, the name can’t contain any non-URL-safe characters.".

Keep in mind though, the user could put whatever they want in that field and Cachito would honor it since just reads the `package.json` file. It might be best to just URL encode to be safe.

cachito/web/models.py
mprahl: Nice, this new commits looks great!

cachito/web/content_manifest.py
```diff
@@ -42,8 +45,11 @@ def process_go_package(self, package, dependency):
         :param Package package: the go-package package to process
         :param Dependency dependency: the go-package package dependency to process
         """
-        purl = package.to_purl()
         if dependency.type == "go-package":
+            purl = package.to_purl()
+            self._gopkg_data.setdefault(
+                purl, {"name": package.name, "purl": purl, "dependencies": [], "sources": []}
```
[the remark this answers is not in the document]
athos-ribeiro: This was there to aid testing. I should have added a comment in there.

I removed the line in all process methods and am simulating the to_json behavior prior testing this.

cachito/web/models.py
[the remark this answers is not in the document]
athos-ribeiro: I encoded it (and the go one as well) to account for any unexpected special characters.

cachito/web/content_manifest.py
athos-ribeiro: Yes, the test for this method has checks for `sources` and `dependencies`

tests/test_content_manifest.py
```diff
@@ -0,0 +1,262 @@
...
+            },
+            "pkg:github/open-cluster-management/security-middleware@i0am0a0commit0hash",
+            True,
+            True,
+        ],
+        [
+            {
+                "name": "security-middleware",
+                "type": "npm",
+                "version": f"gitlab:deep/nested/repo/security-middleware#i0am0a0commit0hash",
```
mprahl, line 199:
```suggestion
                "version": "gitlab:deep/nested/repo/security-middleware#i0am0a0commit0hash",
```

tests/test_content_manifest.py
```diff
@@ -0,0 +1,262 @@
...
+            "pkg:golang/example.com%2Forg%2Fproject@1.1.1",
+            True,
+            True,
+        ],
+        [{"name": "grc-ui", "type": "npm", "version": "1.0.0"}, "pkg:npm/grc-ui@1.0.0", True, True],
+        [
+            {
+                "name": "security-middleware",
+                "type": "npm",
+                "version": f"github:open-cluster-management/security-middleware#i0am0a0commit0hash",
```
mprahl, line 189:
```suggestion
                "version": "github:open-cluster-management/security-middleware#i0am0a0commit0hash",
```

tests/test_content_manifest.py
```diff
@@ -0,0 +1,262 @@
...
+            },
+            "pkg:gitlab/deep/nested/repo/security-middleware@i0am0a0commit0hash",
+            True,
+            True,
+        ],
+        [
+            {
+                "name": "fromgit",
+                "type": "npm",
+                "version": f"git://some.domain/my/project/repo.git#i0am0a0commit0hash",
```
mprahl, line 209:
```suggestion
                "version": "git://some.domain/my/project/repo.git#i0am0a0commit0hash",
```

tests/test_content_manifest.py
```diff
@@ -0,0 +1,262 @@
...
+                "%23i0am0a0commit0hash"
+            ),
+            True,
+            True,
+        ],
+        [
+            {
+                "name": "fromweb",
+                "type": "npm",
+                "version": f"https://some.domain/my/project/package.tar.gz",
```
mprahl, line 222:
```suggestion
                "version": "https://some.domain/my/project/package.tar.gz",
```

tests/test_content_manifest.py
```diff
@@ -0,0 +1,262 @@
...
+                "%2Fpackage.tar.gz"
+            ),
+            True,
+            True,
+        ],
+        [
+            {
+                "name": "fromunknown",
+                "type": "npm",
+                "version": f"unknown://some.domain/my/project/package.tar.gz",
```
mprahl, line 235:
```suggestion
                "version": "unknown://some.domain/my/project/package.tar.gz",
```

athos-ribeiro:
Updated README.md file to communicate content manifest support for npm

athos-ribeiro:
@lcarva, can we merge this one?

PR #203, opened 2020-07-20
Process Go modules without package in the root dir

athos-ribeiro:
- Process all go sub-packages present in the directory hierarchy

### DIFF
--- a/cachito/web/content_manifest.py
+++ b/cachito/web/content_manifest.py
@@ -21,21 +21,55 @@ class ContentManifest:
         self.request = request
         # dict to store go package level data; uses the package purl as key to identify a package
         self._gopkg_data = {}
+        # dict to store go module level purl dependencies. Module names are used as keys
+        self._gomod_data = {}
+
+    def process_gomod(self, package, dependency):
+        """
+        Process gomod package.
+
+        :param Package package: the gomod package to process
+        :param Dependency dependency: the gomod package dependency to process
+        """
+        if dependency.type == "gomod":
+            icm_source = {"purl": dependency.to_purl()}
+            self._gomod_data[package.name].append(icm_source)
 
     def process_go_package(self, package, dependency):
         """
-        Process gomod or go-package package.
+        Process go-package package.
 
-        :param Package package: the gomod or go-package package to process
-        :param Dependency dependency: the gomod or go-package package dependency to process
+        :param Package package: the go-package package to process
+        :param Dependency dependency: the go-package package dependency to process
         """
         purl = package.to_purl()
         if dependency.type == "go-package":
             icm_dependency = {"purl": dependency.to_purl()}
             self._gopkg_data[purl]["dependencies"].append(icm_dependency)
-        elif dependency.type == "gomod":
-            icm_source = {"purl": dependency.to_purl()}
-            self._gopkg_data[purl]["sources"].append(icm_source)
+
+    def set_go_package_sources(self):
+        """
+        Adjust source level dependencies for go packages.
+
+        Go packages are not related to Go modules in cachito's DB. However, Go
+        sources are retreived in a per module basis. To set the proper source
+        in each content manifest entry, we associate each Go package to a Go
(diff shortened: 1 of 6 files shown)


cachito/workers/pkg_managers/gomod.py
```diff
@@ -191,50 +191,62 @@ def resolve_gomod(app_source_path, request, dep_replacements=None):
...
-
-        pkg_level_deps = []
-        for line in go_list_deps_output.splitlines():
-            # line example: pkg.io/foo/bar pkg.io/foo/var v1.0.0 => pkg.io/foo/bar v1.0.1
-            parts = [part for part in line.split(" ")]
+        list_pkgs_cmd = ("go", "list", "-find", "./...")
+        go_list_pkgs_output = run_gomod_cmd(list_pkgs_cmd, run_params)
+        packages = []
+        for package in go_list_pkgs_output.splitlines():
+            if next((d for p in packages for d in p["pkg_deps"] if d["name"] == package), False):
```
mprahl, line 198: This is hard to read and iterates through the same data repeatedly. How about a set called `processed_pkg_deps` and then you can change this line to the following?
```suggestion
            if package in processed_pkg_deps:
```

It does use more memory, but then for every package in your iteration, it's only an O(1) operation and it's easier to read.

cachito/workers/pkg_managers/gomod.py
```diff
@@ -191,50 +191,62 @@ def resolve_gomod(app_source_path, request, dep_replacements=None):
...
+                # Go searches for packages in directories through a top-down approach. If a toplevel
+                # package is already listed as a dependency, we do not list it here, since its
+                # dependencies would also be listed in the parent package
+                log.debug("Package %s already listed. Skipping...", package)
```
mprahl, line 202:
```suggestion
                log.debug("Package %s is already listed as a package dependency. Skipping...", package)
```

cachito/web/content_manifest.py
```diff
@@ -21,21 +21,54 @@ def __init__(self, request=None):
...
             self._gopkg_data[purl]["dependencies"].append(icm_dependency)
-        elif dependency.type == "gomod":
-            icm_source = {"purl": dependency.to_purl()}
-            self._gopkg_data[purl]["sources"].append(icm_source)
+
+    def set_go_package_sources(self):
+        """
+        Adjust source level dependencies for go packages.
+
+        Go packages are not related to Go modules in cachito DB. However, Go
```
mprahl, line 54:
```suggestion
        Go packages are not related to Go modules in Cachito's DB. However, Go
```

cachito/web/content_manifest.py
```diff
@@ -21,21 +21,54 @@ def __init__(self, request=None):
...
+        in each content manifest entry, we associate each Go package to a Go
+        module based on their names.
+        """
+        for purl, pkg_data in self._gopkg_data.items():
+            if pkg_data["name"] in self._gomod_data:
+                self._gopkg_data[purl]["sources"] = self._gomod_data[pkg_data["name"]]
+            else:
+                previous_length = 0
+                for mod_name, sources in self._gomod_data.items():
+                    if pkg_data["name"].startswith(mod_name) and len(mod_name) > previous_length:
```
mprahl, line 65: To clarify, is the logic `len(mod_name) > previous_length` just to handle submodules/nested modules? If so, could you please add a comment explaining this for loop/if statement?

tests/test_api_v1.py
```diff
@@ -1512,16 +1512,6 @@ def test_fetch_request_content_manifest_non_implemented_type(
         p["type"] = pkg_type
 
     image_contents = []
-    if pkg_type == "gomod":
-        # gomod is a special case where pre-icm requests will have ICMs with
```
athos-ribeiro, line 1516: Note that this is no longer true. We can not rely on the existence of a package with the same name as the module. This could be wrong since the toplevel purl is for a package.

PR #206, opened 2020-07-23
Add the request ID to the Cachito worker logs

mprahl:
This will allow for easier debugging when multiple workers are running and the logs are aggregated.

### DIFF
--- a/cachito/workers/celery_logging.py
+++ b/cachito/workers/celery_logging.py
@@ -1,0 +1,111 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+import inspect
+import logging
+
+from celery.utils.log import ColorFormatter
+
+from cachito.workers.config import get_worker_config
+
+
+class AddRequestIDFilter(logging.Filter):
+    """A log filter that sets ``request_id`` on the log record."""
+
+    def __init__(self, request_id, *args, **kwargs):
+        """
+        Initialize a filter that sets ``request_id`` on the log record.
+
+        :param request_id: the request ID to set on the record
+        """
+        super().__init__(*args, **kwargs)
+        self._request_id = request_id
+
+    def filter(self, record):
+        """
+        Set ``request_id`` on the log record.
+
+        :param logging.LogRecord record: the log record
+        :return: always returns ``True`` so the log is not filtered out
+        :rtype: bool
+        """
+        record.request_id = self._request_id
+        return True
+
+
+def _get_function_arg_value(arg_name, func, args, kwargs):
+    """
+    Get the value of the given argument name.
+
+    :param str arg_name: the name of the argument to get
+    :param function func: the function the arguments are for
+    :param list args: the list of arguments passed to the function
+    :param dict kwargs: the dictionary or keyword arguments passed to the function
+    :return: the argument value or ``None``
+    """
+    original_func = func
+    while getattr(original_func, "__wrapped__", None):
+        original_func = original_func.__wrapped__
+    argspec = inspect.getfullargspec(original_func).args
+
+    arg_index = argspec.index(arg_name)
+    arg_value = kwargs.get(arg_name, None)
+    if arg_value is None and len(args) > arg_index:
+        arg_value = args[arg_index]
+
+    return arg_value
+
+
+def setup_task_logging(task_id, task, **kwargs):
+    """
+    Customize the logging for the task.
+
+    This adds a filter that sets ``request_id`` on the log record. If the request ID
+    cannot be determined, "unknown" is set instead. This also sets the log format to
+    the ``cachito_task_log_format`` config.
+
+    :param str task_id: the task ID
+    :param class task: the class of the task being executed
+    """
+    conf = get_worker_config()
+
+    request_id = _get_function_arg_value("request_id", task, kwargs["args"], kwargs["kwargs"])
+    log_filter = AddRequestIDFilter(str(request_id) or "unknown")
+    root_logger = logging.getLogger()
+    for handler in root_logger.handlers:
+        if not isinstance(handler, logging.StreamHandler):
+            continue
+
+        handler.addFilter(log_filter)
+        if isinstance(handler.formatter, ColorFormatter):
+            formatter = ColorFormatter(
+                conf.cachito_task_log_format, use_color=handler.formatter.use_color
+            )
+        else:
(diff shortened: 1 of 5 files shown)


mprahl:
@nirzari FYI, this might help you with your task. Instead of using a decorator like IIB does, you can use the Celery signals I'm using here.

mprahl:
FYI, if we stop using `logging.getLogger` in `cachito.workers` and use `cachito.utils.log.get_task_logger` instead, then we can use the `worker_task_log_format` configuration and stop modifying the log formatter in these signal handlers. The signal handlers would just then add and remove the new log filter.

cachito/workers/celery_logging.py
```diff
@@ -0,0 +1,111 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+import inspect
+import logging
+
+from celery.utils.log import ColorFormatter
+
+
+class RequestFilter(logging.Filter):
+    """A log filter that sets ``request_id`` on the log record."""
```
nirzari, line 9: Why are we using  `` (two of them) here ?

mprahl: This is because the docstrings are in the reStructuredText format:
https://docutils.sourceforge.io/docs/user/rst/quickref.html#inline-markup

cachito/workers/celery_logging.py
```diff
@@ -0,0 +1,111 @@
...
+    def __init__(self, request_id, *args, **kwargs):
+        """
+        Initialize a filter that sets ``request_id`` on the log record.
+
+        :param request_id: the request ID to set on the record
+        """
+        super().__init__(*args, **kwargs)
+        self._request_id = request_id
+
+    def filter(self, record):
```
nirzari, line 20: We discussed the other day how we are using the filter method to set the request id, but should we rename it to something better?

cachito/workers/celery_logging.py
```diff
@@ -0,0 +1,111 @@
...
+    Customize the logging for the task.
+
+    This adds a filter that sets ``request_id`` on the log record. If the request ID
+    cannot be determined, "unknown" is set instead. This also sets the log format to
+    the ``cachito_task_log_format`` config.
+
+    :param str task_id: the task ID
+    :param class task: the class of the task being executed
+    """
+    from cachito.workers.config import get_worker_config
```
nirzari, line 62: As discussed, should we move this to the top of the file?

cachito/workers/celery_logging.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def cleanup_task_logging(task_id, task, **kwargs):
+    """
+    Clean up any logging customizations that were set in ``setup_task_logging``.
+
+    :param str task_id: the task ID
+    :param class task: the class of the task being executed
+    """
+    from cachito.workers.config import get_worker_config
```
nirzari, line 92: As discussed, should we move this to the top of the file?

cachito/workers/celery_logging.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    Get the value of the given argument name.
+
+    :param str arg_name: the name of the argument to get
+    :param function func: the function the arguments are for
+    :param list args: the list of arguments passed to the function
+    :param dict kwargs: the dictionary or keyword arguments passed to the function
+    :return: the argument value or ``None``
+    """
+    argspec = inspect.getfullargspec(func).args
```
lcarva, line 42: Don't we need to handle the case where the actual function is decorated multiple times? I don't think accessing just the immediate `func.__wrapped__` is sufficient. It may work now, but as soon as we decorate it with another decorator it's going to fail.

```python
original_func = func                                
while getattr(func, '__wrapped__', None):           
    original_func = func.__wrapped__                
argspec = inspect.getfullargspec(original_func).args

```

cachito/workers/celery_logging.py
lcarva: > You don’t need to create specialized Filter classes, or use other classes with a filter method: you can use a function (or other callable) as a filter. The filtering logic will check to see if the filter object has a filter attribute: if it does, it’s assumed to be a Filter and its filter() method is called. Otherwise, it’s assumed to be a callable and called with the record as the single parameter. The returned value should conform to that returned by filter().

From https://docs.python.org/3/library/logging.html#logging.Filter

A simple callable would be nice, but we'll have to use closures to store the value of the `request_id`. Something like this:

```python
def make_filter(request_id):
  def set_request_id(record):
    record.request_id = request_id
    return True
  return set_request_id
```

Then use it like:
```python
log_filter = make_filter(str(request_id) or "unknown")
```

It'd be tricky to remove it, but certainly doable.

The class approach seems pretty straightforward to me as well.

cachito/workers/celery_logging.py
mprahl: You're right, but the suggestion you proposed is an infinite loop in Cachito. I filed a PR to fix this in IIB:
https://github.com/release-engineering/iib/pull/133

cachito/workers/celery_logging.py
mprahl: @lcarva I'm going to name it `AddRequestIDFilter`. It makes it easier to know what it is by the name. Thanks for the suggestion.

mprahl:
@athos-ribeiro and @lcarva thanks for the reviews. This is ready for another round.

cachito/workers/celery_logging.py
lcarva: Thanks for catching this!

tests/test_workers/test_celery_logging.py
```diff
@@ -0,0 +1,73 @@
...
+        original_formatters.append(handler.formatter)
+
+    color_handler = logging.StreamHandler()
+    color_handler.setFormatter(ColorFormatter("%(message)s"))
+    log_filter = celery_logging.AddRequestIDFilter(5)
+    color_handler.addFilter(log_filter)
+    root_logger.addHandler(color_handler)
+
+    try:
+        celery_logging.cleanup_task_logging(mock.Mock(), mock.Mock())
```
nirzari, line 65: Why aren't we testing in the try block if handler is removed?

mprahl: `cleanup_task_logging` doesn't remove the handler, it just replaces the formatter and adds a filter.

nirzari:
LGTM

PR #212, opened 2020-07-29
Verify checksum of external NPM dependencies

lcarva:
* CLOUDBLD-1890

### DIFF
--- a/cachito/workers/pkg_managers/general_js.py
+++ b/cachito/workers/pkg_managers/general_js.py
@@ -1,5 +1,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import base64
+import collections
+import hashlib
 import io
 import json
 import logging
@@ -32,6 +34,9 @@ __all__ = [
 ]
 
 log = logging.getLogger(__name__)
+
+
+ChecksumInfo = collections.namedtuple("ChecksumInfo", "algorithm hexdigest")
 
 
 def download_dependencies(request_id, deps, proxy_repo_url, skip_deps=None):
@@ -352,7 +357,9 @@ def prepare_nexus_for_js_request(repo_name):
         raise CachitoError("Failed to prepare Nexus for Cachito to stage JavaScript content")
 
 
-def upload_non_registry_dependency(dep_identifier, version_suffix, verify_scripts=False):
+def upload_non_registry_dependency(
+    dep_identifier, version_suffix, verify_scripts=False, checksum_info=None
+):
     """
     Upload the non-registry npm dependency to the Nexus hosted repository with a custom version.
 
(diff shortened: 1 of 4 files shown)


cachito/workers/pkg_managers/general_js.py
```diff
@@ -451,3 +462,30 @@ def upload_non_registry_dependency(dep_identifier, version_suffix, verify_script
...
         repo_name = get_js_hosted_repo_name()
         nexus.upload_artifact(repo_name, "npm", modified_dep_archive)
+
+
+def _verify_checksum(file_path, checksum_info, chunk_size=10240):
+    """
+    Verify the checksum of the file at the given path matches the expected checksum info.
+
+    :param str file_path: the path to the file to be verified
+    :param ChecksumInfo checksum_info: the expected checksum information
```
nirzari, line 472: Missing param `chunk_size` from the doc string

nirzari:
Left a small comment. Otherwise LGTM :+1:

cachito/workers/pkg_managers/general_js.py
lcarva: Good catch! Fixed it.

cachito/workers/pkg_managers/general_js.py
```diff
@@ -451,3 +462,31 @@ def upload_non_registry_dependency(dep_identifier, version_suffix, verify_script
...
+    :param int chunk_size: the amount of bytes to read at a time
+    :raise CachitoError: if the checksum is not as expected
+    """
+    hasher = hashlib.new(checksum_info.algorithm)
```
mprahl, line 476: Since the input to this comes from the package-lock.json file, in theory, the algorithm could be an arbitrary value. I think you should raise a nice error saying that the hash algorithm is not recognized.

tests/test_workers/test_pkg_managers/test_general_js.py
```diff
@@ -484,3 +494,28 @@ def test_upload_non_registry_dependency_invalid_package_json(
     expected = "The dependency star-wars@5.0.0 does not have a valid package.json file"
     with pytest.raises(CachitoError, match=expected):
         general_js.upload_non_registry_dependency("star-wars@5.0.0", "-the-empire-strikes-back")
+
+
+def test_verify_checksum(tmpdir):
+    file = tmpdir.join("spells.txt")
+    file.write("Beetlejuice! Beetlejuice! Beetlejuice!")
```
mprahl, line 501: Scary!

cachito/workers/pkg_managers/general_js.py
lcarva: That's a good point. Wrapped the error.

PR #213, opened 2020-08-03
Implement setup.cfg handling for Pip

chmeliik:
* CLOUDBLD-2053

Aims for near parity with the setuptools way of handling setup.cfg files
with the notable exception of never executing unknown Python modules.

There may be other minor differences, but should cover 99% of the
typical use cases.

### DIFF
--- a/cachito/workers/pkg_managers/pip.py
+++ b/cachito/workers/pkg_managers/pip.py
@@ -1,0 +1,334 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+import ast
+import configparser
+import logging
+import re
+from abc import ABC, abstractmethod
+from pathlib import Path
+
+import pkg_resources
+
+from cachito.errors import ValidationError
+
+
+log = logging.getLogger(__name__)
+
+NOTHING = object()  # A None replacement for cases where the distinction is needed
+
+
+def any_to_version(obj):
+    """
+    Convert any python object to a version string.
+
+    https://github.com/pypa/setuptools/blob/ba209a15247b9578d565b7491f88dc1142ba29e4/setuptools/config.py#L535
+
+    :param any obj: object to convert to version
+    :rtype: str
+    """
+    version = obj
+
+    if not isinstance(version, str):
+        if hasattr(version, "__iter__"):
+            version = ".".join(map(str, version))
+        else:
+            version = str(version)
+
+    return pkg_resources.safe_version(version)
+
+
+def get_top_level_attr(block_node, attr_name, before_line=None):
+    """
+    Get attribute from module if it is defined at top level and assigned to a literal expression.
+
+    https://github.com/pypa/setuptools/blob/ba209a15247b9578d565b7491f88dc1142ba29e4/setuptools/config.py#L36
+
+    Note that this approach is not equivalent to the setuptools one - setuptools looks for the
+    attribute starting from the top, we start at the bottom. Arguably, starting at the bottom
+    makes more sense, but it should not make any real difference in practice.
+
+    :param ast.AST block_node: Any AST node that has a 'body' attribute
+    :param str attr_name: Name of attribute to search for
+    :param int before_line: Only look for attributes defined before this line
+
+    :rtype: anything that can be expressed as a literal ("primitive" types, collections)
+    :raises AttributeError: If attribute not found or not assigned to a literal expression
+    """
+    if before_line is None:
+        before_line = float("inf")
+    try:
+        return next(
+            ast.literal_eval(node.value)
+            for node in reversed(block_node.body)
+            if node.lineno < before_line and isinstance(node, ast.Assign)
+            for target in node.targets
+            if isinstance(target, ast.Name) and target.id == attr_name
+        )
+    except ValueError:
+        raise AttributeError(f"Attribute {attr_name!r} is not assigned to a literal expression")
+    except StopIteration:
+        raise AttributeError(f"Attribute {attr_name!r} not found")
+
+
+class SetupFile(ABC):
+    """Abstract base class for setup.cfg and setup.py handling."""
+
+    def __init__(self, top_dir, file_name):
+        """
+        Initialize a SetupFile.
+
+        :param str top_dir: Path to root of project directory
+        :param str file_name: Name of Python setup file, expected to be in the root directory
+        """
+        self._top_dir = Path(top_dir).resolve()
+        self._path = self._top_dir / file_name
+
+    def exists(self):
+        """Check if file exists."""
+        return self._path.is_file()
+
+    @abstractmethod
+    def get_name(self):
+        """Attempt to determine the package name. Should only be called if file exists."""
+
+    @abstractmethod
+    def get_version(self):
+        """Attempt to determine the package version. Should only be called if file exists."""
+
+
+class SetupCFG(SetupFile):
+    """
+    Parse metadata.name and metadata.version from a setup.cfg file.
+
+    Aims to match setuptools behaviour as closely as possible, but does make
+    some compromises (such as never executing arbitrary Python code).
+    """
+
+    # Valid Python name - any sequence of \w characters that does not start with a number
+    _name_re = re.compile(r"[^\W\d]\w*")
+
+    def __init__(self, top_dir):
+        """
+        Initialize a SetupCFG.
+
+        :param str top_dir: Path to root of project directory
+        """
+        super().__init__(top_dir, "setup.cfg")
+        self.__parsed = NOTHING
+
+    def get_name(self):
+        """
+        Get metadata.name if present.
+
+        :rtype: str or None
+        """
+        name = self._get_option("metadata", "name")
+        if name is None:
+            log.debug("No metadata.name in setup.cfg")
+            return None
+
+        log.debug("Found metadata.name in setup.cfg: %r", name)
+        return name
+
+    def get_version(self):
+        """
+        Get metadata.version if present.
+
+        Partially supports the file: directive (setuptools supports multiple files
+        as an argument to file:, this makes no sense for version).
+
+        Partially supports the attr: directive (will only work if the attribute
+        being referenced is assigned to a literal expression).
+
+        :rtype: str or None
+        """
+        version = self._get_option("metadata", "version")
+        if version is None:
+            log.debug("No metadata.version in setup.cfg")
+            return None
+
+        log.debug("Resolving metadata.version in setup.cfg from %r", version)
+        version = self._resolve_version(version)
+        if version is None:
+            log.debug("Failed to resolve metadata.version in setup.cfg")
+            return None
+
+        version = any_to_version(version)
+        log.debug("Found metadata.version in setup.cfg: %r", version)
+        return version
+
+    @property
+    def _parsed(self):
+        """
+        Try to parse config file, return None if parsing failed.
+
+        Will not parse file (or try to) more than once.
+        """
+        if self.__parsed is NOTHING:  # Have not tried to parse file yet
+            log.debug("Parsing setup.cfg at %r", str(self._path))
+            parsed = configparser.ConfigParser()
+
+            with self._path.open() as f:
+                try:
+                    parsed.read_file(f)
+                    self.__parsed = parsed
+                except configparser.Error as e:
+                    log.debug("Failed to parse setup.cfg: %s", e)
+                    self.__parsed = None  # Tried to parse file and failed
+
+        return self.__parsed
+
+    def _get_option(self, section, option):
+        """Get option from config section, return None if option missing or file invalid."""
+        if self._parsed is None:
+            return None
+        try:
+            return self._parsed.get(section, option)
+        except (configparser.NoSectionError, configparser.NoOptionError):
+            return None
+
+    def _resolve_version(self, version):
+        """Attempt to resolve the version attribute."""
+        if version.startswith("file:"):
+            file_arg = version[len("file:") :].strip()
+            version = self._read_version_from_file(file_arg)
+        elif version.startswith("attr:"):
+            attr_arg = version[len("attr:") :].strip()
+            version = self._read_version_from_attr(attr_arg)
+        return version
+
+    def _read_version_from_file(self, file_path):
+        """Read version from file after making sure file is a subpath of project dir."""
+        full_file_path = self._ensure_local(file_path)
+        if full_file_path.is_file():
+            version = full_file_path.read_text().strip()
+            log.debug("Read version from %r: %r", file_path, version)
+            return version
+        else:
+            log.debug("Version file %r does not exist or is not a file", file_path)
+            return None
+
+    def _ensure_local(self, path):
+        """Check that path is a subpath of project directory, return resolved path."""
+        full_path = (self._top_dir / path).resolve()
+        try:
+            full_path.relative_to(self._top_dir)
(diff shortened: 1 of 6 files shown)


cachito/workers/pkg_managers/pip.py
```diff
@@ -0,0 +1,321 @@
...
+        except (configparser.NoSectionError, configparser.NoOptionError):
+            return None
+
+    def _resolve_version(self, version):
+        """Attempt to resolve the version attribute."""
+        if version.startswith("file:"):
+            file_arg = version[len("file:") :].strip()
+            version = self._read_version_from_file(file_arg)
+        elif version.startswith("attr:"):
+            attr_arg = version[len("attr:") :].strip()
```
chmeliik, line 182: This is the CI failure (`black` and `flake8` don't agree on the space before `:`) - permission to silence flake8?

chmeliik:
Might want to add logging checks to unit tests, other than that should be pretty much done

cachito/workers/pkg_managers/pip.py
MartinBasti: why space before `:` ?

chmeliik: IDK, that's how `black` wants to format it

MartinBasti: ok, I disagree with black

chmeliik: Silenced `black` instead, if anyone disagrees let me know :-)

lcarva: Can we adjust the configuration instead?

chmeliik: `black` doesn't have any configuration, from the [README](https://github.com/psf/black/blob/master/README.md):

> Black is a PEP 8 compliant opinionated formatter. Black reformats entire files in place. It is not configurable. It doesn't take previous formatting into account. Your main option of configuring Black is that it doesn't reformat blocks that start with # fmt: off and end with # fmt: on.

We could adjust the `flake8` config, not sure what would be the preference

lcarva: The rule of thumb on this project so far has been to always agree with `black`. Even though we may disagree with some of its choices, it's still worth for all the other parts that it gets correct. I suggest we disable the `E203` check in flake8.

chmeliik: Ok, I'll go with `black`. E203 seems valid outside of this specific scenario, I'll just disable the `flake8` check inline

chmeliik: Nevermind, `black` will catch the valid cases, disabling in config

chmeliik:
Removed duplicate test case

docker/Dockerfile-api
```diff
@@ -21,6 +21,7 @@ RUN dnf -y install \
     python3-requests \
     python3-requests-kerberos \
     python3-semver \
+    python3-setuptools \
```
athos-ribeiro, line 24: Do we really need this here as well?

chmeliik:
Added logging checks and finally figured out why they were not working

docker/Dockerfile-api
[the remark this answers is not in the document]
chmeliik: Not sure, probably not. As far as I can tell, `Dockerfile-api` installs everything in `requirements.txt + requirements-web.txt`, and `Dockerfile-workers` installs only `requirements.txt`, so I wanted to keep it that way

cachito/workers/pkg_managers/pip.py
```diff
@@ -0,0 +1,321 @@
...
+    :param any obj: object to convert to version
+    :rtype: str
+    """
+    version = obj
+
+    if not isinstance(version, str):
+        if hasattr(version, "__iter__"):
+            version = ".".join(map(str, version))
+        else:
+            version = str(version)
```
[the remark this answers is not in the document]
chmeliik: Added, thanks!

cachito/workers/pkg_managers/pip.py
```diff
@@ -0,0 +1,321 @@
...
+        :rtype: dict[str, str] or None
+        """
+        package_dir_value = self._get_option("options", "package_dir")
+        if package_dir_value is None:
+            return None
+
+        if "\n" in package_dir_value:
+            package_items = package_dir_value.splitlines()
+        else:
+            package_items = package_dir_value.split(",")
```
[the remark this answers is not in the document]
chmeliik: Added, thanks!

docker/Dockerfile-api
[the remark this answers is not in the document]
athos-ribeiro: Fair enough

cachito/workers/pkg_managers/pip.py
```diff
@@ -0,0 +1,321 @@
...
+        super().__init__(top_dir, "setup.cfg")
+        self._parsed = None
+
+    def get_name(self):
+        """
+        Get metadata.name if present.
+
+        :rtype: str or None
+        """
+        self._parse()
```
athos-ribeiro, line 122: Why are we calling `self._parse()` both here and in `get_version` and using `self._parsed` in `get_option` if `self._parse()` sets and returns `self._parsed`? i.e., why don't we use `self._parse()` in `get_option` and remove the other self._parse() calls?

Alternatively, we could remove the `self._parse()` return value since it is not being used.

lcarva: Conversely, change `_parsed` to `__parsed`, and create a property `_parsed`. That way it gets parsed when, and only when, it's needed.

cachito/workers/pkg_managers/pip.py
```diff
@@ -0,0 +1,321 @@
...
+        if package_dir_value is None:
+            return None
+
+        if "\n" in package_dir_value:
+            package_items = package_dir_value.splitlines()
+        else:
+            package_items = package_dir_value.split(",")
+
+        # Strip whitespace and discard empty values
+        package_items = filter(bool, (p.strip() for p in package_items))
```
lcarva, line 312: This is neat! It avoids the repetition seen in a pure list comprehension approach: 
```python
(p.strip() for p in package_items if p.strip())
```

cachito/workers/pkg_managers/pip.py
```diff
@@ -0,0 +1,321 @@
...
+class SetupCFG(SetupFile):
+    """
+    Parse metadata.name and metadata.version from a setup.cfg file.
+
+    Aims to match setuptools behaviour as closely as possible, but does make
+    some compromises (such as never executing arbitrary Python code).
+    """
+
+    # Valid Python name - any sequence of \w characters that does not start with a number
+    _name_re = re.compile(r"[^\W\d]\w*")
```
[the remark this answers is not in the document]
chmeliik: FWIW, the pattern does also support non-ASCII names:
```python
In [7]: pat = re.compile(r"[^\W\d]\w*")                                                       

In [8]: pat.fullmatch('Žluťoučký')                                                            
Out[8]: <re.Match object; span=(0, 9), match='Žluťoučký'>

In [9]: pat.fullmatch('こんにちは')                                                           
Out[9]: <re.Match object; span=(0, 5), match='こんにちは'>
```
So we might get false positives (at least on py2), but unless the user is actively trying to fail their build, it should be fine :D

cachito/workers/pkg_managers/pip.py
```diff
@@ -0,0 +1,321 @@
...
+    def get_name(self):
+        """
+        Get metadata.name if present.
+
+        :rtype: str or None
+        """
+        self._parse()
+        name = self._get_option("metadata", "name")
+        if name is None:
+            log.debug("No metadata.name in setup.cfg")
```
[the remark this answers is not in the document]
chmeliik: I suppose I could use `self._path.name` instead, but that will just end up being more typing with not that much upside. I was thinking of somehow injecting the filename into the log messages automatically, but I don't have much experience with more advanced Python logging. Ideally, it would be something like
```python
self.log.debug("Found something in %(filename)s: %s", some_value)  # automatically injects the `filename` arg
```
But so far, I haven't figured out how to do that properly, any ideas?

Worst case scenario, I can fall back to using `self._path.name` everywhere if desirable

cachito/workers/pkg_managers/pip.py
chmeliik: Right? That's pretty much exactly what setuptools does in their code, so of course I thought: "Ha! I can do you one better."

cachito/workers/pkg_managers/pip.py
```diff
@@ -0,0 +1,321 @@
...
+
+class SetupFile(ABC):
+    """Abstract base class for setup.cfg and setup.py handling."""
+
+    def __init__(self, top_dir, file_name):
+        """
+        Initialize a SetupFile.
+
+        :param str top_dir: Path to root of project directory
+        :param str file_name: Either "setup.cfg" or "setup.py"
```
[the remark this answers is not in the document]
chmeliik: Good point, changed

cachito/workers/pkg_managers/pip.py
chmeliik: That is better, thanks! I also realized I'm currently ignoring the possibility that setup.cfg parsing would fail, not sure if that should be `ValidationError` or a "failed to resolve", I'm thinking the latter

chmeliik:
Added handling for malformed `.cfg` files and the corresponding unit test scenarios

tox.ini
```diff
@@ -42,7 +42,8 @@ show-source = True
 max-line-length = 100
 exclude = venv,.git,.tox,dist,*egg,cachito/web/migrations,.env
 # W503 line break before binary operator
-ignore = D100,D104,D105,W503
+# E203 whitespace before ':'
+ignore = D100,D104,D105,W503,E203
```
athos-ribeiro, line 46: The reason this is not disabled inline, is that black will also deal with the same issue. Is this right? If so, can we add a comment on that?

cachito/workers/pkg_managers/pip.py
```diff
@@ -0,0 +1,320 @@
...
+        the module if this fails.
+
+        https://github.com/pypa/setuptools/blob/ba209a15247b9578d565b7491f88dc1142ba29e4/setuptools/config.py#L354
+
+        :param str attr_spec: "import path" of attribute, e.g. package.version.__version__
+        :rtype: str or None
+        """
+        module_name, _, attr_name = attr_spec.rpartition(".")
+        if not module_name:
+            # Assume current directory is a module, look for attribute in __init__.py
```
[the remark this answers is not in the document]
chmeliik: Yes it should! Thanks :-)

tox.ini
chmeliik: Exactly, added comment

tests/test_workers/test_pkg_managers/test_pip.py
```diff
@@ -0,0 +1,587 @@
...
+            (
+                "",
+                None,
+                ["Parsing setup.cfg at '{tmpdir}/setup.cfg'", "No metadata.name in setup.cfg"],
+            ),
+            ("[metadata]", None, ["No metadata.name in setup.cfg"]),
+            (
+                dedent(
+                    """
+            [metadata]
```
lcarva, line 92: Is Black pushing for thins indentation? If so, try adding a backslash after `"""`:
```python
dedent(
  """\
  [metadata]
  ...
  """
)

File: tests/test_workers/test_pkg_managers/test_pip.py [hunk 13]
```diff
@@ -0,0 +1,587 @@
...
+    Create a file tree for the duration of the context.
+
+    :param dict tree_def: Definition of file tree, see usage for intuitive examples
+    :param (str | Path) rooted_at: Root of file tree
+    """
+    write_file_tree(tree_def, rooted_at)
+    try:
+        yield
+    finally:
+        remove_file_tree(tree_def, rooted_at)
```
lcarva, line 65: Why not just `shutil.rmtree`? Or, better yet, always use `tmpdir` and let pytest clean up these files automatically.

tests/test_workers/test_pkg_managers/test_pip.py
chmeliik: Oh, that feels *so* much better :D

tests/test_workers/test_pkg_managers/test_pip.py
chmeliik: Good point, I had some issues with leftover files when I was trying to use `importlib.util.find_spec()` for finding modules (which also involved temporarily modifying `sys.path`). In the end, I never got it to work and the current solution no longer has this issue.

tests/test_workers/test_pkg_managers/test_pip.py
```diff
@@ -0,0 +1,587 @@
...
+                        """
+                [metadata]
+                version = file: version.txt
+                """
+                    ),
+                    "version.txt": "\n1.0.0\n",
+                },
+                "1.0.0",
+                None,
+                [],
```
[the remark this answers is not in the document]
chmeliik: That's fair, I'll make the logging checks as complete as possible

chmeliik:
Added missing checks for log messages. Moved the function which actually runs the tests from the bottom of the file to where it makes more sense.

No clue why CI is failing, pytest says
```
  File "/home/travis/build/release-engineering/cachito/.tox/py37/lib/python3.7/site-packages/celery/utils/log.py", line 13, in <module>
    from kombu.five import PY3, values
ModuleNotFoundError: No module named 'kombu.five'
```

tests/test_workers/test_pkg_managers/test_pip.py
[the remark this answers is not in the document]
chmeliik: Added

chmeliik:
Figured out what was wrong with CI, see 2nd commit:
```
Celery requires kombu>=4.6.10,<4.7; this is not respected during
installation (probably because we specify 'kombu' explicitly).
```
 Not sure if that is the right way to fix it though, is there any better way to get around the issue?

athos-ribeiro:
> Figured out what was wrong with CI, see 2nd commit:
> 
> ```
> Celery requires kombu>=4.6.10,<4.7; this is not respected during
> installation (probably because we specify 'kombu' explicitly).
> ```
> 
> Not sure if that is the right way to fix it though, is there any better way to get around the issue?

Note that kombu 5.0.0 was released just a few hours ago, but celery 5 is still in alpha.

For now, we user Celery from Fedora 31 in our container images, which is python-celery-4.3.0-2.fc31. This relies con komub < 5, so we can just set the requirements versions as so. (both celery and kombu should be set to be < 5). 

Then we should file a ticket to support Celery 5 later, whenever it lands in Fedora.

WDYT?

chmeliik:
> Note that kombu 5.0.0 was released just a few hours ago, but celery 5 is still in alpha.
> 
> For now, we use Celery from Fedora 31 in our container images, which is python-celery-4.3.0-2.fc31. This relies on komub < 5, so we can just set the requirements versions as so. (both celery and kombu should be set to be < 5).
> 
> Then we should file a ticket to support Celery 5 later, whenever it lands in Fedora.
> 
> WDYT?

Sounds good to me, I'll make a separate PR for the requirements to unblock other people's PRs (someone else will have to merge it though)

athos-ribeiro:
Nice! Could you remove the workaround commit now?

PR #216, opened 2020-08-05
Make timeout for integration tests configurable 

ejegrova:
Variable timeout is added to test_env_vars.yaml and used in integration tests.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -34,7 +34,7 @@ def default_request(test_env):
     :return: a dict of packages with initial and completed responses from the Cachito API
     :rtype: dict
     """
-    client = utils.Client(test_env["api_url"], test_env["api_auth_type"])
+    client = utils.Client(test_env["api_url"], test_env["api_auth_type"], test_env.get("timeout"))
     default_requests = {}
     packages = test_env["packages"]
     for package_name in packages:
--- a/tests/integration/utils.py
+++ b/tests/integration/utils.py
@@ -14,15 +14,17 @@ Response = namedtuple("Response", "data id status")
 class Client:
     """Simplifies access to the Cachito API."""
 
-    def __init__(self, cachito_api_url, cachito_api_auth_type):
+    def __init__(self, cachito_api_url, cachito_api_auth_type, timeout=15):
         """
         Initialize the Client class.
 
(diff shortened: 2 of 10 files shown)


ejegrova:
> Nice!
> 
> Note that the same commit as the one in #211 is applied here as well.
> 
> I believe we should either have this new patch for configurable timeouts in that same PR (#211) or remove the npm integration tests commit from this PR.

I know there's the same commit, but I don't know how to properly deal with that. Chages in test_npm_basic depends on that commit. Maybe I should just put it in draft and rebase when the first commit is merged.

athos-ribeiro:
> > Nice!
> > Note that the same commit as the one in #211 is applied here as well.
> > I believe we should either have this new patch for configurable timeouts in that same PR (#211) or remove the npm integration tests commit from this PR.
> 
> I know there's the same commit, but I don't know how to properly deal with that. Chages in test_npm_basic depends on that commit. Maybe I should just put it in draft and rebase when the first commit is merged.

Sounds good.

I wouldn't mind having this commit in that PR as well. The code looks good and the re-factoring would make sense. Whatever works for you :)

tests/integration/conftest.py
```diff
@@ -34,7 +34,7 @@ def default_request(test_env):
     :return: a dict of packages with initial and completed responses from the Cachito API
     :rtype: dict
     """
-    client = utils.Client(test_env["api_url"], test_env["api_auth_type"])
+    client = utils.Client(test_env["api_url"], test_env["api_auth_type"], test_env["timeout"])
```
athos-ribeiro, line 37: Won't this break current deployments without a timeout configured? Could we use `test_env.get("timeout)` instead?

tests/integration/utils.py
```diff
@@ -14,15 +14,17 @@
 class Client:
     """Simplifies access to the Cachito API."""
 
-    def __init__(self, cachito_api_url, cachito_api_auth_type):
+    def __init__(self, cachito_api_url, cachito_api_auth_type, timeout):
```
athos-ribeiro, line 17: Then here we could have timeout=SOME_DEFAULT_VALUE

ejegrova:
made the changes, @athos-ribeiro, can you check please?

PR #217, opened 2020-08-05
Pin celery and kombu versions to 4.x.x

chmeliik:
Kombu 5.0.0 has been released while the corresponding Celery version is
still in alpha, this means we end up installing an unsupported version
of Kombu for tests.

Pin both to 4.x.x for now, upgrade to 5.x.x when they land in Fedora.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
-celery
+celery<5
 gitpython
-kombu # A celery dependency but it's directly imported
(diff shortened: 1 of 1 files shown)


chmeliik:
PTAL @athos-ribeiro @lcarva @nirzari 

And @lcarva, could you please add me to the github org so that I have permissions to ask for reviews and such?

athos-ribeiro:
I am merging this one to unblock other PRs

PR #221, opened 2020-08-06
Use local git repositories for integration tests

athos-ribeiro:
When this option is set, we must make sure the repository will be
available for the Cachito workers. For local tests we do this by
creating the repository in the host volume mounted in the Cachito worker
container in the docker-compose.yml file.

### DIFF
--- a/test_env_vars.yaml
+++ b/test_env_vars.yaml
@@ -75,10 +75,17 @@ cached_dependencies:
     https_url: "https://github.com/release-engineering/retrodep.git"
   # Repository that will be used for the test
   test_repo:
+    # Use local bare remote repository
+    # If this is used, the git urls below should be paths and should be available for the workers
+    use_local: True
     # The tracked repository (remote)
-    ssh_url: "git@github.com:user/cachito_test_repo.git"
+    ssh_url: "./tmp/cachito-archives/integration-tests-git-repo.git"
     # The URL of the repository
-    https_url: "https://github.com/user/cachito_test_repo.git"
+    https_url: "./tmp/cachito-archives/integration-tests-git-repo.git"
+    # test repo user
+    git_user: "Arthur Dent"
+    # test repo user email
+    git_email: "dent42@cachito.rocks"
     # Package managers
     pkg_managers: ["gomod"]
 # Test data for the HTTP GET_ALL test
--- a/tests/integration/test_using_cached_dependencies.py
+++ b/tests/integration/test_using_cached_dependencies.py
@@ -1,86 +1,137 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
+from pathlib import Path
 import random
+import shutil
 import string
 
 import git
+import pytest
 
 import utils
 
 
-def test_using_cached_dependencies(test_env, tmpdir):
+def create_local_repository(repo_path):
     """
-    Check that the cached dependencies are used instead of downloading them from repo again.
+    Create a local git repoitory.
 
-    Preconditions:
-    * On git instance prepare an empty repository
-
-    Process:
-    * Clone the package from the upstream repository
-    * Create empty commit on new test branch and push it to prepared repository
-    * Send new request to Cachito API which would fetch data from the prepared repository
-    * Delete branch with the corresponding commit
-    * Send the same request to Cachito API
-
-    Checks:
-    * Check that the state of the first request is complete
-    * Check that the commit is not available in the repository after the branch is deleted
-    * Check that the state of the second request is complete
+    :param str repo_path: path to new bare git repository
+    :return: normalized bare git repository path
+    :rtype: str
     """
-    generated_suffix = "".join(
-        random.choice(string.ascii_letters + string.digits) for x in range(10)
-    )
-    branch_name = f"test-{generated_suffix}"
-    repo = git.repo.Repo.clone_from(
-        test_env["cached_dependencies"]["seed_repo"]["https_url"], tmpdir
-    )
-    remote = repo.create_remote("test", url=test_env["cached_dependencies"]["test_repo"]["ssh_url"])
-    assert remote.exists()
+    bare_repo_dir = Path(repo_path)
+    bare_repo = git.Repo.init(str(bare_repo_dir), bare=True)
+    assert bare_repo.bare
+    # We need to expand this for later usage from the original repo directory
+    return str(bare_repo_dir.resolve())
 
-    try:
-        repo.create_head(branch_name).checkout()
-        repo.git.commit("--allow-empty", m="Commit created in integration test for Cachito")
-        repo.git.push("-u", remote.name, branch_name)
-        commit = repo.head.commit.hexsha
 
-        client = utils.Client(
-            test_env["api_url"], test_env["api_auth_type"], test_env.get("timeout")
+class TestCachedDependencies:
+    """Test class for cached dependencies."""
+
+    @pytest.fixture(autouse=True)
+    def setup_method_fixture(self, test_env):
+        """Create bare git repo and a pool for removing shared directories."""
+        self.directories = []
+        self.git_user = test_env["cached_dependencies"]["test_repo"].get("git_user")
+        self.git_email = test_env["cached_dependencies"]["test_repo"].get("git_email")
+        if test_env["cached_dependencies"]["test_repo"].get("use_local"):
+            repo_path = create_local_repository(
+                test_env["cached_dependencies"]["test_repo"]["ssh_url"]
+            )
+            test_env["cached_dependencies"]["test_repo"]["ssh_url"] = repo_path
+            # Defer cleanups
+            self.directories.append(repo_path)
+
+        self.test_env = test_env
+
+    def teardown_method(self, method):
+        """Remove shared directories in the pool."""
+        for directory in self.directories:
+            shutil.rmtree(directory)
+
+    def test_using_cached_dependencies(self, tmpdir):
+        """
+        Check that the cached dependencies are used instead of downloading them from repo again.
+
+        Preconditions:
+        * On git instance prepare an empty repository
+
+        Process:
+        * Clone the package from the upstream repository
+        * Create empty commit on new test branch and push it to the prepared repository
+        * Send new request to Cachito API which would fetch data from the prepared repository
+        * Delete branch with the corresponding commit
+        * Send the same request to Cachito API
+
+        Checks:
+        * Check that the state of the first request is complete
+        * Check that the commit is not available in the repository after the branch is deleted
+        * Check that the state of the second request is complete
+        """
+        generated_suffix = "".join(
+            random.choice(string.ascii_letters + string.digits) for x in range(10)
         )
+        branch_name = f"test-{generated_suffix}"
+        repo = git.repo.Repo.clone_from(
+            self.test_env["cached_dependencies"]["seed_repo"]["https_url"], tmpdir
+        )
+        remote = repo.create_remote(
+            "test", url=self.test_env["cached_dependencies"]["test_repo"]["ssh_url"]
+        )
+        assert remote.exists()
+
+        # set user configuration, if available
+        if self.git_user:
+            repo.config_writer().set_value("user", "name", self.git_user).release()
+        if self.git_email:
+            repo.config_writer().set_value("user", "email", self.git_email).release()
+
+        try:
+            repo.create_head(branch_name).checkout()
+            repo.git.commit("--allow-empty", m="Commit created in integration test for Cachito")
+            repo.git.push("-u", remote.name, branch_name)
+            commit = repo.head.commit.hexsha
+
+            client = utils.Client(
+                self.test_env["api_url"],
+                self.test_env["api_auth_type"],
+                self.test_env.get("timeout"),
+            )
+            response = client.create_new_request(
+                payload={
+                    "repo": self.test_env["cached_dependencies"]["test_repo"]["https_url"],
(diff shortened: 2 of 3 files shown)


athos-ribeiro:
@ejegrova, this is what I had in mind for handling local integration tests. Let me know what you think about it.

I do not like how this PR sneaks into the cachito worker volume to create a repository to be shared among the host and the worker, but it does the trick so anyone can run integration tests locally without forking retrodep.

athos-ribeiro:
Cc @chmeliik (still cannot add you as a reviewer :/ )

tests/integration/test_using_cached_dependencies.py
```diff
@@ -82,3 +98,7 @@ def test_using_cached_dependencies(test_env, tmpdir):
     first_deps = utils.make_list_of_packages_hashable(first_response.data["dependencies"])
     second_deps = utils.make_list_of_packages_hashable(second_response.data["dependencies"])
     assert first_deps == second_deps
+    # clean ups
+    if test_env["cached_dependencies"]["test_repo"].get("use_local"):
+        # this has been converted to a absolute path string
+        shutil.rmtree(test_env["cached_dependencies"]["test_repo"]["ssh_url"])
```
ejegrova, line 104: If the test fails on some assertion this clean up won't happen. Maybe we can use try/finally block for this too?

athos-ribeiro: @ejegrova I moved the test into a test class and added proper setup and teardown methods.

test_env_vars.yaml
```diff
@@ -74,10 +74,17 @@ cached_dependencies:
...
+    # If this is used, the git urls below should be paths and should be available for the workers
+    use_local: True
     # The tracked repository (remote)
-    ssh_url: "git@github.com:user/cachito_test_repo.git"
+    ssh_url: "./tmp/cachito-archives/integration-tests-git-repo.git"
     # The URL of the repository
-    https_url: "https://github.com/user/cachito_test_repo.git"
+    https_url: "./tmp/cachito-archives/integration-tests-git-repo.git"
+    # test repo user
+    git_user: "Arthur Dent"
```
nirzari, line 85: Are these values to be replaced by an actual user?
If so, you'd have to replace them with the actual ones at https://gitlab.cee.redhat.com/osbs/cachito-tests/-/tree/master/jenkins because any changes done to `test_env_vars.yaml` will affect how we run tests in jenkins.
(It's a different story that currently jenkins tests are broken due to unavailability of nexus host on the test environment)

test_env_vars.yaml
```diff
@@ -74,10 +74,17 @@ cached_dependencies:
     https_url: "https://github.com/release-engineering/retrodep.git"
   # Repository that will be used for the test
   test_repo:
+    # Use local bare remote repository
+    # If this is used, the git urls below should be paths and should be available for the workers
+    use_local: True
     # The tracked repository (remote)
-    ssh_url: "git@github.com:user/cachito_test_repo.git"
+    ssh_url: "./tmp/cachito-archives/integration-tests-git-repo.git"
```
nirzari, line 81: How would this change affect running tests in jenkins?

test_env_vars.yaml
```diff
@@ -74,10 +74,17 @@ cached_dependencies:
...
     # The URL of the repository
-    https_url: "https://github.com/user/cachito_test_repo.git"
+    https_url: "./tmp/cachito-archives/integration-tests-git-repo.git"
```
nirzari, line 83: Same as above!

test_env_vars.yaml
nirzari: If we must make these changes temporarily until jenkins tests are fixed, I recommend creating a temp. `test_env_vars_local.yaml` instead of tempering with `test_env_vars.yaml`. Let me know your thoughts.

tests/integration/test_using_cached_dependencies.py
```diff
@@ -1,84 +1,116 @@
...
+    def test_using_cached_dependencies(self, test_env, tmpdir):
+        """
+        Check that the cached dependencies are used instead of downloading them from repo again.
+
+        Preconditions:
+        * On git instance prepare an empty repository
+
+        Process:
+        * Clone the package from the upstream repository
+        * Create empty commit on new test branch and push it to prepared repository
```
nirzari, line 34:
```suggestion
        * Create empty commit on new test branch and push it to the prepared repository
```

tests/integration/test_using_cached_dependencies.py
```diff
@@ -1,84 +1,116 @@
...
+            response = client.create_new_request(
+                payload={
+                    "repo": test_env["cached_dependencies"]["test_repo"]["https_url"],
+                    "ref": commit,
+                    "pkg_managers": test_env["cached_dependencies"]["test_repo"]["pkg_managers"],
+                },
+            )
+            first_response = client.wait_for_complete_request(response)
+            assert first_response.data["state"] == "complete"
+
```
nirzari, line 88:
```suggestion

```

tests/integration/test_using_cached_dependencies.py
```diff
@@ -1,84 +1,116 @@
...
+            assert repo.git.branch("-a", "--contains", commit)
+
```
nirzari, line 90:
```suggestion

```

tests/integration/test_using_cached_dependencies.py
```diff
@@ -1,84 +1,116 @@
...
-    assert set(first_response.data["pkg_managers"]) == set(second_response.data["pkg_managers"])
-    first_pkgs = utils.make_list_of_packages_hashable(first_response.data["packages"])
-    second_pkgs = utils.make_list_of_packages_hashable(second_response.data["packages"])
-    assert first_pkgs == second_pkgs
-    first_deps = utils.make_list_of_packages_hashable(first_response.data["dependencies"])
-    second_deps = utils.make_list_of_packages_hashable(second_response.data["dependencies"])
-    assert first_deps == second_deps
+        second_response = client.wait_for_complete_request(response)
+        assert second_response.data["state"] == "complete"
+
```
nirzari, line 107:
```suggestion

```

test_env_vars.yaml
athos-ribeiro: It should not unless test_env["cached_dependencies"]["test_repo"]["use_local"] is set there.

test_env_vars.yaml
athos-ribeiro: replied above

test_env_vars.yaml
athos-ribeiro: > Are these values to be replaced by an actual user?

Not for local tests.

> If so, you'd have to replace them with the actual ones at https://gitlab.cee.redhat.com/osbs/cachito-tests/-/tree/master/jenkins

I don't think we should mix upstream with downstream discussions. You are right though. The work will need to be done in the Jenkins side.

> If we must make these changes temporarily until jenkins tests are fixed, I recommend creating a temp. test_env_vars_local.yaml instead of tempering with test_env_vars.yaml

I don't think the configuration file should reflect whatever is used in jenkins. Instead, this should be either a generic configuration file or one set to work with local test runs.

nirzari: I got rid of the url from my comment above. You're right about not mixing both discussions but if the changes proposed here could potentially break how we run tests in jenkins, it's important to acknowledge them here before we merge them.

I agree that the configuration file does not reflect whatever is used in jenkins but it is directly fed to jenkins as it is for running the tests. Hence, let's just keep in mind while changing anything in configurations here, that all necessary changes required in downstream jenkins configuration must be done simultaneously (as part of the same ticket).

test_env_vars.yaml
nirzari: We are definitely not mixing both work but as I mentioned above, it is important to acknowledge the concerns. Let me know if you prefer discussing this in JIRA or someplace else.

@ejegrova Could you please help confirm this as you are familiar with how we run tests locally as well as within jenkins. (whether this change needs additional work downstream)

test_env_vars.yaml
athos-ribeiro: You are right. I made an additional change to make sure this patch will not break custom existing configuration files

test_env_vars.yaml
```diff
@@ -76,10 +76,17 @@ cached_dependencies:
     https_url: "https://github.com/release-engineering/retrodep.git"
   # Repository that will be used for the test
   test_repo:
+    # Use local bare remote repository
+    # If this is used, the git urls below should be paths and should be available for the workers
+    use_local: True
     # The tracked repository (remote)
-    ssh_url: "git@github.com:user/cachito_test_repo.git"
+    ssh_url: "./tmp/cachito-archives/integration-tests-git-repo.git"
```
chmeliik, line 83: It seems that when running tests against the `docker-compose` deployment, you must manually create `./tmp/cachito-archives` prior to running `docker-compose up`.

Otherwise, docker automatically creates this directory as root and the test for cached dependencies fails with
```
>           bare_repo = git.Repo.init(str(bare_repo_dir), bare=True)
...
E           PermissionError: [Errno 13] Permission denied: '<cachito repo>/tmp/cachito-archives/integration-tests-git-repo.git'
```

Is this something that should be mentioned as a comment? Or is that general knowledge for running integration tests? Note that other tests work even if you let the directory be created automatically.

test_env_vars.yaml
ejegrova: I am not entirely sure if I understand the discussion. These changes are needed so the tests will run locally without having to configure github repo, etc. Jenkins configuration is something that will still need to happen because this test is causing our issues on Jenkins. But we definitely shouldn't change upstream so it will fit our needs downstream.

test_env_vars.yaml
athos-ribeiro: Thanks for checking this!

I have not run through this issue because I am using podman to run the tests (and for now, I do have to manually create those directories).

We should either introduce changes to the Makefile to prepare the environment (creating the directories) and update the docs, or just drop this patch. Creating the directory was also part of #225, but should be deprecated by https://github.com/containers/podman-compose/pull/194 for that specific case.

Thoughts?

chmeliik: What were the prerequisites for running this test before? Create any empty repo on github and specify it in the config file?

This does seem like a pretty nice quality-of-life change, so I wouldn't have anything against updating the Makefile and docs, but I'd let more experienced members weigh in on this.

athos-ribeiro: > What were the prerequisites for running this test before? Create any empty repo on github and specify it in the config file?

Yes.

> This does seem like a pretty nice quality-of-life change, so I wouldn't have anything against updating the Makefile and docs, but I'd let more experienced members weigh in on this.

Sounds good. Let's see what others think as well and then I will proceed to making these additional changes.

lcarva: Using the `Makefile` to perform setup actions prior to running `docker-compose up` sounds like a good idea.

tests/integration/test_using_cached_dependencies.py
```diff
@@ -1,86 +1,117 @@
...
+        * Check that the commit is not available in the repository after the branch is deleted
+        * Check that the state of the second request is complete
+        """
+        if test_env["cached_dependencies"]["test_repo"].get("use_local"):
+            # create bare git repo
+            bare_repo_dir = Path(test_env["cached_dependencies"]["test_repo"]["ssh_url"])
+            bare_repo = git.Repo.init(str(bare_repo_dir), bare=True)
+            assert bare_repo.bare
+            # We need to expand this for later usage from the original repo directory
+            test_env["cached_dependencies"]["test_repo"]["ssh_url"] = str(bare_repo_dir.resolve())
```
lcarva, line 50: If we're replacing the `ssh_url` anyways, could we just do this in the `setup_method`?

tests/integration/test_using_cached_dependencies.py
```diff
@@ -1,86 +1,117 @@
...
+        branch_name = f"test-{generated_suffix}"
+        repo = git.repo.Repo.clone_from(
+            test_env["cached_dependencies"]["seed_repo"]["https_url"], tmpdir
+        )
+        remote = repo.create_remote(
+            "test", url=test_env["cached_dependencies"]["test_repo"]["ssh_url"]
+        )
+        assert remote.exists()
+
+        # set user configuration, if available
```
lcarva, line 66: Do this in the `setup_method` as well?

tests/integration/test_using_cached_dependencies.py
athos-ribeiro: We would need to create additional fixtures to use `test_env` in the setup method (and use a custom setup method). That is why I ended up not manipulating test_env in the setup. Do you think it is worth to have this additional fixture here?

tests/integration/test_using_cached_dependencies.py
athos-ribeiro: See discussion above.

test_env_vars.yaml
athos-ribeiro: Done! I also added a comment in the Makefile to explain why we are doing it.

tests/integration/test_using_cached_dependencies.py
lcarva: Oh, I see what you mean. I think a new fixture would make it cleaner, yes. This could be defined in this file so it's not accidentally used by other tests.

athos-ribeiro: Done. I also extracted the function to create the bare git repository.

PR #223, opened 2020-08-07
Update .dockerignore file

athos-ribeiro:
* Add tmp directory with development container volumes
* Add venv directory used in the Makefile

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,4 +7,6 @@
 build/
 dist/
 htmlcov/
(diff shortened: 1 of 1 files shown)


athos-ribeiro:
I was wondering why building the images locally was taking so long. Then I realized we copy the whole project directory into the image. So if we are rebuilding the image after running tests, we end up copying the worker volumes into the new images as well.

PR #225, opened 2020-08-07
Support podman based development environment

athos-ribeiro:
* Use host volumes
* Create shared directories with proper permissions
* Document the need to loosen SElinux restrictions


athos-ribeiro:
I created this PR to show how I am running integration tests with podman-compose instead of docker.

If we were to merge this, we would need to move the entries in the shell script to the Makefile; losen specific selinux configurations instead of setting it to permissive mode; and possibly rebuild the Nexus image to avoid the need of changing permissions on the nexus-volume directory.

athos-ribeiro:
Cc @lcarva, @nirzari, @chmeliik, @akhmanova, @ejegrova

prepare_podman_compose.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# make sure selinux is off
+sestatus | grep '^Current mode:.*enforcing$'
+if [ $? -eq 0 ]; then
+  echo "Selinux is enforcing.Set it to permissive mode before using podman-compose"
```
lcarva, line 6: Yeah, I don't think podman-compose handles the `:z` modifier on volumes properly.

prepare_podman_compose.sh
```diff
@@ -0,0 +1,17 @@
...
+  exit 1
+fi
+
+mkdir -p ./tmp/cachito-archives
```
lcarva, line 10: Apparently this is an explicit backwards compatibility breakage, see https://github.com/containers/podman-compose/issues/185

It looks like a compatibility flag could be introduced to handle this.

athos-ribeiro: Will this podman-compose patch also eliminate the need to change the nexus-volume permissions below?

lcarva: Hmm unclear, but I was able to share named volumes between two containers in iib.

athos-ribeiro:
- Rebased on master
- Moved setup script to Makefile

Makefile
```diff
@@ -1,18 +1,35 @@
+export COMPOSE_ENGINE=docker-compose
```
lcarva, line 1: Can we call this something like `CACHITO_COMPOSE_ENGINE` so I can set it in my `.bashrc` and forget about it? Then in this file we can set a default value if needed.

README.md
```diff
@@ -77,6 +77,15 @@ Run the application locally (requires [docker-compose](https://docs.docker.com/c
 make run
 ```
 
+Alternatively, you could also run the application with
+[podman-compose](https://github.com/containers/podman-compose) by setting the
+`COMPOSE_ENGINE` variable (for now SELinux must be set to the **Permissive**
```
lcarva, line 82: Can we add a warning to indicate this is dangerous?

Makefile
athos-ribeiro: Done

README.md
athos-ribeiro: Done. Also added it to the Makefile

athos-ribeiro:
@lcarva, do you think this is good to be merged or should I close it and link this PR in the README file for whoever wants to run podman?

athos-ribeiro:
Rebased on master

Makefile
```diff
@@ -1,18 +1,35 @@
+CACHITO_COMPOSE_ENGINE ?= docker-compose
+ifeq ($(CACHITO_COMPOSE_ENGINE), docker-compose)
+	DOWN_OPTS=-v
```
lcarva, line 3: If we're not using named volumes, you probably don't need `-v`, right? But it doesn't hurt.

lcarva (approved):
I think this is fine to be merged. The `podman-compose` approach is clearly documented.

PR #226, opened 2020-08-07
Implement setup.py handling for Pip

chmeliik:
* CLOUDBLD-2053

Supports only basic use cases, see docstrings in SetupPY class for more
details.

### DIFF
--- a/cachito/workers/pkg_managers/pip.py
+++ b/cachito/workers/pkg_managers/pip.py
@@ -124,7 +124,7 @@ class SetupCFG(SetupFile):
         :rtype: str or None
         """
         name = self._get_option("metadata", "name")
-        if name is None:
+        if not name:
             log.debug("No metadata.name in setup.cfg")
             return None
 
@@ -144,13 +144,14 @@ class SetupCFG(SetupFile):
         :rtype: str or None
         """
         version = self._get_option("metadata", "version")
-        if version is None:
+        if not version:
             log.debug("No metadata.version in setup.cfg")
             return None
 
         log.debug("Resolving metadata.version in setup.cfg from %r", version)
         version = self._resolve_version(version)
-        if version is None:
+        if not version:
+            # Falsy values also count as "failed to resolve" (0, None, "", ...)
             log.debug("Failed to resolve metadata.version in setup.cfg")
             return None
 
--- a/cachito/workers/pkg_managers/pip.py
+++ b/cachito/workers/pkg_managers/pip.py
@@ -125,10 +125,10 @@ class SetupCFG(SetupFile):
         """
         name = self._get_option("metadata", "name")
         if not name:
-            log.debug("No metadata.name in setup.cfg")
+            log.info("No metadata.name in setup.cfg")
             return None
 
-        log.debug("Found metadata.name in setup.cfg: %r", name)
+        log.info("Found metadata.name in setup.cfg: %r", name)
         return name
 
     def get_version(self):
@@ -145,18 +145,18 @@ class SetupCFG(SetupFile):
         """
         version = self._get_option("metadata", "version")
         if not version:
-            log.debug("No metadata.version in setup.cfg")
+            log.info("No metadata.version in setup.cfg")
             return None
 
         log.debug("Resolving metadata.version in setup.cfg from %r", version)
         version = self._resolve_version(version)
         if not version:
             # Falsy values also count as "failed to resolve" (0, None, "", ...)
-            log.debug("Failed to resolve metadata.version in setup.cfg")
+            log.info("Failed to resolve metadata.version in setup.cfg")
             return None
 
         version = any_to_version(version)
-        log.debug("Found metadata.version in setup.cfg: %r", version)
+        log.info("Found metadata.version in setup.cfg: %r", version)
         return version
 
     @property
@@ -175,7 +175,7 @@ class SetupCFG(SetupFile):
                     parsed.read_file(f)
                     self.__parsed = parsed
                 except configparser.Error as e:
-                    log.debug("Failed to parse setup.cfg: %s", e)
+                    log.error("Failed to parse setup.cfg: %s", e)
                     self.__parsed = None  # Tried to parse file and failed
 
         return self.__parsed
@@ -207,7 +207,7 @@ class SetupCFG(SetupFile):
             log.debug("Read version from %r: %r", file_path, version)
             return version
         else:
-            log.debug("Version file %r does not exist or is not a file", file_path)
+            log.error("Version file %r does not exist or is not a file", file_path)
             return None
 
     def _ensure_local(self, path):
@@ -243,13 +243,13 @@ class SetupCFG(SetupFile):
         if module_file is not None:
             log.debug("Found module %r at %r", module_name, str(module_file))
         else:
-            log.debug("Module %r not found", module_name)
+            log.error("Module %r not found", module_name)
             return None
 
         try:
             module_ast = ast.parse(module_file.read_text(), module_file.name)
         except SyntaxError as e:
-            log.debug("Syntax error when parsing module: %s", e)
+            log.error("Syntax error when parsing module: %s", e)
             return None
 
         try:
@@ -257,7 +257,7 @@ class SetupCFG(SetupFile):
             log.debug("Found attribute %r in %r: %r", attr_name, module_name, version)
             return version
         except (AttributeError, ValueError) as e:
-            log.debug("Could not find attribute in %r: %s", module_name, e)
+            log.error("Could not find attribute in %r: %s", module_name, e)
             return None
 
     def _find_module(self, module_name, package_dir=None):
@@ -443,12 +443,12 @@ class SetupPY(SetupFile):
         """
         name = self._get_setup_kwarg("name")
         if not name or not isinstance(name, str):
-            log.debug(
+            log.info(
                 "Name in setup.py was either not found, or failed to resolve to a valid string"
             )
             return None
 
-        log.debug("Found name in setup.py: %r", name)
+        log.info("Found name in setup.py: %r", name)
         return name
 
     def get_version(self):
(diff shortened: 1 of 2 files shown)


chmeliik:
It's another huge one, **yay!**

PR 2 out of 3 for CLOUDBLD-2053, the last one will be a function which takes a repo and resolves name, version from  setup.py -> setup.cfg -> git data

cachito/workers/pkg_managers/pip.py
```diff
@@ -332,3 +335,296 @@ def _get_package_dirs(self):
...
+
+        log.debug("setup kwarg %r not found", arg_name)
+        return None
+
+    def _get_variable(self, var_name):
+        """Walk back up the AST along setup branch, look for first assignment of variable."""
+        lineno = self._setup_branch.call_node.lineno
+        node_path = self._setup_branch.node_path
+
+        log.debug("Backtracking up the AST from line %s to find variable %r", lineno, var_name)
```
nirzari, line 616: Why `var_name` is treated as printable representation of an object? (Out of curiosity: trying to understand the difference)

cachito/workers/pkg_managers/pip.py
```diff
@@ -332,3 +335,296 @@ def _get_package_dirs(self):
...
+
+        for elem in filter(ASTpathelem.field_is_body, reversed(node_path)):
+            try:
+                value = get_top_level_attr(elem.field, var_name, lineno)
+                log.debug("Found variable %r: %r", var_name, value)
+                return value
+            except ValueError as e:
+                log.debug("Variable cannot be resolved: %s", e)
```
nirzari, line 624:
```suggestion
                log.error("Variable cannot be resolved: %s", e)
```

tests/test_workers/test_pkg_managers/test_pip.py
```diff
@@ -654,3 +654,461 @@ def test_get_version_attr_with_package_dir(
...
+        else:
+            assert False, "'what' must be one of 'name', 'version'"
+
+        assert value == expect_val
+
+        logs = expect_logs.copy()
+        # Does not actually have to be at index 0, this is just to be more obvious
+        logs.insert(0, f"Parsing setup.py at '{tmpdir.join('setup.py')}'")
+        if expect_val is None:
+            logs.append(f"No {what} in setup.py, failed to resolve or invalid")
```
nirzari, line 689: @mprahl Would your recommendation to use the different style formatting for logs instead of f strings apply here as well?

cachito/workers/pkg_managers/pip.py
```diff
@@ -332,3 +335,296 @@ def _get_package_dirs(self):
...
+        return self.__ast
+
+    @property
+    def _setup_branch(self):
+        """
+        Find setup() call anywhere in the file, return setup branch.
+
+        If setup call not found, return None. Will not search more than once.
+        """
+        if self._ast is None:
```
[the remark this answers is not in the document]
chmeliik: The change that you are referring to was specifically for the value of `name` and `version` in setup.cfg (it is a separate commit). The commit message should hopefully explain why the change was done, I would rather use `is None` for the other cases.

cachito/workers/pkg_managers/pip.py
chmeliik: I like to use repr for strings because they may fade into the surrounding text otherwise. Basically, I'd rather have
`find variable 'version'` than `find variable version`.

Another option would be to use `'%s'`, but in some cases, repr provides more readable output, so I just use `%r` everywhere for consistency.

cachito/workers/pkg_managers/pip.py
```diff
@@ -332,3 +335,296 @@ def _get_package_dirs(self):
...
+
+        Will not parse file (or try to) more than once.
+        """
+        if self.__ast is NOTHING:
+            log.debug("Parsing setup.py at %r", str(self._path))
+
+            try:
+                self.__ast = ast.parse(self._path.read_text(), self._path.name)
+            except SyntaxError as e:
+                log.debug("Syntax error when parsing setup.py: %s", e)
```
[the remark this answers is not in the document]
chmeliik: I guess it should. In that case, taking into account your other log suggestion, I guess the following log messages could be `error` as well:

setup.cfg:
- [Failed to parse setup.cfg](https://github.com/chmeliik/cachito/blob/d55853ece62bb3f0b3f1767075895fe1b996e95b/cachito/workers/pkg_managers/pip.py#L178)
- [Version file not found](https://github.com/chmeliik/cachito/blob/d55853ece62bb3f0b3f1767075895fe1b996e95b/cachito/workers/pkg_managers/pip.py#L210)
- [Module not found](https://github.com/chmeliik/cachito/blob/d55853ece62bb3f0b3f1767075895fe1b996e95b/cachito/workers/pkg_managers/pip.py#L246)
- [Syntax error when parsing module](https://github.com/chmeliik/cachito/blob/d55853ece62bb3f0b3f1767075895fe1b996e95b/cachito/workers/pkg_managers/pip.py#L252)
- [Could not find attribute in module](https://github.com/chmeliik/cachito/blob/d55853ece62bb3f0b3f1767075895fe1b996e95b/cachito/workers/pkg_managers/pip.py#L260)

setup.py
- Failed to parse setup.py (this)
- [Setup call not found](https://github.com/chmeliik/cachito/blob/d55853ece62bb3f0b3f1767075895fe1b996e95b/cachito/workers/pkg_managers/pip.py#L510)
- [Setup kwarg is an unsupported expression](https://github.com/chmeliik/cachito/blob/d55853ece62bb3f0b3f1767075895fe1b996e95b/cachito/workers/pkg_managers/pip.py#L605)
- Variable cannot be resolved ([your other comment](https://github.com/release-engineering/cachito/pull/226#discussion_r467245837))

If you agree on these, I would make the change in a separate commit.
Another option would be to use `warning`, since none of these errors are considered fatal. What do you think?

cachito/workers/pkg_managers/pip.py
```diff
@@ -332,3 +335,296 @@ def _get_package_dirs(self):
...
+    def get_name(self):
+        """
+        Attempt to extract package name from setup.py.
+
+        :rtype: str or None
+        """
+        name = self._get_setup_kwarg("name")
+        if not name or not isinstance(name, str):
+            # Invalid means empty string or not a string
+            log.debug("No name in setup.py, failed to resolve or invalid")
```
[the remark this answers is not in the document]
chmeliik: Changed, thanks!

cachito/workers/pkg_managers/pip.py
```diff
@@ -332,3 +335,296 @@ def _get_package_dirs(self):
...
+
+        Rather than trying to keep edge cases consistent with setuptools, treat them
+        consistently within Cachito.
+
+        :rtype: str or None
+        """
+        version = self._get_setup_kwarg("version")
+        if not version:
+            # Invalid means any falsy value (0, None, "", ...)
+            log.debug("No version in setup.py, failed to resolve or invalid")
```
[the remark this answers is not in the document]
chmeliik: Changed (with slightly modified wording, to be more consistent with the message for `name`)

cachito/workers/pkg_managers/pip.py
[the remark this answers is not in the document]
nirzari: +1 to the separate commit idea.

`Warn` seems to be used to indicate potentially harmful situations, whereas `error` seems to be used to indicate the error events that might still allow the application to continue running. Both seem to be fitting for these scenarios. I will leave the best judgement to you while making the changes!

cachito/workers/pkg_managers/pip.py
nirzari: This is to be addressed in the previous comment, hence resolving.

tests/test_workers/test_pkg_managers/test_pip.py
mprahl: It wouldn't apply here since the code is comparing the already formatted log messages captured by pytest in `caplog` and not what was input to the respective log method.

cachito/workers/pkg_managers/pip.py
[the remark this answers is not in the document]
chmeliik: You're right, both seem applicable here. I went with `error` in the end

chmeliik:
Changed the level some messages are logged at (3rd commit):
- `error` for (non-fatal) errors
- `info` for the most important messages (whether name/version was found)

tests/test_workers/test_pkg_managers/test_pip.py
```diff
@@ -654,3 +654,466 @@ def test_get_version_attr_with_package_dir(
...
+                    "Variable 'foo' not found along the setup call branch",
+                ],
+            ),
+            (
+                dedent(
+                    """\
+                    setup({what}=foo)
+
+                    foo = "bar"
+                    """
```
athos-ribeiro, line 833: nice case

tests/test_workers/test_pkg_managers/test_pip.py
```diff
@@ -654,3 +654,466 @@ def test_get_version_attr_with_package_dir(
...
+            (
+                dedent(
+                    """\
+                    try:
+                        pass
+                    except A:
+                        pass
+                    except B:
+                        foo = "bar"
+                        setup({what}=foo)
```
athos-ribeiro, line 972: interesting. What happens when another branch has a similar block (let's say this is replicated in except A with another value attributed to foo)? We will just get the last occurrence, is that it?

This sounds good for now (hoping there is no sane use  cases for this), we just need to make that clear somewhere.

chmeliik: Do you mean something like this:
```python
try:
    pass
except A:
    foo = "1.0.0"
    setup(version=foo)
except B:
    foo = "2.0.0"
    setup(version=foo)
```

In this case we will find the first setup call (in `except A`) and use the value of foo defined in the same block (or any parent block). I'm hoping there is no legitimate use case for calling setup() in 2 or more places in the file, we would have no way to determine which one gets called anyway.

So it basically comes down to: We expect the file to have exactly one setup() call, if there are 2 or more we will only process the first one

I'll add this to the class docstring if this is what you meant

chmeliik: Added to the docstring of `_setup_branch` property

PR #227, opened 2020-08-10
Set state complete in the `create_bundle_archive`

ejegrova:
Set the request state after the bundle is created rather than a separate task

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -238,12 +238,7 @@ def create_request():
             tasks.fetch_npm_source.si(request.id, npm_package_configs).on_error(error_callback)
         )
 
-    chain_tasks.extend(
-        [
-            tasks.create_bundle_archive.si(request.id).on_error(error_callback),
-            tasks.set_request_state.si(request.id, "complete", "Completed successfully"),
-        ]
-    )
+    chain_tasks.extend([tasks.create_bundle_archive.si(request.id).on_error(error_callback)])
 
     try:
         chain(chain_tasks).delay()
--- a/tests/test_workers/test_tasks/test_general.py
+++ b/tests/test_workers/test_tasks/test_general.py
@@ -140,7 +140,7 @@ def test_create_bundle_archive(mock_gwc, mock_set_request, deps_present, tmpdir)
         expected |= set(deps_archive_contents.keys())
 
     assert bundle_contents == expected
-
-    mock_set_request.assert_called_once_with(
-        request_id, "in_progress", "Assembling the bundle archive"
-    )
+    call1 = (request_id, "in_progress", "Assembling the bundle archive")
+    call2 = (request_id, "complete", "Completed successfully")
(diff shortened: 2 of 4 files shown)


tests/test_workers/test_tasks/test_general.py
```diff
@@ -140,7 +140,6 @@ def test_create_bundle_archive(mock_gwc, mock_set_request, deps_present, tmpdir)
         expected |= set(deps_archive_contents.keys())
 
     assert bundle_contents == expected
-
-    mock_set_request.assert_called_once_with(
-        request_id, "in_progress", "Assembling the bundle archive"
-    )
+    call1 = (request_id, "in_progress", "Assembling the bundle archive")
+    call2 = (request_id, "complete", "Completed successfully")
+    mock_set_request.assert_has_calls(mock_set_request(call1), mock_set_request(call2))
```
athos-ribeiro, line 145: nitpick: should we also check mock_set_request was called twice (and only twice) here (previously, it was ensuring it was called only once)?

ejegrova: @athos-ribeiro I didn't find out how to do that. Maybe add the same function call with any_order=True? I would appreciate any suggestion :)

athos-ribeiro: This is how we did it in the gomod tests:

https://github.com/release-engineering/cachito/blob/master/tests/test_workers/test_tasks/test_gomod.py#L66-L67

call_count docs: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_count

chmeliik: @ejegrova looks like mock objects have a `call_count` attribute: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_count - should be what you need :)

ejegrova: Thanks a lot. I changed that, is it ok like this? @athos-ribeiro

cachito/web/api_v1.py
```diff
@@ -238,12 +238,7 @@ def create_request():
             tasks.fetch_npm_source.si(request.id, npm_package_configs).on_error(error_callback)
         )
 
-    chain_tasks.extend(
-        [
-            tasks.create_bundle_archive.si(request.id).on_error(error_callback),
-            tasks.set_request_state.si(request.id, "complete", "Completed successfully"),
-        ]
-    )
+    chain_tasks.extend([tasks.create_bundle_archive.si(request.id).on_error(error_callback)])
```
mprahl, line 241: Could you please use append here instead of extend since you are only adding a single element?

ejegrova: ok, I'll submit new PR for that

ejegrova: https://github.com/release-engineering/cachito/pull/235

PR #229, opened 2020-08-10
Handle PURL for npm dependencies from file

lcarva:
Due to cachito_npm_file_deps_allowlist, Cachito allows NPM dependencies
to be embedded in the application repository. For this reason, purl
conversion for these dependencies must also be supported. Othewise, the
content manifests cannot be generated for the whole request.

* CLOUDBLD-2223

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -198,7 +198,10 @@ class Package(db.Model):
             protocol = match.group("protocol")
             suffix = match.group("suffix")
             has_authority = match.group("has_authority")
-            if not has_authority:
+            if protocol == "file":
+                qualifier = urllib.parse.quote(self.version, safe="")
+                return f"generic/{purl_name}?{qualifier}"
+            elif not has_authority:
                 # github:namespace/name#ref or gitlab:ns1/ns2/name#ref
                 match_forge = re.match(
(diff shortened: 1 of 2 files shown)


lcarva:
@athos-ribeiro, I'm merging and deploying this right now because it's currently blocking a product team. Please, let me know if this needs to be tweaked.

cachito/web/models.py
```diff
@@ -198,7 +198,10 @@ def to_purl(self):
             protocol = match.group("protocol")
             suffix = match.group("suffix")
             has_authority = match.group("has_authority")
-            if not has_authority:
+            if protocol == "file":
+                qualifier = urllib.parse.quote(self.version, safe="")
+                return f"generic/{purl_name}?{qualifier}"
```
lcarva, line 203: @mprahl has pointed out that I'm adding a query parameter without `=`. This should probably say something like `f"generic/{purl_name}?file={qualifier}"`

nirzari: The request was successful even without it. So maybe you can just create another PR to update this but there shouldn't be a rush to deploy that right away again? is that correct?

athos-ribeiro:
> @athos-ribeiro, I'm merging and deploying this right now because it's currently blocking a product team. Please, let me know if this needs to be tweaked.

I opened https://github.com/release-engineering/cachito/pull/230 with a suggestion on how that PURL should be assembled

PR #230, opened 2020-08-11
Improve npm file dependencies PURL generation

athos-ribeiro:
The npm file dependencies "file" descriptor are relative to their parent
package root path.

Considering a PURL should be used to proper locate a package, this
information should be present in such dependencies PURLs.

This patch proposes including the parent package PURL in the
download_url qualifier and pointing to the proper package file through
the file_name qualifier.

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -208,10 +208,14 @@ class Package(db.Model):
             protocol = match.group("protocol")
             suffix = match.group("suffix")
             has_authority = match.group("has_authority")
-            if protocol == "file":
-                qualifier = urllib.parse.quote(self.version, safe="")
-                return f"generic/{purl_name}?{qualifier}"
-            elif not has_authority:
+            if not has_authority:
+                if protocol == "file":
+                    path = self.version.split(":", 1)[1]
+                    safe_path = urllib.parse.quote(path, safe="")
+                    # this path is relative to the parent package's package.json path
+                    # let's mark this purl for later substitution
+                    marker = content_manifest.PARENT_PURL_PLACEHOLDER
+                    return f"pkg:generic/{purl_name}?download_url={marker}&file_name={safe_path}"
                 # github:namespace/name#ref or gitlab:ns1/ns2/name#ref
                 match_forge = re.match(
                     r"(?P<namespace>.+)/(?P<name>[^#/]+)#(?P<version>.+)$", suffix
--- a/cachito/web/content_manifest.py
+++ b/cachito/web/content_manifest.py
@@ -1,5 +1,10 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
+import urllib.parse
+
 import flask
+
+
+PARENT_PURL_PLACEHOLDER = "PARENT_PURL"
 
 
 class ContentManifest:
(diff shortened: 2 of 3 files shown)


cachito/web/models.py
```diff
@@ -199,8 +199,12 @@ def to_purl(self):
             suffix = match.group("suffix")
             has_authority = match.group("has_authority")
             if protocol == "file":
```
chmeliik, line 201: Should this be nested under the `not has_authority` branch? Or do we want to accept `file://`?

athos-ribeiro: AFAIK, just `file:` is supported [1]. I will move it there.

[1] https://docs.npmjs.com/files/package.json#local-paths

athos-ribeiro:
> Looks reasonable, just one comment related to the previous change.
> 
> Out of curiosity, why is that the version of an NPM dependency can be such a complex construct?

My guess is the amount of content sources it supports [1]. We will see how different that is from python on the next Lui's PR =D 

https://docs.npmjs.com/files/package.json#git-urls-as-dependencies

cachito/web/models.py
athos-ribeiro: Done. Also rebased on master

athos-ribeiro:
Before merging this, I'd like to hear @lcarva or Matt's take on using a PURL as a download_url qualifier. The issue here is that a PURL parsing/downloading tool could possibly break on that qualifier, if it just try to, for instance, use python Requests to fetch the content in the URL.

I also opened https://github.com/package-url/purl-spec/issues/89 so the matter can be openly discussed within the PURL community.

chmeliik:
> Before merging this, I'd like to hear @lcarva or Matt's take on using a PURL as a download_url qualifier. The issue here is that a PURL parsing/downloading tool could possibly break on that qualifier, if it just try to, for instance, use python Requests to fetch the content in the URL.
> 
> I also opened [package-url/purl-spec#89](https://github.com/package-url/purl-spec/issues/89) so the matter can be openly discussed within the PURL community.

My take is that, as PURL themselves state, [a purl is a URL](https://github.com/package-url/purl-spec#a-purl-is-a-url) and being used for downloading seems to be one of the main purposes, but it is a good idea to get the opinions of the PURL community and others.

athos-ribeiro:
> My take is that, as PURL themselves state, [a purl is a URL](https://github.com/package-url/purl-spec#a-purl-is-a-url) and being used for downloading seems to be one of the main purposes

This is why I added it in the download URL. My concern is that parsing a PURL is not supported for most URL parsers/resolvers out there :(

But we could do it and add documentaiton bits in OSBS and here.

cachito/web/models.py
```diff
@@ -208,10 +208,14 @@ def to_purl(self):
...
-                return f"generic/{purl_name}?{qualifier}"
-            elif not has_authority:
+            if not has_authority:
+                if protocol == "file":
+                    path = self.version.split(":", 1)[1]
+                    safe_path = urllib.parse.quote(path, safe="")
+                    # this path is relative to the parent package's package.json path
+                    # let's mark this purl for later substitution
+                    marker = content_manifest.PARENT_PURL_PLACEHOLDER
+                    return f"pkg:generic/{purl_name}?download_url={marker}&file_name={safe_path}"
```
mprahl, line 218: Keep in mind that `safe_path` is from the perspective of the `package.json` file of the main package. The main package could be in a subdirectory and thus the relative path of `../client` does not make sense in the PURL.

lcarva: Hmm that's a good point. We should normalize the path then.

athos-ribeiro: For this, we would need to know the parent's package.json path, which we do not at this point. How about including the path for the directory containing the package.json in the parent's PURL? Then we could keep the file_name as a relative path (this would likely require changing the version field to include such path).

lcarva: I see. I'm afraid I don't know the full implication of making such a change. Could we normalize `file_name` before it's saved as a dependency instead? That seems like the right thing to do, but it may be more intrusive.

chmeliik:
@athos-ribeiro what should we do with this one? I just noticed the broken purls for NPM while testing the Yarn CM.

athos-ribeiro:
@chmeliik, would you mind filing an issue downstream to track this one as well?

This is quite outdated and would require some re-work. I am closing this one.

chmeliik:
@athos-ribeiro ack, will do

PR #233, opened 2020-08-12
Implement function that resolves Pip metadata

chmeliik:
* CLOUDBLD-2053

Takes a path to the root directory of a Pip package, attempts to extract
name and version from setup.py or setup.cfg.

### DIFF
--- a/cachito/workers/pkg_managers/pip.py
+++ b/cachito/workers/pkg_managers/pip.py
@@ -20,6 +20,59 @@ log = logging.getLogger(__name__)
 NOTHING = object()  # A None replacement for cases where the distinction is needed
 
 
+def get_pip_metadata(package_dir):
+    """
+    Attempt to get the name and and version of a Pip package.
+
+    First, try to parse the setup.py script (if present) and extract name and version
+    from keyword arguments to the setuptools.setup() call. If either name or version
+    could not be resolved and there is a setup.cfg file, try to fill in the missing
+    values from metadata.name and metadata.version in the .cfg file.
+
+    If either name or version could not be resolved, raise an error.
+
+    :param str package_dir: Path to the root directory of a Pip package
+    :return: Tuple of strings (name, version)
+    :raises CachitoError: If either name or version could not be resolved
+    """
+    name = None
+    version = None
+
+    setup_py = SetupPY(package_dir)
+    setup_cfg = SetupCFG(package_dir)
+
+    if setup_py.exists():
+        log.info("Extracting metadata from setup.py")
+        name = setup_py.get_name()
+        version = setup_py.get_version()
+    else:
+        log.warning("No setup.py in directory, package is likely not Pip compatible")
+
+    if not (name and version) and setup_cfg.exists():
+        log.info("Filling in missing metadata from setup.cfg")
+        name = name or setup_cfg.get_name()
+        version = version or setup_cfg.get_version()
+
+    missing = []
+
+    if name:
+        log.info("Resolved package name: %r", name)
+    else:
+        log.error("Could not resolve package name")
+        missing.append("name")
+
+    if version:
+        log.info("Resolved package version: %r", version)
+    else:
+        log.error("Could not resolve package version")
+        missing.append("version")
+
+    if missing:
+        raise CachitoError(f"Could not resolve package metadata: {', '.join(missing)}")
+
+    return name, version
+
+
 def any_to_version(obj):
     """
     Convert any python object to a version string.
--- a/tests/test_workers/test_pkg_managers/test_pip.py
+++ b/tests/test_workers/test_pkg_managers/test_pip.py
@@ -35,6 +35,93 @@ def write_file_tree(tree_def, rooted_at):
             write_file_tree(value, entry_path)
 
 
+@pytest.mark.parametrize("py_exists", [True, False])
+@pytest.mark.parametrize("py_name", ["name_in_setup_py", None])
+@pytest.mark.parametrize("py_version", ["version_in_setup_py", None])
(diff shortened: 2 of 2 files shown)


cachito/workers/pkg_managers/pip.py
```diff
@@ -17,6 +17,53 @@
 NOTHING = object()  # A None replacement for cases where the distinction is needed
 
 
+def get_pip_metadata(package_dir):
```
chmeliik, line 20: This will probably need to take an instance of `cachito.workers.scm.SCM` instead, it seems that the name of the cloned directory is not expected to match the repo name (important for the fallback case).

How do want to handle the case where pip has a custom path defined, e.g.
```yaml
packages:
   pip:
     - path: marketplaces/tatooine
     - path: yoda
```

I guess we would use the repo name if the package is at top level and directory name otherwise (again, fallback case)?

cachito/workers/pkg_managers/pip.py
```diff
@@ -17,6 +17,53 @@
...
+        log.warning("No setup.py in directory, package is likely not Pip compatible")
+
+    if not (name and version) and setup_cfg.exists():
+        log.info("Filling in missing metadata from setup.cfg")
+        name = name or setup_cfg.get_name()
+        version = version or setup_cfg.get_version()
+
+    if not name:
+        log.warning("Falling back to using directory name as package name")
+        name = Path(package_dir).name
```
chmeliik, line 56: **Fallback case**

Not entirely sure about this, can the directory name be relied on to be the same as the repo name? I'm guessing yes if `package_dir` is what you pass in: https://github.com/release-engineering/cachito/blob/d3c0f9162b60e3c2b2e07be9ac694fd845ddecf6/cachito/workers/paths.py#L49-L54

Also keep in mind that we will want to support subpaths, e.g.:
```yaml
 packages:
   pip:
     - path: marketplaces/tatooine
     - path: yoda
```
So directory name seems like the correct solution

cachito/workers/pkg_managers/pip.py
```diff
@@ -17,6 +17,53 @@
...
+
+    log.info("Resolved package name: %r", name)
+    if version:
+        log.info("Resolved package version: %r", version)
+    else:
+        log.warning("Could not resolve package version")
```
chmeliik, line 62: Should this function deal with git data? From what I understand, the git ref (which would be the fallback version) will already be present in the request, so this function does not seem like the right place to deal with it.

athos-ribeiro: +1; This should be resolved later, when the function is called.

cachito/workers/pkg_managers/pip.py
athos-ribeiro: This seems to be reasonable. Although in such cases we have to have in mind to have the version to be explicit on where to fetch the packages from, as with npm packages. Otherwise, PURLs will not be able to identify the packages and there may be conflicts, since the unique key for the packages in the DB is (name, type, version).

Moreover, what if, for some reason, the version is inferred but the name is not (and then set to be the directory name)? I don't think this should be allowed here, since it could lead to conflicting package data. The opposite seems reasonable though (inferring the name but not a version).

Finally (and offtopic), since we are inferring the version from the package here, the top level package PURL should always be a "forge" PURL and never a pypi one.

mprahl: Just quickly scanning the PR here. I might have missed something but isn't the default case of `package_dir` the root of the repository? Since this is run in the Cachito bundle being constructed, the `package_dir` would be `app`.

I was originally thinking of just falling back to the Git repo's name, but I guess this only works when the Python package is in the root of the repository. So perhaps if this is in a subdirectory, then use the directory name, but if it's the root directory, then use the repo name. What do you think?

chmeliik: @mprahl looks like I was looking at the wrong directory class, I thought we would be interacting with the `sources_dir` attribute of an `SCM` instance, e.g.:
```python
repo = scm.Git(url, ref)
name, version = pip.get_pip_metadata(repo.sources_dir.package_dir)  # or a subpath
```
But upon closer inspection, it looks like `package_dir` will contain only the TAR archive, so this is probably not the right use case.

It seems that we will want to interact with a `RequestBundleDir` instead, and we definitely lose the repo name in the process of creating one.

@athos-ribeiro  @mprahl  Is there an object that this function could accept that contains both unpacked sources *and* information about the repo name? If not, would it be reasonable to defer resolution of fallback cases until we implement more of the Python pipeline?

lcarva: I think it's perfectly acceptable to raise an error if both the name and version cannot be resolved from either setup.py or setup.cfg. However, it looks like quay itself doesn't provide either :disappointed: 
Let me reach out to them to see if they're opposed to adding one of these file.

athos-ribeiro: > @athos-ribeiro @mprahl Is there an object that this function could accept that contains both unpacked sources and information about the repo name? If not, would it be reasonable to defer resolution of fallback cases until we implement more of the Python pipeline?

Not that I know of. I'd be OK with deferring resolution for now.

chmeliik:
Removed the fallback case for package name, now raises an error if metadata could not be resolved.

chmeliik:
@athos-ribeiro @lcarva @nirzari this should be ready for (another) review, if I correctly understood Lui's comment about raising an error when both version and name cannot be resolved.

cachito/workers/pkg_managers/pip.py
athos-ribeiro: I do not see a reason for being strict for both cases. +1 for having it strict and changing later when/if needed.

PR #234, opened 2020-08-12
Add support for parsing python's requirements

lcarva:
* CLOUDBLD-2054

### DIFF
--- a/cachito/workers/pkg_managers/pip.py
+++ b/cachito/workers/pkg_managers/pip.py
@@ -6,6 +6,7 @@ import re
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from pathlib import Path
+from urllib.parse import urlparse
 
 import pkg_resources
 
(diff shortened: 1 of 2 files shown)


cachito/workers/pkg_managers/pip.py
```diff
@@ -635,3 +637,347 @@ def _get_variable(self, var_name):
 
         log.error("Variable %r not found along the setup call branch", var_name)
         return None
+
+
+class PipRequirementsFile:
+    """Parse requirements from a pip requirements file."""
+
+    # Comment lines start with optional leading spaces followed by "#"
+    LINE_COMMENT = re.compile(r"(^|\s)+#.*$")
```
athos-ribeiro, line 646: Do we need the `+` quantifier there? If not need, removing it would make the regex faster.

cachito/workers/pkg_managers/pip.py
```diff
@@ -635,3 +637,347 @@ def _get_variable(self, var_name):
...
+        """Read and yield the lines from the requirements file.
+
+        Lines ending in the line continuation character are joined with the next line.
+        Comment lines are ignored.
+        """
+        buffered_line = []
+
+        with open(self.file_path) as f:
+            for line in f.read().splitlines():
+                if self.LINE_COMMENT.match(line):
```
athos-ribeiro, line 733: This should be `search` instead of `match` to capture comments in the end of a non-comment line. It would be nice to add a unit test for such case as well.

chmeliik: We don't want to ignore such lines though, right? We could do something like this instead:
```python
comment =  re.compile(r"\s*#.*")
line = comment.sub("", line)
if not line:
    continue
```

cachito/workers/pkg_managers/pip.py
chmeliik: I think the `+` is needed, but could be here instead: `(^|\s+)`. Not sure if that makes any difference

Reference: https://pip.pypa.io/en/stable/reference/pip_install/#requirements-file-format
> A line that begins with # is treated as a comment and ignored. Whitespace followed by a # causes the # and the remainder of the line to be treated as a comment.

athos-ribeiro: I am not sure if I follow the need for the `+` quantifier here. The spec says "**whitespece** followed by a #" and not "**whitespaces** followed by a #". Adding the `+` would make the regex engine match additional spaces for no additional benefits.

Are there any examples where the quantifier would make a difference? Note that this should be used in a `search` and not in a `match` since we are not parsing the spaces from the beginning of the line (^) in that regex.

cachito/workers/pkg_managers/pip.py
athos-ribeiro: +1

Note that `re.compile(r"\s*#.*")` may introduce an issue. it would remove `#123` from `foo#123`.  Maybe we could go with `re.compile(r"(^|\s)#.*")`.

The additional spaces before the last one preceding the `#` would need to be stripped later (alternativelly, we could use the - potentially slower - `+` quantifier).

cachito/workers/pkg_managers/pip.py
chmeliik: That's a good point. I suggested below to use `sub` instead of `search` or `match`, because for lines that do not start with a comment, we want to ignore only the comment and not the valid part of the line. But that would also not require removing all the extra whitespace, `str.split()` deals with that.

cachito/workers/pkg_managers/pip.py
```diff
@@ -635,3 +637,347 @@ def _get_variable(self, var_name):
...
+                    # Ignore commment lines
+                    continue
+                if not line.endswith("\\"):
```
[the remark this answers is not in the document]
lcarva: FWIW, both of these requirement files behave the same way:
```
https://github.com/quay/appr/archive/58c88e4952e95935c0dd72d4a24b0c44f2249f5b.tar.gz#egg=cnr_server \\
--hash=sha256:4fd9429bfbb796a48c0bde6bd301ff5b3cc02adb32189d912c7f55ec2e6c70c9
```

```
https://github.com/quay/appr/archive/58c88e4952e95935c0dd72d4a24b0c44f2249f5b.tar.gz#egg=cnr_server \
--hash=sha256:4fd9429bfbb796a48c0bde6bd301ff5b3cc02adb32189d912c7f55ec2e6c70c9
```

cachito/workers/pkg_managers/pip.py
```diff
@@ -635,3 +637,347 @@ def _get_variable(self, var_name):
...
+
+        if cls.HAS_NAME_IN_DIRECT_ACCESS_REQUIREMENT.search(line):
+            package_name, url = line.split("@", 1)
+
+        parsed_url = urlparse(url)
+        if parsed_url.fragment:
+            for section in parsed_url.fragment.split("&"):
+                if "=" in section:
+                    attr, value = section.split("=", 1)
+                    if attr == "egg":
```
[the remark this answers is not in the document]
lcarva: Good catch, I'll add another property, `dict`, to the `PipRequirement` class to capture all the qualifiers.

cachito/workers/pkg_managers/pip.py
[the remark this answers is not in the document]
chmeliik: Ok, I guess the specification should not be trusted too much when it comes to details then

cachito/workers/pkg_managers/pip.py
lcarva: pip treats any number of leading white spaces the same:
```
# no leading white spaces
 # one leading white space
         # many leading white spaces
```
So I think `+` is needed. However, it's at the wrong place on my regex. It should be `"^\s*#"`

lcarva: No, I'm wrong. See the other thread about `search` vs `match`

cachito/workers/pkg_managers/pip.py
lcarva: Thank you for catching this. I was not handling inline comments properly, e.g. `spam==1.2.3 # ignore me`.

cachito/workers/pkg_managers/pip.py
```diff
@@ -635,3 +637,347 @@ def _get_variable(self, var_name):
...
+            direct access requirement
+        """
+        direct_access_kind = None
+
+        if ":" not in line:
+            return None, False
+        scheme = line.split(":", 1)[0].lower()
+        if "@" in scheme:
+            # Strip off the package name from the scheme, e.g. "name @ https"
+            scheme = scheme.rsplit("@", 1)[-1].strip()
```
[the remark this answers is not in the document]
lcarva: pip would raise an error in this case. I'm doing the same now in this method so such issues are caught early.

cachito/workers/pkg_managers/pip.py
```diff
@@ -6,6 +6,8 @@
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from pathlib import Path
+from pkg_resources import parse_requirements, RequirementParseError
+from urllib.parse import urlparse
 
 import pkg_resources
```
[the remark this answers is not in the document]
lcarva: Missed that on the rebase. Good eye :)

cachito/workers/pkg_managers/pip.py
lcarva: This is now handled, let me know what you think.

cachito/workers/pkg_managers/pip.py
lcarva: This is now handled properly, let me know what you think.

lcarva:
@athos-ribeiro, @chmeliik, this is ready for another round. Thanks!

PR #237, opened 2020-08-18
Configure Python repositories for final usage

athos-ribeiro:
A part of supporting Python packages in Cachito, we need to configure
the temporary Nexus repositories, which will be available for a Cachito
request while assembling the bundle.

Here, we allow creating a user for a given request, who will have the
proper permissions to read from both the PyPI hosted and the raw hosted
Nexus request repositories.

### DIFF
--- a/cachito/workers/nexus_scripts/pip_after_content_staged.groovy
+++ b/cachito/workers/nexus_scripts/pip_after_content_staged.groovy
@@ -1,0 +1,93 @@
+/*
+This script configures Nexus so that a temporary user for the request is created and is given
+permission to access the PyPI hosted repository and the raw hosted repository. It'd be preferable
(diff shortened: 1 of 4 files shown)


cachito/workers/nexus_scripts/pip_after_content_staged.groovy
```diff
@@ -0,0 +1,88 @@
+/*
```
chmeliik, line 1: Can you clarify why we do not need to (or cannot) block outbound connections the same way we do for NPM?

athos-ribeiro: Sure! I included an explanation in this comment in the header of the file

PR #239, opened 2020-08-20
Download PyPI dependencies

chmeliik:
* CLOUDBLD-2055

### DIFF
--- a/cachito/workers/pkg_managers/pip.py
+++ b/cachito/workers/pkg_managers/pip.py
@@ -11,10 +11,15 @@ from pathlib import Path
 from urllib.parse import urlparse
 
 import pkg_resources
+import requests
 
 from cachito.errors import CachitoError, ValidationError
 from cachito.workers import nexus
+from cachito.workers.config import get_worker_config
 from cachito.workers.errors import NexusScriptError
+from cachito.workers.paths import RequestBundleDir
+from cachito.workers.pkg_managers.general import ChecksumInfo, verify_checksum
+from cachito.workers.requests import requests_session
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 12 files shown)


chmeliik:
While I work on adding necessary unit tests and other details, please take a look at how global options in requirements.txt. More specifically, please see the 2nd commit - do we want to fail the build for some options? Would you agree on the list of Rejected vs. Ignored options?

Also, if you want to try out the PyPI download against a real PyPI (proxy) server, you can use this script: https://gist.github.com/chmeliik/5be2888273af3c6ddbfd382be672c1a4

athos-ribeiro:
> Resolved comments related to `_process_options()`

That bit LGTM!

cachito/workers/pkg_managers/pip.py
```diff
@@ -1203,3 +1208,228 @@ def finalize_nexus_for_pip_request(pip_repo_name, raw_repo_name, username):
...
+
+    _validate_requirements(requirements_file.requirements, require_hashes)
+
+    bundle_dir = RequestBundleDir(request_id)
+    bundle_dir.pip_deps_dir.mkdir(parents=True, exist_ok=True)
+
+    config = get_worker_config()
+    pypi_proxy_url = config.cachito_nexus_pypi_proxy_url
+    pypi_proxy_auth = requests.auth.HTTPBasicAuth(
+        config.cachito_nexus_username, config.cachito_nexus_password
```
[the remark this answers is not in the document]
chmeliik: Done

chmeliik:
Made get_nexus_hoster_credentials() public, now using it for PyPI proxy auth

athos-ribeiro (approved):

chmeliik:
Removed a leftover `TODO`

cachito/workers/pkg_managers/pip.py
```diff
@@ -1203,3 +1208,264 @@ def finalize_nexus_for_pip_request(pip_repo_name, raw_repo_name, username):
...
+    # See https://warehouse.readthedocs.io/api-reference/json/
+    package_url = f"{pypi_url.rstrip('/')}/pypi/{package}/{version}/json"
+    try:
+        pypi_resp = requests_session.get(package_url, auth=pypi_auth)
+        pypi_resp.raise_for_status()
+    except requests.RequestException as e:
+        raise CachitoError(f"PyPI query failed: {e}")
+
+    data = pypi_resp.json()
+    sdists = [pkg for pkg in data["urls"] if pkg["packagetype"] == "sdist"]
```
nirzari, line 1391: Should we be using ` sdists = [pkg for pkg in data.get("urls", []) if pkg.get("packagetype", None) == "sdist"]` just to be safer as this is a response coming down from a request?

cachito/workers/pkg_managers/pip.py
```diff
@@ -1203,3 +1208,264 @@ def finalize_nexus_for_pip_request(pip_repo_name, raw_repo_name, username):
...
+    if sdist["yanked"]:
+        raise CachitoError(f"All sdists for package {package}=={version} are yanked")
+
+    try:
+        file_resp = requests_session.get(sdist["url"], stream=True)
+        file_resp.raise_for_status()
+    except requests.RequestException as e:
+        raise CachitoError(f"Could not download {sdist['filename']}: {e}")
+
+    package_dir = pip_deps_dir / package
```
nirzari, line 1406: Just wondering about this syntax out of curiosity. What do these extra spaces around `/` indicate if we are defining a simple path?

athos-ribeiro: @nirzari, the slash operator is special for pathlib path objects. It works like os.path.join.

See https://docs.python.org/3/library/pathlib.html#operators

cachito/workers/pkg_managers/pip.py
athos-ribeiro: We are already calling `raise_for_status()`  for the response and the call to `json()` should raise a ValueError` in case there is no JSON object in the response.

+1 to being safe here (we are also relying on the Nexus implementation here)

chmeliik: Done, also made a similar change for `sdist["yanked"]`. Will leave `sdist["url"]` and `sdist["filename"]` as is, we have no choice but to rely on them

athos-ribeiro (approved):

PR #246, opened 2020-08-25
Add groovy script to clean pip volatile resources

athos-ribeiro:
After a request is moved to the stale state, we should cleanup the
temporary Nexus repositories, user, and role created for the request.
This requires a new nexus script to be registered. This patch introduces
such script.

Note that the current script makes Nexus raise exceptions when removing
resources. This is a known issue, already present in the JS analogous
script. The exceptions are safe to ignore and should be fixed in Nexus
3.22. See https://issues.sonatype.org/browse/NEXUS-22829 for further
information.

### DIFF
--- a/cachito/workers/nexus_scripts/pip_cleanup.groovy
+++ b/cachito/workers/nexus_scripts/pip_cleanup.groovy
@@ -1,0 +1,55 @@
+/*
+This script deletes the PyPI hosted repositories (PyPI and raw), user, and role for the Cachito request.
+
+This script should get executed when the Cachito request is set to the stale state.
+ */
+import groovy.json.JsonSlurper
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import org.sonatype.nexus.security.authz.AuthorizationManager
+import org.sonatype.nexus.security.role.NoSuchRoleException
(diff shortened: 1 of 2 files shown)


akhmanova:
@athos-ribeiro Can we somehow test it locally with the `docker-compose` setup?

athos-ribeiro:
You could do something like

```
from cachito.workers.pkg_managers import pip
from cachito.workers import nexus

pypi_repo = "test-pip-hosted-1"
raw_repo = "test-raw-hosted-1"
username = "test-user"

pip.prepare_nexus_for_pip_request(pypi_repo, raw_repo)
pip.finalize_nexus_for_pip_request(pypi_repo, raw_repo, username)
```
Go to nexus at `localhost:8082` and check that the test repos, user, and role were created.

Then
```
payload = {
    "pip_repository_name": pypi_repo,
    "raw_repository_name": raw_repo,
    "username": username,
}
nexus.execute_script("pip_cleanup", payload)
```

And make sure they were removed.

chmeliik (approved):
LGTM :+1:

Tested with Athos' example, worked perfectly

akhmanova:
@athos-ribeiro Thank you for example! Good job!

PR #248, opened 2020-08-25
Get index url to be set as PIP_INDEX_URL

nirzari:
CLOUDBLD-2060

### DIFF
--- a/cachito/workers/pkg_managers/pip.py
+++ b/cachito/workers/pkg_managers/pip.py
@@ -1727,3 +1727,30 @@ def get_hosted_repositories_username(request_id):
     :rtype: str
     """
     return f"cachito-pip-{request_id}"
+
+
+def get_index_url(nexus_pypi_hosted_repo_url, username, password):
+    """
+    Get index url for the nexus pypi hosted repository for the request.
+
+    This is to be set as PIP_INDEX_URL.
+
+    :param str nexus_pypi_hosted_repo_url: the URL for the Nexus PyPI hosted repository
+        for the request
+    :param str username: the username of the user to be created for the Cachito pip request
+    :param str password: the password of the Nexus user that has access to the request's Python
+        repositories
+    :return: the index url
+    :rtype: str
+    """
+    # Instead of getting the token from Nexus, use basic authentication as supported by Pip:
+    # https://pip.pypa.io/en/stable/user_guide/#basic-authentication-credentials
+    # Create the request specific nexus repo url to be passed as `index-url` within pip.conf
+    if "://" in nexus_pypi_hosted_repo_url:
+        index_url = nexus_pypi_hosted_repo_url.replace("://", f"://{username}:{password}@", 1)
+    else:
+        raise CachitoError(
+            f"Nexus PyPI hosted repo URL: {nexus_pypi_hosted_repo_url} is not a valid URL"
+        )
+
+    return index_url
--- a/cachito/workers/pkg_managers/pip.py
+++ b/cachito/workers/pkg_managers/pip.py
@@ -348,12 +348,12 @@ class SetupCFG(SetupFile):
 
         if root_module in package_dir:
             custom_path = Path(package_dir[root_module])
-            log.debug(f"Custom path set for root module %r: %r", root_module, str(custom_path))
+            log.debug("Custom path set for root module %r: %r", root_module, str(custom_path))
             # Custom path replaces the root module
             module_path = custom_path.joinpath(*module_path.parts[1:])
         elif "" in package_dir:
             custom_path = Path(package_dir[""])
-            log.debug(f"Custom path set for all root modules: %r", str(custom_path))
+            log.debug("Custom path set for all root modules: %r", str(custom_path))
             # Custom path does not replace the root module
             module_path = custom_path / module_path
 
--- a/tests/test_workers/test_pkg_managers/test_pip.py
+++ b/tests/test_workers/test_pkg_managers/test_pip.py
@@ -2911,3 +2911,23 @@ def test_get_raw_hosted_repo_url():
 
 def test_get_hosted_repositories_username():
     assert pip.get_hosted_repositories_username(42) == "cachito-pip-42"
+
+
+def test_get_index_url():
+    index_url = pip.get_index_url(
+        "https://repository/cachito-pip-hosted-5/simple", "admin", "admin123",
+    )
+
+    expected = "https://admin:admin123@repository/cachito-pip-hosted-5/simple"
+
+    assert index_url == expected
+
+
+def test_get_index_url_invalid_url():
+    expected = (
+        "Nexus PyPI hosted repo URL: repository/cachito-pip-hosted-5/simple is not a valid URL"
+    )
+    with pytest.raises(CachitoError, match=expected):
+        pip.get_index_url(
+            "repository/cachito-pip-hosted-5/simple", "admin", "admin123",
+        )


cachito/workers/pkg_managers/pip.py
```diff
@@ -1514,3 +1517,101 @@ def upload_raw_package(repo_name, artifact_path, dest_dir, filename, is_request_
...
+        to the main Cachito repositories
+    :param str password: the password of the unprivileged user that has read access
+        to the main Cachito repositories
+    :param str custom_ca_path: the path to set ``cert`` to in the pip.conf file; if not provided,
+        this option will be omitted
+    :return: the contents of the pip.conf file
+    :rtype: str
+    """
+    # Instead of getting the token from Nexus, use basic authentication as supported by Nexus:
+    # https://help.sonatype.com/repomanager3/formats/npm-registry#npmRegistry-AuthenticationUsingBasicAuth
```
chmeliik, line 1538: This links to documentation about NPM, here is the relevant part of PyPI documentation: https://help.sonatype.com/repomanager3/formats/pypi-repositories#PyPIRepositories-ConfiguringPyPIClientTools

The link above does not mention anything specific about authentication, so I'm assuming we should just treat it as regular PyPI, making this the most relevant  document: https://pip.pypa.io/en/stable/user_guide/#basic-authentication-credentials

cachito/workers/pkg_managers/pip.py
```diff
@@ -1514,3 +1517,101 @@ def upload_raw_package(repo_name, artifact_path, dest_dir, filename, is_request_
...
+    Create a custom pip.conf file.
+
+    It should configure pip to point to the request specific repository in Nexus.
+
+    :param int request_id: the ID of the Cachito request
+    :param list package_configs: the list of optional package configurations submitted by
+    the user
+    :rtype: str
+    :return: The file path for pip configuration file that points to the request
+    specific repository in Nexus
```
chmeliik, line 1570: This seems to return a list of dicts rather than a string

cachito/workers/pkg_managers/pip.py
```diff
@@ -1514,3 +1517,101 @@ def upload_raw_package(repo_name, artifact_path, dest_dir, filename, is_request_
...
+        to the main Cachito repositories
+    :param str custom_ca_path: the path to set ``cert`` to in the pip.conf file; if not provided,
+        this option will be omitted
+    :return: the contents of the pip.conf file
+    :rtype: str
+    """
+    # Instead of getting the token from Nexus, use basic authentication as supported by Nexus:
+    # https://help.sonatype.com/repomanager3/formats/npm-registry#npmRegistry-AuthenticationUsingBasicAuth
+    # Create the request specific nexus repo url to be passed as `index-url` within pip.conf
+    nexus_repo_url = base64.b64encode(
```
chmeliik, line 1540: I don't think this should be base64 encoded

cachito/workers/pkg_managers/pip.py
```diff
@@ -1514,3 +1517,101 @@ def upload_raw_package(repo_name, artifact_path, dest_dir, filename, is_request_
...
+        f"https://[{username}[:{password}]@]{nexus_repo_identifier}".encode("utf-8")
```
chmeliik, line 1541: `nexus_repo_identifier` is just the repo name, right? This need to be the full URL, something like (not exactly) `https://username:password@localhost:8081/repository/cachito-pip-hosted-1/simple`.

The square brackets around username and password should also be removed, they are usually used in documentation to indicate that something is optional.

cachito/workers/pkg_managers/pip.py
nirzari: Yes, I was only returning the single pip file earlier but then later changed the logic and forgot to update the doc strings. Thanks for catching this.

cachito/workers/pkg_managers/pip.py
nirzari: Oops, thanks for noticing the brackets! I forgot to remove them after copying from documentation. I planned on discussing the exact URL needed. @athos-ribeiro Do you know the exact URL for request specific nexus hosted repository ?

athos-ribeiro: This is the `cachito_nexus_url` value in the `cachito/workers/config.py` for the hosted repositories. It would be the hoster url for the proxy and the permanent raw repository.

cachito/workers/pkg_managers/pip.py
```diff
@@ -1575,3 +1578,104 @@ def get_hosted_repositories_username(request_id):
...
+    pip_conf_content = textwrap.dedent(
+        f"""\
+        [global]
+        index-url = {nexus_pypi_hosted_repo_url}
+        """
+    )
+    if custom_ca_path:
+        # The CA could be embedded in the actual pip.conf file with the `cert` option, but this
+        # would mean that the CA file contents would be duplicated in the Cachito database for
+        # every request that uses a pip conf file
```
chmeliik, line 1619: Is this true? I know this works for NPM, but I haven't found any Pip documentation about this. If you have a source for this, it would be nice to include it, otherwise I would drop the comment.

cachito/workers/pkg_managers/pip.py
```diff
@@ -1575,3 +1578,104 @@ def get_hosted_repositories_username(request_id):
...
+    :param str username: the username of the unprivileged user that has read access
+        to the main Cachito repositories
+    :param str password: the password of the unprivileged user that has read access
+        to the main Cachito repositories
+    :param str custom_ca_path: the path to set ``cert`` to in the pip.conf file; if not provided,
+        this option will be omitted
+    :return: the contents of the pip.conf file
+    :rtype: str
+    """
+    # Instead of getting the token from Nexus, use basic authentication as supported by Nexus:
```
chmeliik, line 1598: Nitpick: Pip is what supports the basic authentication, not Nexus (well Nexus too, but here Pip is more relevant)

cachito/workers/pkg_managers/pip.py
```diff
@@ -1575,3 +1578,104 @@ def get_hosted_repositories_username(request_id):
...
+    if custom_ca_path:
+        # The CA could be embedded in the actual pip.conf file with the `cert` option, but this
+        # would mean that the CA file contents would be duplicated in the Cachito database for
+        # every request that uses a pip conf file
+        pip_conf_content += f'cert = "{custom_ca_path}"\n'
+
+    return pip_conf_content
+
+
+def create_custom_pip_conf_file(request_id, package_configs=None):
```
chmeliik, line 1625: I found the NPM equivalent of this code in `tasks.npm.fetch_sources`. Is this module the right place for this function? Should it be in `tasks.pip` instead?

(or, taking this a step further, is this the right PR to add this functionality?)

athos-ribeiro: The fetch sources will be a task for python as well in `tasks.pip`. However, the `generate_npmrc_content` is in `pkg_managers.general_js`.

To have an analogous implementation, generating the file contents should be in pkg_managers and this will be used in a task.

Thoughts?

nirzari: >The fetch sources will be a task for python as well in tasks.pip. However, the generate_npmrc_content is in pkg_managers.general_js.

>To have an analogous implementation, generating the file contents should be in pkg_managers and this will be used in a task.

I agree! By `pkg_managers`, did you mean `pkg_managers/pip.py` ?

>Thoughts?

chmeliik: Right, `generate_pip_conf_content` belongs here for sure, what I'm concerned about is the second function, which generates data to put in the Cachito response. I'm not *that* opposed to having it here, fair warning though: if we're keeping it, I will have a couple more comments about the body of the function :D

athos-ribeiro: Yes, @chmeliik has a point  here.

This should either go to the tasks.pip module or be left to be implemented in the fetch_sources work as in the npm module.

nirzari: Since we already have it implemented, I might just move it to `tasks/pip.py` over removing it and then it can be used within fetch_sources there.

@chmeliik I'm waiting for your comments on the body of the function :)

cachito/workers/pkg_managers/pip.py
```diff
@@ -1575,3 +1578,104 @@ def get_hosted_repositories_username(request_id):
...
+        # The custom CA will be called ca.cert in the "app" directory
+        pip_config_files.append(
+            {
+                "content": base64.b64encode(ca_cert.encode("utf-8")).decode("utf-8"),
+                "path": os.path.join("app", "ca.cert"),
+                "type": "base64",
+            }
+        )
+
+    for subpath in subpaths:
```
chmeliik, line 1661: Do we need to generate a config file for each of the subpaths? Unlike `.npmrc`, the location of `pip.conf` should not matter as long as we set the `PIP_CONFIG_FILE` environment variable. I believe we need to set the variable anyway, unless we can somehow put the file in one of the default locations, see https://pip.pypa.io/en/stable/user_guide/#config-file

I think what we need is the cert, a global (for the request) `pip.conf` file and a `PIP_CONFIG_FILE` env var

cachito/workers/pkg_managers/pip.py
```diff
@@ -1575,3 +1578,104 @@ def get_hosted_repositories_username(request_id):
...
+        if ca_cert:
+            # Determine the relative path to the ca.cert file
+            custom_ca_path = os.path.relpath("ca.cert", start=subpath)
+        else:
+            custom_ca_path = None
+        pip_conf_content = generate_pip_conf_content(
+            nexus_pypi_hosted_repo_url,
+            config.cachito_nexus_proxy_username,
+            config.cachito_nexus_proxy_password,
```
chmeliik, line 1670: I'm not completely sure about this, but maybe these should be
- username: what you pass to `finalize_nexus_for_pip_request()`
- password: what you get back from `finalize_nexus_for_pip_request()`

https://github.com/release-engineering/cachito/blob/a25aca2642767192ed1c2df3d86cd0118d4adbd8/cachito/workers/pkg_managers/pip.py#L1208

@athos-ribeiro ^ is this correct?

If it is correct, it would probably be better to accept username and password as params for this function.

athos-ribeiro: +1

The username and password here should be the temporary ones, for the request volatile repositories.

The fetch_sources function will need the auth sdata and will fetch it already. Receiving the params here should be enough, as Adam suggested.

cachito/workers/pkg_managers/pip.py
nirzari: I'm not sure how will we be configuring the `PIP_CONFIG_FILE` env var here to have it all accessible but according to the Cachito Python Support doc: "multiple paths should be supported like the “npm” package manager." and hence, we are generating config file for each of the sub paths.

chmeliik: There is a difference between `npm` and `pip` when it comes to handling config files. When executing `npm`, it will look for `<current_directory>/.npmrc`. Pip, however, will *not* look for `<current_directory>/pip.conf`, so even if we generate one such file for each subpath, they will all be ignored. 

If I'm not mistaken, the only difference between the N generated pip.conf files for each of the subpaths would be the `cert` value, which is a relative path from the subdirectory to the CA cert. If that's the case, then we only need one pip.conf with the correct cert path, and we need to correctly set the `PIP_CONFIG_FILE` environment variable.

chmeliik:
### Summary of offline discussion

1. We do not want one `pip.conf` for each of the subpaths
    - Pip does not support reading config files from `<current_directory>/pip.conf`
    - We need to provide a `PIP_CONFIG_FILE` env var, which will point to the pip.conf file
        - We can only provide one such variable, so multiple config files are not even an option

2. We cannot use a relative path for the CA cert in `pip.conf`
    - The `pip` command would only work when running from the same directory in which we put the pip.conf file
    
3. Solution for the above
    - Generate one `pip.conf` file containing only the `index-url` option
    - Provide the `PIP_CONFIG_FILE` env var with a relative path to this file (OSBS will make it absolute)
    - Provide the `PIP_CERT` env var with a relative path to the CA cert (OSBS will make it absolute)

3.5. Simplified solution
- Provide the `PIP_INDEX_URL` and `PIP_CERT` env vars, no config file needed

---

CC @nirzari @lcarva @athos-ribeiro 
WDYT about 3.5.? Should we drop the config file altogether?

cachito/workers/pkg_managers/pip.py
chmeliik: Continued in https://github.com/release-engineering/cachito/pull/248#issuecomment-684902738

nirzari:
> ### Summary of offline discussion
> 1. We do not want one `pip.conf` for each of the subpaths
>    
>    * Pip does not support reading config files from `<current_directory>/pip.conf`
>    * We need to provide a `PIP_CONFIG_FILE` env var, which will point to the pip.conf file
>      
>      * We can only provide one such variable, so multiple config files are not even an option
> 2. We cannot use a relative path for the CA cert in `pip.conf`
>    
>    * The `pip` command would only work when running from the same directory in which we put the pip.conf file
> 3. Solution for the above
>    
>    * Generate one `pip.conf` file containing only the `index-url` option
>    * Provide the `PIP_CONFIG_FILE` env var with a relative path to this file (OSBS will make it absolute)
>    * Provide the `PIP_CERT` env var with a relative path to the CA cert (OSBS will make it absolute)
> 
> 3.5. Simplified solution
> 
> * Provide the `PIP_INDEX_URL` and `PIP_CERT` env vars, no config file needed
> 
> CC @nirzari @lcarva @athos-ribeiro
> WDYT about 3.5.? Should we drop the config file altogether?

3 & 3.5 both sound like good solutions. The only concern with 3.5 is exposing creds in a env var as part of `PIP_INDEX_URL` instead of keeping them in the conf file but I'm okay with either of them since we don't encrypt the conf file anyway.

If 3.5, I can just alter this PR to have a function to build `index_url` value for `pip.conf` file.

@athos-ribeiro @chmeliik Thoughts?

chmeliik:
> 3 & 3.5 both sound like good solutions. The only concern with 3.5 is exposing creds in a env var as part of `PIP_INDEX_URL` instead of keeping them in the conf file but I'm okay with either of them since we don't encrypt the conf file anyway.
> 
> If 3.5, I can just alter this PR to have a function to build `index_url` value for `pip.conf` file.
> 
> @athos-ribeiro @chmeliik Thoughts?

Not sure if exposing credentials is something we have to worry about, my guess is we don't (since we *want* users to be able to use them). @athos-ribeiro ?
If I'm right about that, I'd vote for 3.5. @nirzari your suggestion about altering this PR makes sense to me.

athos-ribeiro:
The repositories for which the credentials are created are temporary (they will be cleaned up as soon as a request state is changed to "stale"). There is a technical reason for the repositories not being open for anonymous requests, which is explained in the groovy scripts that create such repositories (it is due to a Nexus limitation).

The purpose of these credentials we want to expose is to make sure the user who requested the repositories will be able to read from them. The credentials do not allow anything else other than reading from the pypi and the raw temporary repositories. 

That said, using the environment variables (3.5) will expose the repository credentials in the OSBS logs. I am not against it since ideally these repos would be open for anonymous reading and this is indeed a much simpler approach. It would just be fair to add a comment on that and/or document it and make sure the OSBS folks are aware that this is intentional. Cc @MartinBasti

cachito/workers/tasks/pip.py
```diff
@@ -8,6 +11,9 @@
 from cachito.workers.tasks.celery import app
 
 
+log = logging.getLogger(__name__)
```
nirzari, line 14: I'm not using it for my code, but I'll just leave it here because I'm sure we'll be using it in the other parts of the file.

athos-ribeiro (commented):
LGTM.

I am wondering if the new function shouldn't be in the pkg_managers package instead though given the recent changes.

nirzari:
@athos-ribeiro Oops, I think you're right. I moved it.

athos-ribeiro (approved):

chmeliik (approved):
LGTM :+1:

PR #251, opened 2020-08-27
 Implement support for downloading VCS pip deps 

chmeliik:
* CLOUDBLD-2056

pip.download_dependencies() will no longer ignore VCS dependencies

The only supported VCS is Git.

After downloading from git, will upload the archive to Nexus.

### DIFF
--- a/cachito/workers/pkg_managers/pip.py
+++ b/cachito/workers/pkg_managers/pip.py
@@ -1239,6 +1239,9 @@ def download_dependencies(request_id, requirements_file):
 
     :param int request_id: ID of the request these dependencies are being downloaded for
     :param PipRequirementsFile requirements_file: A requirements.txt file
+    :return: Info about downloaded packages; all items will contain "kind" and "path" keys
+        (and more based on kind, see _download_*_package functions for more details)
+    :rtype: list[dict]
     """
     options = _process_options(requirements_file.options)
 
@@ -1274,11 +1277,11 @@ def download_dependencies(request_id, requirements_file):
         log.info("Downloading %s", req.download_line)
 
         if req.kind == "pypi":
-            download_path = _download_pypi_package(
+            download_info = _download_pypi_package(
                 req, bundle_dir.pip_deps_dir, pypi_proxy_url, pypi_proxy_auth
             )
         elif req.kind == "vcs":
-            download_path = _download_vcs_package(
+            download_info = _download_vcs_package(
                 req, bundle_dir.pip_deps_dir, pip_raw_repo_name, nexus_auth
             )
         else:
@@ -1288,13 +1291,16 @@ def download_dependencies(request_id, requirements_file):
         log.info(
             "Successfully downloaded %s to %s",
             req.download_line,
-            download_path.relative_to(bundle_dir),
+            download_info["path"].relative_to(bundle_dir),
         )
-        downloads.append((req, download_path))
 
-    if require_hashes:
-        for req, download_path in downloads:
-            _verify_hash(download_path, req.hashes)
+        if require_hashes:
+            _verify_hash(download_info["path"], req.hashes)
+
+        download_info["kind"] = req.kind
+        downloads.append(download_info)
+
+    return downloads
 
 
 def _process_options(options):
@@ -1421,7 +1427,7 @@ def _download_pypi_package(requirement, pip_deps_dir, pypi_url, pypi_auth):
     :param str pypi_url: URL of PyPI (proxy) server
     :param requests.auth.AuthBase pypi_auth: Authorization for the PyPI server
 
-    :return: Path to downloaded file
+    :return: Dict with package name, version and download path
     """
     package = requirement.package
     version = requirement.version_specs[0][1]
@@ -1450,7 +1456,14 @@ def _download_pypi_package(requirement, pip_deps_dir, pypi_url, pypi_auth):
 
     download_binary_file(sdist["url"], download_path)
 
-    return download_path
+    package_info = data.get("info", {})
+
+    return {
+        # Use canonical package name and version from PyPI response (if present)
+        "package": package_info.get("name") or package,
+        "version": package_info.get("version") or version,
+        "path": download_path,
+    }
 
 
 def _sdist_preference(sdist_pkg):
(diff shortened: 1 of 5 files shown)


chmeliik:
How  to test everything that you can download so far

step 1
```shell
sed 's;http://localhost:8081/repository/cachito-pip-proxy/;http://nexus:8081/repository/cachito-pip-proxy/;' \
    cachito/workers/config.py --in-place
```
step 2
```shell
docker-compose exec cachito-worker bash -c '                                      
cat << EOF > /tmp/requirements.txt
cachito @ git+https://github.com/release-engineering/cachito@63c5c32b361918730ba3ac06a115be18ece3e6b0
flexmock==0.10.4
EOF
'
```
step 3
```shell
docker-compose exec cachito-worker python3 -c '                                   
from cachito.workers.pkg_managers import pip

requirements_file = pip.PipRequirementsFile("/tmp/requirements.txt")
pip.download_dependencies(100_000, requirements_file)
'
```
step 4: check that it worked
```shell
docker-compose exec cachito-worker ls /tmp/cachito-archives/bundles/temp/100000 -R
curl -u admin:admin 'http://localhost:8082/service/rest/v1/search?repository=cachito-pip-raw'
```

chmeliik:
Next, I will make the `download_dependencies()` function return something useful, either in another commit or another PR (but it would be after this one)

chmeliik:
> Next, I will make the `download_dependencies()` function return something useful, either in another commit or another PR (but it would be after this one)

Done that

cachito/workers/pkg_managers/pip.py
```diff
@@ -1448,6 +1477,115 @@ def _sdist_preference(sdist_pkg):
     return yanked_pref, filetype_pref
 
 
+def _download_vcs_package(requirement, pip_deps_dir, pip_raw_repo_name, nexus_auth):
+    """
+    Fetch the source for a Python package from VCS (only git is supported).
+
+    After downloading, upload this package to the Nexus raw repository and on subsequent
```
athos-ribeiro, line 1497: can we also say we will upload the package to the Nexus hoster instance if 2 nexus instances are being used?

cachito/workers/pkg_managers/pip.py
```diff
@@ -1448,6 +1477,115 @@ def _sdist_preference(sdist_pkg):
...
+    downloads, reuse the uploaded asset instead of fetching from VCS again.
+
+    :param PipRequirement requirement: VCS requirement from a requirements.txt file
+    :param Path pip_deps_dir: The deps/pip directory in a Cachito request bundle
+    :param str pip_raw_repo_name: Name of the Nexus raw repository for Pip
```
athos-ribeiro, line 1502: I am wondering if this should be a param, given the raw hosted repository in the hoster Nexus instance should be a constant (probably in the config module).

cachito/workers/pkg_managers/pip.py
```diff
@@ -1448,6 +1477,115 @@ def _sdist_preference(sdist_pkg):
...
+    :param requests.auth.AuthBase nexus_auth: Authorization for the Nexus raw repo
```
athos-ribeiro, line 1503: This could also be fetched from config, if the case above applies here.

cachito/workers/pkg_managers/pip.py
```diff
@@ -1448,6 +1477,115 @@ def _sdist_preference(sdist_pkg):
...
+    package_dir = pip_deps_dir.joinpath(git_info["host"], *namespace_parts, repo_name)
+    package_dir.mkdir(parents=True, exist_ok=True)
+
+    filename = f"{repo_name}-external-gitcommit-{ref}.tar.gz"
+    download_path = package_dir / filename
+
+    # Check if we already have the raw component
+    raw_component_name = f"{repo_name}/{filename}"
+    log.debug("Looking for raw component %r in %r repo", raw_component_name, pip_raw_repo_name)
+    download_url = nexus.get_raw_component_asset_url(pip_raw_repo_name, raw_component_name)
```
athos-ribeiro, line 1523: Note for future self: this gets the component in the nexus hoster instance, if available.

cachito/workers/pkg_managers/pip.py
```diff
@@ -1448,6 +1477,115 @@ def _sdist_preference(sdist_pkg):
...
+        log.debug("Fetched package, uploading as %r to %r", raw_component_name, pip_raw_repo_name)
+        upload_raw_package(
+            pip_raw_repo_name,
+            repo.sources_dir.archive_path,
+            dest_dir=repo_name,
+            filename=filename,
+            is_request_repository=False,
+        )
+        # Move downloaded archive to expected download path
+        shutil.move(repo.sources_dir.archive_path, download_path)
```
athos-ribeiro, line 1541: Interesting. So, in this case, we would not be using the current feature to re-use sources avaiable in `fetch_source`?

chmeliik: Well, we only want to fetch from git the first time, after that we want to fetch from Nexus right? I can use `copy()` here, though it might be a minor waste of space, WDYT?

cachito/workers/pkg_managers/pip.py
chmeliik: I fetch all the relevant values from config once in `download_dependencies()` and then pass them to the helper functions as params to avoid fetching them multiple times. I can change it to always fetch config values in helper functions if that would be preferrable.

cachito/workers/pkg_managers/pip.py
chmeliik: Nevermind, this is a bad idea, if someone uses X as their app and someone else uses X as a dependency, this could delete the archive that was created for the app. I will use `copy()` instead

cachito/workers/pkg_managers/pip.py
chmeliik: Yup, done. Side question, what exactly does the term "hoster" mean in this context?

athos-ribeiro: That is how we named all the variables that refer to the cachito instance that hosts the permanent Cachito repositories

chmeliik: Makes sense, thanks :-)

cachito/workers/pkg_managers/pip.py
athos-ribeiro: No need! This looks nice the way it is. I was just wondering if we needed the additional params since they are in fact constant values. Making the change would not reduce complexity anyway.

chmeliik: Should I take the thumbs up as "keep it this way" or "change it"? :D

athos-ribeiro (approved):

PR #257, opened 2020-09-03
Fix downloading VCS Pip dependencies

chmeliik:
The URL passed to 'git clone' must not contain any params, query or
fragment.

### DIFF
--- a/tests/test_workers/test_pkg_managers/test_pip.py
+++ b/tests/test_workers/test_pkg_managers/test_pip.py
@@ -2521,25 +2521,38 @@ class TestDownload:
     @pytest.mark.parametrize(
         "url, nonstandard_info",  # See body of function for what is standard info
         [
-            (f"git+https://github.com/monty/python@{GIT_REF}", None),
             (
+                # Standard case
+                f"git+https://github.com/monty/python@{GIT_REF}",
+                None,
+            ),
+            (
+                # Ref should be converted to lowercase
                 f"git+https://github.com/monty/python@{GIT_REF.upper()}",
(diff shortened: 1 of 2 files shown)


tests/test_workers/test_pkg_managers/test_pip.py
```diff
@@ -2548,23 +2561,33 @@ def test_download_vcs_package(
                 },
             ),
             (
+                # Port should (?) be part of host
```
chmeliik, line 2564: Yes or no? This would mean we save the file in `deps/pip/github.com:443/namespace/repo/repo-external-gitcommit-...` (or `deps/pip/my-private-git-server:8083/foo/bar/...`)

chmeliik:
Replaced `None` in `url._replace` with `""` (you're not supposed to mix strings and non-strings in `urllib.parse.ParseResult`

tests/test_workers/test_pkg_managers/test_pip.py
athos-ribeiro: This should be important information for locating a resource. Are we storing ports for npm?

athos-ribeiro (commented):
LGTM, not sure about having ports in the directory names. I am not against it though.

tests/test_workers/test_pkg_managers/test_pip.py
chmeliik: I'm going to keep the port number, like you said it may be important info for locating a resource

athos-ribeiro (approved):

PR #241, opened 2020-08-20
Add function to upload Python packages to Nexus

athos-ribeiro:
Signed-off-by: Athos Ribeiro <athos@redhat.com>

PR #259, opened 2020-09-03
Manually set Nexus asset component file names

athos-ribeiro:
After [1], the file binary data is passed in a dict to the proper
component upload function. As a side effect, Requests is unable to infer
a file name with [2]. This would make the file name to be set as the
dict key carrying the binary data when uploading python assets. As a
side effect, pip would be unable to fetch contents from the hosted
python repository.

This patch sets the file names for the asset only components manually,
instead of deferring it to python requests.

[1] https://github.com/release-engineering/cachito/pull/241
[2] https://github.com/psf/requests/blob/master/requests/utils.py#L220


chmeliik (approved):
LGTM :+1:

Tested locally, the PyPI package now has the proper name and version.

PR #261, opened 2020-09-03
Add validation function for pip configurations

athos-ribeiro:

### DIFF
--- a/cachito/workers/config.py
+++ b/cachito/workers/config.py
@@ -268,6 +268,21 @@ def validate_npm_config():
         )
 
 
+def validate_pip_config():
+    """
+    Perform validation on the Celery configuration for the pip package manager.
+
+    :raise ConfigError: if the Celery configuration isn't configured for pip
+    """
+    validate_nexus_config()
+    conf = get_worker_config()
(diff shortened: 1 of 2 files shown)


cachito/workers/config.py
```diff
@@ -268,6 +268,21 @@ def validate_npm_config():
...
 
+def validate_pip_config():
+    """
+    Perform validation on the Celery configuration for the pip package manager.
+
+    :raise ConfigError: if the Celery configuration isn't configured for pip
+    """
+    validate_nexus_config()
+    conf = get_worker_config()
+    for pip_conf in ("cachito_nexus_pypi_proxy_url", "cachito_nexus_pip_raw_repo_name"):
```
nirzari, line 279: I understand why we are naming variable as `pip_conf` and the function name `validate_pip_config`, but for a second I was confused with `pip.conf`. I am wondering if we should adapt to a different naming here.

Otherwise validation function LGTM!

athos-ribeiro: Any suggestions for the function name?

nirzari: Yes, that's less confusing than the previous one. Thanks for fixing that.

nirzari (approved):

PR #262, opened 2020-09-04
Add pip-dev-preview flag

nirzari:
The pip package manager can only be used if this flag is used. Otherwise, the request should fail. The flag should be removed after UAT.

### DIFF
--- a/cachito/web/migrations/versions/044548f3f83a_add_pip_dev_preview_flag.py
+++ b/cachito/web/migrations/versions/044548f3f83a_add_pip_dev_preview_flag.py
@@ -1,0 +1,42 @@
+"""Add the pip-dev-preview flag
+
+Revision ID: 044548f3f83a
+Revises: cb6bdbe533cc
+Create Date: 2020-09-03 02:16:57.554550
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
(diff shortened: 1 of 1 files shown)


cachito/web/migrations/versions/044548f3f83a_add_pip_dev_preview_flag.py
```diff
@@ -0,0 +1,42 @@
+"""empty message
```
chmeliik, line 1: Nitpick: I'm comparing this file with the `add_gomod_vendor_flag.py` one. That one says "Add the gomod-vendor flag" in the comment, could this have a similar message?

Out of curiosity, what tool (if any) did you use to generate this file?

athos-ribeiro: @chmeliik : https://github.com/release-engineering/cachito/blob/master/README.md#database-migrations

nirzari: Aaah :/ How did I forget that. Updated! Thanks

athos-ribeiro (approved):

chmeliik (approved):
LGTM :+1:

PR #256, opened 2020-09-02
Resolve pip dependencies

athos-ribeiro:
This is a PoC for pip dependency resolving for visibility and is not intended to be merger.

We will use this as a tracker for what needs to be implemented.

PR #263, opened 2020-09-06
Resolve and fetch pip dependencies

athos-ribeiro:


athos-ribeiro:
We are still missing

- injecting the new (modified) requirement files through adding configuration files to the request
- handling `url` dependencies (depends on #255)

These can be done in a separate PR.

tests/test_workers/test_pkg_managers/test_pip.py
```diff
@@ -2971,3 +2999,114 @@ def test_get_index_url_invalid_url():
...
+    build_req_file = tmp_path / pip.DEFAULT_BUILD_REQUIREMENTS_FILE
+    req_file.write_text("bar==2.1")
+    build_req_file.write_text("baz==0.0.5")
+    mock_metadata.return_value = ("foo", "1.0")
+    mock_download.side_effect = [
+        [{"kind": "pypi", "path": "some/path", "package": "bar", "version": "2.1"}],
+        [{"kind": "pypi", "path": "another/path", "package": "baz", "version": "0.0.5"}],
+    ]
+    request = {"id": 1}
+    pkg_info = pip.resolve_pip(tmp_path, request)
```
chmeliik, line 3098: Could we also have a test where requirement files are passed in as a param?

chmeliik (commented):
Looks good! Just concerned about absolute vs. relative paths.

If I were to test this locally, what would I have to do to  get around the following response from the API?
```json
{
  "error": "The following package managers are invalid: pip"
}
```

athos-ribeiro:
> Looks good! Just concerned about absolute vs. relative paths.
> 
> If I were to test this locally, what would I have to do to get around the following response from the API?
> 
> ```json
> {
>   "error": "The following package managers are invalid: pip"
> }
> ```

You may want to check the additional changes I applied in https://github.com/release-engineering/cachito/pull/256. Namely, 
- https://github.com/release-engineering/cachito/pull/256/files#diff-a2ad776d633edd377eef975f3de24505R242
- https://github.com/release-engineering/cachito/pull/256/files#diff-64f5179147532e68ca805eb0ef1e2b35R38
- https://github.com/release-engineering/cachito/pull/256/files#diff-7cbde6013c2753c0bb969ab72bd9d02eR245

Finally, we are still missing the migration for adding the `pip` pkg manager to the DB. for now, you could just duplicate this line in this migration: https://github.com/release-engineering/cachito/blob/master/cachito/web/migrations/versions/615c19a1cee1_add_npm.py#L29 substituting `npm` for `pip` and it should do the trick (given the changes listed above are also applied).

cachito/workers/tasks/pip.py
```diff
@@ -17,3 +35,61 @@ def cleanup_pip_request(request_id):
...
+
+    log.info("Finalizing the Nexus configuration for pip for the request %d", request_id)
+    set_request_state(request_id, "in_progress", "Finalizing the Nexus configuration for pip")
+    username = get_hosted_repositories_username(request_id)
+    password = finalize_nexus_for_pip_request(pip_repo_name, raw_repo_name, username)
+
+    # Set environment variables
+    raw_url = get_pypi_hosted_repo_url(request_id)
+    pip_index_url = get_index_url(raw_url, username, password)
+    env_vars = {"PIP_INDEX_URL": {"value": pip_index_url, "kind": "literal"}}
```
chmeliik, line 84: When testing locally, I realized that with the `DevelopmentConfig`, this URL will not be accessible externally:
```
$ curl localhost:8080/api/v1/requests/1/environment-variables
{
  "PIP_INDEX_URL": {
    "kind": "literal", 
    "value": "http://cachito-pip-1:2f072c2325fa53d2299259004caf980d@nexus:8081/repository/cachito-pip-hosted-1/"
  }
}
```
nexus:8081 would have to be localhost:8082 instead

I assume this will only be an issue for the dev environment, not for any real environment. Is this something we need to worry about?

athos-ribeiro: This should come from `cachito_nexus_url`. In production environments they should contain a fqdn, so this would not be an issue in such environments.

tests/test_workers/test_pkg_managers/test_pip.py
athos-ribeiro: Done

athos-ribeiro:
Nice catch on the absolute paths, @chmeliik. Thanks :)

Fixed the issues and added the additional test.

chmeliik:
> You may want to check the additional changes I applied in #256. Namely,
> 
>     * https://github.com/release-engineering/cachito/pull/256/files#diff-a2ad776d633edd377eef975f3de24505R242
> 
>     * https://github.com/release-engineering/cachito/pull/256/files#diff-64f5179147532e68ca805eb0ef1e2b35R38
> 
>     * https://github.com/release-engineering/cachito/pull/256/files#diff-7cbde6013c2753c0bb969ab72bd9d02eR245
> 
> 
> Finally, we are still missing the migration for adding the `pip` pkg manager to the DB. for now, you could just duplicate this line in this migration: https://github.com/release-engineering/cachito/blob/master/cachito/web/migrations/versions/615c19a1cee1_add_npm.py#L29 substituting `npm` for `pip` and it should do the trick (given the changes listed above are also applied).

We will also need to update this: 
https://github.com/release-engineering/cachito/blob/bd49c6747b1351156c92319776c7c70b1bd33bc7/cachito/web/models.py#L410
https://github.com/release-engineering/cachito/blob/bd49c6747b1351156c92319776c7c70b1bd33bc7/cachito/web/models.py#L419

That should probably be left for a different PR, just making sure we're aware of it.

cachito/workers/pkg_managers/pip.py
```diff
@@ -1921,3 +1926,132 @@ def get_index_url(nexus_pypi_hosted_repo_url, username, password):
...
+    if requirement["kind"] == "pypi":
+        upload_pypi_package(pip_repo_name, requirement["path"])
+        dep = {"name": requirement["package"], "version": requirement["version"], "type": "pip"}
+    else:
+        dest_dir, filename = requirement["raw_component_name"].rsplit("/", 1)
+        upload_raw_package(raw_repo_name, requirement["path"], dest_dir, filename, True)
+        version = (
+            f"vcs:{requirement['host']}/{requirement['namespace']}/{requirement['repo']}@"
+            f"{requirement['ref']}"
+        )
```
chmeliik, line 1983: Now that #255 is merged, do you want to add "url" handling here? Or in another PR? Currently, url dependencies woud fail here with a `KeyError: 'host'`

cachito/workers/pkg_managers/pip.py
```diff
@@ -1921,3 +1926,138 @@ def get_index_url(nexus_pypi_hosted_repo_url, username, password):
...
+    """
+    Get the absolute paths for a list of files in a package.
+
+    :param Path path: the root path for the package containing the files in relative_paths
+    :param list relative_paths: paths to be converted to absolute paths
+    :return: absolute paths for the items in relative_paths
+    :rtype: list
+    """
+    # account for empty lists
+    if relative_paths is not None:
```
chmeliik, line 2062: I don't think we need to check for None here, this function is only called if the paths are not None

cachito/workers/tasks/pip.py
```diff
@@ -17,3 +35,61 @@ def cleanup_pip_request(request_id):
...
+    ca_cert = nexus.get_ca_cert()
+    if ca_cert:
+        env_vars["PIP_CERT"] = {"value": ca_cert or "", "kind": "literal"}
```
chmeliik, line 87: I think we don't need the `or ""`, the ca_cert already has a truthy value here

tests/test_workers/test_pkg_managers/test_pip.py
```diff
@@ -3236,3 +3265,153 @@ def test_get_index_url_invalid_url():
...
+    expected_dependency = {"name": name, "version": version, "type": "pip", "dev": dev}
+    dependency = pip._push_downloaded_requirement(req, pip_repo_name, raw_repo_name)
+    mock_upload.assert_called_once_with(pip_repo_name, path)
+    assert dependency == expected_dependency
+
+
+@pytest.mark.parametrize("dev", [True, False])
+@pytest.mark.parametrize("kind", ["url", "vcs"])
+@mock.patch("cachito.workers.pkg_managers.pip.upload_raw_package")
+def test_push_downloaded_requirement_vcs(mock_upload, dev, kind):
```
chmeliik, line 3304:
```suggestion
def test_push_downloaded_requirement_vcs_or_url(mock_upload, dev, kind):
```

chmeliik (approved):
LGTM :+1:

Left a few nitpicks, but that's all they are (not blockers for me)

chmeliik (approved):

cachito/workers/tasks/pip.py
```diff
@@ -17,3 +35,61 @@ def cleanup_pip_request(request_id):
...
+    username = get_hosted_repositories_username(request_id)
+    password = finalize_nexus_for_pip_request(pip_repo_name, raw_repo_name, username)
+
+    # Set environment variables
+    raw_url = get_pypi_hosted_repo_url(request_id)
+    pip_index_url = get_index_url(raw_url, username, password)
+    env_vars = {"PIP_INDEX_URL": {"value": pip_index_url, "kind": "literal"}}
+    ca_cert = nexus.get_ca_cert()
+    if ca_cert:
+        env_vars["PIP_CERT"] = {"value": ca_cert, "kind": "literal"}
```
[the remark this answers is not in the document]
athos-ribeiro: @chmeliik I was thinking about it earler. Sounds good. I will generate the file with the requirement file as well

PR #264, opened 2020-09-08
Add pip migration

nirzari:


cachito/web/models.py
```diff
@@ -231,6 +232,39 @@ def to_purl(self):
                     f"Unknown protocol in npm package version: {self.version}"
                 )
 
+        elif self.type == "pypi":
```
athos-ribeiro, line 235: Do we need this as part of this PR? This seems part of the ICM work @chmeliik is working on.

nirzari: I have marked this as WIP since I wanted to discuss couple of things regarding this today. I'll bring this up after the mtg.

chmeliik (approved):
LGTM :+1:

As @lcarva noted, running the downgrade when there are already pip packages in the DB would likely break because of FOREIGN KEY constraints, but the npm migration also does not handle this, so that is okay for now.

PR #265, opened 2020-09-09
Generate content manifest content for Pip packages

chmeliik:


chmeliik:
PTAL @ the format of PURLs for pypi, vcs and url deps. If you agree on the format, it would be good to get the 1st commit merged before the next deployment, otherwise we would not be able to generate content manifests for requests handled before this change is deployed.

cachito/web/models.py
```diff
@@ -231,6 +232,26 @@ def to_purl(self):
...
 
+        elif self.type == "pip":
+            # As per the purl spec, PyPI names should be normalized by lowercasing and
+            # converting '_' to '-'. The safe_name() function does the latter but not the
+            # former. It is not necessary to escape characters in the name, safe_name()
+            # also replaces everything except alphanumeric chars and '.' with '-'.
+            name = pkg_resources.safe_name(self.name.lower())
+            version = self.version
+
+            if version.startswith("vcs:"):
```
nirzari, line 243: Could you please give an example here for a sample `version` (just out of curiosity)?

nirzari (changes requested):

cachito/workers/pkg_managers/pip.py
```diff
@@ -1759,15 +1759,29 @@ def _download_url_package(requirement, pip_deps_dir, pip_raw_repo_name, nexus_au
...
         "path": download_path,
         "original_url": requirement.url,
+        "url_with_hash": url_with_hash,
         "raw_component_name": raw_component_name,
         "have_raw_component": have_raw_component,
     }
 
 
+def _add_cachito_hash_to_url(parsed_url, hash_spec):
+    """Add the #cachito_hash fragment to URL."""
```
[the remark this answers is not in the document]
chmeliik: Good point, will do

cachito/web/models.py
```diff
@@ -231,6 +232,26 @@ def to_purl(self):
...
+                quoted_url = urllib.parse.quote(version[len("vcs:") :], safe="")
+                return f"pkg:generic/{name}?vcs_url={quoted_url}"
```
chmeliik, line 245: In the design doc, the format for Git dependencies is defined as `pkg:/github/...`.

However, for npm, we use `pkg:generic` with `vcs_url` if the dependency uses `git+<protocol>://`, see here: https://github.com/release-engineering/cachito/blob/28df0cfc3b6c0d5f09a1792fc3fa82ca7945d9d2/cachito/web/models.py#L223-L225

For Pip, all Git dependencies will use `git+<protocol>://`, so I figured I would keep this consistent with npm, WDYT?

cachito/workers/pkg_managers/pip.py
```diff
@@ -1978,12 +1992,11 @@ def _push_downloaded_requirement(requirement, pip_repo_name, raw_repo_name):
...
-            version = (
-                f"vcs:{requirement['host']}/{requirement['namespace']}/{requirement['repo']}@"
-                f"{requirement['ref']}"
-            )
+            # Version is "vcs:" followed by the value for the vcs_url purl qualifier
+            version = f"vcs:git+{requirement['url']}@{requirement['ref']}"
         else:
-            version = requirement["original_url"]
+            # Version is "url:" followed by the value for the download_url purl qualifier
+            version = f"url:{requirement['url_with_hash']}"
```
chmeliik, line 1999: Am I right in thinking that it does not matter what we save in the database as `version` as long as we can properly construct purls from it?

chmeliik (commented):
Got two more questions of my own, PTAL @athos-ribeiro @nirzari @lcarva

cachito/workers/pkg_managers/pip.py
athos-ribeiro: Yes!
Note that those versions will be available to the users. However, a user should only care for the ICM data when checking for component versions.

chmeliik: Gotcha, thanks!

I might drop the prefixes anyway, there should be no ambiguity because a VCS dep is going to start with `git+<protocol>://`, a URL dep is going to start with `<protocol>://` and PyPI deps will have no scheme

athos-ribeiro: Sounds good to me

chmeliik:
Resolved my comments and added unit tests, one is still missing though (in test_api_v1)

chmeliik:
Rebased on master, added missing test, ready for review.

cachito/workers/pkg_managers/pip.py
nirzari: If VCS dep is going to start with `git+<protocol>://` as mentioned here , why have we mentioned the other vsc formats for `VCS_SCHEMES` in https://github.com/release-engineering/cachito/blob/0408601daca7af2e6a4e5f9a4f52ee60f5aa37df/cachito/workers/pkg_managers/pip.py#L921

chmeliik: @nirzari  Pip does support all those other schemes, see here: https://pip.pypa.io/en/stable/reference/pip_install/#vcs-support

I'd say it is reasonable to support all those for *parsing* requirements files. However, for *downloading* the requirements, we only support `git`. Any of the other VCS prefixes would cause the build to fail.

athos-ribeiro:
Tested locally. Works for:
[x] vcs dependencies
[x] url dependencies
[x] pypi dependencies
[x] differs build and runtime dependencies properly
[ ] main component's PURL.

The main component purl is being listed as a pypi dependency here. I should be a VCS one.

chmeliik:
> The main component purl is being listed as a pypi dependency here. I should be a VCS one.

Ooh, thanks for catching that, I'll have to figure out how to fix this :D

athos-ribeiro:
> Upon further investigation, I found that NPM purls are also generated incorrectly (the main component is `pkg:npm`).
> @athos-ribeiro @nirzari thoughts about merging this in broken (but consistent with npm) state and fixing both in a different PR?

I'd be OK with that.

athos-ribeiro (approved):

PR #267, opened 2020-09-10
Pass Nexus PyPI repository cert as config file

athos-ribeiro:
Passing the Nexus PyPI hosted repository certificate as an environment
variable may fail due to the length of the string being passed around.
Instead, we pass the certificate as a configuration file and set the
PIP_CERT environment variable to point to this new configuration file
containing the certificate.


chmeliik (approved):
LGTM :+1:

PR #268, opened 2020-09-10
Change how Nexus PyPI proxy is configured.

chmeliik:
There is no reason to require a full url, the repo name is more than
enough.


chmeliik:
While working on documentation, I realized I have no way to justify why the proxy repo must be a full URL when the raw repo is just a repo name. Originally, this was to be consistent with the npm proxy value, but that one serves a different purpose.

nirzari (approved):

athos-ribeiro (approved):
LGTM

PR #269, opened 2020-09-10
Generate custom pip requirement files

athos-ribeiro:
When a pip request generates content in the raw hosted Nexus repository,
generate custom requirement files to substitute the ones pointing to the
external non-PyPI dependencies. The new files will be added to the
request as configuration files.

### DIFF
--- a/cachito/workers/tasks/pip.py
+++ b/cachito/workers/tasks/pip.py
@@ -1,4 +1,5 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
+from pathlib import Path
 import base64
 import logging
 import os
@@ -12,10 +13,12 @@ from cachito.workers.pkg_managers.general import (
     update_request_with_package,
 )
 from cachito.workers.pkg_managers.pip import (
+    PipRequirementsFile,
     finalize_nexus_for_pip_request,
     get_index_url,
     get_pypi_hosted_repo_name,
     get_pypi_hosted_repo_url,
+    get_raw_component_name,
     get_raw_hosted_repo_name,
     get_hosted_repositories_username,
     prepare_nexus_for_pip_request,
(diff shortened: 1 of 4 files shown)


nirzari (approved):
I haven't run it but semantically and syntactically makes sense to me. . LGTM!

cachito/workers/tasks/pip.py
```diff
@@ -108,3 +119,55 @@ def fetch_pip_source(request_id, package_configs=None):
...
+    :rtype: dict
+    """
+    original_requirement_file = PipRequirementsFile(requirement_file_path)
+    cachito_requirements = []
+    differs_from_original = False
+    for requirement in original_requirement_file.requirements:
+        raw_component_name = get_raw_component_name(requirement)
+        if raw_component_name:
+            # increase max_attempts to make sure the package upload/setup is complete
+            new_url = nexus.get_raw_component_asset_url(
```
chmeliik, line 145: I'm wondering if we need to include some of the parts from the original URL, in case it contains, for example, the `#egg` fragment. We may need to add the `fragment` part, and possibly also the `query` part of the original url. Not sure about the `params` part (I don't know what those are)

chmeliik: Nevermind, the `copy()` method already takes care of  this.

cachito/workers/tasks/pip.py
```diff
@@ -108,3 +119,55 @@ def fetch_pip_source(request_id, package_configs=None):
...
+    )
+    final_contents = []
+    if cachito_requirement_file.options:
+        final_contents.append(" ".join(cachito_requirement_file.options))
+
+    final_contents.extend(
+        [str(requirement) for requirement in cachito_requirement_file.requirements]
+    )
+    req_str = "\n".join(final_contents)
+    final_path = Path("app") / Path(requirement_file_path).relative_to(source_dir)
```
chmeliik, line 168: Is `source_dir` here the root of the repo? If the user specifies a subpath, wouldn't this be incorrect? I'm trying this with a repo where everything is under `subpath/`, and the config file I get has `"path": "app/requirements.txt"` when it should be `app/subpath/requirements.txt`

athos-ribeiro: source_dir here was the root of the package. I changed it to be the root of the repository so the `final_path` will be correct.

athos-ribeiro:
rebased on master and fixed an issue rebasing introduced.

chmeliik (approved):
LGTM :+1:

PR #271, opened 2020-09-14
Handle multiple pip packages

athos-ribeiro:


chmeliik (approved):
LGTM :+1:

If we run into issues where the `get_component_info_from_nexus()` fails because of a race condition, we could look into keeping track of uploaded components instead of checking this way, but this approach (with retries) should be good enough for now.

PR #272, opened 2020-09-14
Fix function for sorting ICMs

chmeliik:
In some cases, the function could fail when sorting lists of dicts.
List of dicts are sorted by the `items()` of those dicts, which may
result in comparisons such as this one:

```python
("dependencies", list_of_dicts) < ("dependencies", list_of_dicts)

TypeError: '<' not supported between instances of
'collections.OrderedDict' and 'collections.OrderedDict'
```

Instead of sorting lists of dicts by `items()`, sort by `"purl"`.

### DIFF
--- a/cachito/web/utils.py
+++ b/cachito/web/utils.py
@@ -10,18 +10,23 @@ def deep_sort_icm(orig_item):
 
     The function for sorting image content manifests
 
+    All dicts in the content manifest will be turned into OrderedDicts with alphabetically
+    sorted keys. All lists of dicts with a "purl" key will be sorted alphabetically by the
+    "purl" value. Any other objects will be left as is.
+
     :param orig_item: Original content manifest to be sorted
     :return: Recursively sorted dict or list according to orig_item
     :rtype: Any
     """
     if isinstance(orig_item, list):
         sorted_item = [deep_sort_icm(item) for item in orig_item]
-        if len(sorted_item) and isinstance(sorted_item[0], OrderedDict):
-            sorted_item = sorted(sorted_item, key=lambda ordered_dict: list(ordered_dict.items()))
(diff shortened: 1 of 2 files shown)


cachito/web/utils.py
```diff
@@ -10,18 +10,23 @@ def deep_sort_icm(orig_item):
...
+        # If item is a list of dicts with the "purl" key, sort by the "purl" value
+        if sorted_item and isinstance(sorted_item[0], OrderedDict) and "purl" in sorted_item[0]:
+            sorted_item = sorted(sorted_item, key=lambda item: item["purl"])
     elif isinstance(orig_item, dict):
-        sorted_item = OrderedDict(
-            sorted({k: deep_sort_icm(v) for k, v in orig_item.items()}.items())
-        )
+        items = ((k, deep_sort_icm(v)) for k, v in orig_item.items())
+        # Sort items by key
+        sorted_item = OrderedDict(sorted(items, key=lambda keyval: keyval[0]))
```
akhmanova, line 29: So we will use only `key` part to sort OrderedDicts?

chmeliik: I think this does not make any real difference. When you sort a dict by `items()`, you are sorting this structure:
```python
[
    (key1, val1),
    (key2, val2),
    ...,
    (keyN, valN),
]
```
But keep in mind that all keys in a dict are unique, so sorting by `items()` is effectively the same as sorting by `keys()`

I only made this change to make it more obvious that we are sorting by keys.

akhmanova (approved):
LGTM :)

athos-ribeiro (approved):
LGTM

Tested with the new python PR for handling multiple packages and works fine with it.

Thanks for the fix :+1:

PR #273, opened 2020-09-14
Document Cachito support for Pip

chmeliik:
* CLOUDBLD-2521


chmeliik:
6th commit has some info copied from the `npm` section which may or may not also apply to `pip`. Other than that, this should be mostly done (though I expect I must have forgotten something)

docs/pip.md
```diff
@@ -0,0 +1,163 @@
...
+idna==2.10                # via requests
+https://github.com/containerbuildsystem/operator-manifest/archive/v0.0.3.zip  # via -r requirements.in
+requests==2.24.0          # via -r requirements.in
+ruamel.yaml.clib==0.2.2   # via ruamel.yaml
+ruamel.yaml==0.16.12      # via operator-manifest
+six==1.15.0               # via dockerfile-parse
+urllib3==1.25.10          # via requests
+```
+
+As you can see, `pip-compile` gathered all the recursive dependencies and pinned al PyPI packages.
```
nirzari, line 79:
```suggestion
As you can see, `pip-compile` gathered all the recursive dependencies and pinned all PyPI packages.
```

docs/pip.md
```diff
@@ -0,0 +1,163 @@
...
+It did not pin external dependencies, the mechanism for doing so is specific to Cachito. You can pin
+these beforehand in the `requirements.in` file, but if any of the recursive dependencies are
+external, you may need to edit the generated file anyway.
+
+### Explicit package names
+
+Cachito needs to know the package name of all your dependencies. For PyPI dependencies, this is
```
nirzari, line 86:
```suggestion
Cachito needs to know the package name for all of your dependencies. For PyPI dependencies, this is
```

docs/pip.md
```diff
@@ -0,0 +1,163 @@
...
+trivial, as the name is already present in the requirements file. For external dependencies,
+Resolving the name may require executing the `setup.py` file. Cachito does have a mechanism for
```
nirzari, line 88:
```suggestion
resolving the name may require executing the `setup.py` file. Cachito does have a mechanism for
```

docs/pip.md
```diff
@@ -0,0 +1,163 @@
...
+extracting package metadata from `setup.py`, but it is very limited. That is why, for external
+dependencies, you will need to explicitly specify package names using one of the mechanisms that
+pip supports.
+
+1. use '@': `dockerfile-parse @ git+https://github.com/containerbuildsystem/dockerfile-parse`
```
nirzari, line 93: Is there supposed to be a space before and after `@` ? (or that is done for readability?)

docs/pip.md
```diff
@@ -0,0 +1,163 @@
...
+2. use "#egg": `git+https://github.com/containerbuildsystem/dockerfile-parse#egg=dockerfile-parse`
+
+Note that `pip-compile` seems to ignore the `@` mechanism, so using `#egg` may be preferrable.
```
nirzari, line 96: In that case, why are we not enforcing users to use `#egg` ?

docs/pip.md
```diff
@@ -0,0 +1,163 @@
...
+dependencies. If that is not desirable, use the Cachito-specific `#cachito_hash` URL fragment as
+mentioned in the "Pinning versions" section.
+
+### Build dependencies
+
+Setuptools provides a way to specify build dependencies via the [setup_requires][setuptools-docs]
+keyword argument. It is deprecated in favor of the [PEP-518][pep-518] approach but, for similar
+reasons as mentioned in the sections above, Cachito supports neither. If you have any build-only
+dependencies, you will need to put them in a `requirements-build.txt` file which follows the same
+rules as `requirements.txt`.
```
nirzari, line 115: (Note to self): Update this section when we achieve the stretch goal of _Determine build dependencies for user runtime dependencies_

docs/pip.md
```diff
@@ -0,0 +1,163 @@
...
+
+## `setup.py`, `setup.cfg`
+
+Pip packages define their metadata in one of two files, either [setup.py][setup-py-docs] or
```
nirzari, line 119:
```suggestion
Pip packages define their metadata in one of the two files, either [setup.py][setup-py-docs] or
```

README.md
```diff
@@ -543,3 +572,45 @@ Nexus instead. The modified files will be accessible at the
...
+a Git repository or an HTTP(S) URL. After downloading such a dependency, Cachito will upload it to
+the Nexus instance used for hosting permanent content. If Cachito encounters this same dependency
+again in a future request, it will use it directly from Nexus rather than downloading it and
+uploading it again. This guarantees that any dependency used for a Cachito request can be used again
+in a future Cachito request.
+
+Please note that if a dependency is from a private repository, set the
+[.netrc](https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html) and
+`known_hosts` files for the Cachito workers. If the dependency location is not supported, Cachito
+will fail the request. (TODO: does this apply?)
```
nirzari, line 611: AFAIK this applies to npm only?

README.md
```diff
@@ -543,3 +572,45 @@ Nexus instead. The modified files will be accessible at the
...
+`/api/v1/requests/<id>/configuration-files` endpoint.
+
+Cachito will produce a bundle that is downloadable at `/api/v1/requests/<id>/download`. This
+bundle will contain the application source code in the `app` directory and individual source
+archives of all the dependencies in the `deps/pip` directory. These archives are not meant to be
+used to build the application. They are there for convenience so that the dependency sources can be
+published alongside your application sources.
+
+In addition, they can be used to populate a local PyPI repository in the event that the application
+needs to be built without Cachito and the Nexus instance it manages. (TODO: does this apply?)
```
nirzari, line 599: I don't know how to test this but IIUC this seems doable even for pip.

nirzari (changes requested):

README.md
```diff
@@ -478,8 +507,8 @@ Offline Installations   | ✓     | x
   For example, `npm` allows `file` dependencies with the `cachito_npm_file_deps_allowlist`
   configuration. `gomod` allows this through the `go.mod` replace directive.
 * **Offline Installations** - The dependencies can be installed solely with the contents of the
-  bundle. This is true for the `gomod` package manager, however, the `npm` package manager relies
-  on Nexus to be online and properly configured by Cachito.
+  bundle. This is true for the `gomod` package manager, however, the `npm` and `pip` package
+  managers rely on Nexus to be online and properly configured by Cachito.
```
nirzari, line 511: (Optional) Should we give an example for pip for `External Dependency` on https://github.com/release-engineering/cachito/pull/273/files#diff-04c6e90faac2675aa89e2176d2eec7d8R501?

docs/pip.md
```diff
@@ -0,0 +1,163 @@
...
+  }
+}
+```
+
+[setuptools-docs]: https://setuptools.readthedocs.io/en/latest/setuptools.html
+[pep-518]: https://www.python.org/dev/peps/pep-0518/
+[setup-py-docs]: https://setuptools.readthedocs.io/en/latest/setuptools.html#basic-use
+[setup-cfg-docs]: https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files
+[cachito-pip-py]: ../cachito/workers/pkg_managers/pip.py
+>>>>>>> bc46a14... Document additional details of pip support
```
nirzari, line 163: Is this kept here intentionally?

docs/pip.md
```diff
@@ -0,0 +1,163 @@
...
+
+This document describes some of the more intricate details of Cachito support for pip. For a high
+level overview, look [here in the README](../README.md#pip).
+
+Cachito has a number of specific requirements when it comes pip packages. Some of those stem from
+the general ideas behind Cachito (e.g. reproducibility), some from the technical challenges of
+supporting a packaging system which defines most metadata through a Python executable. Read on for
+more details.
+
+## `requirements.txt`
```
nirzari, line 11: (optional): I do understand the reason behind having quotes here to make them codable but when you look at the README, it just looks bit off appearance wise. I'd rather just have it as regular title or put them in normal non-codable quotes.

docs/pip.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Build dependencies
+
+Setuptools provides a way to specify build dependencies via the [setup_requires][setuptools-docs]
+keyword argument. It is deprecated in favor of the [PEP-518][pep-518] approach but, for similar
+reasons as mentioned in the sections above, Cachito supports neither. If you have any build-only
+dependencies, you will need to put them in a `requirements-build.txt` file which follows the same
+rules as `requirements.txt`.
+
+## `setup.py`, `setup.cfg`
```
nirzari, line 117: (optional): I do understand the reason behind having quotes here to make them codable but when you look at the README, it just looks bit off appearance wise. I'd rather just have it as regular title or put them in normal non-codable quotes.

docs/pip.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Recursive dependencies
+
+You will need to include not only the direct dependencies for your application, but all the
+dependencies of those dependencies (recursively) as well. There are two reasons for this. The first
+one is reproducibility (you will need to pin the versions of recursive dependencies as well). The
+second reason is that, to determine the dependencies of your dependencies, Cachito would have to
+execute their `setup.py` files, which is not desirable.
+
+### `pip-compile`
```
nirzari, line 49: (optional): I do understand the reason behind having quotes here to make them codable but when you look at the README, it just looks bit off appearance wise. I'd rather just have it as regular title or put them in normal non-codable quotes.

docs/pip.md
chmeliik: Oops, must have missed that :D

docs/pip.md
chmeliik: Yeah it looks better as a regular title, thanks!

docs/pip.md
chmeliik: AFAIK the space is valid (but not required) in a requirements.txt file, so since both options are valid I chose this one 'cause it looks nicer

docs/pip.md
chmeliik: From our point of view both options are equally valid, using `#egg` may make the user's life easier if they use `pip-compile` though

README.md
chmeliik: It does seem doable for PyPI dependencies, but I'm not sure about external dependencies, it may be difficult to upload those to a PyPI repository.

I'm pretty sure there is a way to install the packages directly from the file system using pip, maybe that would be more relevant if I figure out how to do that.

chmeliik: Updated, I mentioned the option to install from filesystem instead

README.md
chmeliik: IMO that is not necessary, I'd say the npm example is generic enough to also apply to pip (and potentially other package managers in the future)

README.md
```diff
@@ -478,8 +507,8 @@ Offline Installations   | ✓     | x
...
 
 ### gomod
```
arewm, line 513: do you ever provide information on the offline installation feature for the gomod package manager? I assume that this is achieved with the `gomod-vendor` flag (i.e. replacing the vendor directory)... but I have not seen it yet. Since this is unique to golang, I would be interested to know a little background behind _why_ it is.

README.md
```diff
@@ -543,3 +572,44 @@ Nexus instead. The modified files will be accessible at the
...
+corresponding URLs from the raw repository. The modified requirements files are accessible at the
+`/api/v1/requests/<id>/configuration-files` endpoint.
+
+Cachito will produce a bundle that is downloadable at `/api/v1/requests/<id>/download`. This
+bundle will contain the application source code in the `app` directory and individual source
+archives of all the dependencies in the `deps/pip` directory. These archives are not meant to be
+used to build the application. They are there for convenience so that the dependency sources can be
+published alongside your application sources. In addition, they can be used to to install packages
+directly from the filesystem with `pip install --no-index --no-deps <path/to/archive>` in the event
+that the application needs to be built without Cachito and the Nexus instance it manages.
```
arewm, line 598: Is this different from python being able to install in an offline mode?

arewm (commented):
I have only looked at the README, but I wanted to submit these comments before I left so that they didn't get lost. :)

README.md
chmeliik: Hmm, true, this is a way to install in offline mode (but, when building with OSBS, it would require changes in the Dockerfile). Then again, for npm we have this, which would also suggest the possibility:
> In addition, they can be used to populate a local npm
registry in the event that the application needs to be built without Cachito and the Nexus instance
it manages.

@athos-ribeiro @lcarva what exactly do we mean by "Offline Installations"?

arewm: if the `gomod-vendor` flag is the golang's version of offline installation, then that still requires a change as well... just not to the Dockerfile.

lcarva: Offline installation means that you can take the tarball created by Cachito and use those dependencies without any network connection at all. This is true for gomod. It's not for NPM because network access to the Nexus repo is required. 

In either case, the Cachito tarball always includes all sources. This is to fulfill source compliance requirements. It could also be used to perform offline operations, but that would require changes to the pkg manager commands to use them directly.

chmeliik: Would it be fair to say that offline installations for gomod "just work" (i.e. the user does not have to change anything) while npm and pip would require significant changes in the user's workflow to install offline? Therefore we say that only gomod supports offline installations?

README.md
chmeliik: Removed this, if it ends up being  relevant we can always copy it from the npm section

chmeliik:
Removed a potentially irrelevant paragraph, updated worker configuration section to reflect that we configure the pip proxy using a full URL rather than just a repo name

README.md
chmeliik: @arewm I updated the Offline Installations section to reflect what I've learned here, let me know if it's more clear now (thanks for the review BTW)

docs/pip.md
```diff
@@ -0,0 +1,162 @@
...
+* for Git dependencies, specify the commit hash:
+  * `git+https://github.com/containerbuildsystem/dockerfile-parse@ed38dc2eadf6890b524340f45a7d172d02196e61`
+* for HTTP(S) dependencies, include the hash of the source archive (truncated for readability):
+  * `https://github.com/containerbuildsystem/operator-manifest/archive/v0.0.3.zip#cachito_hash=sha256:7774db0`
+
+### Recursive dependencies
+
+You will need to include not only the direct dependencies for your application, but all the
+dependencies of those dependencies (recursively) as well. There are two reasons for this. The first
+one is reproducibility (you will need to pin the versions of recursive dependencies as well). The
```
arewm, line 45: I do not think that this parenthetical phrase is needed here
> (you will need to pin the versions of recursive dependencies as well)

docs/pip.md
```diff
@@ -0,0 +1,162 @@
...
+execute their `setup.py` files, which is not desirable.
+
+### pip-compile
+
+Good news, there is a tool which does part of the required setup for you. That tool is `pip-compile`
+from the [pip-tools](https://github.com/jazzband/pip-tools) Python package. Below is an example of
+how you would use it.
+
+1. rename your `requirements.txt` to `requirements.in` (convention)
+2. run `pip-compile requirements.in -o requirements.txt`
```
arewm, line 56: I think that this can be simplified a little bit.. maybe something like

### Specifying dependencies

In addition to specifying direct dependencies, recursive dependencies also need to be explicitly defined for two reasons:

1. Further enable reproducibility by explicitly specifying _every_ needed package
2. Prevent the need for remote execution of `setup.py`

#### pip-compile

While this might be onerous to manually maintain, `pip-compile` from [pip-tools](https://github.com/jazzband/pip-tools) can be used to automate this process for you using the following procedure.

1. rename `requirements.txt` to `requirements.in` (by convention)
2. run `pip-compile requirements.in -o requirements.txt`

arewm: If you want to put more information into the first 1/2, you can. I _assume_ that most people understand how remote execution of code is generally not desired, but if that is not  widely known, feel free to add more there.

docs/pip.md
```diff
@@ -0,0 +1,162 @@
...
+trivial, as the name is already present in the requirements file. For external dependencies,
+resolving the name may require executing the `setup.py` file. Cachito does have a mechanism for
+extracting package metadata from `setup.py`, but it is very limited. That is why, for external
+dependencies, you will need to explicitly specify package names using one of the mechanisms that
+pip supports.
+
+1. use '@': `dockerfile-parse @ git+https://github.com/containerbuildsystem/dockerfile-parse`
+2. use "#egg": `git+https://github.com/containerbuildsystem/dockerfile-parse#egg=dockerfile-parse`
+
+Note that `pip-compile` seems to ignore the `@` mechanism, so using `#egg` may be preferrable.
```
arewm, line 96: It was not immediately clear to me that you are trying to illustrate two ways to specify the `dockerfile-parse` package name. Can you either present these two examples as two ways to achieve this for a specific package or generalize it with something like `<package-name>`? Maybe don't use inline formatting but use code blocks?

arewm: Based on this, the examples above in the pinning versions section is incomplete. Can you include a complete example of a `requrements.txt` file that includes all of these referenced parts?

docs/pip.md
```diff
@@ -0,0 +1,162 @@
...
+### Hash checking
+
+In general, Cachito handles hash checking the same way that pip does. If `--require-hashes` is
+present in the requirements file, or if any dependency uses the `--hash` option, Cachito will
+require that all dependencies specify a hash and will check that the hashes are valid.
+
+For HTTP(S) dependencies, Cachito will always require a hash and will always validate it. You can
+provide it using `--hash`, but as mentioned above, that will turn on hash checking for *all* your
+dependencies. If that is not desirable, use the Cachito-specific `#cachito_hash` URL fragment as
+mentioned in the "Pinning versions" section.
```
arewm, line 107: You should link to any anchors that are available to make navigation easier. This was not mentioned so much as it was shown... 

> If that is not desirable, use the Cachito-specific `#cachito_hash` URL fragment as shown in the HTTP(S) dependencies example in the [Pinning versions](#pinning-versions) section.

docs/pip.md
```diff
@@ -0,0 +1,162 @@
...
+Setuptools provides a way to specify build dependencies via the [setup_requires][setuptools-docs]
+keyword argument. It is deprecated in favor of the [PEP-518][pep-518] approach but, for similar
+reasons as mentioned in the sections above, Cachito supports neither. If you have any build-only
+dependencies, you will need to put them in a `requirements-build.txt` file which follows the same
+rules as `requirements.txt`.
+
+## setup.py, setup.cfg
+
+Pip packages define their metadata in one of the two files, either [setup.py][setup-py-docs] or
+[setup.cfg][setup-cfg-docs] (or a combination of the two). Cachito will scan both of these files
```
arewm, line 120: If a combination of the two is valid, then the metadata is not defined in _one_ of the two.

> Pip packages can define their metadata in two files -- [setup.py][setup-py-docs] and [setup.cfg][setup-cfg-docs]

README.md
arewm: Thanks, looks good.

README.md
chmeliik: resolved in the other thread

docs/pip.md
chmeliik: Thanks, I used your version :-)

docs/pip.md
chmeliik: Fixed and fixed (and I also used code blocks for the examples in `Pinning versions`). I included an additional example of `requirements.in` rather than the full `requirements.txt`, I hope that is enough

docs/pip.md
chmeliik: You're right, fixed

docs/pip.md
```diff
@@ -0,0 +1,195 @@
...
+git+https://github.com/namespace/repo#egg=<package-name>
+```
+
+Similarly to the procedure used for pinning external dependency versions, you can specify explicit
+package names in `requirements.in` to avoid having to edit the file generated by `pip-compile`.
+However, `pip-compile` seems to ignore the `@` mechanism, so using `#egg` may be preferrable.
+
+After pinning versions and specifying package names for external dependencies, the `requirements.in`
+file at the [top of this section](#requirementstxt) would look like this (commit ref and hash value
+truncated for readability):
```
arewm, line 123: Can the hashes be truncated in Cachito? If not, I would argue that you should do something like one of these

* Put the full digest here and take the hit to readability
* Replace it with a placeholder like `<insert-full-sha-digest>`
* Make it _explicit_ i the paragraph that the full sha is needed and that the sample is not considered valid because the full sha is not used.

arewm: and I guess something might also need to be clarified when you use the same standard above...

chmeliik: Fair point, I replaced the digest and commit ref with placeholders (they must both be full length)

chmeliik:
Replaced truncated ref and digest examples with placeholders, documented additional findings about `pip-compile` and Build dependencies

arewm (approved):
LGTM, thanks!

README.md
```diff
@@ -564,3 +596,40 @@ Nexus instead. The modified files will be accessible at the
 again in a future request, it will use it directly from Nexus rather than downloading it and
 uploading it again. This guarantees that any dependency used for a Cachito request can be used again
 in a future Cachito request.
+
+### pip
+
+The pip package manager works by parsing the `requirements.txt` and `requirements-build.txt` files
```
athos-ribeiro, line 602: Should we say that it can also parse additional (custom) requirement files upo request?

README.md
```diff
@@ -564,3 +596,40 @@ Nexus instead. The modified files will be accessible at the
...
+as the files use the expected format.
+
+Cachito then creates two repositories in an instance of Nexus it manages that contain just the
+dependencies discovered in the requirements files. PyPI dependencies are uploaded to a PyPI hosted
+repository, external dependencies are uploaded to a raw repository. Connection information for the
+hosted repository is provided as the `PIP_INDEX_URL` environment variable accessible at the
+`/api/v1/requests/<id>/environment-variables` endpoint. To make external dependencies available,
+Cachito modifies the requirements files for the request by replacing relevant entries with their
+corresponding URLs from the raw repository. The modified requirements files are accessible at the
+`/api/v1/requests/<id>/configuration-files` endpoint.
```
athos-ribeiro, line 614: It would be nice to have a note/warning saying that credentials for the temporary repositories will be exposed here. We could also mention this is not an issue, since the reason for them not being open for anonymous access is due to nexus technical limitations.

chmeliik: Good idea, added

athos-ribeiro (approved):

athos-ribeiro (approved):

PR #274, opened 2020-09-15
Revert "Change how Nexus PyPI proxy is configured."

chmeliik:
This reverts commit b381abce93025094dad59dbc84b5830a97b8e73a.

There *is* a reason to use the full URL after all, which is that the
PyPI proxy may be hosted at a different URL than the other repositories
in the Nexus hoster instance.

### DIFF
--- a/cachito/workers/config.py
+++ b/cachito/workers/config.py
@@ -40,8 +40,8 @@ class Config(object):
     cachito_nexus_hoster_username = None
     cachito_nexus_js_hosted_repo_name = "cachito-js-hosted"
     cachito_nexus_npm_proxy_repo_url = "http://localhost:8081/repository/cachito-js/"
-    cachito_nexus_pip_proxy_repo_name = "cachito-pip-proxy"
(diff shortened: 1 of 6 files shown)


chmeliik:
> PyPI proxy may be hosted at a different URL than the other repositories
> in the Nexus hoster instance.

This sounds like something that should not be the case, or be temporary at best.
@athos-ribeiro @nirzari @lcarva what do you think?

athos-ribeiro:
> > PyPI proxy may be hosted at a different URL than the other repositories
> > in the Nexus hoster instance.
> 
> This sounds like something that should not be the case, or be temporary at best.
> @athos-ribeiro @nirzari @lcarva what do you think?

Agreed. We need to figure out why there are differences in the internal instances we have been using in production before moving forward here, unfortunately :(

athos-ribeiro (approved):

chmeliik:
Filed a story to follow up on the inconsistent URLs. For now this is a good enough workaround.

PR #278, opened 2020-09-17
Get non-PyPI component URLs from self-hosted Nexus

athos-ribeiro:


chmeliik (approved):
LGTM :+1:

PR #280, opened 2020-09-18
Fix host directory permissions for docker-compose

chmeliik:
By changing the nexus volume from a named volume to a host directory, we prevented the Nexus container from writing in its own data directory.


chmeliik:
I wasn't able to use the local environment with docker after the podman-compose fix, this should hopefully work with both docker and podman (please test)

nirzari (approved):

athos-ribeiro (approved):

akhmanova (approved):

PR #281, opened 2020-09-22
Fallback to cloning repository

athos-ribeiro:
This is the first PR for the series of changes on handling corrupted archives.

When Cachito does not succeed on re-using an existing source archive to
retrieve the project's git history, try using other existing archives.
Additionally, if no archives can be used, fall back to cloning the
repository as it is done when an archive does not exist.

### DIFF
--- a/cachito/workers/scm.py
+++ b/cachito/workers/scm.py
@@ -133,14 +133,21 @@ class Git(SCM):
         # The previous archive does not mean the one just before the request that
         # schedules current task. The only reason for finding out such a file is
         # to access the git history. So, anyone is ok.
-        previous_archive = max(
-            self.sources_dir.package_dir.glob("*.tar.gz"), key=os.path.getctime, default=None
+        previous_archives = sorted(
+            self.sources_dir.package_dir.glob("*.tar.gz"), key=os.path.getctime, reverse=True
         )
+        for previous_archive in previous_archives:
+            try:
+                self.update_and_archive(previous_archive)
+                return
+            except (
+                git.exc.InvalidGitRepositoryError,
+                tarfile.ExtractError,
+                OSError,  # raised by tarfile when an FS operation fails
+            ) as exc:
+                log.warning("Error handling archived artifact '%s': %s", previous_archive, exc)
 
-        if previous_archive:
-            self.update_and_archive(previous_archive)
-        else:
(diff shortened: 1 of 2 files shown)


nirzari (approved):
LGTM

tests/test_workers/test_scm.py
```diff
@@ -122,6 +123,42 @@ def test_fetch_source_by_pull(mock_update_and_archive, mock_getctime):
...
+    scm_git = scm.Git(url, ref)
+
+    po = mock.patch.object
+    with po(scm_git.sources_dir.archive_path, "exists", return_value=False):
+        with po(
+            scm_git.sources_dir.package_dir, "glob", return_value=["29eh2a.tar.gz", "a8c2d2.tar.gz"]
+        ):
+            scm_git.fetch_source()
+
+    mock_update_and_archive.call_count == 2
```
chmeliik, line 153: Is the `assert` keyword missing here?

athos-ribeiro: fixed

cachito/workers/scm.py
```diff
@@ -133,14 +133,21 @@ def fetch_source(self):
...
-        else:
-            self.clone_and_archive()
+        for previous_archive in previous_archives:
+            try:
+                self.update_and_archive(previous_archive)
+                return
+            except (
+                git.exc.InvalidGitRepositoryError,
+                tarfile.ExtractError,
+                EnvironmentError,
```
chmeliik, line 146: I was confused  about `EnvironmentError` for a second, turns out on python3.3 and later it is just an alias of `OSError`: https://docs.python.org/3.8/library/exceptions.html#EnvironmentError

Should we use `OSError` here instead? Would it also be possible to add a comment explaining when an `OSError` might occur?

athos-ribeiro: I checked the wrong reference when looking for exceptions to capture here. I checked https://github.com/python/cpython/blob/2.7/Lib/tarfile.py when I wanted to check https://github.com/python/cpython/blob/3.7/Lib/tarfile.py, where the EnvironmentError occurrences were substituted for ether OSError or ExtractError ones.

I will change it to `OSError`

chmeliik (approved):
LGTM :+1:

PR #283, opened 2020-09-23
Add git-submodule migration

nirzari:
Running the downgrade when there are already git-submodule packages in the DB would likely break because of FOREIGN KEY constraints, but the npm and pip migration also do not handle this, so that is okay for now.


chmeliik (approved):
LGTM :+1:

athos-ribeiro (approved):

PR #287, opened 2020-09-29
Clone submodules to make them part of the archive

nirzari:
If `git-submodule` is passed in the pkg_managers list for the Cachito request, the git submodules within the requested repo will be cloned into it's respective location

### DIFF
--- a/cachito/workers/scm.py
+++ b/cachito/workers/scm.py
@@ -111,10 +111,11 @@ class Git(SCM):
             os.unlink(self.sources_dir.archive_path)
             raise
 
-    def clone_and_archive(self):
+    def clone_and_archive(self, gitsubmodule=False):
         """
         Clone the git repository and create the compressed source archive.
 
+        :param bool gitsubmodule: a bool to determine whether git submodules need to be processed.
         :raises CachitoError: if cloning the repository fails or if the archive can't be created
         """
         with tempfile.TemporaryDirectory(prefix="cachito-") as temp_dir:
@@ -133,13 +134,18 @@ class Git(SCM):
                 raise CachitoError("Cloning the Git repository failed")
 
             self._reset_git_head(repo)
+
+            if gitsubmodule:
+                self.update_git_submodules(repo)
+
             self._create_archive(repo.working_dir)
 
-    def update_and_archive(self, previous_archive):
+    def update_and_archive(self, previous_archive, gitsubmodule=False):
         """
         Update the existing Git history and create a source archive.
 
         :param str previous_archive: path to an archive file created before.
+        :param bool gitsubmodule: a bool to determine whether git submodules need to be processed.
         :raises CachitoError: if pulling the Git history from the remote repo or
             the checkout of the target Git ref fails.
         """
@@ -157,10 +163,19 @@ class Git(SCM):
                 raise CachitoError("Failed to fetch from the remote Git repository")
 
             self._reset_git_head(repo)
+            if gitsubmodule:
+                self.update_git_submodules(repo)
+
             self._create_archive(repo.working_dir)
 
-    def fetch_source(self):
-        """Fetch the repo, create a compressed tar file, and put it in long-term storage."""
+    def fetch_source(self, gitsubmodule=False):
+        """Fetch the repo, create a compressed tar file, and put it in long-term storage.
+
+        :param bool gitsubmodule: a bool to determine whether git submodules need to be processed.
+        """
+        if gitsubmodule:
+            self.sources_dir = SourcesDir(self.repo_name, f"{self.ref}-with-submodules")
+
         # If it already exists and isn't corrupt, don't download it again
         archive_path = self.sources_dir.archive_path
         if archive_path.exists():
@@ -180,8 +195,13 @@ class Git(SCM):
             self.sources_dir.package_dir.glob("*.tar.gz"), key=os.path.getctime, reverse=True
         )
         for previous_archive in previous_archives:
+            # Excluding previous archives with submodules since there is
+            # no straight-forward way to get rid of the previously included
+            # submodules
+            if "-with-submodules" in str(previous_archive):
+                continue
             try:
-                self.update_and_archive(previous_archive)
+                self.update_and_archive(previous_archive, gitsubmodule=gitsubmodule)
                 return
             except (
                 git.exc.InvalidGitRepositoryError,
@@ -190,7 +210,24 @@ class Git(SCM):
             ) as exc:
                 log.warning("Error handling archived artifact '%s': %s", previous_archive, exc)
 
-        self.clone_and_archive()
+        self.clone_and_archive(gitsubmodule=gitsubmodule)
+
+    def update_git_submodules(self, repo):
+        """Update git submodules.
+
+        For the given repo, update submodules. For Cachito request for `retrodep` repo and
+        `go-github` submodule, the dir structure would look like,
+        retrodep/go-github/<content_of_go-github_repo>
+
+        :param git.Repo repo: the repository object.
+        :raises CachitoError: if updating the git submodules fail.
+        """
+        try:
+            log.debug(f"Git submodules for the requested repo are: {repo.submodules}")
+            repo.submodule_update(recursive=False)
+        except Exception as e:
+            log.exception("Updating the Git submodule(s) from %s failed %s", self.url, e)
+            raise CachitoError("Updating the Git submodule(s) failed")
 
     @property
     def repo_name(self):
--- a/tests/test_api_v1.py
+++ b/tests/test_api_v1.py
@@ -32,16 +32,17 @@ RE_INVALID_PACKAGES_VALUE = (
 
 
 @pytest.mark.parametrize(
-    "dependency_replacements, pkg_managers, user, expected_pkg_managers, flags",
+    "dependency_replacements, pkg_managers, user, expected_pkg_managers, flags, gitsubmodule",
     (
-        ([], [], None, [], None),
-        ([], ["gomod"], None, ["gomod"], None),
+        ([], [], None, [], None, False),
+        ([], ["gomod", "git-submodule"], None, ["gomod"], None, True),
         (
             [{"name": "github.com/pkg/errors", "type": "gomod", "version": "v0.8.1"}],
             ["gomod"],
             None,
             ["gomod"],
             None,
+            False,
         ),
         (
             [
@@ -52,14 +53,15 @@ RE_INVALID_PACKAGES_VALUE = (
                     "version": "v0.8.1",
                 }
             ],
-            ["gomod"],
+            ["gomod", "git-submodule"],
             None,
             ["gomod"],
             None,
+            True,
         ),
-        ([], [], "tom_hanks@DOMAIN.LOCAL", [], None),
-        ([], ["npm"], None, ["npm"], None),
-        ([], ["pip"], None, ["pip"], None),
+        ([], [], "tom_hanks@DOMAIN.LOCAL", [], None, False),
+        ([], ["npm"], None, ["npm"], None, False),
+        ([], ["pip"], None, ["pip"], None, False),
     ),
 )
 @mock.patch("cachito.web.api_v1.chain")
@@ -74,6 +76,7 @@ def test_create_and_fetch_request(
     client,
     db,
     flags,
+    gitsubmodule,
 ):
     data = {
         "repo": "https://github.com/release-engineering/retrodep.git",
@@ -109,6 +112,7 @@ def test_create_and_fetch_request(
             "https://github.com/release-engineering/retrodep.git",
             "c50b93a32df1c9d700e3e80996845bc2e13be848",
             1,
+            gitsubmodule,
         ).on_error(error_callback)
     ]
     if "gomod" in expected_pkg_managers:
@@ -155,6 +159,7 @@ def test_create_and_fetch_request_npm_package_configs(
             "https://github.com/release-engineering/web-terminal.git",
             "c50b93a32df1c9d700e3e80996845bc2e13be848",
             1,
+            False,
         ).on_error(error_callback),
         fetch_npm_source.si(1, package_value["npm"]).on_error(error_callback),
         create_bundle_archive.si(1).on_error(error_callback),
@@ -195,6 +200,7 @@ def test_create_and_fetch_request_pip_package_configs(
             "https://github.com/release-engineering/web-terminal.git",
             "c50b93a32df1c9d700e3e80996845bc2e13be848",
             1,
+            False,
         ).on_error(error_callback),
         fetch_pip_source.si(1, package_value["pip"]).on_error(error_callback),
         create_bundle_archive.si(1).on_error(error_callback),
@@ -245,6 +251,7 @@ def test_create_and_fetch_request_with_flag(mock_chain, app, auth_env, client, d
                 "https://github.com/release-engineering/retrodep.git",
                 "c50b93a32df1c9d700e3e80996845bc2e13be848",
                 1,
+                False,
             ).on_error(error_callback),
             fetch_gomod_source.si(1, []).on_error(error_callback),
             create_bundle_archive.si(1).on_error(error_callback),
@@ -1811,17 +1818,32 @@ def test_get_request_logs_not_configured(app, client, db, worker_auth_env):
     assert "logs" not in rv.json
 
 
+@pytest.mark.parametrize(
+    "data, gitsubmodule",
+    (
+        (
+            {
+                "repo": "https://github.com/release-engineering/retrodep.git",
+                "ref": "c50b93a32df1c9d700e3e80996845bc2e13be848",
+                "pkg_managers": ["pip"],
+            },
+            False,
+        ),
+        (
+            {
+                "repo": "https://github.com/release-engineering/retrodep.git",
+                "ref": "c50b93a32df1c9d700e3e80996845bc2e13be848",
+                "pkg_managers": ["pip", "git-submodule"],
+            },
+            True,
+        ),
+    ),
+)
 @mock.patch("cachito.web.api_v1.chain")
 def test_create_and_fetch_request_with_pip_preview(
-    mock_chain, app, auth_env, client, db,
+    mock_chain, app, auth_env, client, db, data, gitsubmodule,
 ):
     db.session.commit()
-
-    data = {
-        "repo": "https://github.com/release-engineering/retrodep.git",
-        "ref": "c50b93a32df1c9d700e3e80996845bc2e13be848",
-        "pkg_managers": ["pip"],
-    }
 
     rv = client.post("/api/v1/requests", json=data, environ_base=auth_env)
     assert rv.status_code == 201
@@ -1837,6 +1859,7 @@ def test_create_and_fetch_request_with_pip_preview(
                 "https://github.com/release-engineering/retrodep.git",
                 "c50b93a32df1c9d700e3e80996845bc2e13be848",
                 1,
+                gitsubmodule,
             ).on_error(error_callback),
             fetch_pip_source.si(1, {}).on_error(error_callback),
             create_bundle_archive.si(1).on_error(error_callback),
--- a/tests/test_workers/test_scm.py
+++ b/tests/test_workers/test_scm.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from unittest import mock
 import logging
 import os
+import re
 import tarfile
 
 import git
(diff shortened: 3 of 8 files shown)


nirzari:
Notes:
Updating a git submodule seems to be very tricky.
- We would probably not have the previous ref for the submodule just by traversing through the repo from `previous_archive`.
- Also, if a submodule was previously part of the `previous_archive`, trying to re-clone it seems to be resulting into an error (already exists). 
- There is no obvious way to find out if a submodule was part of the `previous_archive` and hence, there is no straightforward way to implement that check (except probably a hacky way to try to check the directory for empty content)
- if a previous_archive with submodules is being used, then the subsequent requests don't get rid of the associated submodules. This could lead to dependency issues in the next part of the story.
Hence, for the reasons stated above, the best option seems to be to exclude the previous_archives with submodules while updating and archiving repo for a new Cachito request.

Also, it is assumed that users would throw in "git-submodule" package manager only if the submodules need to be processed. Nothing harmful can happen if no submodules are found for the same but that archive won't be used for any subsequent requests and it would be stored in a path that has "with-submodule" in it.

There is a lot of caching issue with testing this PR. I tested for almost an entire day with multiple orders of commits, only to find out that my local browser was caching it badly and making submodules present/absent in my downloaded archives when they were not supposed to. (I'm slightly skeptical if the same behavior is possible when run non-locally)

I will create a separate PR for documenting this package manager as I get more progress with deps implementation.

You can use https://github.com/nirzari/retrodep/commits/master to test this PR.

chmeliik:
> Notes:
> Updating a git submodule seems to be very tricky.
>
> . . .
>
> Also, it is assumed that users would throw in "git-submodule" package manager only if the submodules need to be processed. Nothing harmful can happen if no submodules are found for the same but that archive won't be used for any subsequent requests and it would be stored in a path that has "with-submodule" in it.
>
> . . .

Excluding archives with submodules from the `update_and_archive` mechanism seems reasonable to me. To be honest, I don't really understand the purpose of that function. All I found is this comment:
https://github.com/release-engineering/cachito/blob/004a554d4f2c931dd242b8f95096307c368c0749/cachito/workers/scm.py#L131-L138

Can someone elaborate on why (if) we need to access the git history as the comment says?

cachito/workers/scm.py
```diff
@@ -147,7 +171,30 @@ def fetch_source(self):
...
+        try:
+            repo = git.repo.Repo.clone_from(
+                submodule.url,
+                submodule.abspath,
+                no_checkout=True,
+                # Don't allow git to prompt for a username if we don't have access
+                env={"GIT_TERMINAL_PROMPT": "0"},
+            )
+
+            self._reset_git_head(repo, submodule.hexsha)
```
chmeliik, line 194: If I'm not mistaken, you could replace all of this with `submodule.update()`

athos-ribeiro:
> > Notes:
> > Updating a git submodule seems to be very tricky.
> > . . .
> > Also, it is assumed that users would throw in "git-submodule" package manager only if the submodules need to be processed. Nothing harmful can happen if no submodules are found for the same but that archive won't be used for any subsequent requests and it would be stored in a path that has "with-submodule" in it.
> > . . .
> 
> Excluding archives with submodules from the `update_and_archive` mechanism seems reasonable to me. To be honest, I don't really understand the purpose of that function. All I found is this comment:
> 
> https://github.com/release-engineering/cachito/blob/004a554d4f2c931dd242b8f95096307c368c0749/cachito/workers/scm.py#L131-L138
> 
> Can someone elaborate on why (if) we need to access the git history as the comment says?

We use that function to avoid cloning the repository all over again, i.e., we get an archived repository (if one exists) and change to the requested ref. Additional data is fetched if not present already.

cachito/workers/scm.py
```diff
@@ -117,11 +125,22 @@ def update_and_archive(self, previous_archive):
...
 
-    def fetch_source(self):
-        """Fetch the repo, create a compressed tar file, and put it in long-term storage."""
+    def fetch_source(self, gitsubmodule=False):
+        """Fetch the repo, create a compressed tar file, and put it in long-term storage.
+
+        :param bool gitsubmodule: a bool to determine whether git submodules need to be processed.
+        """
+        if gitsubmodule:
+            self.sources_dir = SourcesDir(self.repo_name, f"{self.ref}-with-submodule")
```
chmeliik, line 142: If we're not going to use the `-with-submodule` archives for updating (which seems reasonable), then I would suggest storing them separately from the normal archives. For example, you could add something like `submodule_archives_dir` to the `SourcesDir` class, and you would store the archives with submodules there. Example file structure:

```
.
└── some-namespace
    └── retrodep
        ├── 1735db74fe4834fb0b95cc78d5aa3e3ef1dfbb73.tar.gz
        └── with-submodules
            └── 1735db74fe4834fb0b95cc78d5aa3e3ef1dfbb73.tar.gz
```

In the `fetch_source` function, when `gitsubmodule = True`, you would then do the following:
1. If archive with submodules already exists, return
2. Fetch repo and create normal archive (without submodules) as usual
3. Process submodules and create archive with submodules

This way, the submodules workflow would be more separated from the normal git workflow, which may make things less complicated. You would just need to make sure to extract the correct archive to the `RequestBundleDir` later in `fetch_app_source`.

chmeliik:
> We use that function to avoid cloning the repository all over again, i.e., we get an archived repository (if one exists) and change to the requested ref. Additional data is fetched if not present already.

@athos-ribeiro right, so it is really just a performance optimization, not something that Cachito depends on, right? Would it be OK to not support this for git submodules? Or do we want to find a way to support it? (We can always just file a story to do it later if it is not strictly necessary)

cachito/workers/scm.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, url, ref):
         self.sources_dir = SourcesDir(self.repo_name, ref)
 
     @abstractmethod
-    def fetch_source(self):
+    def fetch_source(self, gitsubmodule=False):
```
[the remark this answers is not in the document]
nirzari: Okay, will take it off. I wasn't sure anyway!

cachito/workers/scm.py
nirzari: Ummm, that may make things less complicated from the implementation point of view or enhance the readability but what is the other benefit of doing so over what we are doing rn except having a separate workflow?

The advantage of current workflow is that, even for a Cachtio request with submodules, in case of previous archive being present, we're not cloning the repo all over again but only adjusting the `ref` and cloning just the submodules. This makes it faster than having to clone every-time. I was trying to utilize the existing update functionality as much as possible.

cachito/workers/scm.py
nirzari: This works. How come I didn't come across that option. Where did you find it? -_-

chmeliik: I found it here: https://gitpython.readthedocs.io/en/stable/tutorial.html#submodule-handling

 But honestly, the only reason I knew `update()` would do what you need is that I have used submodules before and I was confused why the command used for fetching was `git submodule update` :D

cachito/workers/scm.py
chmeliik: You don't have to give up the advantages of the current workflow. You would first fetch the repo using the existing mechanism (re-use existing / update previous / clone from scratch), then you would extract the archive somewhere, fetch submodules and create new archive. It could even be a separate method.

But you're right in that there are no real advantages except maybe ease of implementation/understanding, so I'm ok with keeping it as it is (but having a separate directory for archives with submodules could still be relevant)

nirzari: I took up on your suggestion and tried with a separate directory for submodules, to make it look like `<repo_name>/submodule/<ref>.tar.gz`, but there are more places than just this class where we are accessing `sources_dir.archive_path`, e.g. https://github.com/release-engineering/cachito/blob/ad3a2b94bb62607e1b328a24ba0b4b148fb769d0/cachito/workers/pkg_managers/pip.py#L1672 and so just to simplify the implementation/coding here in this class, the value of `gitsubmodule` bool would need to be propagated within multiple other classes/functions. IMO, that would just make everything else more complex. With your other suggestion, the current implementation seems to be redable and understandable and so I'd like to keep it as it is. lmk wdyt!

cachito/workers/scm.py
```diff
@@ -72,10 +72,11 @@ def _create_archive(self, from_dir):
         with tarfile.open(self.sources_dir.archive_path, mode="w:gz") as bundle_archive:
             bundle_archive.add(from_dir, "app")
 
-    def clone_and_archive(self):
+    def clone_and_archive(self, gitsubmodule):
```
chmeliik, line 75: The other functions have `gitsubmodule` as a keyword arg, here it is a positional arg. I'd like to keep it consistent (my preference is keyword arg)

cachito/workers/scm.py
```diff
@@ -147,7 +167,23 @@ def fetch_source(self):
...
+
+        For the given repo, update submodules. For Cachito request for `retrodep` repo and
+        `go-github` submodule, the dir structure would look like,
+        retrodep/go-github/<content_of_go-github_repo>
+
+        :param git.Repo repo: the repository object.
+        """
+        try:
+            log.debug(f"Git submodules for the requested repo are: {repo.submodules}")
+            repo.submodule_update()
```
chmeliik, line 183: If I'm not mistaken, this function ends up calling [RootModule.update()](https://gitpython.readthedocs.io/en/stable/reference.html#git.objects.submodule.root.RootModule.update), which is recursive by default (fetches submodules inside submodules). Do we want that or not?

cachito/workers/scm.py
```diff
@@ -118,10 +124,19 @@ def update_and_archive(self, previous_archive):
...
 
-    def fetch_source(self):
-        """Fetch the repo, create a compressed tar file, and put it in long-term storage."""
+    def fetch_source(self, gitsubmodule=False):
+        """Fetch the repo, create a compressed tar file, and put it in long-term storage.
+
+        :param bool gitsubmodule: a bool to determine whether git submodules need to be processed.
+        """
+        if gitsubmodule:
+            self.sources_dir = SourcesDir(self.repo_name, f"{self.ref}-with-submodule")
```
chmeliik, line 138: (optional): Should this be `-with-submodules` (plural) since we expect there could be more than one?

cachito/workers/scm.py
```diff
@@ -147,7 +167,23 @@ def fetch_source(self):
             ) as exc:
                 log.warning("Error handling archived artifact '%s': %s", previous_archive, exc)
 
-        self.clone_and_archive()
+        self.clone_and_archive(gitsubmodule)
+
+    def update_git_submodule(self, repo):
```
chmeliik, line 172: (optional): Should this be `update_git_submodules` (plural) since we expect there could be more than one?

cachito/workers/scm.py
```diff
@@ -138,7 +153,12 @@ def fetch_source(self):
         )
         for previous_archive in previous_archives:
             try:
-                self.update_and_archive(previous_archive)
+                # Excluding previous archives with submodules since there is
+                # no straight-forward way to get rid of the previously included
+                # submodules
+                if "-with-submodule" in str(previous_archive):
```
chmeliik, line 159: Is `previous_archive` not a string? Do we need to call `str()` on it?

chmeliik (commented):
Found a few more things to nitpick about, but overall looks good :-)
I still have to go through unit tests.

cachito/workers/scm.py
```diff
@@ -147,7 +167,23 @@ def fetch_source(self):
...
+        For the given repo, update submodules. For Cachito request for `retrodep` repo and
+        `go-github` submodule, the dir structure would look like,
+        retrodep/go-github/<content_of_go-github_repo>
+
+        :param git.Repo repo: the repository object.
+        """
+        try:
+            log.debug(f"Git submodules for the requested repo are: {repo.submodules}")
+            repo.submodule_update()
+        except:  # noqa E722
```
athos-ribeiro, line 184: Can we capture specific exceptions here?

cachito/workers/scm.py
nirzari: AFAIR, We decided to not support nested gitsubmodules as of now. Adding a flag like that might make it more complicated when it comes to dependency handling. I'd refrain from opting that for now. 

Regarding the recursive thing, I did find that in the documentation initially but had thought of leaving it as it is since it's not adversely affecting something. but if there is a good reason to set it to `recursive=False`, I don't mind doing that either (if it is supported here)

cachito/workers/scm.py
nirzari: Yeah, it's not. It's of type `<class 'cachito.workers.paths.SourcesDir'>`

chmeliik: Ah gotcha :+1:

cachito/workers/scm.py
nirzari: None of the methods in the file has exception type defined. I'll add them all at once in a separate commit/PR

cachito/workers/scm.py
```diff
@@ -147,7 +167,23 @@ def fetch_source(self):
...
+        self.clone_and_archive(gitsubmodule)
+
+    def update_git_submodule(self, repo):
+        """Update git submodules.
+
+        For the given repo, update submodules. For Cachito request for `retrodep` repo and
+        `go-github` submodule, the dir structure would look like,
+        retrodep/go-github/<content_of_go-github_repo>
+
+        :param git.Repo repo: the repository object.
```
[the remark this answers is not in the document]
nirzari: Added

cachito/workers/scm.py
chmeliik: I'm not sure actually, would it be a problem if we cloned all submodules recursively but only included the top-level ones in the content manifest? My gut feeling says we should only clone what we declare in the content manifest, but IDK

nirzari: That seems fair. Let me test with setting `recursive=False`

nirzari: I updated it, however FWIW, `recursive=False` didn't make a difference. The submodules within submodules were not cloned anyhow. One could test it with,
`curl -X POST -H "Content-Type: application/json" http://localhost:8080/api/v1/requests -d '{"repo": "https://github.com/nirzari/retrodep.git", "ref": "d2a9933ca704a5b2120cd493af08d2d3aaa47ec2", "pkg_managers": ["gomod", "git-submodule"]}'`

chmeliik: It looks like you did not update the "tour" submodule to the latest commit in your repo, you can do that using `git submodule update --remote -- tour` (you may need to run `git submodule init` first).

When I updated the submodule to the latest commit, the recursive option behaved as expected.

tests/test_workers/test_scm.py
```diff
@@ -116,11 +120,27 @@ def test_fetch_source_by_pull(mock_update_and_archive, mock_getctime):
     po = mock.patch.object
     with po(scm_git.sources_dir.archive_path, "exists", return_value=False):
         with po(
-            scm_git.sources_dir.package_dir, "glob", return_value=["29eh2a.tar.gz", "a8c2d2.tar.gz"]
+            scm_git.sources_dir.package_dir,
+            "glob",
+            return_value=["a8c2d2.tar.gz", "a8c2d2-with-submodules.tar.gz"],
```
chmeliik, line 125: Should we use 3 different archives here?
1. no submodules, oldest
2. no submodules, second oldest
3. with submodules, newest

Archive 2 should be used (to test that newest goes first but `-with-submodules` is ignored)

tests/test_workers/test_scm.py
```diff
@@ -116,11 +120,27 @@ def test_fetch_source_by_pull(mock_update_and_archive, mock_getctime):
...
+    scm_git_submodule = scm.Git(url, f"{ref}-with-submodules")
+    mock_scr.return_value = scm_git_submodule.sources_dir
 
-    mock_update_and_archive.assert_called_once_with("a8c2d2.tar.gz")
+    po = mock.patch.object
+    with po(scm_git_submodule.sources_dir.archive_path, "exists", return_value=False):
+        with po(scm_git_submodule.sources_dir.package_dir, "glob", return_value=[]):
+            scm_git_submodule.fetch_source(True)
+
+    mock_clone_and_archive.assert_called_once_with(gitsubmodule=True)
```
chmeliik, line 143: This test could be dropped if you parametrize the one above instead

tests/test_workers/test_scm.py
```diff
@@ -194,4 +217,27 @@ def test_update_and_archive_pull_error(mock_repo, mock_tarfile_open):
...
+
+@mock.patch("git.Repo")
+def test_update_git_submodules(mock_repo):
+    repo = mock_repo.return_value
+    git_obj = scm.Git(url, ref)
+
+    git_obj.update_git_submodules(repo)
+
+    # Verify the git submodule update was called
+    repo.submodule_update.assert_called_once()
```
chmeliik, line 231: Should we also check that `recursive=False` was used?

tests/test_api_v1.py
```diff
@@ -109,6 +109,7 @@ def test_create_and_fetch_request(
             "https://github.com/release-engineering/retrodep.git",
             "c50b93a32df1c9d700e3e80996845bc2e13be848",
             1,
+            False,
```
chmeliik, line 112: Could we also have a test where gitsubmodule is expected to be True?

cachito/workers/scm.py
```diff
@@ -138,7 +153,13 @@ def fetch_source(self):
         )
         for previous_archive in previous_archives:
             try:
-                self.update_and_archive(previous_archive)
+                # Excluding previous archives with submodules since there is
+                # no straight-forward way to get rid of the previously included
+                # submodules
+                print(type(previous_archive))
```
chmeliik, line 159: I think you forgot this here :-)

cachito/workers/scm.py
```diff
@@ -181,7 +196,12 @@ def fetch_source(self):
         )
         for previous_archive in previous_archives:
             try:
-                self.update_and_archive(previous_archive)
+                # Excluding previous archives with submodules since there is
+                # no straight-forward way to get rid of the previously included
+                # submodules
+                if "-with-submodules" in str(previous_archive):
+                    continue
```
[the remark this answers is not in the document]
athos-ribeiro: Also, would it be possible to re-use the git history for the submodules as we do with the main repositories? In other words, in case the request includes the submodules pkg manager, could we prefer an archive with submodules?

I'd be OK leaving this for a separate PR, but we would keep consistency on how Cachito fetches sources by doing so (and would pull less data)

cachito/workers/scm.py
athos-ribeiro: I really think we should capture specific exceptions here. Handling errors like this could hide unrelated issues/bugs in our code.

From https://github.com/gitpython-developers/GitPython/blob/e4a83ff7910dc3617583da7e0965cd48a69bb669/git/objects/submodule/root.py, I see it raises at least `git.InvalidGitRepositoryError` and `ValueError`. However, it does capture all other exceptions that could be raised in a huge try/except block and raises.

If we are just catching any exception here, we should at least log the real exception that was raised instead of concealing it in a CachitoError. For instance, we could have sth like `except Exception as e` and log whatever we have in `e`.

cachito/workers/scm.py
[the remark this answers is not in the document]
nirzari: We tried to achieve that only at first but it was not possible as I had mentioned above.

> Notes:
> Updating a git submodule seems to be very tricky.
> 
> * We would probably not have the previous ref for the submodule just by traversing through the repo from `previous_archive`.
> * Also, if a submodule was previously part of the `previous_archive`, trying to re-clone it seems to be resulting into an error (already exists).
> * There is no obvious way to find out if a submodule was part of the `previous_archive` and hence, there is no straightforward way to implement that check (except probably a hacky way to try to check the directory for empty content)
> * if a previous_archive with submodules is being used, then the subsequent requests don't get rid of the associated submodules. This could lead to dependency issues in the next part of the story.
>   Hence, for the reasons stated above, the best option seems to be to exclude the previous_archives with submodules while updating and archiving repo for a new Cachito request.

Hence, we ended up adding these conditions.

If we really want to achieve that, there should be a spike created to investigate the feasibility first.

athos-ribeiro:
I wonder if we need setup/teardown methods for the scm tests due to those fale_repo files (tests are failing with permission errors)

athos-ribeiro (approved):
No regressions for integration tests.

I also tried your repository with the app submodule; tried a request without sub-modules in the same repo and finally tried a new clean request to check for the behavior when trying to re-use an archive.

LGTM

PR #288, opened 2020-09-29
Disable pip dev preview flag

nirzari:


cachito/web/migrations/versions/f133002ffdb4_disable_pip_dev_review.py
```diff
@@ -0,0 +1,36 @@
...
+
+def upgrade():
+    connection = op.get_bind()
+    connection.execute(
+        flag_table.update().where(flag_table.c.name == "pip-dev-preview").values(active=False)
+    )
+
+
+def downgrade():
+    pass
```
chmeliik, line 36: Shouldn't the downgrade re-enable the flag?

tests/integration/test_pip_packages.py
```diff
@@ -28,14 +28,7 @@ def test_all_pip_packages(env_name, test_env, tmpdir):
...
     initial_response = client.create_new_request(
-        payload={
-            "repo": env_data["repo"],
-            "ref": env_data["ref"],
-            "pkg_managers": ["pip"],
-            # TODO: delete pip-dev-preview flag when
-            #  the pip package manager will be ready for production usage
-            "flags": ["pip-dev-preview"],
-        },
+        payload={"repo": env_data["repo"], "ref": env_data["ref"], "pkg_managers": ["pip"],}
```
chmeliik, line 31:
```suggestion
        payload={"repo": env_data["repo"], "ref": env_data["ref"], "pkg_managers": ["pip"]}
```
`black` is complaining about this

nirzari: Yes I noticed just now. fixed it

cachito/web/migrations/versions/f133002ffdb4_disable_pip_dev_review.py
nirzari: Should it? :thinking: I thought the idea was to make sure that we don't re-enable it in anyway and hence, we could just leave the downgrade blank

akhmanova (commented):
Everything related to the integration test looks good to me! Nice work. Let wait for approval related other parts :)

cachito/web/migrations/versions/f133002ffdb4_disable_pip_dev_review.py
```diff
@@ -33,4 +33,13 @@ def upgrade():
...
+    connection = op.get_bind()
+    res = connection.execute(
+        flag_table.select().where(flag_table.c.name == "pip-dev-preview")
+    ).fetchone()
+    if res is None:
+        connection.execute(flag_table.insert().values(name="pip-dev-preview", active=True))
+    else:
+        connection.execute(
+            flag_table.update().where(flag_table.c.name == "pip-dev-preview").values(active=True)
+        )
```
chmeliik, line 45: This change is in the wrong commit (should be in "Migration for disabling pip-dev-preview flag")

athos-ribeiro: In this matter, the commit order seem to be inverted as well: We should first remove the filters for the flag and then disable it in the DB.

athos-ribeiro:
Nice work! LGTM, there is just a couple comments on commit order and on which contents should go where (Adam's comment).

cachito/web/migrations/versions/f133002ffdb4_disable_pip_dev_review.py
nirzari: Good catch, Fixed it

akhmanova:
Sorry for making more work for you, could you please delete the flag in test_pip_local_path as well https://github.com/release-engineering/cachito/pull/291 ?

athos-ribeiro (approved):

PR #289, opened 2020-09-29
Verify archived git repositories

athos-ribeiro:
When storing a tarball for a git repository, make sure such tarball is
valid and contains a valid git repository. When invalid, delete the new
archive and fail the request.

Moreover, if a request for an already existing archive is performed,
check the existing archive. In case this archive is invalid, throw a
warning and re-create the archive.

### DIFF
--- a/cachito/workers/scm.py
+++ b/cachito/workers/scm.py
@@ -1,10 +1,11 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
+from abc import ABC, abstractmethod
 import logging
 import os
-from abc import ABC, abstractmethod
-import urllib.parse
+import subprocess
 import tarfile
 import tempfile
+import urllib.parse
 
 import git
 
@@ -62,15 +63,53 @@ class Git(SCM):
                 f'of "{self.ref}" is valid.'
             )
 
+    def _verify_archive(self):
+        """
+        Verify the archive containing the git repository.
+
+        :raises CachitoError: if 'git fsck' fails for the extracted sources
+        """
+        log.debug("Verifying the archive at %s", self.sources_dir.archive_path)
+        if not os.path.exists(self.sources_dir.archive_path) or not tarfile.is_tarfile(
+            self.sources_dir.archive_path
+        ):
+            err_msg = f"No valid archive found at {self.sources_dir.archive_path}"
+            log.exception(err_msg)
+            raise CachitoError(err_msg)
+
+        with tempfile.TemporaryDirectory(prefix="cachito-") as temp_dir:
+            with tarfile.open(self.sources_dir.archive_path, mode="r:gz") as tar:
+                tar.extractall(temp_dir)
+
+            cmd = ["git", "fsck"]
+            repo_path = os.path.join(temp_dir, "app")
+            try:
+                subprocess.run(cmd, cwd=repo_path, check=True, capture_output=True)
+            except subprocess.CalledProcessError as exc:
+                log.error(
+                    "Cachito found an error when verifying the generated archive at '%s': %s - %s",
+                    self.sources_dir.archive_path,
+                    exc,
+                    exc.stderr,
+                )
+                raise CachitoError(f"Invalid archive at {self.sources_dir.archive_path!s}")
+
     def _create_archive(self, from_dir):
         """
-        Create an archive from a specified directory.
+        Create a verified archive from a specified directory.
 
         :param str from_dir: path to a directory from where to create the archive.
+        :raises CachitoError: if the archive verification fails
         """
         log.debug("Creating the archive at %s", self.sources_dir.archive_path)
         with tarfile.open(self.sources_dir.archive_path, mode="w:gz") as bundle_archive:
             bundle_archive.add(from_dir, "app")
+        try:
+            self._verify_archive()
+        except CachitoError:
+            log.debug("Removing invalid archive at %s", self.sources_dir.archive_path)
+            os.unlink(self.sources_dir.archive_path)
+            raise
 
     def clone_and_archive(self):
         """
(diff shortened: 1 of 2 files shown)


cachito/workers/scm.py
```diff
@@ -62,15 +63,50 @@ def _reset_git_head(self, repo):
...
+
+            cmd = ["git", "fsck"]
+            repo_path = os.path.join(temp_dir, "app")
+            try:
+                subprocess.run(cmd, cwd=repo_path, check=True)
+            except subprocess.CalledProcessError as exc:
+                log.error(
+                    "Cachito found an error when verifying the generated archive at '%s': %s",
+                    self.sources_dir.archive_path,
+                    exc,
```
lcarva, line 90: Will this contain the data from stderr to help diagnose issues?

cachito/workers/scm.py
```diff
@@ -126,7 +162,11 @@ def fetch_source(self):
         archive_path = self.sources_dir.archive_path
         if archive_path.exists() and tarfile.is_tarfile(archive_path):
```
lcarva, line 163: Should we remove the `is_tarfile` call from here since it's done during `_verify_archive`?

cachito/workers/scm.py
```diff
@@ -62,15 +63,50 @@ def _reset_git_head(self, repo):
...
+        """
+        Verify the archive containing the git repository.
+
+        :raises CachitoError: if 'git fsck' fails for the extracted sources
+        """
+        log.debug("Verifying the archive at %s", self.sources_dir.archive_path)
+        if not os.path.exists(self.sources_dir.archive_path) or not tarfile.is_tarfile(
+            self.sources_dir.archive_path
+        ):
+            raise CachitoError(f"No valid tar archive at {self.sources_dir.archive_path}")
```
nirzari, line 76:
```suggestion
            log.error(f"No valid archive found at %s, {self.sources_dir.archive_path}")
            raise CachitoError(f"No valid archive found at {self.sources_dir.archive_path}")
```
I think we have only been using the word just `archive` as it is assumed to be a tar only, and the word 'tar' would be more for developers than the users?

Also, should we log the error as well?

cachito/workers/scm.py
```diff
@@ -62,15 +63,50 @@ def _reset_git_head(self, repo):
...
+            repo_path = os.path.join(temp_dir, "app")
+            try:
+                subprocess.run(cmd, cwd=repo_path, check=True)
+            except subprocess.CalledProcessError as exc:
+                log.error(
+                    "Cachito found an error when verifying the generated archive at '%s': %s",
+                    self.sources_dir.archive_path,
+                    exc,
+                )
+                raise CachitoError(f"Invalid tar archive at {self.sources_dir.archive_path!s}")
```
nirzari, line 92: Same as above if you decide to use just the "archive" for user facing error.

nirzari (changes requested):

cachito/workers/scm.py
```diff
@@ -62,15 +63,50 @@ def _reset_git_head(self, repo):
...
         """
         log.debug("Creating the archive at %s", self.sources_dir.archive_path)
         with tarfile.open(self.sources_dir.archive_path, mode="w:gz") as bundle_archive:
             bundle_archive.add(from_dir, "app")
+        try:
+            self._verify_archive()
+        except CachitoError:
+            log.debug("Removing invalid archive at %s", self.sources_dir.archive_path)
+            os.unlink(self.sources_dir.archive_path)
+            raise
```
nirzari, line 109: Should this raise CachitoError or fall back on re-cloning the repo?

athos-ribeiro: It should re-raise the CachitoError at some point.

I thought about retrying the operation, but then I thought these errors would just be forgotten in the logs in case they start happening often and they could hide issues with the file system or the OCP cluster.

I'd be happy to change it to log a warning and retry the operation N times (once should be enough) before raising here.

Thoughts?

cachito/workers/scm.py
athos-ribeiro: Good idea. initially, I did not add it thinking most of the messages in stderr would be just "Not a git repository". On a second thought, we better be safe here and include stderr.

cachito/workers/scm.py
athos-ribeiro: Done

cachito/workers/scm.py
athos-ribeiro: Changed the exception message.

Do we need to repeat ourselves in the logs here? There is no additional information being added in the log message and I do not see how this exception message would be missing from the logs.

cachito/workers/scm.py
athos-ribeiro: Done

cachito/workers/scm.py
chmeliik: IMO if the data is corrupted immediately after creation, something is very wrong and I would rather have the request fail. On the off-chance that a retry would work, it would just hide a potentially serious issue as Athos said.

cachito/workers/scm.py
nirzari: It wouldn't be missed from the visibility in `/<request-id>/logs` url content as that's how the request would fail and it'd point out to this line at the end of the logs but my suggestion was more to be consistent with logging throughout the module. One benefit I could see is, if in the future we are querying from kibana or someplace else through logs, this could be helpful. One such example can be found at https://github.com/release-engineering/cachito/blob/master/cachito/workers/pkg_managers/general.py#L58

cachito/workers/scm.py
```diff
@@ -124,9 +161,13 @@ def fetch_source(self):
...
         archive_path = self.sources_dir.archive_path
-        if archive_path.exists() and tarfile.is_tarfile(archive_path):
+        if archive_path.exists():
             log.debug('The archive already exists at "%s"', archive_path)
-            return
+            try:
+                self._verify_archive()
+                return
+            except CachitoError:
+                log.debug('The archive at "%s" is invalid and will be re-created', archive_path)
```
nirzari, line 170:
```suggestion
                log.error('The archive at "%s" is invalid and will be re-created', archive_path)
```
This could be warning too but I think it should not be `debug`

cachito/workers/scm.py
nirzari: Hmmm. I'm fine with raising this issue without a re-try. If we are checking the verification here, and are planning to clean up the existing archives explicitly, why do we need to have this check for existing archives? (out of curiosity)

athos-ribeiro: We will run a script in our production environment to clean the corrupted files present at the moment. After this is merged, no more corrupted caches should be created. Even if it happens (let's say a worker is killed while writing a file), next time the same reference is requested, the archive will be re-created. If a different reference is requested, the corrupted file will not be used.

chmeliik (approved):
LGTM :+1:

PR #294, opened 2020-10-01
Use vcs purls for top-level packages

chmeliik:

### DIFF
--- a/cachito/web/content_manifest.py
+++ b/cachito/web/content_manifest.py
@@ -49,9 +49,8 @@ class ContentManifest:
         :param Dependency dependency: the go-package package dependency to process
         """
         if dependency.type == "go-package":
-            purl = package.to_top_level_purl(self.request)
             icm_dependency = {"purl": dependency.to_purl()}
-            self._gopkg_data[purl]["dependencies"].append(icm_dependency)
+            self._gopkg_data[package.id]["dependencies"].append(icm_dependency)
 
     def set_go_package_sources(self):
         """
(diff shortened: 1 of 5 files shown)


chmeliik:
Don't mind the failing tests, just trying to figure out if we want this approach.

My concern is that by using `pkg:generic` (or `pkg:github`), we lose info about the type and version of packages.

The solution could be to use the same purls as before and just add the `vcs_url` qualifier, which is [valid for *all* package types][1]:
> With this warning, the known key and value defined here are valid for use in all package types:

However, for dependency purls, we currently use `pkg:generic` or `pkg:github` in case they come from VCS, so this would be inconsistent.

@athos-ribeiro thoughts? Are generic vcs purls good enough for the use cases (what are those, anyway?) of content manifests?

[1]: https://github.com/package-url/purl-spec#known-qualifiers-keyvalue-pairs
[2]: https://github.com/package-url/purl-spec#purl

athos-ribeiro:
> Don't mind the failing tests, just trying to figure out if we want this approach.
> 
> My concern is that by using `pkg:generic` (or `pkg:github`), we lose info about the type and version of packages.
> 
> The solution could be to use the same purls as before and just add the `vcs_url` qualifier, which is [valid for _all_ package types](https://github.com/package-url/purl-spec#known-qualifiers-keyvalue-pairs):
> 
> > With this warning, the known key and value defined here are valid for use in all package types:
> 
> However, for dependency purls, we currently use `pkg:generic` or `pkg:github` in case they come from VCS, so this would be inconsistent.
> 
> @athos-ribeiro thoughts? Are generic vcs purls good enough for the use cases (what are those, anyway?) of content manifests?

Ideally, we would use `github`, `gitlab`, etc for such forges and generic for anything else. I say ideally for readability on the PURL. 

It is important to notethat the only role of these PURL strings is to uniquely locate a source artifact in a standardized fashion. PURL docs and upstream discussions suggest it should be done in the most readable form possible. For instance, using `github` instead of `generic` + `vcs_url` if possible.

In the other hand, generic is already better then what we currently have: At the moment, we are not providing correct means to locate the top level requested packages.

athos-ribeiro (commented):
LGTM!

Thanks for fixing integration tests as well.

Would you mind elaborating on the 3rd commit, for purls not being unique? The commit looks good, I am wondering what cases you had in mind for that though.

chmeliik:
> LGTM!
> 
> Thanks for fixing integration tests as well.
> 
> Would you mind elaborating on the 3rd commit, for purls not being unique? The commit looks good, I am wondering what cases you had in mind for that though.

From now until we implement support for subpaths, all npm and pip packages in a request will have the same purl, e.g. `pkg:github/namespace/repo@commit-id`. Previously, purls likely would have been unique because they typically contain the package name, but in some edge cases there could still have been conflicts (like name normalization in purl)

athos-ribeiro (approved):

PR #299, opened 2020-10-06
Add script for finding python build dependencies

chmeliik:
Given a requirements.txt file with all recursive runtime dependencies,
this script will find all their build dependencies.


athos-ribeiro (commented):
Nice work!

Although we are using a set to avoid duplications, it is possible that a version is specified and and entry ends up being duplicated (that is the case for cachito's requirements.txt file, which yields:

```
# Generated by pip_find_builddeps.py on Oct 06 2020 13:31:28
cffi!=1.11.3,>=1.8
pycparser
setuptools>=40.6.0
setuptools>=40.8.0
wheel
```

Should we handle these cases to output something like

```
# Generated by pip_find_builddeps.py on Oct 06 2020 13:31:28
cffi!=1.11.3,>=1.8
pycparser
setuptools>=40.6.0,>=40.8.0
wheel
```

instead?

bin/pip_find_builddeps.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+@contextlib.contextmanager
+def _remove_on_success(tmpdir):
+    """Remove tmpdir only if no exception is raised during context."""
+    try:
+        yield
+    except:  # noqa: E722
+        # Re-raise the exception but keep the tmpdir
+        raise
```
athos-ribeiro, line 48: I am wondering if it would be better to print the error message to STDERR end exit with a code > 0 here for so it is easier for a human user to parse the output.

Then, for the most common error here (which I suppose to be dependencies not being fetched), the user would see just:

`ERROR: pip download failed, see /tmp/pip_find_builddeps.py-j0o4ghx0/pip-download-output.txt for more info`

Instead of the stack trace.

A debug mode could output the full trace though.

chmeliik: Good idea, I instead added a custom exception type and I'm suppressing it for main() execution. This way, expected errors will look nice and unexpected errors will still produce a full traceback

chmeliik:
> Nice work!
> 
> Although we are using a set to avoid duplications, it is possible that a version is specified and and entry ends up being duplicated (that is the case for cachito's requirements.txt file, which yields:
> 
> ```
> # Generated by pip_find_builddeps.py on Oct 06 2020 13:31:28
> cffi!=1.11.3,>=1.8
> pycparser
> setuptools>=40.6.0
> setuptools>=40.8.0
> wheel
> ```
> 
> Should we handle these cases to output something like
> 
> ```
> # Generated by pip_find_builddeps.py on Oct 06 2020 13:31:28
> cffi!=1.11.3,>=1.8
> pycparser
> setuptools>=40.6.0,>=40.8.0
> wheel
> ```
> 
> instead?

I believe that would only be a cosmetic difference, right? If so, I'd rather not do it because it adds a dependency on `setuptools` for parsing the requirements

athos-ribeiro:
> > Nice work!
> > Although we are using a set to avoid duplications, it is possible that a version is specified and and entry ends up being duplicated (that is the case for cachito's requirements.txt file, which yields:
> > ```
> > # Generated by pip_find_builddeps.py on Oct 06 2020 13:31:28
> > cffi!=1.11.3,>=1.8
> > pycparser
> > setuptools>=40.6.0
> > setuptools>=40.8.0
> > wheel
> > ```
> > 
> > 
> > Should we handle these cases to output something like
> > ```
> > # Generated by pip_find_builddeps.py on Oct 06 2020 13:31:28
> > cffi!=1.11.3,>=1.8
> > pycparser
> > setuptools>=40.6.0,>=40.8.0
> > wheel
> > ```
> > 
> > 
> > instead?
> 
> I believe that would only be a cosmetic difference, right? If so, I'd rather not do it because it adds a dependency on `setuptools` for parsing the requirements

Not really.

pip will detect a duplicated entry in the file for that case:

```
ERROR: Double requirement given: setuptools>=40.8.0 (from -r requirements.txt (line 5)) (already in setuptools>=40.6.0 (from -r requirements.txt (line 4)), name='setuptools')
```

If we are introducing an additional dependency, I wouldn't mind just documenting this and letting users know they should address duplicated entries before submitting a request using the script output as an additional requirement file for example.

athos-ribeiro (approved):

PR #301, opened 2020-10-06
Add submodule as a package to the request

nirzari:
Patch request api is called to add submodule as a package to the request


chmeliik (changes requested):
I would strongly prefer if the package-manager code for git submodules was completely separate from the fetching code. Ideally a new module in `cachito.workers.tasks` with a method for adding all git submodules from repo to the database.

This method would be used in `cachito.web.api_v1.create_request` the same way that other package managers are used there.

cachito/workers/tasks/gitsubmodule.py
```diff
@@ -0,0 +1,50 @@
...
+    :param str url: the source control URL to pull the source from
+    :param str ref: the source control reference
+    :raises CachitoError: if adding submodules as a package fail.
+    """
+    scm = Git(url, ref)
+    sources_dir = SourcesDir(scm.repo_name, f"{scm.ref}-with-submodules")
+    archive_path = sources_dir.archive_path
+    with tempfile.TemporaryDirectory(prefix="cachito-") as temp_dir:
+        with tarfile.open(archive_path, mode="r:gz") as tar:
+            tar.extractall(temp_dir)
```
chmeliik, line 34: I'm looking at this code in `fetch_app_source()`: https://github.com/release-engineering/cachito/blob/58fe40c3710265fd0fc2c9113b1341425ff59685/cachito/workers/tasks/general.py#L46-L51
Maybe we don't even need to extract the source archive? The repo might be accessible like this:
```python
bundle_dir = RequestBundleDir(request_id)
repo = git.Repo(str(bundle_dir.source_root_dir))
```

cachito/workers/tasks/gitsubmodule.py
```diff
@@ -0,0 +1,50 @@
...
+        for sm in repo.submodules:
+            try:
+                # Save package to db
+                package = {
+                    "type": "git-submodule",
+                    "name": sm.name,
+                    "version": f"{sm.url}#{sm.hexsha}",
+                }
+                log.debug("Adding submodule '%s' as a package for Cachito request", sm.name)
+                update_request_with_package(request_id, package)
```
athos-ribeiro, line 45: This is the only line we need in the try/except block, right?

cachito/workers/tasks/gitsubmodule.py
```diff
@@ -0,0 +1,50 @@
...
+            except Exception as e:
```
athos-ribeiro, line 46: `update_request_with_package` raises a CachitoError when an API call to update the request fails. Can we capture specific exceptions here? 

If something else fails for some other unexpected reason, we should consider it to be a bug and tend to it. Letting it fail without capturing it here may help debugging (and should definitely call our attention to it), when/if it happens.

cachito/workers/tasks/gitsubmodule.py
nirzari: TBH I was also thinking about this yesterday but fixing unit tests took up too much time and I forgot to try this way. Will try and update if that works.

chmeliik:
Looks good apart from the current comments :-)

I have not looked too thoroughly at unit tests as those are likely to change soon, but I don't expect I will have too many issues with those

tests/test_workers/test_scm.py
```diff
@@ -294,19 +294,21 @@ def test_create_and_verify_archive(fake_repo, caplog):
     assert f"Verifying the archive at {git_obj.sources_dir.archive_path}" in caplog.text
 
 
-def test_clone_and_verify_archive(fake_repo, caplog):
+@pytest.mark.parametrize("gitsubmodule", [True, False])
+def test_clone_and_verify_archive(fake_repo, caplog, gitsubmodule):
     repo_dir, _ = fake_repo
     git_obj = scm.Git(f"file://{repo_dir}", "master")
-    git_obj.clone_and_archive()
+    git_obj.clone_and_archive(gitsubmodule)
```
athos-ribeiro, line 301: Should we test if update_git_submodules was called here?

tests/test_workers/test_scm.py
```diff
@@ -294,19 +294,21 @@ def test_create_and_verify_archive(fake_repo, caplog):
...
-def test_update_and_verify_archive(fake_repo, caplog):
+@pytest.mark.parametrize("gitsubmodule", [True, False])
+def test_update_and_verify_archive(fake_repo, caplog, gitsubmodule):
     repo_dir, _ = fake_repo
     git_obj = scm.Git(f"file://{repo_dir}", "master")
-    git_obj.clone_and_archive()
+    git_obj.clone_and_archive(gitsubmodule)
     caplog.clear()
-    git_obj.update_and_archive(git_obj.sources_dir.archive_path)
+    git_obj.update_and_archive(git_obj.sources_dir.archive_path, gitsubmodule)
```
athos-ribeiro, line 311: Should we test if update_git_submodules was called here?

tests/test_api_v1.py
```diff
@@ -1816,58 +1816,3 @@ def test_get_request_logs_not_configured(app, client, db, worker_auth_env):
     rv = client.get(f"/api/v1/requests/{request_id}")
     assert rv.status_code == 200
     assert "logs" not in rv.json
-
-
-@pytest.mark.parametrize(
```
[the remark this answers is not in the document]
nirzari: I don't see a value in retaining this. The only reason this was added was for the purpose of testing out the flag. Instead of editing at the time of flag removal, this should have been removed altogether. The scenarios being tested are covered in https://github.com/release-engineering/cachito/blob/master/tests/test_api_v1.py#L64 and removing this doesn't affect coverage either.

This is already being handled in a separate commit as part of this PR.

tests/test_workers/test_scm.py
nirzari: We could technically do but I don't see a value in doing so since that is already being tested at https://github.com/release-engineering/cachito/blob/master/tests/test_workers/test_scm.py#L70

Adding it here won't change the coverage and so, I'd rather stick to the title of the unit test to make tests easy to read.

tests/test_workers/test_scm.py
nirzari: We could technically do but I don't see a value in doing so since that is already being tested at https://github.com/release-engineering/cachito/blob/master/tests/test_workers/test_scm.py#L273

Adding it here won't change the coverage and so, I'd rather stick to the title of the unit test to make tests easy to read.

tests/test_api_v1.py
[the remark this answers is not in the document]
nirzari: I just re-read your comment. I still feel that we don't need a test for something that doesn't exist as part of the codebase/workflow anymore. If we were to follow that approach, we could end up with infinite unit tests.

tests/test_workers/test_scm.py
athos-ribeiro: In this case, why do we need the parametrization for the gitsubmodule there?

Note that in the test you pointed, the update_git_submodules was being mocked whereas here it isn't.

tests/test_api_v1.py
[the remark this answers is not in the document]
athos-ribeiro: The flag still exists, it is just marked as inactive in the database and this test could be used to make sure the migration ran properly and the flag remains deactivated or that its usage introduce breakage.

> If we were to follow that approach, we could potentially end up with infinite unit tests.

Unit test completeness should not be seen as a bad thing.  Since the test is already there, we might as well use it instead of just dropping it. It may show itself to be useless. It may also prevent us introducing bugs in the future.

athos-ribeiro (approved):
Code looks good to me.

I left some nitpicks inline.

I'd rather not see tests being removed due to a feature change, but adapted to cover the new code behavior.

I will leave it to you to decide whether to merge or drop the commit removing unit tests :)

Nice work! :rocket:

tests/test_workers/test_scm.py
nirzari: Fair enough. Will take off the parametrization from this test, or mock the `update_git_submodule`. Not sure how come it works even without mocking though :thinking:

tests/test_api_v1.py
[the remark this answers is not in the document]
nirzari: I see your point, however the test can't be left alone since introducing a new task will require this to be modified as well and upon doing so it's a duplication of an existing test.

If we wanted to test the migration indirectly, as you mentioned it should have been done separately, probably when we disabled the flag.

According to the functionality, passing an inactive flag will fail the request on https://github.com/release-engineering/cachito/blob/master/cachito/web/models.py#L814 and it won't even reach the code we're testing here.

Hence, I'll remove it from this module. If we want to test the same functionality, we can add it back in the future as part of other modules while testing the flag existence.

nirzari:
All integration tests passed

tests/test_workers/test_scm.py
athos-ribeiro: It works because there are no submodules and the `fake_repo` in an actual gi repository with "fake" data

PR #302, opened 2020-10-07
Improve log flow for corrupted cache detection

athos-ribeiro:
The error message logged when a corrupted cache is found may lead a user
to believe something went wrong with the request. This patch adds an
additional log entry to let readers know Cachito is recovering from the
issue.

### DIFF
--- a/cachito/workers/scm.py
+++ b/cachito/workers/scm.py
@@ -208,7 +208,16 @@ class Git(SCM):
                 tarfile.ExtractError,
                 OSError,  # raised by tarfile when an FS operation fails
             ) as exc:
-                log.warning("Error handling archived artifact '%s': %s", previous_archive, exc)
+                log.warning(
+                    "Error handling archived artifact '%s': %s - %s",
+                    previous_archive,
+                    repr(exc),
+                    exc,
+                )
+                log.info(
+                    "Existing archive at '%s' may be corrupted and will not be used. Recovering...",
+                    previous_archive,
+                )
 
         self.clone_and_archive(gitsubmodule=gitsubmodule)
 


cachito/workers/scm.py
```diff
@@ -209,6 +209,10 @@ def fetch_source(self, gitsubmodule=False):
                 OSError,  # raised by tarfile when an FS operation fails
             ) as exc:
                 log.warning("Error handling archived artifact '%s': %s", previous_archive, exc)
```
chmeliik, line 211: (optional): Should we also log the exception type (or use repr)? Sometimes, the type says more than the message

chmeliik (commented):
LGTM, one comment unrelated to the change

cachito/workers/scm.py
```diff
@@ -209,6 +209,10 @@ def fetch_source(self, gitsubmodule=False):
...
+                log.info(
```
nirzari, line 212: Should this also be a `log.warning` ?

nirzari (changes requested):

cachito/workers/scm.py
athos-ribeiro: Not sure, since the idea here is to announce Cachito is recovering from the issue raised in the previous warning...

cachito/workers/scm.py
athos-ribeiro: I added the repr to the msg as well. Not sure if the repr alone should be enough or if we better leave the string msg there as well

chmeliik: Good point, we can keep both

cachito/workers/scm.py
chmeliik: IMO `log.info` is fine

chmeliik (approved):
LGTM :+1:

nirzari (approved):

PR #304, opened 2020-10-09
pip_find_builddeps.py improvements

chmeliik:
Various improvements for the pip_find_builddeps.py script


bin/pip_find_builddeps.py
```diff
@@ -96,18 +132,17 @@ def main():
...
-    lines = [f"# Generated by {SCRIPT_NAME} on {date}"]
-    if builddeps:
-        lines.extend(builddeps)
-    else:
-        lines.append("# <no build dependencies found>")
+    builddeps = find_builddeps(args.requirements_files, no_cache=args.no_cache)
+    file_content = generate_file_content(builddeps)
 
-    file_content = "\n".join(lines)
+    log.info("Will write output to %s", args.output_file or "stdout")
```
athos-ribeiro, line 159: Is this one needed as `info`? The command input/output should say this already (if an output file is given, it will be in the user command, otherwise, the output will be shown in the screen).

bin/pip_find_builddeps.py
```diff
@@ -56,47 +55,48 @@ def _filter_builddeps(pip_download_output_file):
...
+            log.error(msg)
+            log.warning("Ignoring error...")
+            is_partial = True
+        else:
             raise FindBuilddepsError(msg)
 
+    try:
         log.info("Looking for build dependencies in the output of pip download")
-        return _filter_builddeps(pip_output_file)
+        builddeps = _filter_builddeps(pip_output_file)
```
athos-ribeiro, line 88: Not sure if I follow the reasoning behind this last try/finally block: Let's say an encoding error happened while parsing the `pip download` output even if the command succeeded. Then we are still removing the file before raising the exception, right? Is this the intended behaviour?

athos-ribeiro (commented):
LGTM

Nitpicks and questions in line

bin/pip_find_builddeps.py
chmeliik: Hmm, true, if parsing the file fails, it would probably be nice to keep that file. Will change

athos-ribeiro (approved):

PR #308, opened 2020-10-12
Fix cleanup job

chmeliik:
* CLOUDBLD-2629

By marking requests as stale while iterating through the paginated
/requests?state={state} endpoint, we modify the data as we are iterating
over it. This may cause the cleanup job to miss requests that it should
clean up, and may also result in a 404 error where a page stops existing
because the total number of requests in given state is reduced.

To avoid this, first collect all the requests in the specified state and
then start marking them as stale.

### DIFF
--- a/cachito/workers/cleanup_job.py
+++ b/cachito/workers/cleanup_job.py
@@ -23,38 +23,53 @@ payload = {"state": state, "state_reason": state_reason}
 def main():
     """Mark all end of life requests as stale using the REST API."""
     for state in ("complete", "in_progress", "failed"):
-        url = f"{config.cachito_api_url.rstrip('/')}/requests"
-        while url:
-            try:
-                response = session.get(
-                    url, params={"state": state}, timeout=config.cachito_api_timeout
-                )
-            except requests.RequestException:
-                msg = f"The connection failed when querying {url}"
-                log.exception(msg)
-                raise CachitoError(msg)
+        stale_candidate_requests = find_all_requests_in_state(state)
+        identify_and_mark_stale_requests(stale_candidate_requests)
 
-            if not response.ok:
-                log.error(
-                    "The request to %s failed with the status code %d and the following text: %s",
-                    url,
-                    response.status_code,
-                    response.text,
-                )
-                raise CachitoError(
-                    "Could not reach the Cachito API to find the requests to be marked as stale"
-                )
 
-            json_response = response.json()
-            identify_and_mark_stale_requests(json_response["items"])
-            url = json_response["meta"]["next"]
+def find_all_requests_in_state(state):
+    """
+    Find all requests in specified state.
+
+    :param str state: state of request, e.g. 'complete', 'in_progress'
+    :return: list of Cachito requests (as JSON data) in specified state
+    """
+    found_requests = []
+
+    url = f"{config.cachito_api_url.rstrip('/')}/requests"
+    while url:
+        try:
+            response = session.get(url, params={"state": state}, timeout=config.cachito_api_timeout)
+        except requests.RequestException:
+            msg = f"The connection failed when querying {url}"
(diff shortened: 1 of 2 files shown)


chmeliik:
To test locally:
- Clean up any persistent data and redeploy
- Create 11 failed requests: 
    ```shell
    for i in {1..11}; do
            -X POST \
            -d '{"repo": "https://github.com/chmeliik/cachito-sample-pip-package/", "ref": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"}'
    done
    ```
- Inside the cachito-worker container, modify the following line: https://github.com/release-engineering/cachito/blob/c20409547a0e845f4cf5d811cd22c0afa11d36e7/cachito/workers/cleanup_job.py#L64
    - use a small delta, such as `timedelta(seconds=1)`
- Inside the cachito-worker container, run `cachito-cleanup`
    - with code from master, you will get 404, 10 requests will be marked as stale, 1 will stay in failed state
    - with this change, everything should be marked as stale

When testing the change, you may need to rebuild the cachito containers (changes to this script may not be picked up on redeployment)

athos-ribeiro (commented):
Could we cover the former problematic scenario in `tests/test_workers/test_cleanup_job.py`?

chmeliik:
> Could we cover the former problematic scenario in `tests/test_workers/test_cleanup_job.py`?

I tried something, let me know what you think about the code for mocking pagination. Is it too complex to be a valuable test? For what it's worth, the test does fail on master as expected

cachito/workers/cleanup_job.py
```diff
@@ -23,38 +23,51 @@
...
-                    response.text,
-                )
-                raise CachitoError(
-                    "Could not reach the Cachito API to find the requests to be marked as stale"
-                )
-
-            json_response = response.json()
-            identify_and_mark_stale_requests(json_response["items"])
-            url = json_response["meta"]["next"]
+        maybe_stale_requests = find_all_requests_in_state(state)
```
athos-ribeiro, line 26: Would `stale_candidate_requests` be a better name for this?

cachito/workers/cleanup_job.py
```diff
@@ -23,38 +23,51 @@
...
+            )
+            raise CachitoError(
+                "Could not reach the Cachito API to find the requests to be marked as stale"
+            )
+
+        json_response = response.json()
+        found_requests.extend(json_response["items"])
+        url = json_response["meta"]["next"]
+
+    return found_requests
```
athos-ribeiro, line 63: Should we ensure entries in `found_requests` are unique (using ther `id`s)? I suppose the contents in each page may be quite dynamic depending on the load in Chachito at a given hour.

This is a minor issue since the request will not be update if we POST the same data, but still, less requests...

However, it does make the API throw a (somewhat unrelated) error when applying the operation twice in the same request:

```
[2020-10-13 09:42:41 flask.app INFO auth.load_user_from_request] The REMOTE_USER environment variable wasn't set on the request, but the LOGIN_DISABLED configuration is set to True.
[2020-10-13 09:42:41 flask.app ERROR api_v1.patch_request] Failed to delete the log file /var/log/cachito/requests/10.log
  File "/src/cachito/web/api_v1.py", line 431, in patch_request
    os.remove(path_to_file)
FileNotFoundError: [Errno 2] No such file or directory: '/var/log/cachito/requests/10.log'
```

athos-ribeiro (commented):
Looks good! 

These tests should be enough. I do have a minor comment on possible duplicated entries in the stale candidates list.

cachito/workers/cleanup_job.py
chmeliik: Good point, fixed

athos-ribeiro (approved):
Good job! :rocket:

PR #309, opened 2020-10-12
Fix default pkg configuration param values

athos-ribeiro:
The package configurations parameters passed to the fetch_npm_source and
fetch_pip_source tasks should be lists. The API code makes them default
to dicts. This patch fixes the default values for those params.

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -237,7 +237,7 @@ def create_request():
                 "Dependency replacements are not yet supported for the npm package manager"
             )
 
-        npm_package_configs = package_configs.get("npm", {})
+        npm_package_configs = package_configs.get("npm", [])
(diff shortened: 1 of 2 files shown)


mprahl (approved):

nirzari (approved):
LGTM. Feel free to merge once integration tests are run successfully. I was just wondering if we need to let anybody know about this or has impact on users?

Because I saw an ACM request which used ` "packages": {"pip": [{"path": "."}, {"path": "subpackage"}]}` as part of their Cachito request.
[Update]: @arewm  Aah, I just woke up and totally neglected those brackets. lol. :heavy_check_mark:

arewm:
@nirzari, that package config should still be valid even after this PR goes through. It is a list of dictionaries. :)

PR #305, opened 2020-10-09
Support multiple gomod packages in a single repository

arewm:
- CLOUDBLD-1611

Handle multiple gomod packages in a single repo in Cachito

While investigating the code to see if a custom `go.mod` path could be provided to Cachito, I went ahead and performed an initial implementation of supporting multiple gomod packages.

Signed-off-by: arewm <arewm@users.noreply.github.com>

PR #311, opened 2020-10-13
Include package subpaths in purls and /requests JSON

chmeliik:
* CLOUDBLD-2769

Git submodules,  pip packages and npm packages will now have their subpaths stored in the DB.
The subpaths will also be present in content manifests and the /requests/{id} JSON response.

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -442,6 +442,9 @@ class RequestPackage(db.Model):
     package_id = db.Column(
         db.Integer, db.ForeignKey("package.id"), autoincrement=False, index=True, primary_key=True
     )
+    # Each package in a request may have a custom subpath. If the package is located in the root
+    # of the request repo, subpath will be null.
+    subpath = db.Column(db.String, nullable=True)
 
     __table_args__ = (db.UniqueConstraint("request_id", "package_id"),)
 
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -153,18 +153,21 @@ class Package(db.Model):
         cls.validate_json(package)
         return cls(**package)
 
-    def to_json(self, dependencies=None):
+    def to_json(self, dependencies=None, subpath=None):
         """
         Generate the JSON representation of the package.
 
         :param list dependencies: the JSON representation of the dependencies associated with this
             package
+        :param str subpath: relative path to package from root of repository
         :return: the JSON form of the Package object
         :rtype: dict
         """
         rv = {"name": self.name, "type": self.type, "version": self.version}
         if dependencies is not None:
             rv["dependencies"] = dependencies
+        if subpath:
+            rv["path"] = subpath
         return rv
 
     @classmethod
@@ -794,7 +797,10 @@ class Request(db.Model):
                 package_to_deps.setdefault(req_dep.package.id, []).append(dep_json)
 
             rv["packages"] = [
-                req_package.package.to_json(package_to_deps.get(req_package.package_id, []))
+                req_package.package.to_json(
+                    dependencies=package_to_deps.get(req_package.package_id, []),
+                    subpath=req_package.subpath,
+                )
                 for req_package in self.request_packages
             ]
             if flask.current_app.config["CACHITO_REQUEST_FILE_LOGS_DIR"]:
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -446,6 +446,9 @@ class RequestPackage(db.Model):
     # of the request repo, subpath will be null.
     subpath = db.Column(db.String, nullable=True)
 
+    request = db.relationship("Request", backref="request_packages")
+    package = db.relationship("Package", foreign_keys=[package_id], lazy="joined")
+
     __table_args__ = (db.UniqueConstraint("request_id", "package_id"),)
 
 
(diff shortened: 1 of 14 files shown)


chmeliik:
First PR of [at least 2] for adding subpaths to purls. Adds the database column but does not use it yet (so there is no change in functionality)

chmeliik:
Added commits from what would be the second PR, now subpaths for git submodules, pip packages and npm packages will be stored in the database.

They are not yet shown anywhere (neither in the `/request/{id}` JSON nor in a content manifest)

chmeliik:
To test locally:
```bash
curl "localhost:8080/api/v1/requests" \
    -X POST \
    -d '{
          "repo": "https://github.com/chmeliik/cachito-sample-pip-package/",
          "ref": "1ca07be3001450dbc4f0224e0f763c60353d0f01",
          "pkg_managers": ["git-submodule", "pip", "npm"],
          "packages": {
            "pip": [
              {"path": "cachito-pip-with-deps"}
            ],
            "npm": [
              {"path": "cachito-npm-test"}
            ]
          }
        }'
```
```bash
docker-compose exec db psql cachito -U cachito -c '
SELECT package.name, package.version, request_package.subpath
FROM package JOIN request_package ON package.id = request_package.package_id
WHERE request_package.subpath IS NOT NULL
'
```

chmeliik:
Does not support gomod yet, as support for multiple paths is currently being added in #305

chmeliik:
Rebased on master

chmeliik:
Fixed issue introduced by rebasing

chmeliik:
Included final 2 commits. Now, when testing as described above, you can use the `/requests/{id}` and `/requests/{id}/content-manifest` endpoints instead of querying the database.

cachito/web/models.py
```diff
@@ -597,20 +597,36 @@ class Request(db.Model):
...
+
         :param Package package: the Package object to associate with the request
-        :param str subpath: custom subpath for package within request repository
+        :param association_attrs: additional attributes to set on the RequestPackage association
         :raises ValidationError: if the association already exists with a different subpath
         """
+        subpath_provided = "subpath" in association_attrs
+
+        subpath = association_attrs.get("subpath")
+        if not subpath or subpath == os.curdir:
```
nirzari, line 633: I'm slightly confused about this part `subpath == os.curdir`. The description mentions one condition to set subpath to null which is for new package, when subpath is not provided. Why are we setting it to null when subpath matches current directory?

cachito/workers/tasks/gitsubmodule.py
```diff
@@ -29,4 +29,4 @@ def add_git_submodules_as_package(request_id):
             "version": f"{sm.url}#{sm.hexsha}",
         }
         log.debug("Adding submodule '%s' as a package for Cachito request", sm.name)
-        update_request_with_package(request_id, package)
+        update_request_with_package(request_id, package, package_subpath=sm.path)
```
nirzari, line 32: In the later commit it is mentioned that _purls for git submodules point to a different repo, subpath is neither needed nor valid_ then what is the purpose of adding this here?

nirzari (commented):
Overall LGTM. Left some comments for the parts that were bit unclear to me.
Will run the code tomorrow to test locally.
Good Work!

cachito/web/models.py
chmeliik: Confusingly, [os.curdir](https://docs.python.org/3/library/os.html#os.curdir) is not actually the current directory, it is the string that the operating system uses to refer to the current directory (`"."` on all the systems that Python supports). Should I use `subpath == "."` instead to make it easier to understand?

cachito/workers/tasks/gitsubmodule.py
chmeliik: IMO it's worth storing this information even if we don't currently need it. We may need subpaths for something in the future (and they *are* already included in the `.packages` part of a `/requests/{id}` response)

cachito/web/models.py
nirzari: Oh okay. I think just adding a comment there might suffice

cachito/workers/tasks/gitsubmodule.py
nirzari: Okay makes sense. I don't see any harm in storing this additional info anyhow

nirzari (approved):
Tested locally. The request fails towards the end for me while _Downloading aiowsgi==0.7_ due to connection error. (It tries a few times to connect before failing). But if you're getting successful result for the same, I'll +1 this. Looks okay to merge from my end post resolving the open thread.

chmeliik:
> Tested locally. The request fails towards the end for me while _Downloading aiowsgi==0.7_ due to connection error. (It tries a few times to connect before failing). But if you're getting successful result for the same, I'll +1 this. Looks okay to merge from my end post resolving the open thread.

See #319, I've been meaning to fix this for a while but I never got around to it, sorry :D

cachito/web/models.py
athos-ribeiro: +1 to adding a comment there :)

cachito/workers/tasks/gitsubmodule.py
athos-ribeiro: Is this accurate though? Semantically, isn't the subpath of a package related to its own root directory?
It should be OK ot have a different semantic for submodules, but then we should document this somewhere.

chmeliik: Well, since subpaths are kind of a new concept, I guess we get to decide the semantics now. For purls, the semantics are obvious, because you must be able to locate the package with only the purl. For the JSON representation of a request, it probably does not matter that much, the current semantics are "relative path to package from root of request repo."

WDYT?

chmeliik:
Rebased on master

athos-ribeiro (approved):
Nice work!

PR #312, opened 2020-10-13
Add purl support for git submodule

nirzari:
Content manifest endpoint now includes the `purl` value for git submodule if it is available as package for Cachito request

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -256,6 +256,13 @@ class Package(db.Model):
                 quoted_url = urllib.parse.quote(self.version, safe="")
                 return f"pkg:generic/{name}?download_url={quoted_url}&checksum={checksum}"
 
+        elif self.type == "git-submodule":
+            # Version is a submodule repository url followed by `#` separator and
+            # `submodule-commit-ref`, e.g.
+            # https://github.com/org-name/submodule-name.git#522fb816eec295ad58bc488c74b2b46748d471b2
+            repo_url, ref = self.version.rsplit("#", 1)
+            return self.to_vcs_purl(repo_url, ref)
+
         else:
             raise ContentManifestError(f"The PURL spec is not defined for {self.type} packages")
 
@@ -306,7 +313,7 @@ class Package(db.Model):
         :return: the PURL string of the Package object
         :rtype: str
         """
-        if self.type in ("gomod", "go-package"):
+        if self.type in ("gomod", "go-package", "git-submodule"):
             return self.to_purl()
         elif self.type in ("npm", "pip"):
             return self.to_vcs_purl(request.repo, request.ref)
--- a/tests/test_content_manifest.py
+++ b/tests/test_content_manifest.py
@@ -177,6 +177,37 @@ def test_to_json(app, package):
     assert cm.to_json() == expected
 
 
+@pytest.mark.parametrize(
+    "packages",
+    [
+        [
+            {"name": "example.com/org/project", "type": "go-package", "version": "1.1.1"},
+            {
+                "name": "tour",
+                "type": "git-submodule",
+                "version": (
+                    "https://github.com/testrepo/tour.git#58c88e4952e95935c0dd72d4a24b0c44f2249f5b"
+                ),
+            },
+        ]
+    ],
+)
+@mock.patch("cachito.web.content_manifest.ContentManifest.generate_icm")
+def test_to_json_with_multiple_packages(mock_generate_icm, app, packages):
+    request = Request()
+    cm = ContentManifest(request)
+
+    image_contents = []
+    for package in packages:
+        pkg = Package.from_json(package)
+        request.packages.append(pkg)
+        content = {"purl": pkg.to_purl(), "dependencies": [], "sources": []}
+        image_contents.append(content)
+    res = cm.to_json()
+    mock_generate_icm.assert_called_once_with(image_contents)
+    assert res == mock_generate_icm.return_value
+
+
 @pytest.mark.parametrize("contents", [None, [], "foobar", 42, OrderedDict({"egg": "bacon"})])
 def test_generate_icm(contents, default_request):
     cm = ContentManifest(default_request)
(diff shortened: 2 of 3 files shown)


nirzari:
Relevant part of content manifest:

> "image_contents": [
{
"dependencies": [],
"purl": "pkg:github/nirzari/retrodep@abfebf169e8d67fcf5de07d41bbe9b1ea132491b",
"sources": []
},
{
"dependencies": [...],
"purl": "pkg:golang/github.com%2Frelease-engineering%2Fretrodep%2Fv2@v2.1.2-0.20201002070802-abfebf169e8d",
"sources": [...]
}
],

for the Cachito Request,
> curl -X POST -H "Content-Type: application/json" http://localhost:8080/api/v1/requests -d '{"repo": "https://github.com/nirzari/retrodep.git", "ref": "abfebf169e8d67fcf5de07d41bbe9b1ea132491b", "pkg_managers": ["gomod", "git-submodule"]}'

cachito/web/models.py
```diff
@@ -308,7 +308,7 @@ def to_top_level_purl(self, request):
         """
         if self.type in ("gomod", "go-package"):
             return self.to_purl()
-        elif self.type in ("npm", "pip"):
+        elif self.type in ("npm", "pip", "git-submodule"):
             return self.to_vcs_purl(request.repo, request.ref)
```
chmeliik, line 312: IMO there are 2 ways to identify a git submodule using a purl (in this example, the "tour" submodule in https://github.com/nirzari/retrodep):
1. main repo + submodule path: `pkg:github/nirzari/retrodep@abfebf169e8d67fcf5de07d41bbe9b1ea132491b#tour`
2. submodule repo: `pkg:github/nirzari/tour@522fb816eec295ad58bc488c74b2b46748d471b2`

This is option 1, except the subpath is missing (I'm working on that)
As far as I can tell, the only immediate benefit of option 2 would  be that it is correct even without subpath support, so option 1 is probably fine. I could be missing something though, cc @athos-ribeiro

chmeliik: Since we've agreed on option 2 offline, here's some more details:
- the `to_purl()` method should be updated to handle "git-submodule" packages
    - you can use the `to_vcs_purl()` method, but you should parse the `repo_url` and `ref` from `self.version` (as stored in the DB)
    - see the handling for pip vcs dependencies https://github.com/release-engineering/cachito/blob/b61783d1590a220570ad3f1643e4a46f83360b3d/cachito/web/models.py#L246-L251
- in `to_top_level_purl()`, "git-submodule" packages should just call `to_purl()` (like "gomod" and "go-package")

tests/test_content_manifest.py
```diff
@@ -153,28 +153,36 @@ def test_process_pip(default_request, default_toplevel_purl):
...
+        [
+            None,
+            {"name": "example.com/org/project", "type": "go-package", "version": "1.1.1"},
+            {"name": "example.com/org/project", "type": "git-submodule", "version": "1.1.1"},
+        ],
+    ],
 )
-def test_to_json(app, package):
+@mock.patch("cachito.web.content_manifest.ContentManifest.generate_icm")
+def test_to_json(mock_generate_icm, app, packages):
```
chmeliik, line 166: We should also have a test for processing git submodules similar to `test_process_{go,npm,pip}`

chmeliik, line 166: Once `to_purl()` supports git submodules, `test_purl_conversion` should also include a test case for it

tests/test_content_manifest.py
```diff
@@ -152,29 +152,63 @@ def test_process_pip(default_request, default_toplevel_purl):
...
+    # emulate to_json behavior to setup internal packages cache
+    cm._gitsubmodule_data.setdefault(
+        pkg.id, {"purl": expected_purl, "dependencies": [], "sources": []}
+    )
+
+    expected_contents = {pkg.id: {"purl": expected_purl, "dependencies": [], "sources": []}}
+
+    assert cm._gitsubmodule_data
+    assert pkg.id in cm._gitsubmodule_data
+    assert cm._gitsubmodule_data == expected_contents
```
chmeliik, line 178: Oh my bad, there's no `process_gitsubmodule_package()` method so there's nothing to test here. This test can be removed

tests/test_content_manifest.py
```diff
@@ -152,29 +152,63 @@ def test_process_pip(default_request, default_toplevel_purl):
...
-    assert cm.to_json() == expected
+    for package in packages:
+        if package:
+            pkg = Package.from_json(package)
+            request.packages.append(pkg)
+            content = {"purl": pkg.to_purl(), "dependencies": [], "sources": []}
+            image_contents.append(content)
+    res = cm.to_json()
+    mock_generate_icm.assert_called_once_with(image_contents)
+    assert res == mock_generate_icm.return_value
```
chmeliik, line 211: Why were the changes in this test needed?

chmeliik (commented):
Looks good, only comments about unit tests left

tests/test_content_manifest.py
nirzari: Since, I passed an additional package there, I'd need to update the `expected` to have deep sort run on it to make it an Ordered dict probably. And we already have a separate test beneath `test_generate_icm`. And this test is for `to_json` so why not just mock `generate_icm`?

chmeliik: But the `packages` param was originally just `package`, right? It is a good idea to also test with multiple packages, but then I would like to keep the original test cases as well (empty list, list with just 1 package).

+1 to mocking `generate_icm`

tests/test_content_manifest.py
```diff
@@ -177,6 +177,37 @@ def test_to_json(app, package):
...
+                "type": "git-submodule",
+                "version": (
+                    "https://github.com/testrepo/tour.git#58c88e4952e95935c0dd72d4a24b0c44f2249f5b"
+                ),
+            },
+        ]
+    ],
+)
+@mock.patch("cachito.web.content_manifest.ContentManifest.generate_icm")
+def test_to_json_with_muiltiple_packages(mock_generate_icm, app, packages):
```
chmeliik, line 196: typo: "muiltiple"

chmeliik (approved):
LGTM 👍 

Found a typo in test name, should be good to merge apart from that

PR #314, opened 2020-10-14
Document git-submodule pkg manager support

nirzari:


README.md
```diff
@@ -489,16 +489,17 @@ Nexus instance that hosts the permanent content: `cachito_nexus_hoster_username`
...
----                     | ---   | --- | --- |
-Baseline                | ✓     | ✓   | ✓   |
-Content Manifest        | ✓     | ✓   | ✓   |
-Dependency Replacements | ✓     | x   | x   |
-Dev Dependencies        | ✓     | ✓   | ✓   |
-External Dependencies   | N/A   | ✓   | ✓   |
-Multiple Paths          | x     | ✓   | ✓   |
-Nested Dependencies     | ✓     | ✓   | x   |
-Offline Installations   | ✓     | x   | x   |
+Feature                 | gomod | npm | pip | git-submodule |
```
chmeliik, line 492: Not sure if we should include `git-submodule` here, it's not really a full-featured package manager, more like a secondary thing that would usually be used in combination with other package managers.

IMO we could just include the relevant info in the `### git-submodule` section below

README.md
```diff
@@ -489,16 +489,17 @@ Nexus instance that hosts the permanent content: `cachito_nexus_hoster_username`
...
+---                     | ---   | --- | --- |               |
+Baseline                | ✓     | ✓   | ✓   | ✓⁺            |
+Content Manifest        | ✓     | ✓   | ✓   | ✓             |
+Dependency Replacements | ✓     | x   | x   | x             |
+Dev Dependencies        | ✓     | ✓   | ✓   | x             |
+External Dependencies   | N/A   | ✓   | ✓   | N/A           |
+Multiple Paths          | x     | ✓   | ✓   | x             |
```
chmeliik, line 499: We do support multiple paths in the sense that if the repo contains multiple submodules, we will fetch them all. It is just not configurable. Even if we do drop `git-submodule` from this table, it might be worth mentioning elsewhere.

README.md
nirzari: I thought a lot about this as well. The only reason I ended up choosing this approach was to make sure, each possible value for `pkg_managers` is part of the table as in to give an overview, hence I also put a `+` there to explain the specifics of it but the other part of me agrees with you. So, I'll just remove it

README.md
```diff
@@ -636,3 +636,37 @@ Compared to gomod and npm, Cachito support for pip has restrictions and limitati
 not expect. For more details, see the [Cachito pip documentation](docs/pip.md).
 
 [nexus-docs]: https://help.sonatype.com/repomanager3
+
+### git-submodule
+
+Cachito will produce a bundle that is downloadable at `/api/v1/requests/<id>/download`. This
```
chmeliik, line 642: Could we have a sort of introductory first paragraph like we do for the other package managers? It could explain what git-submodule does and mention the fact that submodules are fetched *before* any of the other package managers are processed.

Or you could just reword this paragraph slightly, it currently does not feel like a "1st paragraph"

README.md
```diff
@@ -636,3 +636,37 @@ Compared to gomod and npm, Cachito support for pip has restrictions and limitati
...
+  "repo": "https://github.com/namespace/repo",
+  "ref": "abcdef...",
+  "pkg_managers": ["git-submodule", "pip"],
+  "packages": {
+    "pip": [
+      {"path": "cool_python_submodule"}
+    ]
+  }
+}
+```
```
chmeliik, line 663: Would it be possible to provide a real example instead? Maybe you could link one of the repos you used for testing or create a new one (keep in mind that subpaths for gomod packages are currently not supported).

README.md
nirzari: Okay I agree.I had a few lines as an introductory paragraph in my draft but felt like it is only repeating stuff but I"ll try to do that.

README.md
nirzari: What would be benefit from providing a real repo example? IMO, if I were user, I'd rather like the current example which clearly mentions the components of url too. e.g. namespace, repo

I could maybe update the request to look like, `"repo": "https://github.com/<repo-namespace>/<repo>.git"`

wdyt?

chmeliik: What I had in mind was to use a real value for "repo" and "ref", so a user could actually go to the repo and see what's in it, or maybe a developer could submit the request locally and see how it works.

But IDK how valuable it would be, I personally don't need it, I'm trying to put myself in the shoes of someone who does not already know how it works :D

README.md
```diff
@@ -636,3 +636,63 @@ Compared to gomod and npm, Cachito support for pip has restrictions and limitati
...
+-d '{
+      "repo": "https://github.com/nirzari/retrodep.git",
+      "ref": "18002daac67f82f1a0f3b1f41beb3469f23116ea",
+      "pkg_managers": ["gomod", "git-submodule"]
+    }'
+```
+
+In the above case, submodules `tour` and `go-github` within specified `retrodep` repo are fetched
+as part of the downloadable bundle. They would also be available as packages for Cachito API request
+repsonse. Furhter, they become part of the Content Manifest.
```
chmeliik, line 672:
```suggestion
response. Further, they become part of the Content Manifest.
```

chmeliik (commented):
LGTM, just 2 typos

README.md
nirzari: Aaaghhh. I need to type slower or focus more on right-left hand coordination :/ Thanks for catching these.

chmeliik (approved):

PR #315, opened 2020-10-15
Store subpaths for packages in database

chmeliik:
* CLOUDBLD-2769

Modify the PATCH API to allow setting subpaths for packages. Also modify the helper function that utilizes the PATCH API to add packages to request.

Save subpaths for git submodules, pip and npm packages.


chmeliik:
Will merge this into the first PR

PR #317, opened 2020-10-20
Support arbitrary user ids within container image

nirzari:
This is to avoid hitting an error with `get_user_id()` while using git-python library. We can remove this when an upstream patch is submitted to fix the issue.

### DIFF
--- a/docker/Dockerfile-workers
+++ b/docker/Dockerfile-workers
@@ -6,10 +6,12 @@ RUN dnf -y install \
     --setopt=deltarpm=0 \
     --setopt=install_weak_deps=false \
     --setopt=tsflags=nodocs \
-    golang \
+    gettext \
     git-core \
+    golang \
     mercurial \
     npm \
+    nss_wrapper \
     python3-celery \
     python3-GitPython \
     python3-packaging \
@@ -22,4 +24,5 @@ RUN dnf -y install \
 COPY . .
 RUN pip3 install . --no-deps
 EXPOSE 8080
+ENTRYPOINT ["/src/docker/docker-entrypoint.sh"]
 CMD ["/bin/celery-3", "-A", "cachito.workers.tasks", "worker", "--loglevel=info"]
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+set -e
+
+# The first 6 commands are added here for using NSS_WRAPPER. See "Support Arbitrary User IDs" in
+# https://access.redhat.com/documentation/en-us/openshift_container_platform/3.11/html/creating_images/creating-images-guidelines
+export USER_ID=$(id -u)
+export GROUP_ID=$(id -g)
+envsubst < /src/docker/passwd.template > /tmp/passwd
+export LD_PRELOAD=/usr/lib64/libnss_wrapper.so
+export NSS_WRAPPER_PASSWD=/tmp/passwd
+export NSS_WRAPPER_GROUP=/etc/group
+exec "$@"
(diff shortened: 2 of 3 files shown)


docker/Dockerfile-workers
```diff
@@ -22,4 +24,6 @@ RUN dnf -y install \
 COPY . .
 RUN pip3 install . --no-deps
 EXPOSE 8080
-CMD ["/bin/celery-3", "-A", "cachito.workers.tasks", "worker", "--loglevel=info"]
+CMD ["export USER_ID=$(id -u)", "export GROUP_ID=$(id -g)", "envsubst < ${HOME}/passwd.template > /tmp/passwd", \
+    "export LD_PRELOAD=/usr/lib64/libnss_wrapper.so", "export NSS_WRAPPER_PASSWD=/tmp/passwd", \
+    "export NSS_WRAPPER_GROUP=/etc/group", "/bin/celery-3", "-A", "cachito.workers.tasks", "worker", "--loglevel=info"]
```
athos-ribeiro, line 29: There's no newline break here, could you add one?

athos-ribeiro: Also, should this become an "entrypoint.sh" script?

docker/password.template
```diff
@@ -0,0 +1,14 @@
...
+lp:x:4:7:lp:/var/spool/lpd:/sbin/nologin
+sync:x:5:0:sync:/sbin:/bin/sync
+shutdown:x:6:0:shutdown:/sbin:/sbin/shutdown
+halt:x:7:0:halt:/sbin:/sbin/halt
+mail:x:8:12:mail:/var/spool/mail:/sbin/nologin
+operator:x:11:0:operator:/root:/sbin/nologin
+games:x:12:100:games:/usr/games:/sbin/nologin
+ftp:x:14:50:FTP User:/var/ftp:/sbin/nologin
+nobody:x:99:99:Nobody:/:/sbin/nologin
+cachito:x:${USER_ID}:${GROUP_ID}:Cachito Server:${HOME}:/bin/bash
```
lcarva, line 14: Missing newline at end of file. Not sure if that would actually break the file format though. Probably not.

docker/Dockerfile-workers
nirzari: Any specific benefits for the same? IMO, since this is temp. and we might be removing it soon once the upstream patch is submitted, there might not be a need for it. wdyt?

lcarva: The benefit of entrypoint is that it would probably reduce the size of this command, but the downside is that OpenShift's handling of entrypoint and command is confusing: https://docs.openshift.com/container-platform/3.11/dev_guide/deployments/basic_deployment_operations.html#executing-commands-inside-a-container-deployments

I don't have a strong preference either way.

athos-ribeiro: The definition for `CMD` [1] in the current format is that all other entries after the first one sould be considered parameters for the command being run.

In that list, I see 2 different commands being executed (envsubst and /bin/celery-3).

Moreover, I see no reason for getting rid of this approach in the future, since it may save us from future issues regarding some library consulting /etc/passwd.

[1] https://docs.docker.com/engine/reference/builder/#cmd

lcarva (approved):

docker/Dockerfile-workers
athos-ribeiro: TBH, I do not fully understand why whatever follows `envsubst` is not being processed as parameters for that command.

@lcarva, any ideas?

nirzari: Hmm good point @athos-ribeiro The doc mentions _There can only be one CMD instruction in a Dockerfile. If you list more than one CMD then only the last CMD will take effect._

Doesn't it mean that it's ignoring everything else that we added and only running the last one? :thinking:

nirzari: I got rid of all the additional commands that we added and the git-submodule support still works fine :brain:

lcarva (changes requested):
I didn't catch this at first, but as we've seen, the CMD instruction won't work as is. Let's try using ENTRYPOINT  as suggested.

lcarva (approved):
Nice!

PR #319, opened 2020-10-21
Set correct PyPI proxy url for dev environment

chmeliik:
The default url is never valid, so remove it from Config and add the
correct one to DevelopmentConfig.

### DIFF
--- a/cachito/workers/config.py
+++ b/cachito/workers/config.py
@@ -42,7 +42,6 @@ class Config(object):
     cachito_nexus_max_search_attempts = 5
     cachito_nexus_npm_proxy_repo_url = "http://localhost:8081/repository/cachito-js/"
     cachito_nexus_pip_raw_repo_name = "cachito-pip-raw"
-    cachito_nexus_pypi_proxy_url = "http://localhost:8081/repository/cachito-pip-proxy/"
     cachito_nexus_proxy_password = None
     cachito_nexus_proxy_username = None
     cachito_nexus_request_repo_prefix = "cachito-"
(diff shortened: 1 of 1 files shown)


cachito/workers/config.py
```diff
@@ -42,7 +42,6 @@ class Config(object):
     cachito_nexus_max_search_attempts = 5
     cachito_nexus_npm_proxy_repo_url = "http://localhost:8081/repository/cachito-js/"
     cachito_nexus_pip_raw_repo_name = "cachito-pip-raw"
-    cachito_nexus_pypi_proxy_url = "http://localhost:8081/repository/cachito-pip-proxy/"
```
athos-ribeiro, line 45: Should the same thing be done to the JS one above? We should normalize those names as well (could be in a different PR though; changes to deployments would also be needed for this one.)

chmeliik: Unlike the pypi proxy url, this one is actually valid, relevant part of README:
> cachito_nexus_npm_proxy_repo_url - the URL to the cachito-js repository which is a Nexus group that points to the cachito-js-hosted hosted repository and the cachito-js-proxy proxy repository. This defaults to http://localhost:8081/repository/cachito-js/. This only needs to change if you are using the two Nexus instance approach as described in the "Nexus For npm" section or you use a different name for the repository.

So I would keep it in the main Config, since it is a sane default.

+1 for normalizing names, I'll file something to sync it with the next deployment

athos-ribeiro (approved):

cachito/workers/config.py
athos-ribeiro: It seems to fall in the same case where this is only valid if nexus is hosted in the same machine running Cachito. Even for a single instance of nexus, pointing to localhost would only work if that instance is not external.

I am OK with merging the single change for Pyhton though

athos-ribeiro: NVM, I could reason behind it now (the JS one is only used in a specific case whereas the pip one os always used)

PR #322, opened 2020-10-23
Perform cleanup if a corrupted archive is created

athos-ribeiro:
Cachito performs cleanups when the archive verification method raises a
CachitoError. Currently, the method only raises CachitoError when a git
repository is bogus, but not when a tarball is corrupted. This patch
changes Cachito to also raise the error when a tar archive cannot be
properly extracted.

### DIFF
--- a/cachito/workers/scm.py
+++ b/cachito/workers/scm.py
@@ -6,6 +6,7 @@ import subprocess
 import tarfile
 import tempfile
 import urllib.parse
+import zlib
 
 import git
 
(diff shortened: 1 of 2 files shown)


chmeliik (approved):
LGTM, one optional suggestion

cachito/workers/scm.py
```diff
@@ -77,22 +78,26 @@ def _verify_archive(self):
...
-            except subprocess.CalledProcessError as exc:
-                log.error(
-                    "Cachito found an error when verifying the generated archive at '%s': %s - %s",
-                    self.sources_dir.archive_path,
-                    exc,
-                    exc.stderr,
-                )
-                raise CachitoError(f"Invalid archive at {self.sources_dir.archive_path!s}")
+            except (subprocess.CalledProcessError) as exc:
+                msg = ".".join([err_msg["log"], " STDERR: %s"])
```
chmeliik, line 98:
```suggestion
                msg = f"{err_msg['log']}. STDERR: %s"
```
The suggestion (forgot to add this, oops)

athos-ribeiro: Done

chmeliik (approved):

PR #323, opened 2020-10-26
Explicitly support Python 3.8 and 3.9

athos-ribeiro:
Before moving Cachito container images to new Fedora releases, we should
make sure the code base supports the Python version shipped in those new
releases. Let's start running our unit tests for new Python versions.


athos-ribeiro:
This will require some work in travis. Will defer this for whenever we migrate to f32

PR #324, opened 2020-10-26
Avoid incomplete archive race conditions

athos-ribeiro:
When a Cachito request is writing a tarball for a big git repository and
another request is performed for that same repository, the task for the
latter request will try to re-use the tarball that is still being
written, resulting in an error later during the archive verification.

This patch changes Cachito to save the archive for the first request in
a temporary file in the same filesystem where the final archive will be
stored. Cachito then creates a hard link to the temporary archive in the
final path where the archive is expected to be, and removes
the temporary archive. If the second request is made while the archive
for the first request is being written, Cachito will proceed to create a
new temporary file for the second request. If the requests are for
different references in the same repository, each request will create
its own final archive named after the git commit hash for the requested
revision. However, if the requests are for the same references, the
first request to finish writing the temporary file will claim the hard
link pointing to its temporary file. The second one will log a warning,
discard the temporary file, and proceed, using the archive generated for
the other request.

### DIFF
--- a/cachito/workers/scm.py
+++ b/cachito/workers/scm.py
@@ -103,12 +103,39 @@ class Git(SCM):
         """
         Create a verified archive from a specified directory.
 
+        We first create the archive in a temporary path so other tasks will not
+        use it while it is being written. Note that this operation must be done
+        in the same filesystem to avoid copy operations to be performed, which
+        would result in the same issue as writting the archive directly to the
+        final path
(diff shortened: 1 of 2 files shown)


chmeliik (approved):
Nice! I never thought I would see a legitimate use case for a hardlink, but here we are :D

LGTM :+1:

PR #325, opened 2020-10-27
Add /status endpoint

chmeliik:
* CLOUDBLD-2011

Shows status of Cachito workers and services that Cachito depends on.
Also add /status/short, which returns 200 if everything is OK, 503
otherwise.

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -22,11 +22,30 @@ from cachito.web.models import (
     RequestState,
     RequestStateMapping,
 )
+from cachito.web.status import status
 from cachito.web.utils import pagination_metadata, str_to_bool
 from cachito.workers import tasks
 from cachito.paths import RequestBundleDir
 
 api_v1 = flask.Blueprint("api_v1", __name__)
+
+
+@api_v1.route("/status", methods=["GET"])
+def get_status():
+    """Return status of Cachito workers and services that Cachito depends on."""
+    return flask.jsonify(status())
+
+
+@api_v1.route("/status/short", methods=["GET"])
+def get_status_short():
+    """Return 200 if all workers and services seem to be ok, 503 otherwise."""
+    try:
+        status(short=True)
+        retval = {"ok": True}
+    except CachitoError as e:
+        retval = {"ok": False, "reason": str(e)}
+
+    return flask.jsonify(retval), 200 if retval["ok"] else 503
 
 
 @api_v1.route("/requests", methods=["GET"])
--- a/cachito/web/status.py
+++ b/cachito/web/status.py
@@ -1,0 +1,232 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+import time
+
+import flask
+import kombu
+import requests
+import sqlalchemy.exc
+
+from cachito.errors import CachitoError
+from cachito.web import db
+from cachito.workers.config import get_worker_config
+from cachito.workers.tasks.celery import app
+
+ATHENS = "athens"
+NEXUS = "nexus"
+NEXUS_HOSTER = "nexus-hoster"
+DATABASE = "database"
+RABBITMQ = "rabbitmq"
+
+PKG_MANAGER_REQUIRES = {
+    "gomod": {ATHENS},
+    "npm": {NEXUS, NEXUS_HOSTER},
+    "pip": {NEXUS, NEXUS_HOSTER},
+    "*": {DATABASE, RABBITMQ},
+}
+
+no_retry_session = requests.Session()
+
+
+def _service_ok(url):
+    """
+    Check if a service is reachable and working correctly.
+
+    The status of the service is determined from the status code of the response.
+
+    :param str url: url to fetch
+    :return: tuple (ok: bool, reason: str or None)
+    """
+    try:
+        resp = no_retry_session.get(url)
+    except requests.RequestException:
+        return False, "connection failed"
+
+    if not resp.ok:
+        return False, f"{resp.status_code}: {resp.reason}"
+
+    return True, None
+
+
+def nexus_ok(nexus_url):
+    """
+    Check if Nexus is OK.
+
+    Uses the /status endpoint, specifically /status/writable.
+
+    :param str nexus_url: url of the Nexus host
+    :return: tuple (ok: bool, reason: str or None)
+    """
+    nexus_status_url = f"{nexus_url.rstrip('/')}/service/rest/v1/status/writable"
+    return _service_ok(nexus_status_url)
+
+
+def athens_ok(athens_url):
+    """
+    Check if Athens is OK.
+
+    Simply checks if the root url is reachable (Athens has no status endpoint).
+
+    :param str athens_url: url of the Athens host
+    :return: tuple (ok: bool, reason: str or None)
+    """
+    return _service_ok(athens_url)
+
+
+def database_ok():
+    """
+    Check if the database connection is working.
+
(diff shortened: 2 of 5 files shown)


chmeliik:
Don't let the 550 lines discourage you, 300 of those are unit tests :-)

Note that this will always expose the hostnames (or the RabbitMQ equivalent thereof) of all the Cachito workers, and potentially of other services as well when there is a problem. I don't think that's an issue, but if you can think of a reason why it would be an issue, I can make the output less detailed.

cachito/web/status.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    nexus_status_url = f"{nexus_url.rstrip('/')}/service/rest/v1/status/writable"
+    return _service_ok(nexus_status_url)
+
+
+def athens_ok(athens_url):
+    """
+    Check if Athens is OK.
+
+    Simply checks if the root url is reachable (Athens has no status endpoint).
```
athos-ribeiro, line 50: Note for future selves: This may be useful in the future https://github.com/gomods/athens/pull/1658

chmeliik:
Found an issue where the worker inspection takes a long time (and then fails) if you kill RabbitMQ, will add separate check for RabbitMQ

cachito/web/status.py
```diff
@@ -0,0 +1,142 @@
...
+
+    for worker_name, reply in reply_tuples:
+        # The reply format does not seem to be documented anywhere, but is
+        # likely a dict with one key: "ok"/"error" and a corresponding message.
+        worker = {"name": worker_name, "ok": "ok" in reply}
+
+        if "ok" not in reply:
+            reason = reply.get("error", "unknown reason")
+            if short:
+                raise CachitoError(f"worker {worker_name} unavailable: {reason}")
```
athos-ribeiro, line 115: So, if 1 out of 12 workers is unavailable here, we'd respond with an error, but if 11 workers do not respond to the ping above and one do, we'd say everything is ok. Should we onlyreturn 503 if cachito is actually unable to process requests at the moment?

Would adding more information in a 200 response contents help here? Something like "all_systems_go: true/false"?
This of course would only be really truthful if we also included some sort of configuration to let Cachito know the number of workers expected to exist in a certain deployment (this last part may require further discussion).

athos-ribeiro: In this matter, should we also consider that Cachito may be partially unavailable? For instance, if Athens is unavailable, Cachito is still operational for python and npm requests.

athos-ribeiro (commented):
Looks great!

I added some comments to start some discussions on a couple of topics regarding the new endpoint. I'd be OK with merging this as is and improving the endpoint as the questions and expectations become clearer though.

athos-ribeiro:
> Don't let the 550 lines discourage you, 300 of those are unit tests :-)
> 
> Note that this will always expose the hostnames (or the RabbitMQ equivalent thereof) of all the Cachito workers, and potentially of other services as well when there is a problem. I don't think that's an issue, but if you can think of a reason why it would be an issue, I can make the output less detailed.

Could some of these URLs, e.g., the broker url, include credentials, such as `USERNAME:PASSWORD@HOSTNAME` ?

chmeliik:
> Could some of these URLs, e.g., the broker url, include credentials, such as `USERNAME:PASSWORD@HOSTNAME` ?

Ahh, that is possible, e.g.
```json
{
  "name": "nexus", 
  "ok": false, 
  "reason": "401 Client Error: Unauthorized for url: http://user:pass@nexus:8081/service/rest/v1/status/writable"
}
```

But the only url which contains credentials in our config is the broker one, for that one the error messages do not seem to expose credentials:
```json
{
  "name": "rabbitmq", 
  "ok": false, 
  "reason": "failed to resolve broker hostname"
}
```
```json
{
  "name": "rabbitmq", 
  "ok": false, 
  "reason": "[Errno 111] Connection refused"
}
```
Using invalid credentials in the broker url results in an uncaught exception:
```
amqp.exceptions.AccessRefused: (0, 0): (403) ACCESS_REFUSED - Login was refused using authentication mechanism AMQPLAIN. For details see the broker logfile.
```


But I cannot guarantee that credentials will never be exposed. I see a few options here:
- do not include any error messages, just `"ok": true/false`
- include custom error messages
- censor urls in error messages

athos-ribeiro:
If this is really unknown, I'd rather include custom error messages when needed or censoring the authentication bits of any URLs if they are present. WDYT?

chmeliik:
> If this is really unknown, I'd rather include custom error messages when needed or censoring the authentication bits of any URLs if they are present. WDYT?

Yeah I'm thinking the same thing, currently going with custom, less informative messages

chmeliik:
Updated error messages, using custom format for all errors (better safe then sorry, right?)

cachito/web/status.py
chmeliik: Hmm, all good points. I think I'll replace the `all_ok` attribute with something like `can_serve`, which will be True if all services and at least one worker are ok.

If we decide to add the configuration for expected number of workers, I'll also add `all_ok`.

I like the idea of partial availability, but I imagine that would best be left for another PR

chmeliik:
Found another issue where the `ping()` method would sometimes inexplicably raise a `ConnectionResetError`, wrapped it in a try/except

cachito/web/api_v1.py
```diff
@@ -22,13 +22,36 @@
...
+        status(short=True)
+        content = ""
+        status_code = 200
+    except CachitoError as e:
+        content = str(e)
+        status_code = 503
+
+    response = make_response(content, status_code)
+    response.content_type = "text/plain"
+    return response
```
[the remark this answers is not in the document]
chmeliik: OK, done

cachito/web/status.py
chmeliik: Done, including partial availability (only for /status, not /status/short)

chmeliik:
Added a few thing:
- custom retries on `ConnectionError` when pinging workers (sometimes, the `ping()` function raises `ConnectionResetError` even when everything is working fine)
- per-package-manager availability info in `/status` response
- only fail if no workers are available, rather than if any workers are unavailable

chmeliik:
Minor unit test change (merged unnecessary tests into one)

tests/test_status.py
```diff
@@ -0,0 +1,454 @@
...
+        expect_reason = "503: Service Unavailable"
+    else:
+        expect_reason = None
+
+    ok, reason = status._service_ok(url)
+
+    if fail_reason:
+        assert not ok
+        assert reason == expect_reason
+        assert url not in reason
```
athos-ribeiro, line 44: :rocket:

tests/test_status.py
```diff
@@ -0,0 +1,454 @@
...
+            {"gomod": False, "npm": False, "pip": False, "git-submodule": True},
+        ),
+    ],
+)
+def test_can_process(worker_ok, failing_services, expect_result):
+    services = [{"name": service, "ok": False} for service in failing_services]
+    assert status._can_process(TEST_PACKAGE_MANAGERS, services, worker_ok) == expect_result
+
+
+def mock_worker_config(nexus_hoster=False):
```
athos-ribeiro, line 217: Nitpick: could we move this function to the beginning of the file?

athos-ribeiro (approved):
Great work!!!

Added one comment for moving a mocking function to the beginning of a module. This should not be blocker. Feel free to merge with or without addressing the comment :)

tests/test_status.py
chmeliik: Done

PR #326, opened 2020-10-27
Integration tests: add implicit gomod test

ejegrova:
Added test that uses only repo and ref as payload for request- Package manager is implicitly gomod.

### DIFF
--- a/tests/integration/test_data/gomod_packages.yaml
+++ b/tests/integration/test_data/gomod_packages.yaml
@@ -202,3 +202,87 @@ vendored_without_flag:
   repo: https://github.com/cachito-testing/gomod-vendored.git
   ref:  ff1960095dd158d3d2a4f31d15b244c24930248b
   pkg_managers: ["gomod"]
+# The test data for implicit gomod package manager test
+implicit_gomod:
+  repo: https://github.com/cachito-testing/cachito-gomod-with-deps.git
+  ref: 4c65d49cae6bfbada4d479b321d8c0109fa1aa97
+  response_expectations:
+    packages:
+      - dependencies:
+          - name: "rsc.io/sampler"
+            replaces: null
+            type: "gomod"
+            version: "v1.3.0"
+          - name: "rsc.io/quote"
+            replaces: null
+            type: "gomod"
+            version: "v1.5.2"
+          - name: "golang.org/x/text"
+            replaces: null
+            type: "gomod"
+            version: "v0.0.0-20170915032832-14c0d48ead0c"
+        name: "github.com/cachito-testing/cachito-gomod-with-deps"
+        type: "gomod"
+        version: "v0.0.0-20201001160613-4c65d49cae6b"
+      - dependencies:
+          - name: "rsc.io/quote"
+            replaces: null
+            type: "go-package"
+            version: "v1.5.2"
+          - name: "rsc.io/sampler"
+            replaces: null
+            type: "go-package"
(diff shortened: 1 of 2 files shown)


tests/integration/test_packages.py
```diff
@@ -34,14 +35,21 @@ def test_packages(env_package, env_name, test_env, tmpdir):
...
+    if env_name == "implicit_gomod":
+        payload = {
+            "repo": env_data["repo"],
+            "ref": env_data["ref"],
+        }
+    else:
+        payload = {
             "repo": env_data["repo"],
             "ref": env_data["ref"],
             "pkg_managers": env_data["pkg_managers"],
```
athos-ribeiro, line 48: How about just changing this line to 

`"pkg_managers": env_data.get("pkg_managers", []),`

and adding the following after the payload assignment:

```
if env_name == "implicit_gomod":
    payload.pop(pkg_managers)
```

tests/integration/test_packages.py
```diff
@@ -34,14 +35,21 @@ def test_packages(env_package, env_name, test_env, tmpdir):
...
             "flags": env_data.get("flags", []),
```
athos-ribeiro, line 49: It would also be nice to change this one (in a different commit, could be a different PR.
We are always testing the request with flags. we should just add it later if it exists, like
```
if env_data.get("flags"):
    payload["flags"] = env_data.get["flags"]
```

and then adding a single test passing empty flags just to check Cachito's sanity.

athos-ribeiro (commented):
LGTM!

Just a small nitpick. Feel free to just ignore the comment about the `flags` we can file an issue for that later :)

tests/integration/test_packages.py
ejegrova: oh, I spend some time today trying to figure out how to simplify this but I didn't think of pop :D

tests/integration/test_data/gomod_packages.yaml
```diff
@@ -202,3 +202,86 @@ vendored_without_flag:
   repo: https://github.com/cachito-testing/gomod-vendored.git
   ref:  ff1960095dd158d3d2a4f31d15b244c24930248b
   pkg_managers: ["gomod"]
+implicit_gomod:
```
akhmanova, line 205: Can you please add a comment about the test data?

tests/integration/test_packages.py
akhmanova: Yes, a good idea.
Or like this one: 
```suggestion
           if "flags" in env_data:
               payload["flags"] = env_data["flags"]
```

ejegrova: I think we should file new issue for that and add new test to the testing doc.

tests/integration/test_data/gomod_packages.yaml
ejegrova: done :)

athos-ribeiro (changes requested):
Thanks for the changes!

There is a lint error now. Could you fix that?

athos-ribeiro (approved):

akhmanova:
@ejegrova Looks good, what is the purpose of empty commit?

ejegrova:
> @ejegrova Looks good, what is the purpose of empty commit?

sorry about that, wrong branch for testing Jenkins, it's removed now :)

athos-ribeiro (approved):
Good catch on the empty commit, @akhmanova

PR #328, opened 2020-10-30
Integration tests: Add meaningful messages to assertions

ejegrova:
Provide error messages to assertions so there is a better understanding
what went wrong.

### DIFF
--- a/tests/integration/test_gomod_packages.py
+++ b/tests/integration/test_gomod_packages.py
@@ -26,6 +26,9 @@ def test_gomod_vendor_without_flag(test_env):
         error_msg = (
             'The "gomod-vendor" flag must be set when your repository has vendored dependencies'
         )
-        assert error_msg in completed_response.data["state_reason"]
+        assert error_msg in completed_response.data["state_reason"], (
+            f"#{completed_response.id}: Request failed correctly, but with unexpected message: "
+            f"{completed_response.data['state_reason']}. Expected message was: {error_msg}"
+        )
     else:
         utils.assert_properly_completed_response(completed_response)
--- a/tests/integration/test_http_get_all.py
+++ b/tests/integration/test_http_get_all.py
@@ -74,12 +74,20 @@ def assert_no_requests_in_progress_state(client: utils.Client, requests_amount, 
     response = client.fetch_all_requests(query_params, all_pages=False)
     assert response.status == 200
 
-    assert all(item["state"] == "in_progress" for item in response.data["items"])
+    assert all(
+        item["state"] == "in_progress" for item in response.data["items"]
+    ), f"At least one of requests was not in in_progress state"
 
     response_ids = set(item["id"] for item in response.data["items"])
     # Check there are no completed requests in the response.
-    assert not any(request_id in response_ids for request_id in submitted_requests)
-    assert 0 <= len(response.data["items"]) <= requests_amount
+    assert not any(request_id in response_ids for request_id in submitted_requests), (
+        f"At least one of requests {response_ids} was in completed state. ",
+        f"All requests are expected to be in other state than completed.",
+    )
+    assert 0 <= len(response.data["items"]) <= requests_amount, (
+        f"Number of obtained responses ({len(response.data['items'])}) "
+        f"is not in the range from 0 to {requests_amount}"
+    )
 
 
 def assert_completed_requests(client: utils.Client, submitted_ids):
--- a/tests/integration/test_valid_data_in_request.py
+++ b/tests/integration/test_valid_data_in_request.py
@@ -62,9 +62,15 @@ def test_npm_basic(test_env, default_requests):
 
     for item in response.data["dependencies"]:
         if item["name"] not in test_env["get"]["npm"]["non_dev_dependencies"]:
-            assert item["dev"]
+            assert item["dev"], (
+                f"dependency {item['name']} is not listed as non-dev in the test expectations,"
+                f"but is listed as non-dev in the response"
+            )
         else:
-            assert not item["dev"]
+            assert not item["dev"], (
+                f"dependency {item['name']} is listed as non-dev in the test expectations,"
+                f"but is not listed as non-dev in the response"
+            )
 
     env_variables = response.data["environment_variables"]
     assert env_variables["CHROMEDRIVER_SKIP_DOWNLOAD"] == "true"
--- a/tests/integration/utils.py
+++ b/tests/integration/utils.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 from collections import namedtuple
 import filecmp
+import json
 import os
 import shutil
 import time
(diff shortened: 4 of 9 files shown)


tests/integration/test_http_get_all.py
```diff
@@ -78,7 +78,9 @@ def assert_no_requests_in_progress_state(client: utils.Client, requests_amount,
 
     response_ids = set(item["id"] for item in response.data["items"])
     # Check there are no completed requests in the response.
-    assert not any(request_id in response_ids for request_id in submitted_requests)
+    assert not any(
+        request_id in response_ids for request_id in submitted_requests
+    ), f"At least one of requests {response_ids} was in completed state"
```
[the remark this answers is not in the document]
ejegrova: Is it expected to fail? I thought that all of those requests has to be in other that completed state for assertion to pass. No matter what states they're in.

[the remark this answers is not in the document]
ejegrova: Maybe change it like this:
```f"At least one of requests {response_ids} was in completed state. All requests are expected to be in other state than completed."```

File: tests/integration/test_gomod_packages.py [hunk 2]
```diff
@@ -26,6 +26,9 @@ def test_gomod_vendor_without_flag(test_env):
         error_msg = (
             'The "gomod-vendor" flag must be set when your repository has vendored dependencies'
         )
-        assert error_msg in completed_response.data["state_reason"]
+        assert error_msg in completed_response.data["state_reason"], (
+            f"#{completed_response.id}: Request failed correctly, but with unexpected message: "
+            f"{completed_response.data['state_reason']}"
```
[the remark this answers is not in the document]
ejegrova: added it

tests/integration/test_valid_data_in_request.py
```diff
@@ -62,9 +62,15 @@ def test_npm_basic(test_env, default_requests):
...
+            assert item["dev"], (
+                f"#{response.id}: dependency {item['name']} is not a dev dependency,"
+                f" but it is not listed as non-dev dependency"
+            )
         else:
-            assert not item["dev"]
+            assert not item["dev"], (
+                f"#{response.id}: dependency {item['name']} is listed as non-dev dependency,"
+                f" but it is a dev dependency"
+            )
```
chmeliik, line 73: IMHO these messages could be clearer about what is expected vs. actual, maybe something like `dependency {dep} should be listed as a dev dependency, but is listed as a non-dev dependency`

tests/integration/utils.py
```diff
@@ -50,6 +50,7 @@ def create_new_request(self, payload):
         :rtype: Response
         :raises requests.exceptions.HTTPError: if the request to the Cachito API fails
         """
+        print(payload)
```
chmeliik, line 53: This is intended, right? I would normally suggest to use logging for this instead, but I'd also be fine with adding a comment to explain that it is not just a forgotten debugging print :-)

ejegrova: that's definitely not intended, thanks for noticing :)

tests/integration/test_valid_data_in_request.py
ejegrova: It's hard for me to write these two messages so it makes sense for someone else. We only use `non_dev_dependencies` list in test data, so you're not listing dev dependencies anywhere. I'll try to come up with something better.

ejegrova: changed it, what do you think?

chmeliik: Hmm, I might have misunderstood what "listed" means. If something is listed as a non-dev dependency, it means that it is in the `non_dev_dependencies` list in the test data? The way the error messages are phrased now, it sounds like you should change the test expectations if there is an error :D

Maybe this would work?
`dependency {item['name']} is not listed as non-dev in the test expectations, but is listed as non-dev in the response`
`dependency {item['name']} is listed as non-dev in the test expectations, but is not listed as non-dev in the response`

tests/integration/test_http_get_all.py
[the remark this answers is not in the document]
akhmanova: Looks good :)

tests/integration/test_valid_data_in_request.py
ejegrova: oh, that's true :facepalm: Your suggestion looks great.

akhmanova:
Can we add messages also into: 
https://github.com/release-engineering/cachito/pull/328/files#diff-d9e46323e51c9f9cfa4c9fa1acc40a8cc931da3ba7f82f6867360507590cc671R275
and
https://github.com/release-engineering/cachito/pull/328/files#diff-d9e46323e51c9f9cfa4c9fa1acc40a8cc931da3ba7f82f6867360507590cc671L374
?

ejegrova:
> Can we add messages also into:
> https://github.com/release-engineering/cachito/pull/328/files#diff-d9e46323e51c9f9cfa4c9fa1acc40a8cc931da3ba7f82f6867360507590cc671R275
> and
> https://github.com/release-engineering/cachito/pull/328/files#diff-d9e46323e51c9f9cfa4c9fa1acc40a8cc931da3ba7f82f6867360507590cc671L374
> ?

These are comparing equality so the error should show us what is being compared here.

akhmanova:
> > Can we add messages also into:
> > https://github.com/release-engineering/cachito/pull/328/files#diff-d9e46323e51c9f9cfa4c9fa1acc40a8cc931da3ba7f82f6867360507590cc671R275
> > and
> > https://github.com/release-engineering/cachito/pull/328/files#diff-d9e46323e51c9f9cfa4c9fa1acc40a8cc931da3ba7f82f6867360507590cc671L374
> > ?
> 
> These are comparing equality so the error should show us what is being compared here.

The way assert prints it is not really helpful (It does print partly). Can we add `json.dumps(response, indent=4, sort_keys=True)`?

ejegrova:
I added messages to all assertions in assert functions (utils). For some reason if assert is not done in test file, it will show only something like this:
```
>       assert completed_response.status == 300
E       AssertionError
```

chmeliik (approved):
LGTM :+1:

tests/integration/utils.py
```diff
@@ -270,7 +273,12 @@ def assert_elements_from_response(response_data, expected_response_data):
     if "dependencies" in expected_response_data:
         sort_pkgs_and_deps_in_place(dependencies=expected_response_data["dependencies"])
     for element_name, expected_data in expected_response_data.items():
-        assert response_data[element_name] == expected_data
+        assert response_data[element_name] == expected_data, (
+            f"#{response_data['id']}: elements in reponse differs from test expactations. \n"
+            f"Response elements: "
+            f"{json.dumps(response_data[element_name], indent=4, sort_keys=True)}, \n"
```
akhmanova, line 279: Thank you for adding this!

akhmanova (approved):
LGTM :)

chmeliik:
@ejegrova can this be merged? I'm currently seeing a failure locally and this would be super helpful :D

ejegrova:
> @ejegrova can this be merged? I'm currently seeing a failure locally and this would be super helpful :D

sorry for that :)

PR #330, opened 2020-11-02
Add support for setuptools >= 49

athos-ribeiro:
setuptools 49 removed RequirementParseError in favor of
packaging.requirements.InvalidRequirement. The former name was kept for
backwards compatibility but InvalidRequirement should be used instead.

Not performing this change leads to breakage when using setuptools >=
49, which is shipped in Fedora 33.

Moreover, this change should not break deployments using setuptools <
49, since the InvalidRequirement error is not new in "packaging".

* https://setuptools.readthedocs.io/en/latest/history.html#id67

### DIFF
--- a/cachito/workers/pkg_managers/pip.py
+++ b/cachito/workers/pkg_managers/pip.py
@@ -1066,7 +1066,11 @@ class PipRequirement:
 
         try:
             parsed = list(pkg_resources.parse_requirements(to_be_parsed))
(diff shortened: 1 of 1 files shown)


nirzari (approved):

nirzari:
Although there is a lint failure. Need to break into multiple lines.

chmeliik (approved):
LGTM :+1:

PR #333, opened 2020-11-06
Add mechanism to make package managers mutually exclusive

chmeliik:
* CLOUDBLD-3309

Also add a mechanism for making any pair of package managers mutually
exclusive.

### DIFF
--- a/cachito/web/config.py
+++ b/cachito/web/config.py
@@ -18,7 +18,7 @@ class Config(object):
     CACHITO_LOG_FORMAT = "[%(asctime)s %(name)s %(levelname)s %(module)s.%(funcName)s] %(message)s"
     CACHITO_MAX_PER_PAGE = 100
     # Pairs of mutually exclusive package managers (cannot process the same package)
-    CACHITO_MUTUALLY_EXCLUSIVE_PACKAGE_MANAGERS = []
+    CACHITO_MUTUALLY_EXCLUSIVE_PACKAGE_MANAGERS = [("npm", "yarn")]
     CACHITO_PACKAGE_MANAGERS = ["gomod"]
     CACHITO_REQUEST_FILE_LOGS_DIR = None
     # Users that are allowed to use the "user" property when creating a request
--- a/cachito/web/config.py
+++ b/cachito/web/config.py
@@ -37,6 +37,7 @@ class DevelopmentConfig(Config):
 
     CACHITO_BUNDLES_DIR = os.path.join(tempfile.gettempdir(), "cachito-archives", "bundles")
     CACHITO_LOG_LEVEL = "DEBUG"
+    CACHITO_MUTUALLY_EXCLUSIVE_PACKAGE_MANAGERS = [("npm", "yarn"), ("gomod", "git-submodule")]
     CACHITO_PACKAGE_MANAGERS = ["gomod", "npm", "pip", "git-submodule"]
     CACHITO_REQUEST_FILE_LOGS_DIR = "/var/log/cachito/requests"
     SQLALCHEMY_DATABASE_URI = "postgresql+psycopg2://cachito:cachito@db:5432/cachito"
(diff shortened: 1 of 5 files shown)


cachito/web/models.py
```diff
@@ -948,6 +1015,12 @@ class PackageManager(db.Model):
     id = db.Column(db.Integer, primary_key=True)
     name = db.Column(db.String, nullable=False)
 
+    # Pairs of mutually exclusive package managers (locally - for the same subpath)
+    # This relation is automatically symmetric - (A, B) implies (B, A)
+    LOCALLY_EXCLUSIVE = [
+        ("npm", "yarn"),
+    ]
```
chmeliik, line 1022: Not sure if this class is the right place to declare this, WDYT?

cachito/web/models.py
```diff
@@ -566,6 +566,73 @@ def _validate_request_package_configs(request_kwargs, pkg_managers_names):
...
+        a, b = pkg_manager, "git-submodule"
+        if not ((a, b) in mutually_exclusive or (b, a) in mutually_exclusive):
+            continue
+
+        if any(path != "." for path in paths):
+            msg = (
+                f"Cannot process non-root packages with '{pkg_manager}' "
+                "when 'git-submodule' is also set"
+            )
+            raise ValidationError(msg)
```
chmeliik, line 634: Do we want this? Could come in handy if we decide to disallow gomod+gitsubmodule after #305, but maybe this check should instead be done by a worker, when we know that the subpath is actually a git submodule

cc @athos-ribeiro @arewm

arewm: Even though it would make processing more difficult, I think that the biggest issue between git submodules and other package managers is when the paths coincide. If that is deemed too hard to implement, however, this seems like a reasonable approximation.

cachito/web/models.py
nirzari: Out of all we have currently, this seems a reasonable location as it provides some context.

nirzari (approved):
LGTM

cachito/web/models.py
athos-ribeiro: +1. We can start with this and  pursue an algorithmic approach  based on how users interact with the feature.

cachito/web/models.py
athos-ribeiro: Should this be in the config instead?

This way it would be easier to toggle the validation for git-submodules in the future. i.e, introduce the validation and remove it if need. So deployments won't need a hotfix.

athos-ribeiro (commented):
LGTM!

Just a nitpick about the location of the exclusivity configuration

cachito/web/models.py
chmeliik: In my mind, there was no reason to change exclusivity settings without a corresponding code change, but having it in config would be more flexible. I'll change this.

chmeliik:
Rebased on master, made exclusive package managers configurable, added gomod+gitsubmodule

chmeliik:
Now also accepting 2-item lists as package manager pairs (ansible+yaml+jinja makes things awkward)

cachito/web/config.py
```diff
@@ -17,6 +17,8 @@ class Config(object):
     CACHITO_LOG_LEVEL = "INFO"
     CACHITO_LOG_FORMAT = "[%(asctime)s %(name)s %(levelname)s %(module)s.%(funcName)s] %(message)s"
     CACHITO_MAX_PER_PAGE = 100
+    # Pairs of mutually exclusive package managers (cannot process the same package)
+    CACHITO_MUTUALLY_EXCLUSIVE_PACKAGE_MANAGERS = [("npm", "yarn")]
```
athos-ribeiro, line 21: Should we add some validation for this configuration to ensure the inner tuple/lists always have 2 elements?

chmeliik: Good point, added

arewm: Do you want `("gomod", "git-submodule")` here? It seems like it would be more important to have the mutually exclusive pair on the default class than it would on the `DevelopmentConfig`.

chmeliik: @arewm this is intentional, to differentiate between `("npm", "yarn")`, which will be exclusive by design and `("gomod", "git-submodule")`, which are exclusive until we implement support. In the deployment playbooks, this will be configured properly

athos-ribeiro (approved):

PR #337, opened 2020-11-13
Add API handler for yarn requests

ejegrova:
Add API handler for yarn requests, request to Cachito will return a 501 error (NotImplementedError).

### DIFF
--- a/cachito/errors.py
+++ b/cachito/errors.py
@@ -15,3 +15,7 @@ class ConfigError(CachitoError):
 
 class ContentManifestError(CachitoError, ValueError):
     """An error was encountered during content manifest generation."""
+
+
+class CachitoNotImplementedError(CachitoError, ValueError):
+    """An error was encountered during request validation."""
--- a/tests/test_api_v1.py
+++ b/tests/test_api_v1.py
@@ -2161,3 +2161,15 @@ def test_validate_package_manager_exclusivity(
             _validate_package_manager_exclusivity(pkg_managers, package_configs, mutually_exclusive)
     else:
         _validate_package_manager_exclusivity(pkg_managers, package_configs, mutually_exclusive)
+
+
+def test_create_request_with_yarn(auth_env, client):
+    data = {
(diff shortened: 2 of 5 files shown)


tests/test_api_v1.py
```diff
@@ -2161,3 +2161,14 @@ def test_validate_package_manager_exclusivity(
...
+
+def test_create_request_with_yarn(auth_env, client):
+    data = {
+        "repo": "https://github.com/seriousManual/dedupe",
+        "ref": "955aa2f0d2dedf1b04814e38ad80deb17a602b9c",
+        "pkg_managers": ["yarn"],
+    }
+
+    rv = client.post("/api/v1/requests", json=data, environ_base=auth_env)
+    assert rv.status_code == 501
```
akhmanova, line 2174: It could be nice to check the message as well. Otherwise looks good to me

akhmanova (changes requested):

cachito/errors.py
```diff
@@ -15,3 +15,7 @@ class ConfigError(CachitoError):
 
 class ContentManifestError(CachitoError, ValueError):
     """An error was encountered during content manifest generation."""
+
+
+class NotImplementedError(CachitoError, ValueError):
```
chmeliik, line 20: I think `NotImplementedError` is a Python built-in, can you rename this to avoid confusion?

tests/test_api_v1.py
ejegrova: added

cachito/errors.py
chmeliik: hmm, that sounds a bit weird to me, maybe `CachitoNotImplementedError`? Or just `CachitoNotImplemented`?

ejegrova: changed that to `CachitoNotImplementedError`, thanks :)

chmeliik (approved):
LGTM :+1:

akhmanova (approved):
LGTM!

PR #338, opened 2020-11-13
Set the GOMODCACHE environment variable

arewm:
GOMODCACHE was introduced in go 1.15. We should make sure to set it to
make sure that the cache can be found.

The default value for GOMODCACHE is GOPATH[0]/pkg/mod according to https://golang.org/doc/go1.15

Signed-off-by: arewm <arewm@users.noreply.github.com>


lcarva (approved):
Nice!

athos-ribeiro (approved):

PR #341, opened 2020-11-19
Add pyarn as a dependency

chmeliik:
- reorganize `setup.py`, `tox.ini` and `Makefile` to disconnect `setup.py` from `requirements.txt`
- add `pyarn` as a dependency to `setup.py`, `requirements.txt` and Dockerfiles
- add function and unit test as a sanity-check that it works


nirzari (approved):
Looks Great

PR #344, opened 2020-11-23
Perform git gc in Cachito archives

ejegrova:


chmeliik (commented):
Looks good, @athos-ribeiro was also mentioning we might want to consider the [prune](https://git-scm.com/docs/git-gc#Documentation/git-gc.txt---pruneltdategt) and [aggressive](https://git-scm.com/docs/git-gc#_aggressive) options. The git documentation does not really recommend *aggressive*, but *prune=now* would make sense to me (and it might be worth it to test both in one of those huge repos).

ejegrova:
> Looks good, @athos-ribeiro was also mentioning we might want to consider the [prune](https://git-scm.com/docs/git-gc#Documentation/git-gc.txt---pruneltdategt) and [aggressive](https://git-scm.com/docs/git-gc#_aggressive) options. The git documentation does not really recommend _aggressive_, but _prune=now_ would make sense to me (and it might be worth it to test both in one of those huge repos).
I already tried aggressive mode with small archive and it took 6.5s, so I wouldn't use that. In manual, it says that ` --prune is on by default` but it uses value: two weeks ago. So I'll try it with now.

ejegrova:
> > Looks good, @athos-ribeiro was also mentioning we might want to consider the [prune](https://git-scm.com/docs/git-gc#Documentation/git-gc.txt---pruneltdategt) and [aggressive](https://git-scm.com/docs/git-gc#_aggressive) options. The git documentation does not really recommend _aggressive_, but _prune=now_ would make sense to me (and it might be worth it to test both in one of those huge repos).
> 
> I already tried aggressive mode with small archive and it took 6.5s, so I wouldn't use that. In manual, it says that ` --prune is on by default` but it uses value: two weeks ago. So I'll try it with now.

With or without using `--prune=now`, the result was the same on small archive and also with the big archive in test env. However big archive is from Nov 11, so it could have no effect.

chmeliik:
> With or without using `--prune=now`, the result was the same on small archive and also with the big archive in test env. However big archive is from Nov 11, so it could have no effect.

I see, we could use it anyway, I don't think there are any downsides, The increased risk of corruption mentioned in git docs only applies if two or more process are writing to the same repo concurrently, that is *almost* certainly not the case in Cachito.

@athos-ribeiro @lcarva WDYT?

athos-ribeiro:
> > With or without using `--prune=now`, the result was the same on small archive and also with the big archive in test env. However big archive is from Nov 11, so it could have no effect.
> 
> I see, we could use it anyway, I don't think there are any downsides, The increased risk of corruption mentioned in git docs only applies if two or more process are writing to the same repo concurrently, that is _almost_ certainly not the case in Cachito.
> 
> @athos-ribeiro @lcarva WDYT?

+1

FWIW, if different tasks are handling the same git repository simultaneously, it should be treated as a bug anyway.

athos-ribeiro (commented):
Do we want a small unit test to make sure `git.gc` is being called?

chmeliik (approved):
LGTM :+1:

PR #346, opened 2020-11-27
Resolve yarn dependencies

chmeliik:
* CLOUDBLD-3311
    
Like all the resolve_{pkg_manager} functions, this one downloads
dependencies for a yarn package and returns useful information.


chmeliik:
PTAL
Still in a bunch of TODOs in there, I'm going to need your opinions on those.

You can test with this branch: https://github.com/chmeliik/cachito/tree/fetch-yarn
Using this request:
```shell
    "repo": "https://github.com/chmeliik/cachito-npm-test",
    "ref": "76d6a658f99be99ef6da9455405d874d2d546069",
    "pkg_managers": ["yarn"]
}'
```

chmeliik:
Notes for reviewers:

- this tries to match the npm code structure as close as possible, most helper functions do the same thing as their npm equivalents
- I cannot guarantee that this handles all cases or that it handles them correctly. Yarn documentation is typically too high-level or non-existent. Yarn source code is a mess. It should be mostly consistent with the npm handling though.

cachito/workers/pkg_managers/yarn.py
```diff
@@ -55,3 +64,289 @@ def get_yarn_proxy_repo_username(request_id):
...
+    _sanity_check(app_source_path)
+
+    package_json_path = app_source_path / "package.json"
+    yarn_lock_path = app_source_path / "yarn.lock"
+    package_and_deps_info = _get_package_and_deps(package_json_path, yarn_lock_path)
+
+    # By downloading the dependencies, it stores the tarballs in the bundle and also stages the
+    # content in the npm repository for the request
+    proxy_repo_url = get_yarn_proxy_repo_url(request["id"])
+    package_and_deps_info["downloaded_deps"] = download_dependencies(
```
chmeliik, line 343: I just realized that `download_dependencies` always download to `deps/npm`, we probably want to change that right?

athos-ribeiro: Yes, it would be nice to pass a parameter to it so it can download data to `deps/yarn` upon request

chmeliik: Ok, fixed

cachito/workers/pkg_managers/general_js.py
```diff
@@ -457,3 +460,130 @@ def upload_non_registry_dependency(
...
         nexus.upload_asset_only_component(repo_name, "npm", modified_dep_archive)
+
+
+@dataclass(frozen=True)
+class JSDependency:
+    """Holds package-manager-independent data about a JavaScript dependency."""
+
+    # The name of the dependency
+    # In package-lock.json these are the keys in the "dependencies" object
+    # In yarn.lock, these are the top-level keys in the file (the keys are <name>@<version>)
```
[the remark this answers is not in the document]
chmeliik: Hmm, not really, I just kind of happened to do it like this :D
Would it make more sense in the docstring?

cachito/workers/pkg_managers/general_js.py
```diff
@@ -457,3 +460,130 @@ def upload_non_registry_dependency(
...
+        # When the dependency is uploaded to the Nexus hosted repository, it will be in the format
+        # of `<version>-gitcommit-<commit hash>`
+        version_suffix = f"-external-gitcommit-{commit_hash}"
+        # Dangerous scripts might be required to be executed by `npm pack` since this is a Git
+        # dependency. If those scripts are present, Cachito will fail the request since it will not
+        # execute those scripts when packing the dependency.
+        verify_scripts = True
+    elif any(js_dep.source.startswith(prefix) for prefix in http_prefixes):
+        if not js_dep.integrity:
+            msg = f"The dependency {js_dep.source} is missing the integrity value"
```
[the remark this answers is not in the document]
chmeliik: Should this also reflect if the dependency is coming from package-lock.json or yarn.lock? Because yarn.lock has another possible source for integrity. I kind of wanted to avoid that with `JSDependency`.
Would something like this be good enough?
```
Is the "integrity" key missing in your lockfile?
```
(completely ignoring the fact that in yarn.lock, integrity typically comes from urls)

cachito/workers/pkg_managers/general_js.py
```diff
@@ -457,3 +460,130 @@ def upload_non_registry_dependency(
...
+            log.error(msg)
+            raise CachitoError(msg)
+
+        checksum_info = convert_integrity_to_hex_checksum(js_dep.integrity)
+        # When the dependency is uploaded to the Nexus hosted repository, it will be in the format
+        # of `<version>-external-<checksum algorithm>-<hex checksum>`
+        version_suffix = f"-external-{checksum_info.algorithm}-{checksum_info.hexdigest}"
+    else:
+        raise CachitoError(f"The dependency {js_dep.source} is hosted in an unsupported location")
```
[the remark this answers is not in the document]
chmeliik: **repr** for dataclasses is pretty nice, but I'm not sure if it's nice enough for user-facing errors :thinking: 
Maybe `<name>@<source>` instead?

cachito/workers/pkg_managers/general_js.py
```diff
@@ -457,3 +460,130 @@ def upload_non_registry_dependency(
...
+        "gitlab:",
+    }
+    http_prefixes = {"http://", "https://"}
+    verify_scripts = False
+    checksum_info = None
+    if any(js_dep.source.startswith(prefix) for prefix in git_prefixes):
+        try:
+            _, commit_hash = js_dep.source.rsplit("#", 1)
+        except ValueError:
+            msg = f"The dependency {js_dep.source} was in an unexpected format"
```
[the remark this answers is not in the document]
chmeliik: Added

chmeliik:
Rebased on master, bumping for reviews :)

cachito/workers/pkg_managers/general_js.py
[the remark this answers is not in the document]
athos-ribeiro: Nah, this looks good! :)

cachito/workers/pkg_managers/general_js.py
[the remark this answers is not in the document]
athos-ribeiro: Sounds good!

cachito/workers/pkg_managers/general_js.py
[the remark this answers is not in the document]
athos-ribeiro: Sounds good!

chmeliik:
Improved error messages for non-registry deps (I made a separate commit for it) and removed the hex_sha512 function

cachito/workers/pkg_managers/yarn.py
```diff
@@ -55,3 +64,293 @@ def get_yarn_proxy_repo_username(request_id):
...
+            # dependencies later)
+            if package.relpath and normpath(package.relpath) in file_deps_allowlist:
+                log.info("The dependency %r is an allowed exception", package.name)
+            else:
+                nexus_replacement = _convert_to_nexus_hosted(package.name, source, dep_data)
+
+        dep = {
+            "bundled": False,  # yarn.lock does not seem to contain bundled deps at all
+            "dev": False,  # yarn.lock does not state whether a dependency is dev
+            # TODO: ^ this can be solved by cross-referencing with package.json
```
athos-ribeiro, line 117: It does not :/

We could either try to infer it from the package.json file or just ignore it. This looks good for now, but it would be nice to track this somehow so we can address it in the future.

cachito/workers/pkg_managers/yarn.py
```diff
@@ -55,3 +64,293 @@ def get_yarn_proxy_repo_username(request_id):
...
+        # The dependency identifier is a line that may contain multiple comma-separated
+        # dependencies (different ways to specify the same dependency)
+        deps = map(str.strip, deps_identifier.split(","))
+        for d in deps:
+            expanded_replacements[d] = nexus_replacement
+
+    package_json_new = copy.deepcopy(package_json)
+    modified = False
+
+    # TODO: optionalDependencies also exist, ignored for npm, safe to ignore for yarn?
```
athos-ribeiro, line 250: I believe it would be ok to track this one somehow and address it in the future when/if needed.

@arewm, do you need optional dependencies in any of your projects?

cachito/workers/pkg_managers/yarn.py
```diff
@@ -55,3 +64,293 @@ def get_yarn_proxy_repo_username(request_id):
...
+    :param str dep_source: the source (url or relative path) of the dependency
+    :param dict dep_info: the dependency info from the yarn lock file
+    :return: the dependency information of the Nexus hosted version to use in the yarn lock file
+        instead of the original
+    :raise CachitoError: if the dependency is from an unsupported location or has an unexpected
+        format in the lock file
+    """
+    integrity = dep_info.get("integrity")
+    # TODO: integrity may be a space-separated list of integrity values,
+    #   we ignore the possibility for npm, safe to ignore for yarn as well?
```
athos-ribeiro, line 155: In what cases would yarn v1 generate a separated list here?

cachito/workers/pkg_managers/yarn.py
```diff
@@ -55,3 +64,293 @@ def get_yarn_proxy_repo_username(request_id):
...
+    try:
+        package = {
+            "name": package_json["name"],
+            "version": package_json["version"],
+            "type": "yarn",
+        }
+    except KeyError:
+        raise CachitoError("The package.json file is missing required data (name, version)")
+
+    # TODO: use a different allowlist for yarn? rename allowlist variable?
```
athos-ribeiro, line 207: Good catch. @lcarva do we want to add support for this here now?

If so, let's add a new allowlist for this one.

cachito/workers/pkg_managers/yarn.py
```diff
@@ -55,3 +64,293 @@ def get_yarn_proxy_repo_username(request_id):
...
+
+
+def _sanity_check(app_source_path):
+    """
+    Sanity-check the application directory.
+
+    - package.json must be present
+    - yarn.lock must be present
+    - package-lock.json must not be present
+    # TODO: npm-shrinkwrap.json?
```
athos-ribeiro, line 293: It seems this one should not be present either.

cachito/workers/pkg_managers/yarn.py
arewm: We do not currently have any yarn projects, but we might need to start building grafana in the near future (which uses yarn). I haven't looked at what their build setup is yet.

cachito/workers/pkg_managers/yarn.py
chmeliik: `yarn install --update-checksums`

Here's how to reproduce:
```shell
cd /tmp
git clone https://github.com/chmeliik/cachito-npm-test.git
cd cachito-npm-test
git checkout yarn-test
podman run --rm -ti -v "$PWD:$PWD:z" -w "$PWD" fedora:33
# [inside the container]
curl --silent --location https://dl.yarnpkg.com/rpm/yarn.repo | sudo tee /etc/yum.repos.d/yarn.repo
dnf -y install yarn git-core
yarn install --update-checksums
git diff
```
Here's an excerpt from the diff:
```diff
diff --git a/yarn.lock b/yarn.lock
index 84b5502..012305f 100644
--- a/yarn.lock
+++ b/yarn.lock
@@ -5,12 +5,12 @@
 assertion-error@^1.1.0:
   version "1.1.0"
   resolved "https://registry.yarnpkg.com/assertion-error/-/assertion-error-1.1.0.tgz#e60b6b0e8f301bd97e5375215bda406c85118c0b"
-  integrity sha512-jgsaNduz+ndvGyFt3uSuWqvy4lCnIJiovtouQN5JZHOKCS2QuhEdbcQHFhVksz2N2U9hXJo8odG7ETyWlEeuDw==
+  integrity "sha1-5gtrDo8wG9l+U3UhW9pAbIURjAs= sha512-jgsaNduz+ndvGyFt3uSuWqvy4lCnIJiovtouQN5JZHOKCS2QuhEdbcQHFhVksz2N2U9hXJo8odG7ETyWlEeuDw=="
```

chmeliik: It's been quite some time since I looked into this, but I *believe* what yarn does is that it picks the strongest hash and checks only that one

cachito/workers/pkg_managers/yarn.py
chmeliik: Oh, I was actually going to remove this function altogether in favour of [doing this check in the fetch_yarn_source task](https://github.com/chmeliik/cachito/blob/522560f7abf70ab4c5addd6023775c93d133e162/cachito/workers/tasks/yarn.py#L63). Forgot to do that, my bad (but this answers the question that I would have had in the future anyway :D )

chmeliik:
Rebased on master, fixes incoming

cachito/workers/pkg_managers/yarn.py
athos-ribeiro: Can we do the same, or at least get only one of them when this is the case so our chcks will not break in the future?

Nice catch by the way. I was not aware of this one!

cachito/workers/pkg_managers/yarn.py
athos-ribeiro: @chmeliik, we can ignore these for now since we are ignoring them for npm as well. We may revisit this in the future.

chmeliik:
Resolved all TODOs except for the one about multiple integrity values, gonna mark this as ready for (final) review while I work on that one

cachito/workers/pkg_managers/yarn.py
chmeliik: Added allowlist, we can just keep it empty for now

cachito/workers/pkg_managers/yarn.py
chmeliik: Made a story to track this internally

cachito/workers/pkg_managers/yarn.py
chmeliik: So, I tried to figure out what exactly yarn does. It lead me from [here](https://github.com/yarnpkg/yarn/blob/3119382885ea373d3c13d6a846de743eca8c914b/src/fetchers/tarball-fetcher.js#L349) me to a [poorly documented class](https://github.com/npm/ssri#integrity-stream) of the `ssri` package, and my understanding of the yarn code does not match the observed behaviour anyway.

I'll just make Cachito pick the strongest algorithm, if this fails for users they will probably know better what behaviour they expect from yarn.

Fun fact: for git dependencies, yarn does not seem to check integrity at all. Note the comment: https://github.com/yarnpkg/yarn/blob/3119382885ea373d3c13d6a846de743eca8c914b/src/fetchers/git-fetcher.js#L98

:headdesk:

chmeliik: done

athos-ribeiro (approved):
Great work!

LGTM

Integration tests are green :)

PR #350, opened 2020-12-10
WIP: Validate symlinks

chmeliik:


chmeliik:
Not nearly done, but PTAL @lcarva @athos-ribeiro

chmeliik:
Hmm, scratch that, rather than trying to track down all the files that Cachito may end up reading while processing a package manager, I'll just check every single file in the repo. This probably makes the changes in here irrelevant

chmeliik:
Closing in favor of #352

PR #351, opened 2020-12-10
Migrate Cachito to Fedora 33

ejegrova:
Update base images of cachito-workers and cachito-api to Fedora 33.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,12 @@
 language: python
 jobs:
   include:
+    - python: "3.9"
+      env: TOXENV=py39
     - python: "3.8"
-      env: TOXENV=black,flake8
+      env: TOXENV=black,flake8  # as soon as black > 20.8b1 is released and we start using it, we can move these tests to the python 3.9 job
     - python: "3.8"
       env: TOXENV=py38
-    - python: "3.7"
-      env: TOXENV=py37
 install:
   - pip install tox
 script:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [tool.black]
 line-length = 100
-target-version = ['py37', 'py38']
+target-version = ['py38']  # as soon as black > 20.8b1 is released and we start using it, we can include py39 here
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
-celery<5
+celery>=5
 gitpython
-kombu<5 # A celery dependency but it's directly imported
+kombu>=5 # A celery dependency but it's directly imported
 packaging
 pyarn @ git+https://github.com/containerbuildsystem/pyarn.git
 requests_kerberos
--- a/setup.py
+++ b/setup.py
@@ -10,9 +10,9 @@ setup(
     include_package_data=True,
     zip_safe=False,
     install_requires=[
-        "celery<5",
+        "celery>=5",
         "gitpython",
-        "kombu<5",  # A celery dependency but it's directly imported
+        "kombu>=5",  # A celery dependency but it's directly imported
         "packaging",
         "pyarn",
         "requests_kerberos",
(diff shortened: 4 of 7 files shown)


requirements.txt
```diff
@@ -1,6 +1,6 @@
-celery<5
+celery<=5.0.2
 gitpython
-kombu<5 # A celery dependency but it's directly imported
+kombu<=5.0.2 # A celery dependency but it's directly imported
```
chmeliik, line 3: Should we just pin to `==5.0.2`?

chmeliik (commented):
F33 has `python3.9`, right? Could you update CI to run tests for it? For reference: https://github.com/release-engineering/cachito/commit/54aafebc62907310440cc884313632ba9bab193d

Maybe also drop tests for 3.7, not sure how we should go about keeping/dropping compatibility with older python versions

requirements.txt
athos-ribeiro: Should we pin it to an specific version?
`<=5.0.2` or `==5.0.2` will have our builds breaking as soon as there is a patch version bump in Fedora (which may happen at any time). Should we say we want it to be >=5 instead? We could include saying we want it <6 but at this point I am not sure if it would matter at all.

setup.py
```diff
@@ -10,9 +10,9 @@
     include_package_data=True,
     zip_safe=False,
     install_requires=[
-        "celery<5",
+        "celery<=5.0.2",
         "gitpython",
-        "kombu<5",  # A celery dependency but it's directly imported
+        "kombu<=5.0.2",  # A celery dependency but it's directly imported
```
akhmanova, line 15: Please, change this version as well :)

requirements.txt
ejegrova: In that case, I think >=5 would be the best.
On F33 is Python3.9 as a default version, I'll update CI.

.travis.yml
```diff
@@ -1,12 +1,12 @@
 language: python
 jobs:
   include:
+    - python: "3.9"
+      env: TOXENV=py39
     - python: "3.8"
-      env: TOXENV=black,flake8
+      env: TOXENV=black,flake8  # as soon as version 20.8b1 or higher is released, we can move it to py39
```
athos-ribeiro, line 7: 20.8b1 is already released. We need a higher version than that (https://pypi.org/project/black/20.8b1/)

pyproject.toml
```diff
@@ -1,3 +1,3 @@
 [tool.black]
 line-length = 100
-target-version = ['py37', 'py38']
```
athos-ribeiro, line 3: We should add that comment here as well to include py39 whenever a next version of black is released.

ejegrova: changed the comment

.travis.yml
ejegrova: changed

pyproject.toml
```diff
@@ -1,3 +1,3 @@
...
+target-version = ['py38']  # as soon as higher version than 20.8b1 is released, we can move it to py39
```
athos-ribeiro, line 3: Should we say
`as soon as black > 20.8b1 is released and we start using it, we can move include py39 here` ?

.travis.yml
```diff
@@ -1,12 +1,12 @@
 language: python
 jobs:
   include:
+    - python: "3.9"
+      env: TOXENV=py39
     - python: "3.8"
-      env: TOXENV=black,flake8
+      env: TOXENV=black,flake8  # as soon as higher version than 20.8b1 is released, we can move it to py39
```
athos-ribeiro, line 7: Should we say 

`as soon as black > 20.8b1 is released and we start using it, we can move these tests to the python 3.9 job`

athos-ribeiro (approved):

PR #352, opened 2020-12-11
Check repos for potentially malicious symlinks

chmeliik:
Will fail the request if any symlink in the cloned repository points
outside of its root directory. This check runs after git submodules
are cloned (if there are any), so malicious symlinks in submodules will
also be caught.

### DIFF
--- a/cachito/workers/tasks/general.py
+++ b/cachito/workers/tasks/general.py
@@ -3,10 +3,11 @@ import logging
 import os
 import shutil
 import tarfile
+from pathlib import Path
 
 import requests
 
-from cachito.errors import CachitoError
+from cachito.errors import CachitoError, ValidationError
 from cachito.workers.config import get_worker_config
 from cachito.workers.scm import Git
 from cachito.workers.paths import RequestBundleDir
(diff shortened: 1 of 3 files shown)


cachito/workers/tasks/general.py
```diff
@@ -49,6 +50,30 @@ def fetch_app_source(url, ref, request_id, gitsubmodule=False):
...
+    for dirpath, subdirs, files in os.walk(repo_root):
+        dirpath = Path(dirpath)
+
+        for entry in subdirs + files:
+            full_path = dirpath / entry
+            real_path = full_path.resolve()
+            try:
+                real_path.relative_to(repo_root)
+            except ValueError:
+                # This one cannot fail
```
lcarva, line 72: Right, because we're not using the real path here. Can we improve the comment so readers don't have to think as hard about this? :wink:

cachito/workers/tasks/general.py
```diff
@@ -49,6 +50,30 @@ def fetch_app_source(url, ref, request_id, gitsubmodule=False):
...
+                relative_path = str(full_path.relative_to(repo_root))
+                raise ValidationError(
+                    f"The destination of {relative_path!r} is outside of cloned repository"
```
lcarva, line 75: Why not just use the error message from the `ValueError` exception?

lcarva (commented):
Mostly cosmetic suggestions. +1 on the overall approach.

cachito/workers/tasks/general.py
chmeliik: Good point, fixed :smile:

cachito/workers/tasks/general.py
chmeliik: The error message would be `'<real path>' does not start with '<full path to bundle app dir>'`, e.g. `'/src' does not start with '/tmp/cachito-archives/bundles/temp/2/app'`

IMO, users will care more about the relative path to the problematic file in their repo, and I'd say they don't need to know what it resolved to.

athos-ribeiro (approved):

cachito/workers/tasks/general.py
lcarva: That's a great point. Your approach makes it easier to diagnose the issue.

lcarva (approved):

PR #355, opened 2021-01-05
Use Fedora images from registry.fedoraproject.org

ejegrova:
Changing images to use Fedora registry instead of  DockerHub


chmeliik (approved):
LGTM :+1:

athos-ribeiro (approved):

PR #358, opened 2021-01-21
Add cachito-download.sh script

chmeliik:
Downloads the archive for a request, then injects all the configuration
files and environment variables (as cachito.env). Should match current
OSBS behavior.

It is written in bash and has one non-standard dependency which is jq.

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,6 +309,35 @@ may mount a
 by adding the volume mount `- /path/to/.netrc:/root/.netrc:ro,z` in your `docker-compose.yml`
 file under the `cachito-worker` container.
 
+### Using Cachito Requests Locally
+
+When testing new functionality or debugging a failure, you may want to take a completed Cachito
+request and try to install the processed packages locally (using `pip`, `gomod`, `npm` etc.). You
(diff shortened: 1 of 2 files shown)


chmeliik:
To be honest, I'm not sure if this should live here. Should we consider having a cachito-scripts repo on github (open-source)? Maybe [pip_find_builddeps.py](https://github.com/release-engineering/cachito/blob/master/bin/pip_find_builddeps.py) should not have been here either, but we've already told people to use it so it's too late to move that one.

athos-ribeiro (commented):
Considering your concerns about this being here, should we move it somewhere under the integration tests directory (and document this script somehow)?

Moreover, about the other script (pip related), it is OK to have it here. I do agree that this new script should not be part of the python package (or a possible RPM package in the future)

chmeliik:
> Considering your concerns about this being here, should we move it somewhere under the integration tests directory (and document this script somehow)?

Hmm, that might make it hard to discover. If we're keeping it in this repo, `bin/` would probably be best.
I will add some documentation.

athos-ribeiro (approved):

PR #359, opened 2021-01-25
Add content manifest support for yarn

ejegrova:


cachito/web/models.py
```diff
@@ -203,13 +203,13 @@ def to_purl(self):
...
 
-        elif self.type == "npm":
+        elif self.type == "npm" or self.type == "yarn":
             purl_name = urllib.parse.quote(self.name)
             match = re.match(
                 r"(?P<protocol>[^:]+):(?P<has_authority>//)?(?P<suffix>.+)", self.version
             )
             if not match:
-                return f"pkg:npm/{purl_name}@{self.version}"
+                return f"pkg:{self.type}/{purl_name}@{self.version}"
```
chmeliik, line 212: @athos-ribeiro may correct me on this, but I think we want to use `pkg:npm` for both NPM and Yarn

chmeliik (commented):
One comment on the format for yarn purls. I also found a few unit tests which could be parametrized to include a yarn case (in test_content_manifest.py):
* test_to_json
* test_purl_conversion
* test_top_level_purl_conversion (this one may need an additional change in models.py)

cachito/web/models.py
ejegrova: changed it back to npm

chmeliik: For context, this is the doc that we reference when deciding what a purl should look like: https://github.com/package-url/purl-spec
You will see that npm is defined there but yarn is not, so we're going with npm here.

chmeliik (approved):
LGTM :+1:

Tested locally (by rebasing this on top of https://github.com/chmeliik/cachito/tree/fetch-yarn-test and submitting a yarn request), works as expected

PR #361, opened 2021-01-26
Integration tests: Remove empty flag from request

ejegrova:


chmeliik (approved):
LGTM :+1:

tests/integration/test_packages.py
```diff
@@ -42,8 +42,10 @@ def test_packages(env_package, env_name, test_env, tmpdir):
         "repo": env_data["repo"],
         "ref": env_data["ref"],
         "pkg_managers": env_data.get("pkg_managers", []),
-        "flags": env_data.get("flags", []),
     }
+    if env_data.get("flags"):
```
chmeliik, line 46: Now I'm thinking, do we still want the ability to explicitly pass an empty list? Maybe it the test data file specifies
`flags: []` then the empty list should be passed, if not then nothing should be passed.

To do that, this would have to be `if "flags" in env_data`

ejegrova: We don't need that in any of the integration tests that we'll work on, but you're right that it will be better to have that ability.

akhmanova (approved):
LGTM! Thank you for fixing this :)

chmeliik (approved):

PR #364, opened 2021-02-02
Forward yarn Cachito requests to workers

ejegrova:
Cachito request including the yarn package manager triggers the fetch_yarn_source celery task

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -11,7 +11,7 @@ from flask_login import current_user, login_required
 import kombu.exceptions
 from werkzeug.exceptions import Forbidden, InternalServerError, Gone, NotFound
 
-from cachito.errors import CachitoError, ValidationError, CachitoNotImplementedError
+from cachito.errors import CachitoError, ValidationError
 from cachito.web import db
 from cachito.web.models import (
     ConfigFileBase64,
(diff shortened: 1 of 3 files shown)


chmeliik (commented):
Looks good! Could you add a yarn case to these 2 tests as well?
* [test_create_request_invalid_dependency_replacement](https://github.com/release-engineering/cachito/blob/b5164f37e170f9fee795fab27ebc72a740512729/tests/test_api_v1.py#L610)
* [test_create_request_invalid_packages](https://github.com/release-engineering/cachito/blob/b5164f37e170f9fee795fab27ebc72a740512729/tests/test_api_v1.py#L721)

akhmanova:
Looks good to me as well! 
@chmeliik I am wondering, is it possible to reuse conditions for package managers in [api_v1.py](https://github.com/release-engineering/cachito/pull/364/files#diff-e623547b2cf808002df23175e5c384f66643c7abe9d45ceda9bd2848378710b7R278-R285)? Or do we want to keep it separate?

chmeliik:
> Looks good to me as well!
> @chmeliik I am wondering, is it possible to reuse conditions for package managers in [api_v1.py](https://github.com/release-engineering/cachito/pull/364/files#diff-e623547b2cf808002df23175e5c384f66643c7abe9d45ceda9bd2848378710b7R278-R285)? Or do we want to keep it separate?

That's a good idea! Ideally, I think all these validation should be done in [Request.from_json()](https://github.com/release-engineering/cachito/blob/b5164f37e170f9fee795fab27ebc72a740512729/cachito/web/models.py#L891). Or not necessarily there, but definitely *before* committing the request to the database.

I think it would we be best to do this as part of the story where we fix all the validations to make them happen before commiting to DB (CLOUDBLD-4201 for those with access).

athos-ribeiro (commented):
Thanks!!!

LGTM. Just need to catch the test cases pointed out by Adam!

chmeliik (approved):
LGTM :+1:

athos-ribeiro (approved):

PR #367, opened 2021-02-04
Fix parsing of go list -deps output

chmeliik:
The current parsing code supports 3 of the 6 known forms of a dependency
line. The remaining 3 will all result in an IndexError.

Support all 6 forms, but raise an error if local replacements are used.

### DIFF
--- a/cachito/workers/pkg_managers/gomod.py
+++ b/cachito/workers/pkg_managers/gomod.py
@@ -7,6 +7,8 @@ import os.path
 import re
 import shutil
 import tempfile
+from pathlib import PureWindowsPath
+from typing import Tuple
 
 import git
 import semver
@@ -224,25 +226,16 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
 
             pkg_level_deps = []
             for line in go_list_deps_output.splitlines():
-                # line example: pkg.io/foo/bar pkg.io/foo/var v1.0.0 => pkg.io/foo/bar v1.0.1
-                parts = [part for part in line.split(" ")]
-
-                name = parts[0]
-                # parts[1] is the module owning the package and parts[2] is it's version
-                if len(parts) > 2:
-                    version = parts[2]
-                else:
-                    # This package belongs to the requested module,
-                    # let's use the requested module version
-                    version = module_version
-
-                if len(parts) > 3:
-                    # parts[3] is just an arrow "=>" pointing to the replacements
-                    # parts[4] is the module being replaced
-                    # we do not add a "replacements" field to the packages though
-                    # since the concept of replacements are tied to the modules
-                    # Instead, we just set the proper version
-                    version = parts[5]
+                name, version = _parse_name_and_version(line)
+                # If the line did not contain a version, we'll use the module version
+                version = version or module_version
+                if version.startswith("."):
+                    raise CachitoError(f"Local gomod dependencies are not yet supported: {version}")
+                elif version.startswith("/") or PureWindowsPath(version).root:
+                    # This will disallow paths starting with '/', '\' or '<drive letter>:\'
+                    raise CachitoError(
+                        f"Absolute paths to gomod dependencies are not supported: {version}"
+                    )
 
                 pkg = {
                     "name": name,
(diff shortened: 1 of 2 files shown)


cachito/workers/pkg_managers/gomod.py
```diff
@@ -260,6 +245,59 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
...
+    # line example: pkg.io/foo/bar pkg.io/foo/var v1.0.0 => pkg.io/foo/bar v1.0.1
+    parts = list_deps_line.split(" ")
+
+    # As far as we know, the possible forms of the dependency line are as follows:
+    # | len | semantics                                              |
+    # |-----|--------------------------------------------------------|
+    # |   2 | <package> <module>                                     |
+    # |   3 | <package> <module> <version>                           |
+    # |   4 | <package> <module> => <./local/path>                   |
+    # |   5 | <package> <module> => <module> <version>               |
```
chmeliik, line 258: @athos-ribeiro, how is it possible that we've never encountered this case? Here are some examples from processing kubernetes:
```
go.etcd.io/etcd/pkg/pathutil go.etcd.io/etcd => go.etcd.io/etcd v0.5.0-alpha.5.0.20200910180754-dd1b699fc489
github.com/coreos/go-semver/semver github.com/coreos/go-semver => github.com/coreos/go-semver v0.3.0
google.golang.org/protobuf/internal/detrand google.golang.org/protobuf => google.golang.org/protobuf v1.25.0
golang.org/x/net/context golang.org/x/net => golang.org/x/net v0.0.0-20201110031124-69a78807bb2b
```
And hundreds of other cases.

chmeliik:
Assumptions about format were true for kubernetes and integration tests, removed check. Also, now raising RuntimeError if the line has more than 6 parts. If this breaks something, it's probably for the best. Who knows what Cachito would have parsed as version before.

chmeliik:
Fixed mocking, previously the new test would create weird directories in your workdir

chmeliik:
Found out that file paths can be absolute, also disallowing those: https://golang.org/ref/mod#go-mod-file-replace

akhmanova:
@chmeliik I just checked previous lines of [resolve_gomod()](https://github.com/release-engineering/cachito/pull/367/commits/bc7dcc95bae2ffca17fc4015ac7d6ad70ccaecdf#diff-f787f4cff45f5fc78eeaae68fa9edd9709cbe91fe914ad7ba5a21a91ec4e1772R217-R225) . And I am interested, can we just use: 

 `go list -deps -f '{{if not .Standard}}{{.Module}}{{end}}'`

instead of: 
`go list -deps -f '{{if not .Standard}}{{.ImportPath}} {{.Module}}{{end}}'`

? 

Because for me `.ImportPath` is the same with the first part of `.Module`. In which cases it can differ?

chmeliik:
> Because for me `.ImportPath` is the same with the first part of `.Module`. In which cases it can differ?

`.ImportPath` is the full package name, `.Module` is just the module name. Example:
```
k8s.io/kubectl/pkg/util k8s.io/kubectl => ./staging/src/k8s.io/kubectl
```
`k8s.io/kubectl` is the module, `k8s.io/kubectl/pkg/util` is a package inside that module. Worth noting that the package name will always *start with* the module name, but is not always the same.

cachito/workers/pkg_managers/gomod.py
athos-ribeiro: Could this be a change from something newer than go `1.12` or `1.13`, when we started getting this information?

It would be nice to find some docs for these formates so we can link it here. I could not find any back then.

cachito/workers/pkg_managers/gomod.py
```diff
@@ -260,6 +253,38 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
...
+    # In all the cases above, parts[0] is the package name and parts[1] is the module
+    # that owns the package
+    name = parts[0]
+
+    if len(parts) <= 2:
+        # No version in dependency line
+        # len = 1 should be impossible, but conservatively, let's allow both 1 and 2 here
+        version = None
+    elif len(parts) <= 6:
+        # for all the cases with len in {3..6}, the version that we want is the last part
```
athos-ribeiro, line 280: nice

athos-ribeiro (approved):
LGTM

cachito/workers/pkg_managers/gomod.py
chmeliik: There is some info in `go help list`, but the format for `.Module` is not described in detail. This is the relevant part:
```
The default output is to print the module path and then
information about the version and replacement if any.
For example, 'go list -m all' might print:

    my/main/module
    golang.org/x/text v0.3.0 => /tmp/text
    rsc.io/pdf v0.1.1
```
For what it's worth, when there is a replacement, the format seems to match the `ReplaceSpec` here: https://golang.org/ref/mod#go-mod-file-replace

chmeliik:
@athos-ribeiro, added a comment to point to at least some resources about the format

cachito/workers/pkg_managers/gomod.py
chmeliik: Added a comment

cachito/workers/pkg_managers/gomod.py
```diff
@@ -260,6 +253,39 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
         return {"module": module, "module_deps": module_level_deps, "packages": packages}
 
 
+def _parse_name_and_version(list_deps_line: str) -> Tuple[str, str]:
+    """Parse package name and version (if present) from one line of go list -deps output."""
+    # line example: pkg.io/foo/bar pkg.io/foo/var v1.0.0 => pkg.io/foo/bar v1.0.1
+    parts = list_deps_line.split(" ")
+
+    # As far as we know, the possible forms of the dependency line are as follows (see
+    #   `go help list` and https://golang.org/ref/mod#go-mod-file-replace):
```
akhmanova, line 262: @chmeliik Thank you for the information! In go help list I found only info about graph traversal order.

akhmanova (approved):
@chmeliik Such a thoughtful job! Even with tests. LGTM

PR #370, opened 2021-02-10
Don't fail on local gomod dependencies

omertuc:
We want to add local dependencies in https://github.com/openshift/assisted-service, this "Local gomod dependencies are not yet supported" exception is blocking us from doing so. 

Is there any reason for this exception to be raised? Can it just be ignored like in the diff of this PR?

The local dependencies are always relative to the root of the repo anyway, so they get pulled along with all other repository files - it doesn't require any special handling by cachito, at-least to the best of my knowledge. 

What do you think?

### DIFF
--- a/cachito/workers/pkg_managers/gomod.py
+++ b/cachito/workers/pkg_managers/gomod.py
@@ -230,7 +230,7 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
                 # If the line did not contain a version, we'll use the module version
                 version = version or module_version
                 if version.startswith("."):
-                    raise CachitoError(f"Local gomod dependencies are not yet supported: {version}")
+                    continue
                 elif version.startswith("/") or PureWindowsPath(version).root:
                     # This will disallow paths starting with '/', '\' or '<drive letter>:\'
                     raise CachitoError(
--- a/tests/test_workers/test_pkg_managers/test_gomod.py
+++ b/tests/test_workers/test_pkg_managers/test_gomod.py
@@ -320,42 +320,6 @@ def test_resolve_gomod_unused_dep(mock_run, mock_temp_dir, tmpdir):
 @mock.patch("cachito.workers.pkg_managers.gomod.GoCacheTemporaryDirectory")
 @mock.patch("cachito.workers.pkg_managers.gomod.RequestBundleDir")
 @mock.patch("subprocess.run")
-def test_resolve_gomod_disallow_local_dependencies(
-    mock_run, mock_bundle_dir, mock_temp_dir, mock_golang_version, mock_merge_tree, tmpdir
-):
-    # Mock the tempfile.TemporaryDirectory context manager
-    mock_temp_dir.return_value.__enter__.return_value = str(tmpdir)
-
-    # Mock the "subprocess.run" calls
-    mock_run.side_effect = [
-        # go mod download
-        mock.Mock(returncode=0, stdout=None),
-        # go list -m all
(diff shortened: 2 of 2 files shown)


chmeliik:
@omertuc thanks for your interest in contributing to Cachito!

The reason why local dependencies are not allowed is that Cachito cannot (yet) properly present them in the content manifest. I am currently working on adding support for that.

When the prerequisites are in place, we will add an allowlist for modules that are allowed to use local dependencies.

For more details, check the downstream issue (which I believe you should have access to): CLOUDBLD-3015

chmeliik:
@omertuc additional question: we did not break your build with the recent changes, did we? I believe local dependencies were never supported, they just failed less gracefully before.

omertuc:
> @omertuc additional question: we did not break your build with the recent changes, did we? I believe local dependencies were never supported, they just failed less gracefully before.

That's right, it's just something we were planning to do. Concerns were raised about making sure Cachito will work with those changes, so I tried, and came across this error. 

> The reason why local dependencies are not allowed is that Cachito cannot (yet) properly present them in the content manifest. I am currently working on adding support for that.

That's good to hear. Just checking, because this change is important for my team - is it an option to have this merged until this support is added? Or it is critical for the content manifest to be correct even with regards to local dependencies?

chmeliik:
> That's good to hear. Just checking, because this change is important for my team - is it an option to have this PR merged until this support is added? Or it is critical for the content manifest to be correct even with regards to local dependencies?

That is a good question. @athos-ribeiro, @lcarva, what would be the consequences if we do not record information about local dependencies? I would rather have the full support, but maybe it is fine.

Also note that if we merge this change, it will only make it to the next release (or we would have to do a hotfix). It is likely that my changes will be ready by then.

omertuc:
> Also note that if we merge this change, it will only make it to the next release (or we would have to do a hotfix). It is likely that my changes will be ready by then.

I see. When is the next release? Weekly?

> Edit: and then there's the fact that we only want to allow local deps based on an allowlist, @lcarva would you elaborate on the reasons for that?

Also, where is this whitelist maintained? Who decides what goes in there?

chmeliik:
> I see. When is the next release? Weekly?

We've been sticking to biweekly releases and we're just in the middle of one, so likely 2 weeks from now (but the schedule is not strict, we might be able to do it sooner).

> Also, where is this whitelist maintained? Who decides what goes in there?

Internally in the Cachito deployment playbooks, it would be something similar to the npm allowlist https://github.com/release-engineering/cachito/blob/c8fd79c3af6479eb52ddded9802d11089905623b/cachito/workers/config.py#L114
As for who decides, I believe it's us, the Cachito maintainers, but I'll defer to Luiz for more details.

athos-ribeiro:
Thanks for the PR, @omertuc !!!

If the actual local deps for gomod will be in place in the next 2~3 weeks, merging this one would not make much sense given both changes would go into the same release.

@chmeliik, since you are the one working on the aforementioned change, I'd actually defer it to you, if you want to re-use this PR.

chmeliik:
> @chmeliik, since you are the one working on the aforementioned change, I'd actually defer it to you, if you want to re-use this PR.

I won't be reusing this PR. I think we've come to the conclusion that we want to wait for proper support, @omertuc would you mind if we close this?

omertuc:
> > @chmeliik, since you are the one working on the aforementioned change, I'd actually defer it to you, if you want to re-use this PR.
> 
> I won't be reusing this PR. I think we've come to the conclusion that we want to wait for proper support, @omertuc would you mind if we close this?

Nope

PR #373, opened 2021-02-11
Support local gomod dependencies

chmeliik:
Adds content manifest support and an allowlist for local gomod dependencies.

### DIFF
--- a/cachito/web/content_manifest.py
+++ b/cachito/web/content_manifest.py
@@ -21,17 +21,17 @@ class ContentManifest:
         :param Request request: the request to generate a ContentManifest for
         """
         self.request = request
-        # dict to store go package level data; uses the package purl as key to identify a package
+        # dict to store go package level data; uses the package id as key to identify a package
         self._gopkg_data = {}
         # dict to store go module level purl dependencies. Module names are used as keys
         self._gomod_data = {}
-        # dict to store npm package data; uses the package purl as key to identify a package
+        # dict to store npm package data; uses the package id as key to identify a package
         self._npm_data = {}
-        # dict to store pip package data; uses the package purl as key to identify a package
+        # dict to store pip package data; uses the package id as key to identify a package
         self._pip_data = {}
-        # dict to store yarn package data; uses the package purl as key to identify a package
+        # dict to store yarn package data; uses the package id as key to identify a package
         self._yarn_data = {}
-        # dict to store gitsubmodule package level data; uses the package purl as key to identify a
+        # dict to store gitsubmodule package level data; uses the package id as key to identify a
         # package
         self._gitsubmodule_data = {}
 
--- a/cachito/web/content_manifest.py
+++ b/cachito/web/content_manifest.py
@@ -67,21 +67,24 @@ class ContentManifest:
         module based on their names.
         """
         for package_id, pkg_data in self._gopkg_data.items():
-            if pkg_data["name"] in self._gomod_data:
-                self._gopkg_data[package_id]["sources"] = self._gomod_data[pkg_data["name"]]
+            pkg_name = pkg_data.pop("name")
+
+            if pkg_name in self._gomod_data:
+                module_name = pkg_name
             else:
                 # We use the longest module available in the request that matches the package name
-                previous_length = 0
-                for mod_name, sources in self._gomod_data.items():
-                    if pkg_data["name"].startswith(mod_name) and len(mod_name) > previous_length:
-                        self._gopkg_data[package_id]["sources"] = sources
-                        previous_length = len(mod_name)
+                module_name = max(
+                    (mod_name for mod_name in self._gomod_data if pkg_name.startswith(mod_name)),
+                    key=len,
+                    default=None,
+                )
 
-                if not previous_length:
-                    flask.current_app.logger.warning(
-                        "Could not find a Go module for %s", pkg_data["purl"]
-                    )
-            pkg_data.pop("name")
+            if module_name is not None:
+                self._gopkg_data[package_id]["sources"] = self._gomod_data[module_name]
+            else:
+                flask.current_app.logger.warning(
+                    "Could not find a Go module for %s", pkg_data["purl"]
+                )
 
     def process_npm_package(self, package, dependency):
         """
--- a/cachito/web/content_manifest.py
+++ b/cachito/web/content_manifest.py
@@ -4,6 +4,9 @@ import flask
 from cachito.web.utils import deep_sort_icm
 
 
+PARENT_PURL_PLACEHOLDER = "PARENT_PURL"
+
+
 class ContentManifest:
     """A content manifest associated with a Cacihto request."""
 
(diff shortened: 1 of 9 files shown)


chmeliik:
To test:
```shell
          "repo": "https://github.com/cachito-testing/cachito-gomod-local-deps.git",
          "ref": "ef4adaa69f735dba85378301126ac49981143ead",
          "pkg_managers": ["gomod"]
        }'
```
Excerpt from the content manifest:
```json
{
  "dependencies": [
    {
      "purl": "pkg:golang/github.com%2Fcachito-testing%2Fcachito-gomod-local-deps@v0.0.0-20210211104706-ef4adaa69f73#staging/src/some-module"
    }
  ],
  "purl": "pkg:golang/github.com%2Fcachito-testing%2Fcachito-gomod-local-deps@v0.0.0-20210211104706-ef4adaa69f73",
  "sources": [
    {
      "purl": "pkg:golang/github.com%2Fcachito-testing%2Fcachito-gomod-local-deps@v0.0.0-20210211104706-ef4adaa69f73#staging/src/some-module"
    }
  ]
}
```
i.e. local deps are presented as `<module_purl>#<subpath>`. I believe this is valid and correct according to the [purl spec](https://github.com/package-url/purl-spec).

chmeliik:
BTW, please do not be discouraged by the size of this PR, 75% of the diff is unit tests :-)

As an (in)sanity check, I also processed kubernetes with these changes. It took 40 minutes in the dev env and the content manifest is 17 megabytes, but it seems to be working.

cachito/web/content_manifest.py
```diff
@@ -21,17 +21,17 @@ def __init__(self, request):
         :param Request request: the request to generate a ContentManifest for
         """
         self.request = request
-        # dict to store go package level data; uses the package purl as key to identify a package
+        # dict to store go package level data; uses the package id as key to identify a package
```
akhmanova, line 27: @chmeliik  Oh, looks like we will need to change integration tests that check content manifest. Because they keep old key data, am I right?

cachito/web/content_manifest.py
```diff
@@ -67,21 +67,24 @@ def set_go_package_sources(self):
...
         """
         for package_id, pkg_data in self._gopkg_data.items():
-            if pkg_data["name"] in self._gomod_data:
-                self._gopkg_data[package_id]["sources"] = self._gomod_data[pkg_data["name"]]
+            pkg_name = pkg_data.pop("name")
+
+            if pkg_name in self._gomod_data:
+                module_name = pkg_name
             else:
                 # We use the longest module available in the request that matches the package name
```
akhmanova, line 80: Why do we need to use a module with the longest name?

akhmanova: I like new way of handling empty names now more. It is much more logical

cachito/web/content_manifest.py
chmeliik: Integration tests should be fine, the change only affects purls for local dependencies (we do not test those). Content manifests for everything else should stay exactly the same.

Also, I ran `tox -e integration` and it worked, so it should be fine :D

cachito/web/content_manifest.py
chmeliik: This code is older than my time in Cachito so I can only guess, but I think it's something like this.
Imagine you have a main module called `example.org/some-module`. This module has one submodule: `example.org/some-module/submodule`. Both modules have their own package as illustrated below:
- example.org/some-module
  - example.org/some-module/some-package
  - example.org/some-module/submodule
    - example.org/some-module/submodule/other-package

The first package, `example.org/some-module/some-package`, only matches the main module - `example.org/some-module`. We want to use that one as the parent module.

The second package, `example.org/some-module/submodule/other-package`, matches both modules, but it's immediate parent is `example.org/some-module/submodule` (`example.org/some-module` is a "grandparent", if you will).

That's why we take the longest matching module name.

cachito/web/content_manifest.py
akhmanova: Perfect, thank you!

cachito/web/content_manifest.py
```diff
@@ -43,8 +46,10 @@ def process_gomod(self, package, dependency):
         :param Dependency dependency: the gomod package dependency to process
         """
         if dependency.type == "gomod":
-            icm_source = {"purl": dependency.to_purl()}
-            self._gomod_data[package.name].append(icm_source)
+            parent_purl = self._gomod_data[package.name]["purl"]
+            dep_purl = dependency.to_purl().replace(PARENT_PURL_PLACEHOLDER, parent_purl)
+            icm_source = {"purl": dep_purl}
+            self._gomod_data[package.name]["dependencies"].append(icm_source)
```
athos-ribeiro, line 52: nitpick: It would be nice if the structure change (using `"dependencies"`) was in a separate commit here.

cachito/workers/pkg_managers/gomod.py
```diff
@@ -286,6 +287,67 @@ def _parse_name_and_version(list_deps_line: str) -> Tuple[str, str]:
...
+            raise CachitoError(f"Absolute paths to gomod dependencies are not supported: {version}")
+
+
+def _fail_unless_allowlisted(module_name: str, package_name: str, allowed_patterns: List[str]):
+    """Fail unless the module is allowed to replace the package with a local dependency."""
+    if not any(fnmatch.fnmatch(package_name, pat) for pat in allowed_patterns):
+        raise CachitoError(
+            f"The module {module_name} is not allowed to replace {package_name} with a local "
+            f"dependency. Please contact the maintainers of this Cachito instance about adding "
+            "an exception."
```
athos-ribeiro, line 328: Nice msg :)

cachito/web/models.py
```diff
@@ -198,6 +198,11 @@ def to_purl(self):
         :raise ContentManifestError: if the there is no implementation for the package type
         """
         if self.type in ("go-package", "gomod"):
+            if self.version.startswith("."):
+                # Package is relative to the parent module
+                normpath = os.path.normpath(self.version)
+                return f"{content_manifest.PARENT_PURL_PLACEHOLDER}#{normpath}"
```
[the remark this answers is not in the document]
chmeliik: I believe those are not possible: https://golang.org/ref/mod#go-mod-file-replace
```
replace directives only apply in the main module's go.mod file and are ignored in other modules.
See Minimal version selection for details.
```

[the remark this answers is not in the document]
athos-ribeiro: This was my only concertn here. In this case, LGTM

athos-ribeiro (approved):

cachito/web/content_manifest.py
chmeliik: Broke up into 2 commits

cachito/workers/pkg_managers/gomod.py
```diff
@@ -286,6 +287,67 @@ def _parse_name_and_version(list_deps_line: str) -> Tuple[str, str]:
...
+                log.debug("Allowed local packages for %s: %s", module_name, allowed_patterns)
+
+            log.debug(
+                "Module %s wants to replace %s with a local dependency: %s",
+                module_name,
+                name,
+                version,
+            )
+            _fail_unless_allowlisted(module_name, name, allowed_patterns)
+            _fail_if_outside_repo(version, module_path, repo_root)
```
chmeliik, line 316: I found this in the purl spec:
```
When percent-decoded, a segment:

    must not contain a '/'
    must not be any of '..' or '.'
    must not be empty
```
We need a stricter check, not only must the path be inside the repo, it must not start with `..` at all

chmeliik:
Changed the check to disallow all paths with `..` in them. Also, the allowlist will now be logged only once.

nirzari (approved):
Looks Good :ship: :it:

chmeliik:
> Looks Good ship it

@nirzari I was wondering what the italian flag was about until I saw the text representation :smile:

PR #374, opened 2021-02-11
Commit DB operations after celery errors

ejegrova:
In case of celery errors, create_request function commit operations to
DB and mark request as failed.

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -294,6 +294,7 @@ def create_request():
         )
         error = "Failed to schedule the task to the workers. Please try again."
         request.add_state("failed", error)
+        db.session.commit()
         raise CachitoError(error)
 
     flask.current_app.logger.debug("Successfully scheduled request %d", request.id)
--- a/tests/test_api_v1.py
+++ b/tests/test_api_v1.py
@@ -936,8 +936,20 @@ def test_create_request_connection_error(mock_chain, app, auth_env, client, db):
     )
     rv = client.post("/api/v1/requests", json=data, environ_base=auth_env)
 
+    engine = db.create_engine(app.config["SQLALCHEMY_DATABASE_URI"], {})
+    connection = engine.connect()
+    request_state = db.Table("request_state", db.MetaData(), autoload=True, autoload_with=engine)
+    query = db.select([request_state]).where(request_state.columns.request_id == 1)
+
+    state_reasons = []
+    for res in connection.execute(query):
+        state_reasons.append(res[2])
+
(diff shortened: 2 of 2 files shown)


cachito/web/api_v1.py
```diff
@@ -294,6 +294,8 @@ def create_request():
         )
         error = "Failed to schedule the task to the workers. Please try again."
         request.add_state("failed", error)
+        db.session.add(request)
```
athos-ribeiro, line 297: Do we really need to add the request again here? Note that the add_state call will call `db.session.add` for the state.

athos-ribeiro, line 297: Do we really need to add the request again here? Note that the add_state call will call db.session.add for the state.

ejegrova: Oh right, I had db.session.commit on wrong place before and it didn't work without this call. Removing that.

tests/test_api_v1.py
```diff
@@ -936,8 +936,20 @@ def test_create_request_connection_error(mock_chain, app, auth_env, client, db):
...
+    connection = engine.connect()
+    request_state = db.Table("request_state", db.MetaData(), autoload=True, autoload_with=engine)
+    query = db.select([request_state]).where(request_state.columns.request_id == 1)
+
+    state_reasons = []
+    for res in connection.execute(query):
+        state_reasons.append(res[2])
+
+    error = "Failed to schedule the task to the workers. Please try again."
+    assert any(elem == error for elem in state_reasons)
```
chmeliik, line 949: This test already queries the database for the request, see below
```python
assert Request.query.get(1).state.state_name == "failed"
```
You can use the same approach to test for the state reason

ejegrova: Ok, I didn't know what this line is about because I didn't find query in Request class. Now I get where it's coming from. In that case there's something odd with this test or I just don't understand how that works. But this assertion is passing in both cases: when you run db.session.commit() in create_request() and also if you don't run it there. I think this shouldn't be happening.
I tried to check state with:
```assert Request.query.get(1).state.state_reason == "Failed to schedule the task to the workers. Please try again."```
This also does not reflect if db.session.commit() run and always passes.

chmeliik: Can confirm that this code is a proper test while `Request.query` is not, very interesting :D
We can keep your version and later investigate why the previous test does not work.

chmeliik (approved):
LGTM :+1:

Tested locally by killing rabbitmq (`docker-compose kill rabbitmq`), submitting a request and verifying that it is marked as failed in the `/requests` response

athos-ribeiro (approved):

PR #380, opened 2021-02-17
Fall back to versionless go modules in allowlist

chmeliik:
If module name contains a version and is not present in the allowlist,
also try matching without the version. E.g. if example.org/module/v2 is
not present in the allowlist, return allowed deps for example.org/module


akhmanova:
The change looks good to me. Running https://master-jenkins-csb-osbs.cloud.paas.psi.redhat.com/view/%20%20Cachito/job/cachito-test-integration-tests/290/

athos-ribeiro (approved):

PR #382, opened 2021-02-22
Set task priorities

athos-ribeiro:
Cachito tasks are processed in a breadth first approach. This is not
optimal because starting the next steps after fetching the requested app
sources requires that no other requests to fetch app sources (for a
different Cachito request) is queued. As a consequence, Cachito could
get completely blocked (no Requests being completed) if a user submits a
large number of requests (and if new requests keep coming).

This patch changes Cachito's approach by introducing priorities on how
tasks in our queues get processed. The tasks shall be processed in a
depth first approach, where the first task for a new request will only
be processed after there are no other tasks (from requests being
executed) left to be picked up.

### DIFF
--- a/cachito/workers/config.py
+++ b/cachito/workers/config.py
@@ -76,6 +76,8 @@ class Config(object):
         kombu.Queue("cachito_gomod", routing_key="cachito.gomod"),
         kombu.Queue("cachito_npm", routing_key="cachito.npm"),
     )
+    task_queue_max_priority = 10
+    task_default_priority = 5
     # Requeue the message if the worker abruptly exits or is signaled
     task_reject_on_worker_lost = True
     # Route gomod tasks and npm tasks to separate queues. This is useful if workers are dedicated
--- a/cachito/workers/tasks/general.py
+++ b/cachito/workers/tasks/general.py
@@ -22,7 +22,7 @@ __all__ = [
 log = logging.getLogger(__name__)
 
 
-@app.task
+@app.task(priority=0)
 def fetch_app_source(url, ref, request_id, gitsubmodule=False):
     """
     Fetch the application source code that was requested and put it in long-term storage.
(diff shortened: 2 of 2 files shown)


lcarva (approved):
wow this turned out to be simpler than what I had expected! Were you able to verify the behavior in the local development environment?

athos-ribeiro:
> wow this turned out to be simpler than what I had expected! Were you able to verify the behavior in the local development environment?

Yes. This is the original behaviour: https://athoscr.fedorapeople.org/without_priority.png

This is how it behaves after the change: https://athoscr.fedorapeople.org/with_priority.png

This is the script I used to test it:

```
#!/bin/bash

function request {
  curl -s -X POST -H "Content-Type: application/json" localhost:8080/api/v1/requests -d \
    '{
        "repo": "https://github.com/athos-ribeiro/cachito-yarn-test",
        "ref": "e9f1fb707f77a2f1819da0bfe1a6b545aaf7e3be",
        "pkg_managers": ["yarn"]
      }' | jq '.id' 
}

function check {
  while true; do
    STATE=$(curl -s -H "Content-Type: application/json" localhost:8080/api/v1/requests/$1 | jq -r '.state')
    if [ "$STATE" == "complete" ]; then
      FIRST=$(curl -s -H "Content-Type: application/json" localhost:8080/api/v1/requests/$1 | jq -r '.state_history | .[-1].updated')
      LAST=$(curl -s -H "Content-Type: application/json" localhost:8080/api/v1/requests/$1 | jq -r '.state_history | .[0].updated')
      FIRST=`date '+%s' --date="$FIRST"`
      LAST=`date '+%s' --date="$LAST"`
      echo $(( $LAST - $FIRST ))
      break
    elif [ "$STATE" == "failed" ]; then
      echo REQUEST $1 FAILED
      break
    else
      sleep 1
    fi
  done
}

REQUESTS=()
for i in {1..100}; do
  REQ=`request`
  echo created request $REQ
  REQUESTS+=($REQ)
done

for r in ${REQUESTS[@]}; do
  echo Request $r took `check $r` seconds to complete
done
```

PR #388, opened 2021-03-12
WIP: store and present subpaths when multiple gomod packages specified

chmeliik:


chmeliik:
@lcarva @athos-ribeiro I tested with this repo: https://github.com/chmeliik/gomod-test (at the latest commit)
And this is the content manifest:
```json
{
  "image_contents": [
    {
      "dependencies": [
        {
          "purl": "pkg:golang/github.com%2Fchmeliik%2Fgomod-test%2Fc%2Fpackage-b@v0.0.0-20210312155304-bcd209154943"
        }
      ], 
      "purl": "pkg:github/chmeliik/gomod-test@bcd20915494373adb88d34c4b8e82e3786757725#b", 
      "sources": []
    }, 
    {
      "dependencies": [
        {
          "purl": "pkg:golang/github.com%2Fchmeliik%2Fgomod-test%2Fa%2Fv2%2Fpackage-a@v2.0.0-20210312155304-bcd209154943"
        }
      ], 
      "purl": "pkg:golang/github.com%2Fchmeliik%2Fgomod-test%2Fa%2Fv2@v2.0.0-20210312155304-bcd209154943", 
      "sources": []
    }
  ], 
  "metadata": {
    "icm_spec": "https://raw.githubusercontent.com/containerbuildsystem/atomic-reactor/f4abcfdaf8247a6b074f94fa84f3846f82d781c6/atomic_reactor/schemas/content_manifest.json", 
    "icm_version": 1, 
    "image_layer_index": -1
  }
}
```
Module `a` has the correct path, `pkg:golang` is used. Module `c` is in directory `b`, so `pkg:github` is used. The problem, as you can see, is the package inside module `c`, which is listed as a dependency.

To fix the purl for the dependency, we would have to:
* match it to its parent module
* check the purl of its parent module
  * if `pkg:golang`, use `pkg:golang`
  * if `pkg:github`, parse the subpath from the module purl, get the subpath from the module to the package, use `pkg:github`

The sad part is that the `Package` class, responsible for generating purls, has none of the information necessary to construct a `pkg:github` purl for a dependency. It is doable, but the purl generation would be entirely up to the `content_manifest` code.

lcarva:
@chmeliik, what's the expected behavior? I think you're trying to highlight that the package stored in `github.com/chmeliik/gomod-test/b` is actually named `github.com/chmeliik/gomod-test/c`, right?

If that's the case, I do think the name of the package is `github.com/chmeliik/gomod-test/c` but it's only stored at an odd location.  Wouldn't a package named this way be very difficult to use from another gomod application? You'd have to add a replace directive such as:
```
replace github.com/chmeliik/gomod-test/c => github.com/chmeliik/gomod-test/b v1.2.3
```

I'm not even sure that would work:

```
$ go get github.com/chmeliik/gomod-test/c
package github.com/chmeliik/gomod-test/c: cannot find package "github.com/chmeliik/gomod-test/c" in any of:
	/usr/lib/golang/src/github.com/chmeliik/gomod-test/c (from $GOROOT)
	/home/lucarval/gocode/src/github.com/chmeliik/gomod-test/c (from $GOPATH)
lucarval@localhost /tmp/yolo 
(1)

$ go get github.com/chmeliik/gomod-test/b
package github.com/chmeliik/gomod-test/c/package-b: cannot find package "github.com/chmeliik/gomod-test/c/package-b" in any of:
	/usr/lib/golang/src/github.com/chmeliik/gomod-test/c/package-b (from $GOROOT)
	/home/lucarval/gocode/src/github.com/chmeliik/gomod-test/c/package-b (from $GOPATH)
(1)

$ go get github.com/chmeliik/gomod-test/a
(0)  
```

chmeliik:
@lcarva true, if the module name vs. subpath do not match, it probably cannot be used as a dependency. It can still be used as the main application though. Unfortunately, Cachito users are free to store their modules in the wrong place. We do have a way to generate correct purls even for incorrect go modules. But given the amount of duct tape it would require, I'm not sure I want to do it. Also, I've never seen any real consequences of a broken purl, so I'm leaning towards not worth it.

I'll add the rest of the code changes and then we can decide if we want just the first 2 commits or the whole thing.

chmeliik:
Tested with 2 different repos:
- https://github.com/chmeliik/gomod-test
```shell
          "repo": "https://github.com/chmeliik/gomod-test",
          "ref": "bcd20915494373adb88d34c4b8e82e3786757725",
          "pkg_managers": ["gomod"],
          "packages": {
            "gomod": [
              {"path": "a"},
              {"path": "b"}
            ]
          }
        }'
```
- https://github.com/chmeliik/gomod-multi
```shell
          "repo": "https://github.com/chmeliik/gomod-multi",
          "ref": "250077667d19039fd7cb5913b819ce07d162a152",
          "pkg_managers": ["gomod"],
          "packages": {
            "gomod": [
              {"path": "correctly_named"},
              {"path": "incorrectly_named"},
              {"path": "correct_v2"},
              {"path": "incorrect_v2"},
              {"path": "correct_with_local"},
              {"path": "incorrect_with_local"},
              {"path": "correct_v2_local"},
              {"path": "incorrect_v2_local"}
            ]
          }
        }'
```
Results seem to be correct.

chmeliik:
Unit tests for WIP commits are missing/broken, please ignore for now

First 3 commits would be worth having either way. WIP commits are debatable, I'm personally leaning more towards not worth it. DROP commits are just there for testing.

@athos-ribeiro @lcarva what do you think, is it worth going to these lengths to protect users from themselves? Is having broken purls for broken go modules even a problem?

lcarva:
@chmeliik, is it fair to say that the malformed golang package names are only an issue for the top level package?

chmeliik:
> @chmeliik, is it fair to say that the malformed golang package names are only an issue for the top level package?

@lcarva by top-level package, you mean the one that has go.mod right? In which case no, all the packages defined in subdirectories of this top-level package will also have malformed names, and we present them in the content manifest as dependencies of the top-level package (if the top-level package imports them).

chmeliik:
Closing, I'll resubmit only the parts that are relevant for now.

PR #389, opened 2021-03-18
Additional changes post pinning requirements.txt

chmeliik:
I randomly noticed that `make venv` was failing and it lead to me making a couple quality-of-life changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 CACHITO_COMPOSE_ENGINE ?= docker-compose
-PYTHON_VERSION_VENV ?= python3.8
-TOX_ENVLIST ?= py38
+PYTHON_VERSION_VENV ?= python3.9
+TOX_ENVLIST ?= py39
 TOX_ARGS ?=
 
 PODMAN_COMPOSE_AUTO_URL ?= https://raw.githubusercontent.com/containers/podman-compose/devel/podman_compose.py
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,11 @@ clean: run-down
 
 .PHONY: venv
 venv:
-	virtualenv --python=${PYTHON_VERSION_VENV} venv && venv/bin/pip install --upgrade pip && venv/bin/pip install -r requirements.txt -r requirements-web.txt tox && venv/bin/python setup.py develop
+	virtualenv --python=${PYTHON_VERSION_VENV} venv
+	venv/bin/pip install --upgrade pip
+	venv/bin/pip install -r requirements.txt -r requirements-web.txt
+	venv/bin/pip install tox
+	venv/bin/python setup.py develop
 
 # Keep run target for backwards compatibility
 run run-start:
--- a/Makefile
+++ b/Makefile
@@ -60,3 +60,13 @@ test-suite test-tox:
 	PATH="${PWD}/venv/bin:${PATH}" tox -e $(TOX_ENVLIST) -- $(TOX_ARGS)
 
 test-all: test-unit test-integration
+
+pip-compile: venv/bin/pip-compile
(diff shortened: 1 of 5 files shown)


akhmanova:
Nice work! Let me test it before the merging

akhmanova (approved):
LGTM! Thank you

Makefile
```diff
@@ -60,3 +60,13 @@ test-suite test-tox:
 	PATH="${PWD}/venv/bin:${PATH}" tox -e $(TOX_ENVLIST) -- $(TOX_ARGS)
 
 test-all: test-unit test-integration
+
+pip-compile: venv/bin/pip-compile
+	venv/bin/pip-compile --allow-unsafe --generate-hashes --output-file=requirements.txt requirements.in
```
lcarva, line 65: It would be nice to add a comment as to why we need to use `--allow-unsafe` here. From what I can tell, it's because we have a dependency on `setuptools` listed in `requirements.in`. For my own understanding, why is that needed?

Makefile
```diff
@@ -60,3 +60,13 @@ test-suite test-tox:
...
+	venv/bin/pip-compile --generate-hashes --output-file=requirements-web.txt requirements-web.in
+	venv/bin/pip-compile --allow-unsafe --generate-hashes --output-file=requirements-test.txt requirements-test.in
+
+venv/bin/pip-compile:
+	# pip-compile>=6 fails with
+	#   "option --use-deprecated: invalid choice: 'legacy-resolver' (choose from )"
```
lcarva, line 71: Do we have a follow up issue for addressing this?

Makefile
chmeliik: Added. It is needed because we use `pkg_resources` in the pip code, and `setuptools` is what provides that package

Makefile
chmeliik: Not yet, in fact I can't seem to reproduce it anymore. I'll try to figure out what changed.

chmeliik: `pip < 20.3` does not support `--use-deprecated legacy-resolver`
`pip >= 20.3` does

pip-tools does specify this dependency properly: https://github.com/jazzband/pip-tools/blob/f49b3fa8aa1b5d8c6dda80f719eaae706feff6b0/setup.cfg#L34

I must have messed something up while testing. I removed the `<6` constraint.

chmeliik:
@akhmanova was the `mod-wsgi` dependency excluded intentionally, or by accident?

akhmanova (commented):
> @akhmanova was the `mod-wsgi` dependency excluded intentionally, or by accident?

@chmeliik Do you mean `cachito-worker`? I think by accident. Could you please add it as it was done with `cachito-api`?

chmeliik:
> @chmeliik Do you mean `cachito-worker`? I think by accident. Could you please add it as it was done with `cachito-api`?

I meant `requirements-text.txt` (see last commit)

akhmanova:
> I meant `requirements-text.txt` (see last commit)

@chmeliik No, it was not done by intention

chmeliik:
@lcarva what do you think about `mod-wsgi`? Should we make an exception an install it from an RPM? Do we even need it at all? Removing the dependency doesn't seem to make a difference and IIB doesn't use it either.

lcarva:
> @lcarva what do you think about `mod-wsgi`? Should we make an exception an install it from an RPM? Do we even need it at all? Removing the dependency doesn't seem to make a difference and IIB doesn't use it either.

Huh odd... I wouldn't think `mod-wsgi` is an application python module. That's the thing that loads the application, right? It sounds like it should be installed from RPM and removed from the pip requirements.

chmeliik:
@lcarva PTAL

lcarva (approved):

PR #390, opened 2021-03-23
Fix how JS code searches for Nexus components

chmeliik:
CLOUDBLD-4938

The Nexus search API does not allow the user to search specifically for
components without groups. If no group is specified, Nexus returns all
the matching components regardless of group.

In some cases, this causes Cachito to fail. For example, if a Yarn
package requires both `some-package` and `@types/some-package` (at the
same version), Cachito will eventually search for `some-package` and
fail because 2 different results are found.

Implement an ad-hoc workaround where the search result is filtered by
None afterwards. Keep the default behaviour the same as before (match
regardless of group).

### DIFF
--- a/cachito/workers/pkg_managers/general_js.py
+++ b/cachito/workers/pkg_managers/general_js.py
@@ -340,7 +340,7 @@ def _get_js_component_info_from_nexus(
         component_group = component_group[1:]
     else:
         component_name = name
-        component_group = None
+        component_group = nexus.NULL_GROUP
 
     return nexus.get_component_info_from_nexus(
         repository,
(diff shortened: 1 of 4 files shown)


chmeliik:
Tested with
```shell
    -X POST \
    -d '{
    "repo": "https://github.com/wtam2018/argo-cd",
    "ref": "8f8cee80523c31c3f399bb4b76dd7782dc969ed3",
    "pkg_managers": ["yarn"],
    "packages": {"yarn": [{"path": "ui"}]}
}'
```
I also checked `tox -e integration`, works fine

cachito/workers/pkg_managers/general_js.py
```diff
@@ -340,7 +340,7 @@ def _get_js_component_info_from_nexus(
         component_group = component_group[1:]
     else:
         component_name = name
-        component_group = None
+        component_group = nexus.NULL_GROUP
```
chmeliik, line 343: I hope this was the original intention and we're not breaking any existing workflows :thinking:

lcarva (approved):

PR #391, opened 2021-03-25
Set proper CA bundle for python requests

chmeliik:
The Fedora package patches requests like this:

```diff
-from certifi import where
+def where():
+    """Return the absolute path to the system CA bundle."""
+    return '/etc/pki/tls/certs/ca-bundle.crt'
```

Now that we are installing requests from PyPI, we will need to set the
REQUESTS_CA_BUNDLE variable to achieve the same results.


ejegrova (approved):
LGTM :)

PR #392, opened 2021-03-26
Integration tests: add Npm/Yarn with/without deps 

ejegrova:
Adding 4 integration tests:
- Npm without dependencies 
- Npm with dependencies 
- Yarn without dependencies
- Yarn with dependencies

### DIFF
--- a/tests/integration/test_data/npm_packages.yaml
+++ b/tests/integration/test_data/npm_packages.yaml
@@ -1,0 +1,191 @@
+# Test data for npm packages
+#
+# npm package without dependencies
+# repo: The URL for the upstream git repository
+# ref: A git reference at the given git repository
+# expected_files: Expected source files <relative_path>: <file_URL>
+# expected_deps_files: Expected dependencies files (empty)
+# response_expectations: Parts of the Cachito response to check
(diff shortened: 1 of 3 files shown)


tests/integration/test_data/npm_packages.yaml
```diff
@@ -0,0 +1,191 @@
...
+          - dev: false
+            name: "get-func-name"
+            replaces: null
+            type: "npm"
+            version: "2.0.0"
+          - dev: false
+            name: "fecha"
+            replaces: null
+            type: "npm"
+            version: "git+https://github.com/taylorhakes/fecha.git#e5f18494528622f7f3c37434b10c6b8d03c9597b"
```
chmeliik, line 69: For the yarn test, this one is a plain https url. I didn't find one for npm, should we change this one to be plain https?

ejegrova: Good catch, I thought that Plain URL will be used by test-npm-dep: https://github.com/cachito-testing/cachito-npm-with-deps/blob/master/package.json
I'll change fecha dep in the same way as it is in yarn test.

chmeliik (approved):
LGTM :+1:

PR #393, opened 2021-03-30
Check request state at the beginning of tasks

thrbkova:
Verify that request state is in the expected state (in_progress), otherwise log a warning.
This ensures the message will not be put back in the celery queue.

### DIFF
--- a/cachito/workers/tasks/general.py
+++ b/cachito/workers/tasks/general.py
@@ -12,6 +12,7 @@ from cachito.workers.config import get_worker_config
 from cachito.workers.scm import Git
 from cachito.workers.paths import RequestBundleDir
 from cachito.workers.tasks.celery import app
+from cachito.workers.tasks.utils import runs_if_request_in_progress
 
 __all__ = [
     "create_bundle_archive",
@@ -23,6 +24,7 @@ log = logging.getLogger(__name__)
 
 
 @app.task(priority=0)
+@runs_if_request_in_progress
 def fetch_app_source(url, ref, request_id, gitsubmodule=False):
     """
     Fetch the application source code that was requested and put it in long-term storage.
@@ -125,6 +127,7 @@ def set_request_state(request_id, state, state_reason):
 
 
 @app.task
+@runs_if_request_in_progress
 def failed_request_callback(context, exc, traceback, request_id):
     """
     Wrap set_request_state for task error callbacks.
@@ -142,6 +145,7 @@ def failed_request_callback(context, exc, traceback, request_id):
 
 
 @app.task(priority=10)
+@runs_if_request_in_progress
 def create_bundle_archive(request_id):
     """
     Create the bundle archive to be downloaded by the user.
--- a/cachito/workers/tasks/pip.py
+++ b/cachito/workers/tasks/pip.py
@@ -26,7 +26,7 @@ from cachito.workers.pkg_managers.pip import (
 )
 from cachito.workers.tasks.celery import app
 from cachito.workers.tasks.general import set_request_state
-from cachito.workers.tasks.utils import make_base64_config_file
+from cachito.workers.tasks.utils import make_base64_config_file, runs_if_request_in_progress
 
 
 log = logging.getLogger(__name__)
@@ -45,6 +45,7 @@ def cleanup_pip_request(request_id):
 
 
 @app.task
+@runs_if_request_in_progress
 def fetch_pip_source(request_id, package_configs=None):
     """
     Resolve and fetch pip dependencies for a given request.
--- a/cachito/workers/tasks/utils.py
+++ b/cachito/workers/tasks/utils.py
@@ -1,11 +1,19 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import base64
+import functools
+import logging
 from pathlib import Path
 from typing import Union, Callable
 
-from cachito.errors import ValidationError
+import requests
+
+from cachito.errors import ValidationError, CachitoError
+from cachito.workers.celery_logging import get_function_arg_value
+from cachito.workers.config import get_worker_config
 
 __all__ = ["make_base64_config_file", "AssertPackageFiles"]
+
+log = logging.getLogger(__name__)
 
 
 def make_base64_config_file(content: str, dest_relpath: Union[str, Path]) -> dict:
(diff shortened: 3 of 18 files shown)


chmeliik (commented):
Added suggestions for unit tests as discussed.

The state check will also have to be done in a bunch of other tasks, I think all the functions decorated with `@app.task` except for the cleanup ones will need it (not just in the general.py module).

It looks like this will be quite a few tasks, maybe implementing the decorator would be worth it?

cachito/workers/tasks/general.py
```diff
@@ -18,10 +18,48 @@
...
+def get_request_state(request_id):
+    """
+    Get the state of the request.
+
+    :param int request_id: the Cachito request ID this is for
+    """
+    from cachito.workers.requests import requests_auth_session
+
+    config = get_worker_config()
+    request_url = f'{config.cachito_api_url.rstrip("/")}/requests/{request_id}'
```
tkdchen, line 36: There is `_get_request_url` doing the same thing. It would be good to make it as public and reuse it to avoid duplicate.

cachito/workers/tasks/general.py
```diff
@@ -18,10 +18,48 @@
...
+
+    rv = requests_auth_session.get(request_url)
```
tkdchen, line 38: Is it expected to separate tasks from the Cachito internal and interact by the Web API?

I'm thinking the state change could reuse existing code as much as possible without the logical separation. For example:

```python
if Request.query.get_or_404(request_id).is_in_progress:
    return
```

On the other hand, if the separation is required, it would be good to have a Cachito lib wrapping the API, then we can:

```python
request = cachito.get_request(request_id)
if request.is_in_progress:
    return
```

cachito/workers/tasks/general.py
```diff
@@ -18,10 +18,48 @@
...
+        rv.raise_for_status()
+    except requests.RequestException:
+        msg = f"An error occured when getting the state of request {request_id}"
+        log.exception(msg)
+        raise CachitoError(msg)
+
+    return rv.json()["state"]
+
+
+def check_state(request_id):
```
tkdchen, line 50: For current implementation, suggest to rename this function name to `is_request_in_progress`.

cachito/workers/tasks/general.py
```diff
@@ -18,10 +18,48 @@
...
+    """
+    Check if the state of request is 'in_progress'.
+
+    :param int request_id: the Cachito request ID this is for
+    """
+    state = get_request_state(request_id)
+    if state != "in_progress":
+        log.error("The state of request %d is not in progress", request_id)
```
tkdchen, line 58: Instead of logging from here, would it be more clear like:

```
Skip fetching app source since the request is in progress.
Skip setting request state to %s with reason "%s", since the request is in progress.
```

cachito/workers/tasks/general.py
chmeliik: Tasks are executed on workers, workers do not have direct access to the database. That's why it's necessary to query the Cachito API.

chmeliik: I would leave the API wrapper for a separate story, @tkdchen you can file something if you want

cachito/workers/tasks/general.py
chmeliik: This is another thing that could be solved with the decorator (we could log the `__name__` of the function that is being skipped). @thrbkova I can hack together a proof of concept if you'd like

chmeliik:
@thrbkova It seems to me that almost all the changes in the first two commits were reverted in the third one. Would it make sense to squash all three commits into one?

cachito/workers/tasks/pip.py
```diff
@@ -45,6 +45,7 @@ def cleanup_pip_request(request_id):
 
 
 @app.task
+@runs_if_request_in_progress
 def fetch_pip_source(request_id, package_configs=None):
```
chmeliik, line 49: Like `fetch_pip_source`, all the other `fetch_*_source` tasks will need this decorator. Also `add_git_submodules_as_package`

cachito/workers/tasks/utils.py
```diff
@@ -97,3 +105,81 @@ def _assert(
...
+    :param task task_fn: the task to run if its state is in_progress
+    """
+
+    @functools.wraps(task_fn)
+    def task_with_state_check(*args, **kwargs):
+        """Check request state, proceed to execute task if check succeeds."""
+        request_id = get_function_arg_value("request_id", task_fn, args, kwargs)
+        if not request_id:
+            log.error(
+                "Skipping %s task because the request_id was not found", task_fn.__name__,
```
chmeliik, line 125: If there is no request_id, I'd say we want to raise an error (probably TypeError or ValueError) here. Using this decorator with a function that has no request_id would be a developer error.

cachito/workers/tasks/utils.py
```diff
@@ -97,3 +105,81 @@ def _assert(
...
+        rv.raise_for_status()
+    except requests.RequestException:
+        msg = f"An error occured when getting the state of request {request_id}"
+        log.exception(msg)
+        raise CachitoError(msg)
+
+    return rv.json()["state"]
+
+
+def get_function_arg_value(arg_name, func, args, kwargs):
```
chmeliik, line 165: Could you also update the `celery_logging` module to import this function instead of using the `_get_function_arg_value` one that's already defined there?

tests/test_workers/test_tasks/test_utils.py
```diff
@@ -112,3 +115,43 @@ def test_different_pkg_managers(self, pkg_manager, assert_files_testdir):
...
+
+    @utils.runs_if_request_in_progress
+    def dummy_task(request_id):
+        return 42
+
+    if id is None:
+        with pytest.raises(
+            ValueError, match="Failed during state check: no request_id found for dummy_task task"
+        ):
+            assert dummy_task(id) is None
```
chmeliik, line 148:
```suggestion
            dummy_task(id)
```
The `assert` will not do anything here because `dummy_task(id)` will raise an exception. In general, assertions in a `pytest.raises` block will not work because the block is exited as soon as the exception is raised.

Which is fine, you don't have to assert anything here, only check that an exception was raised. To make that clear, you could remove the `assert` as suggested.

chmeliik (commented):
Looks good, just a few nitpicks left :-)

Also, please squash the "mend" commit into the first one and fix the commit message.

cachito/workers/tasks/utils.py
```diff
@@ -124,15 +124,15 @@ def task_with_state_check(*args, **kwargs):
...
+            return
         request_state = get_request_state(request_id)
         if request_state != "in_progress":
             log.error(
                 "Skipping %s task because the request was not in_progress (was: %s)",
                 task_fn.__name__,
                 request_state,
             )
-            return None
+            return
```
chmeliik: Sorry, I was not specific enough, What I meant is that `return` after `raise` is dead code, e.g.
```python
def f():
    raise ValueError
    return None  # unreachable statement
```
`return None` by itself (when the return statement is reachable) is completely fine.

thrbkova: Haha sorry, my bad.

cachito/workers/tasks/utils.py
```diff
@@ -97,3 +105,57 @@ def _assert(
...
+    return task_with_state_check
+
+
+def get_request_state(request_id):
+    """
+    Get the state of the request.
+
+    :param int request_id: the Cachito request ID this is for
+    """
+    from cachito.workers.requests import requests_auth_session
```
chmeliik, line 147: This could be the regular `requests_session` (no auth)

cachito/workers/tasks/utils.py
```diff
@@ -97,3 +105,57 @@ def _assert(
...
+
+    config = get_worker_config()
+    request_url = f'{config.cachito_api_url.rstrip("/")}/requests/{request_id}'
+
+    rv = requests_auth_session.get(request_url)
+
+    try:
+        rv.raise_for_status()
+    except requests.RequestException:
```
chmeliik, line 156: The `get()` method should be called in the try/except block.

`get()` will raise an exception if the connection fails. `raise_for_status()` will raise an exception if the connection works fine but the status code in the response is between 400 and 599.

chmeliik (commented):
Nice! I found 2 final things to nitpick about, but I'm *almost* sure that's everything :P

PR #411, opened 2021-04-07
Improve dev environment compatibility with podman

chmeliik:
These changes should allow us to run docker-compose with podman without
needing podman-compose.

See https://www.redhat.com/sysadmin/podman-docker-compose.


chmeliik:
I could not check if it still works with docker, would appreciate if someone could test

nirzari (approved):

PR #412, opened 2021-04-08
Check PKG-INFO metadata existence in sdist got from PyPI

tkdchen:
CLOUDBLD-2927

This patch ensures the sdist package contains the PKG-INFO metadata. If
no such metadata exists, the whole request will fail. This check only
applies to the sdist package downloaded from PyPI.


chmeliik (commented):
Nice! I left a couple suggestions, but overall I like it :-)

Btw, feel free to add type annotations when updating or adding functions, we've started adding them in newer parts of the codebase.

cachito/workers/pkg_managers/pip.py
```diff
@@ -2175,3 +2185,67 @@ def get_raw_component_name(requirement):
...
+            )
+    except tarfile.ReadError as e:
+        raise ValidationError(f"Cannot open {sdist_path} as a Tar file. Error: {str(e)}")
+    except zipfile.BadZipFile:
+        raise ValidationError(f"Cannot open {sdist_path} as a Zip file. Error: {str(e)}")
+    except subprocess.CalledProcessError as e:
+        raise ValidationError(f"Cannot uncompress {sdist_path} to validate. Error: {str(e)}")
+    if not any(map(is_pkg_info_dir, files_iter)):
+        filename = os.path.basename(sdist_path)
+        raise ValidationError(f"{filename} does not include metadata.")
```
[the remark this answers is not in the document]
chmeliik: > It would be awesome if the error message could include the suggested download url.

Nevermind, I forgot that the Cachito download url points to Nexus, not to PyPI. Telling users to get the package from git or a direct download url will have to be good enough.

cachito/workers/pkg_managers/pip.py
```diff
@@ -2175,3 +2185,67 @@ def get_raw_component_name(requirement):
...
+        return zf.namelist()
+
+
+def iter_tar_z(filename):
+    # uncompress file > another_file
+    with tempfile.TemporaryDirectory() as work_dir:
+        test_file = os.path.join(work_dir, os.path.basename(filename))
+        with open(test_file, "wb") as f:
+            subprocess.check_call(["uncompress", "-c", filename], stdout=f)
+        return iter_tar_file(test_file)
```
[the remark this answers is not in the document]
tkdchen: Don't handle `.tar.Z` now, and a warning will be logged if there is really a special case using the .Z before `.tar.Z` is removed from Python packaging and Cachito.

cachito/workers/pkg_managers/pip.py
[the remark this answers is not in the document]
tkdchen: Updated the message. But, not sure if the update is accurate or informative enough. PTAL.

cachito/workers/pkg_managers/pip.py
```diff
@@ -2175,3 +2185,67 @@ def get_raw_component_name(requirement):
...
+            files_iter = iter_zip_file(sdist_path)
+        elif sdist_path.endswith(COMPRESSED_TAR_EXT):
+            files_iter = iter_tar_z(sdist_path)
+        elif any(map(sdist_path.endswith, SDIST_FILE_EXTENSIONS)):
+            files_iter = iter_tar_file(sdist_path)
+        else:
+            raise ValidationError(
+                f"Cannot check metadata from {sdist_path}, "
+                f"which does not have a known supported extension.",
+            )
```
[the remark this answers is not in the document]
tkdchen: Thanks for the info. I think it would be good to keep it there :) so that the function can cover all cases for itself.

cachito/workers/pkg_managers/pip.py
```diff
@@ -2175,3 +2181,60 @@ def get_raw_component_name(requirement):
...
+        )
+
+    try:
+        if not any(map(is_pkg_info_dir, files_iter)):
+            raise ValidationError(
+                f"{sdist_path.name} does not include metadata. "
+                f"Please download from the package from git or a direct download URL."
+            )
+    except tarfile.ReadError as e:
+        raise ValidationError(f"Cannot open {sdist_path} as a Tar file. Error: {str(e)}")
```
chmeliik, line 2240: It's a bit odd that `_iter_zip_file()` raises `zipfile.BadZipFile` as soon as you call it but `_iter_tar_file()` raises the error when your iterate through the result. Could `_iter_tar_file()` be changed to behave the same way?

cachito/workers/pkg_managers/pip.py
```diff
@@ -2175,3 +2181,60 @@ def get_raw_component_name(requirement):
...
+
+
+def check_metadata_in_sdist(sdist_path: Path):
+    """Check if a downloaded sdist package has metadata.
+
+    :param sdist_path: the path of a sdist package file.
+    :type sdist_path: pathlib.Path
+    :raise ValidationError: if the sdist cannot be validated.
+    """
+    if str(sdist_path).endswith(ZIP_FILE_EXT):
```
chmeliik, line 2217: This could be `sdist_path.name.endswith`, same for the lines below

cachito/workers/pkg_managers/pip.py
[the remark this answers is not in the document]
chmeliik: I would go with something like this:

{sdist_path.name} does not include metadata (there is no PKG-INFO file). It is not a valid sdist and cannot be downloaded from PyPI. Consider editing your requirements file to download the package from git or a direct download URL instead.

cachito/workers/pkg_managers/pip.py
tkdchen: The major consideration behind the current behavior of `_iter_tar_file()` is the comment[1]. I originally wanted the same iteration behavior for zip file, but AFAIK the ZipFile does not support it.

[1] https://github.com/pypa/warehouse/pull/5609/files#diff-bffd49e4f943217b54b4cfcf5835737bec7a145360bd1dc2ec1541d2dcafad16R587

cachito/workers/pkg_managers/pip.py
tkdchen: Fixed

cachito/workers/pkg_managers/pip.py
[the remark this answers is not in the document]
tkdchen: It's clear. Just update it to the code. Thanks.

cachito/workers/pkg_managers/pip.py
chmeliik: Ah, that makes sense. So an invalid tar file will *only* raise an error when you start to iterate it? Is it possible that `tarfile.open()` will raise the error? (that would currently be uncaught)

tests/test_workers/test_pkg_managers/test_pip.py
```diff
@@ -17,6 +18,7 @@
 from tests.helper_utils import write_file_tree
 
 
+THIS_MODULE_DIR = Path(os.path.dirname(os.path.realpath(__file__)))
```
chmeliik, line 21: You could do this without `os.path`:
```suggestion
THIS_MODULE_DIR = Path(__file__).resolve().parent
```

cachito/workers/pkg_managers/pip.py
tkdchen: Just had a quick look. The error is raised just when start the iteration, when us tarfile object as a context manager object. Calling `tarfile.open` directly will raise the error immediately.

tkdchen: Actually function `_iter_tar_file` starts to execute only when the iteration happens. Due to the aforementioned ability of ZipFile, the two `_iter-*` functions returns different type of value, one is a list and another one is iterator.

chmeliik: Oh, that's true! Would it make sense to move the try/except blocks into the `_iter_*` functions? That way, the caller doesn't have to worry about the list vs. generator difference

chmeliik: Alternatively, you could make the zip one a generator with `yield from zf.namelist()`

chmeliik (approved):
Nice! :rocket: 

Tested locally with this request, works great
```shell
curl  http://localhost:8080/api/v1/requests -X POST -H 'content-type: application/json' -d '{
          "repo": "https://github.com/chmeliik/cachito-sample-pip-package",
          "ref": "b8dd1dad9b534aba44b4a511e07863f9bdcb38fa",
          "pkg_managers": ["pip"]
        }'
```
```shell
curl localhost:8080/api/v1/requests/1 | jq '.state_reason'
"PyPDF2-1.26.0.tar.gz does not include metadata (there is no PKG-INFO file). It is not a valid sdist and cannot be downloaded from PyPI. Consider editing your requirements file to download the package from git or a direct download URL instead."
```

PR #413, opened 2021-04-08
Fix how gomod checks if module is a parent module

chmeliik:
Previously, the check could have produced false positives.


chmeliik:
Loosely related to CLOUDBLD-3365

nirzari (approved):
Assuming subpackage name and package name having same values is legit allowed, LGTM

chmeliik:
> Assuming subpackage name and package name having same values is legit allowed, LGTM

At that point they are the same package, which is supposed to count as true. Maybe the function needs a better name though?

chmeliik:
@nirzari renamed the function to better reflect purpose

lcarva (approved):

PR #421, opened 2021-04-14
New API endpoint /content-manifest

tkdchen:
CLOUDBLD-4451

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 from collections import OrderedDict
+from copy import deepcopy
 import copy
 import functools
 import os
@@ -9,10 +10,11 @@ import flask
 from flask import stream_with_context
 from flask_login import current_user, login_required
 import kombu.exceptions
-from werkzeug.exceptions import Forbidden, InternalServerError, Gone, NotFound
+from werkzeug.exceptions import Forbidden, InternalServerError, Gone, NotFound, BadRequest
 
 from cachito.errors import CachitoError, ValidationError
 from cachito.web import db
+from cachito.web.content_manifest import BASE_ICM
 from cachito.web.models import (
     ConfigFileBase64,
     Dependency,
@@ -23,7 +25,7 @@ from cachito.web.models import (
     RequestStateMapping,
 )
 from cachito.web.status import status
-from cachito.web.utils import pagination_metadata, str_to_bool
+from cachito.web.utils import deep_sort_icm, pagination_metadata, str_to_bool
 from cachito.workers import tasks
 from cachito.paths import RequestBundleDir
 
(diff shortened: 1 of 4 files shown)


tests/test_api_v1.py
```diff
@@ -2249,3 +2251,82 @@ def test_validate_package_manager_exclusivity(
...
+    db.session.add(pkg2)
+    db.session.add(pkg3)
+
+    request1.add_package(pkg1)
+    request2.add_package(pkg2)
+    request3.add_package(pkg3)
+
+    db.session.commit()
+
+    test_data = [
```
akhmanova, line 2300: LGTM! I have only two suggestions about `data` & `test_data`: 
1. Can we move all `data` to `@pytest.mark.parametrize`?
2. If it's possible, we can move `test_data` to `@pytest.mark.parametrize` as well, and update only request ids. (perhaps it doesn't worth it).

akhmanova (changes requested):

tests/test_api_v1.py
tkdchen: By moving to the `pytest.mark.parametrize`, the test_data cannot reference the `request.id` created in the test body.

akhmanova: Yes, I see. It is why I mentioned that it may not worth moving `test_data`, and fill request ids later. Let's keep `test_data` in the test body :)

cachito/web/api_v1.py
```diff
@@ -588,3 +588,40 @@ def get_request_logs(request_id):
     return flask.Response(
         stream_with_context(generate_stream_response(log_file_path)), mimetype="text/plain"
     )
+
+
+@api_v1.route("/requests/content-manifest", methods=["GET"])
```
chmeliik, line 593: This should be just `/content-manifest`

cachito/web/api_v1.py
```diff
@@ -588,3 +588,40 @@ def get_request_logs(request_id):
...
+    Retrieve the content manifest associated with the given requests.
+
+    :return: a Flask JSON response
+    :rtype: flask.Response
+    :raise NotFound: if one of the given request is not found
+    :raise ValidationError: if any of the given request is not in the "complete" or "stale" state
+    """
+    arg = flask.request.args.get("requests")
+    if not arg:
+        return flask.jsonify([])
```
chmeliik, line 605: This should be an empty content manifest, i.e. the return value of `ContentManifest.generate_icm(image_contents=None)`

cachito/web/api_v1.py
```diff
@@ -588,3 +588,40 @@ def get_request_logs(request_id):
...
+        if request.state.state_name not in (
+            RequestStateMapping.complete.name,
+            RequestStateMapping.stale.name,
+        ):
+            raise BadRequest(
+                f"Request {request.id} is in state {request.state.state_name}, "
+                f"not complete or stale."
+            )
+        item_contents.extend(request.content_manifest.to_json()["image_contents"])
+    return flask.jsonify(item_contents)
```
chmeliik, line 627: This should be the full content manifest, not just the image contents.

cachito/web/api_v1.py
```diff
@@ -588,3 +588,40 @@ def get_request_logs(request_id):
...
+
+@api_v1.route("/requests/content-manifest", methods=["GET"])
+def get_content_manifest_by_requests():
+    """
+    Retrieve the content manifest associated with the given requests.
+
+    :return: a Flask JSON response
+    :rtype: flask.Response
+    :raise NotFound: if one of the given request is not found
+    :raise ValidationError: if any of the given request is not in the "complete" or "stale" state
```
chmeliik, line 601: This would be a `BadRequest`, right?

cachito/web/api_v1.py
```diff
@@ -588,3 +588,40 @@ def get_request_logs(request_id):
...
+        return flask.jsonify([])
+    request_ids = []
+    item: str
+    for item in arg.split(","):
+        if not item.strip():
+            continue
+        if not item.strip().isdigit():
+            raise BadRequest(f"{item} is not an integer.")
+        request_ids.append(int(item))
+    requests = Request.query.filter(Request.id.in_(request_ids))
```
chmeliik, line 614: Is this guaranteed to return the requests in a specific order? We should make sure that the generated content manifest will always be in the same order. IMO the response should also be the same for `/content-manifest?requests=1,2,3` and `/content-manifest?requests=3,2,1`. Maybe we can just `deep_sort_icm()` The final content manifest.

cachito/web/api_v1.py
```diff
@@ -588,3 +590,41 @@ def get_request_logs(request_id):
...
+    for request in requests:
+        if request.state.state_name not in (
+            RequestStateMapping.complete.name,
+            RequestStateMapping.stale.name,
+        ):
+            raise BadRequest(
+                f"Request {request.id} is in state {request.state.state_name}, "
+                f"not complete or stale."
+            )
+        assembled_icm["image_contents"].extend(request.content_manifest.to_json()["image_contents"])
```
chmeliik, line 629: @tkdchen as we discussed, I've been thinking about removing duplicates and unfortunately, I don't think de-duplicating by `purl` would be enough. There are various cases in which different packages can have the same purl.

For example, if your repo contains both a `pip` package and an `npm` package in the root directory, their purls will both be `pkg:github/<repo>@<ref>`.
Other example, if two different repos contain gomod packages with exactly the same name and version, their purls will also be the same. This *shouldn't* happen, but nothing actually prevents the users from doing it.
Both of these are cases are bad and Cachito should handle them better, but purls in Cachito are kind of a mess at the moment.

We can remove duplicates only if all the attributes (`purl`, `dependencies`, `sources`) are exactly the same. Maybe that would be worth doing?

cachito/web/api_v1.py
```diff
@@ -588,3 +590,41 @@ def get_request_logs(request_id):
...
+    request_ids = []
+    item: str
+    for item in arg.split(","):
+        if not item.strip():
+            continue
+        if not item.strip().isdigit():
+            raise BadRequest(f"{item} is not an integer.")
+        request_ids.append(int(item))
+
+    requests = Request.query.filter(Request.id.in_(request_ids))
```
chmeliik, line 617: This will ignore missing requests, but the docstring says that missing requests will cause a `NotFound` error (and it should raise the error IMO).

chmeliik (approved):
LGTM, nice!

PR #422, opened 2021-04-15
Set Cachito upstream versions

ejegrova:
Setting the version with https://pypi.org/project/setuptools-git-versioning/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,6 @@ from setuptools import setup, find_packages
 
 setup(
     name="cachito",
-    version="1.0",
     long_description=__doc__,
     packages=find_packages(),
     include_package_data=True,
@@ -43,4 +42,8 @@ setup(
     ],
     license="GPLv3+",
     python_requires=">=3.6",
+    setup_requires=['setuptools-git-versioning'],
+    version_config={
+        "dev_template": "{tag}.post{ccount}-git.{sha}",
+    },
 )


setup.py
```diff
@@ -43,4 +43,8 @@
     ],
     license="GPLv3+",
     python_requires=">=3.6",
+    setup_requires=['setuptools-git-versioning'],
+    version_config={
+        "dirty_template": "{tag}.rc{ccount}-git.{sha}",
```
ejegrova, line 48: This is giving us version in form: `sprint-25.rc2-git.5ec90cf7`, there's a difference between this and what was suggested in jira story: `sprint25.rc1+git.77269de5`. When trying out the plus sign before git in the template, it gives me a dash instead of a plus as a result. Do you think it's ok to have a dash here?

chmeliik: Doesn't matter IMO, it's not going to be a valid PyPI version either way (and that's fine).

BTW, what exactly is the difference between `dirty_template` and `dev_template`?

ejegrova: The template is reflecting the state of repo. It will first check which template to use and then if there is `version_config` for that template it will change the string accordingly. I'll change it to `dev_template` because `dirty_template` is expecting unstashed changes (which was case for my local repo).

setup.py
```diff
@@ -43,4 +43,8 @@
     ],
     license="GPLv3+",
     python_requires=">=3.6",
+    setup_requires=['setuptools-git-versioning'],
+    version_config={
+        "dev_template": "{tag}.rc{ccount}-git.{sha}",
+    },
```
tkdchen, line 49: Is the original issue "Set Cachito upstream versions" only for setting the version before a new release?

By using setuptools-git-versioning, should we also remove the `version=`? When releasing a new version, after tagging the new version by `git tag sprintN`, the version can be generated by the tool automatically.

setup.py
chmeliik: It wasn't working for me until I found out that `dirty_template` also applies if you have untracked files. When I fixed that, everything worked as expected :+1:

setup.py
```diff
@@ -43,4 +43,8 @@
     ],
     license="GPLv3+",
     python_requires=">=3.6",
+    setup_requires=['setuptools-git-versioning'],
+    version_config={
+        "dev_template": "{tag}.rc{ccount}-git.{sha}",
```
chmeliik, line 48: What do you think about using `post` (the default) instead of `rc`? Rc stands for release candidate and normally, `sprint-N.rc` should come *before* `sprint-N`. Post would make more sense to me.

setup.py
ejegrova: Not sure if we'll be setting the version only before release. Version will be used when building images and I'm not sure if that's gonna happen only when we're releasing or every time new commit is merged to master. @chmeliik 

Either way, I'm removing `version=` line.

setup.py
ejegrova: It does make more sence to me, too.

chmeliik (approved):
LGTM :+1:

setup.py
chmeliik: The end goal is that every commit merged to master will trigger an automatic deployment and yeah, building images is part of that.

PR #425, opened 2021-04-16
Check for changes after vendoring go deps

chmeliik:
CLOUDBLD-3282

When the gomod-vendor-check flag is used, Cachito will check the vendor
directory after running `go mod vendor`. If Cachito made any changes,
the request will fail with a ValidationError.

### DIFF
--- a/cachito/web/migrations/versions/c6ac095d8e9f_add_gomod_vendor_check_flag.py
+++ b/cachito/web/migrations/versions/c6ac095d8e9f_add_gomod_vendor_check_flag.py
@@ -1,0 +1,43 @@
+"""Add the gomod-vendor-check flag
+
+Revision ID: c6ac095d8e9f
+Revises: 07d89c5778f2
+Create Date: 2021-04-15 13:45:44.761284
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "c6ac095d8e9f"
+down_revision = "07d89c5778f2"
+branch_labels = None
+depends_on = None
+
+
+flag_table = sa.Table(
+    "flag",
+    sa.MetaData(),
+    sa.Column("id", sa.Integer(), primary_key=True),
+    sa.Column("name", sa.String(), nullable=False),
+    sa.Column("active", sa.Boolean(), nullable=False, default=True),
+)
+
+
+def upgrade():
+    connection = op.get_bind()
+    res = connection.execute(
+        flag_table.select().where(flag_table.c.name == "gomod-vendor-check")
+    ).fetchone()
+    if res is None:
+        connection.execute(flag_table.insert().values(name="gomod-vendor-check", active=True))
+    else:
+        connection.execute(
+            flag_table.update().where(flag_table.c.name == "gomod-vendor-check").values(active=True)
+        )
+
+
+def downgrade():
+    connection = op.get_bind()
+    connection.execute(flag_table.update().values(name="gomod-vendor-check", active=False))
--- a/cachito/workers/pkg_managers/gomod.py
+++ b/cachito/workers/pkg_managers/gomod.py
@@ -14,7 +14,7 @@ from typing import Tuple, List, Iterable, Optional
 import git
 import semver
 
-from cachito.errors import CachitoError
+from cachito.errors import CachitoError, ValidationError
 from cachito.workers.config import get_worker_config
 from cachito.workers.paths import RequestBundleDir
 from cachito.workers.pkg_managers.general import run_cmd
@@ -151,16 +151,11 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
             )
         # Vendor dependencies if the gomod-vendor flag is set
         flags = request.get("flags", [])
-        if "gomod-vendor" in flags:
-            log.info("Vendoring the gomod dependencies")
-            run_gomod_cmd(("go", "mod", "vendor"), run_params)
-        elif worker_config.cachito_gomod_strict_vendor and os.path.isdir(
-            os.path.join(app_source_path, "vendor")
-        ):
-            raise CachitoError(
-                'The "gomod-vendor" flag must be set when your repository has vendored'
-                " dependencies."
-            )
+        should_vendor, can_make_changes = _should_vendor_deps(
+            flags, app_source_path, worker_config.cachito_gomod_strict_vendor
+        )
+        if should_vendor:
+            _vendor_deps(run_params, can_make_changes, git_dir_path)
         else:
             log.info("Downloading the gomod dependencies")
             run_gomod_cmd(("go", "mod", "download"), run_params)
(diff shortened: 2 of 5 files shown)


chmeliik:
Tested with
```shell
    -X POST \
    -d '{
"repo": "https://github.com/cachito-testing/gomod-vendored.git",
"ref": "3604a4d7990f5c9279f26dda783b7b9bce617ed4",
"flags": ["gomod-vendor-check"],
"pkg_managers": ["gomod"]
}'
```

cachito/web/migrations/versions/c6ac095d8e9f_add_gomod_vendor_check_flag.py
```diff
@@ -0,0 +1,43 @@
+"""Add the gomod-vendor-check flag
```
akhmanova, line 1: It was autogenerated with `cachito db migrate`, right?

chmeliik: Autogenerated, renamed and edited, yeah. The title, `flag_table`, `upgrade()` and `downgrade()` were edited manually.

cachito/workers/pkg_managers/gomod.py
```diff
@@ -316,6 +311,82 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
...
+    """
+    Determine if Cachito should vendor dependencies and if it is allowed to make changes.
+
+    This is based on the presence of flags:
+    - gomod-vendor-check => should vendor, can only make changes if vendor dir does not exist
+    - gomod-vendor => should vendor, can make changes
+
+    :param flags: flags from the Cachito request
+    :param app_dir: absolute path to the app directory
+    :param strict: fail the request if the vendor dir is present but the flags are not?
```
akhmanova, line 324: I am curious why does it have `?` in the end :D

akhmanova: Oh, I see! So `true/false` will be the answer. Great idea!

cachito/workers/pkg_managers/gomod.py
```diff
@@ -316,6 +311,82 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
...
+    :raise ValidationError: if the vendor dir is present, the flags are not and we are strict
+    """
+    vendor = Path(app_dir) / "vendor"
+
+    if "gomod-vendor-check" in flags:
+        return True, not vendor.exists()
+    elif "gomod-vendor" in flags:
+        return True, True
+
+    if strict and vendor.is_dir():
```
tkdchen, line 335: If `worker_config.cachito_gomod_strict_vendor` is False, Cachito will run `go mod download` eventually. What and how will it affect a repository having the `vendor` directory?

cachito/workers/pkg_managers/gomod.py
```diff
@@ -316,6 +311,82 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
...
+    Determine if Cachito should vendor dependencies and if it is allowed to make changes.
+
+    This is based on the presence of flags:
+    - gomod-vendor-check => should vendor, can only make changes if vendor dir does not exist
+    - gomod-vendor => should vendor, can make changes
+
+    :param flags: flags from the Cachito request
+    :param app_dir: absolute path to the app directory
+    :param strict: fail the request if the vendor dir is present but the flags are not?
+    :return: (should vendor: bool, allowed to make changes: bool)
```
tkdchen, line 325: Make what changes?

cachito/workers/pkg_managers/gomod.py
```diff
@@ -316,6 +311,82 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
...
+    :raise ValidationError: if the vendor dir is present, the flags are not and we are strict
```
tkdchen, line 326: > the flags are not and we are strict

Missing something after "flags are not"?

cachito/workers/pkg_managers/gomod.py
```diff
@@ -316,6 +311,82 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
...
+
+    :param run_params: common params for the subprocess calls to `go`
+    :param can_make_changes: is Cachito allowed to make changes?
+    :param git_dir: path to the repository root
+    :raise ValidationError: if vendor directory changed and Cachito is not allowed to make changes
+    """
+    log.info("Vendoring the gomod dependencies")
+    run_gomod_cmd(("go", "mod", "vendor"), run_params)
+    app_dir = run_params["cwd"]
+    if not can_make_changes and _vendor_changed(git_dir, app_dir):
```
tkdchen, line 359: Can you help me understand what the purpose of variable `can_make_changes`?

cachito/workers/pkg_managers/gomod.py
```diff
@@ -316,6 +311,82 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
...
+            "The content of the vendor directory is not consistent with go.mod. Run "
+            "`go mod vendor` locally to fix this problem. See the logs for more details."
+        )
+
+
+def _vendor_changed(git_dir: str, app_dir: str) -> bool:
+    """Check for changes in the vendor directory."""
+    repo = git.Repo(git_dir)
+    # Add untracked files but do not stage them
+    repo.git.add("--intent-to-add", "--force", "--", app_dir)
```
tkdchen, line 370: Is it necessary to revert this `git add` operation at the end of this function?

cachito/workers/pkg_managers/gomod.py
```diff
@@ -316,6 +311,82 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
...
+    :raise ValidationError: if vendor directory changed and Cachito is not allowed to make changes
+    """
+    log.info("Vendoring the gomod dependencies")
+    run_gomod_cmd(("go", "mod", "vendor"), run_params)
+    app_dir = run_params["cwd"]
+    if not can_make_changes and _vendor_changed(git_dir, app_dir):
+        raise ValidationError(
+            "The content of the vendor directory is not consistent with go.mod. Run "
+            "`go mod vendor` locally to fix this problem. See the logs for more details."
+        )
```
tkdchen, line 363: Why not just run these lines inside the `if` directly? Any consideration of putting these lines into a separate function?

cachito/workers/pkg_managers/gomod.py
chmeliik: If `can_make_changes` is True, then it is OK for Cachito to change the content of the `vendor/` directory. Otherwise, Cachito has to check that calling `go mod vendor` did not change anything.

cachito/workers/pkg_managers/gomod.py
chmeliik: Mostly just to make unit testing easier. The `resolve_gomod` function is huge and adding unit tests for it is always a mess.

cachito/workers/pkg_managers/gomod.py
chmeliik: That's a good point. It's not strictly necessary, because if this operation adds anything, the request is guaranteed to fail, but this function does not really know that. I'll add a revert.

cachito/workers/pkg_managers/gomod.py
chmeliik: Changes in the vendor directory, I updated the docstring. Is that good enough or does it need better explanation?

cachito/workers/pkg_managers/gomod.py
chmeliik: I meant it as "flags are not present", but I changed it to "flags are not used"

cachito/workers/pkg_managers/gomod.py
chmeliik: I believe `go mod download` does not make changes in the repository. I could be wrong, but that's my understanding.

cachito/workers/pkg_managers/gomod.py
chmeliik: Fixed

akhmanova (approved):
LGTM. I believe without Athos reviews can take a long time. Sorry for it :)

chmeliik:
@akhmanova no problem, thanks for the review :)

@tkdchen have I addressed all of your concerns?

tkdchen (approved):
Sorry for late response. LGTM.

PR #428, opened 2021-04-19
Search requests by repo, ref and pkg_manager

tkdchen:

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -10,6 +10,7 @@ import flask
 from flask import stream_with_context
 from flask_login import current_user, login_required
 import kombu.exceptions
+from sqlalchemy import func
 from werkzeug.exceptions import Forbidden, InternalServerError, Gone, NotFound, BadRequest
 
 from cachito.errors import CachitoError, ValidationError
@@ -20,9 +21,11 @@ from cachito.web.models import (
     Dependency,
     EnvironmentVariable,
     Package,
+    PackageManager,
     Request,
     RequestState,
     RequestStateMapping,
+    is_request_ref_valid,
 )
 from cachito.web.status import status
 from cachito.web.utils import deep_sort_icm, pagination_metadata, str_to_bool
@@ -74,6 +77,32 @@ def get_requests():
         state_int = RequestStateMapping.__members__[state].value
         query = query.join(RequestState, Request.request_state_id == RequestState.id)
         query = query.filter(RequestState.state == state_int)
+    repo = flask.request.args.get("repo")
+    if repo:
+        query = query.filter(Request.repo == repo)
+    ref = flask.request.args.get("ref")
+    if ref:
+        if not is_request_ref_valid(ref):
+            raise ValidationError(f"{ref} is not a valid ref.")
+        query = query.filter(Request.ref == ref)
+    pkg_managers = flask.request.args.getlist("pkg_manager")
+    if pkg_managers:
+        pkg_manager_ids = []
+        for name in pkg_managers:
+            if not name:
+                # Ignore if pkg_manager= presents in the querystring
+                continue
+            pkg_manager: PackageManager = PackageManager.get_by_name(name)
+            if pkg_manager is None:
+                raise ValidationError(f"Cachito does not have package manager {name}.")
+            pkg_manager_ids.append(pkg_manager.id)
+        if pkg_manager_ids:
+            query = (
+                query.join(PackageManager, Request.pkg_managers)
+                .filter(PackageManager.id.in_(pkg_manager_ids))
+                .group_by(Request.id)
+                .having(func.count(PackageManager.id) == len(pkg_manager_ids))
+            )
     try:
         per_page = int(flask.request.args.get("per_page", 10))
     except ValueError:
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -2,6 +2,7 @@
 from collections import OrderedDict
 from copy import deepcopy
 from enum import Enum
+import functools
 import itertools
 import os
 import re
@@ -17,6 +18,11 @@ from werkzeug.exceptions import Forbidden
 from cachito.web import content_manifest
 from cachito.errors import ContentManifestError, ValidationError
 from cachito.web import db
+
+
+def is_request_ref_valid(ref: str) -> bool:
+    """Check if a string is a valid git ref in the expected format."""
+    return re.match(r"^[a-f0-9]{40}$", ref) is not None
 
 
 request_pkg_manager_table = db.Table(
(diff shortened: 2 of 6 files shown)


cachito/web/api_v1.py
```diff
@@ -74,6 +76,18 @@ def get_requests():
...
+        query = query.filter(Request.repo == repo)
+    ref = flask.request.args.get("ref")
+    if ref:
+        if not is_request_ref_valid(ref):
+            raise ValidationError(f"{ref} is not a valid ref.")
+        query = query.filter(Request.ref == ref)
+    pkg_managers = flask.request.args.getlist("pkg_manager")
+    if pkg_managers:
+        query = query.join(PackageManager, Request.pkg_managers)
+        query = query.filter(PackageManager.name.in_(pkg_managers))
```
chmeliik, line 90: When multiple `pkg_manager` values are used, the endpoint should return requests that use *all* of the specified package managers. If I'm not mistaken, this will find requests with *any* of the package managers.

cachito/web/models.py
```diff
@@ -19,6 +19,11 @@
 from cachito.web import db
 
 
+def is_request_ref_valid(ref: str) -> bool:
+    """Check if a string is valid in the expected format."""
```
chmeliik, line 23:
```suggestion
    """Check if a string is a valid git ref in the expected format."""
```

cachito/web/api_v1.py
```diff
@@ -74,6 +77,28 @@ def get_requests():
...
+            raise ValidationError(f"{ref} is not a valid ref.")
+        query = query.filter(Request.ref == ref)
+    pkg_managers = flask.request.args.getlist("pkg_manager")
+    if pkg_managers:
+        qr = (
+            PackageManager.query.filter(PackageManager.name.in_(pkg_managers))
+            .with_entities(PackageManager.id)
+            .all()
+        )
+        pkg_manager_ids = [pm_id for pm_id, in qr]
```
chmeliik, line 95: Could this be a method of the `PackageManager` class? Something like `pkg_manager_ids = [PackageManager.get_id(pm_name) for pm_name in pkg_managers]`. You could even decorate that method with `functools.cache` to make sure the DB will never be queried more than once for the same package manager.

cachito/web/api_v1.py
```diff
@@ -74,6 +77,28 @@ def get_requests():
...
+    repo = flask.request.args.get("repo")
+    if repo:
+        query = query.filter(Request.repo == repo)
+    ref = flask.request.args.get("ref")
+    if ref:
+        if not is_request_ref_valid(ref):
+            raise ValidationError(f"{ref} is not a valid ref.")
+        query = query.filter(Request.ref == ref)
+    pkg_managers = flask.request.args.getlist("pkg_manager")
+    if pkg_managers:
```
MartinBasti, line 89: can be `repo`, `ref` and `pkg_managers` defined/used together?

tkdchen: @MartinBasti Yes.

MartinBasti: ah yes, nvm :D

cachito/web/models.py
```diff
@@ -1073,6 +1079,12 @@ def get_pkg_managers(cls, pkg_managers):
 
         return found_pkg_managers
 
+    @classmethod
+    @functools.cache
```
chmeliik, line 1083: This is new in python 3.9, my bad. I think `lru_cache(maxsize=None)` does the same thing.

cachito/web/models.py
```diff
@@ -1073,6 +1079,12 @@ def get_pkg_managers(cls, pkg_managers):
...
+    def get_by_name(cls, name: str):
```
chmeliik, line 1084: Could we have a unit test for this?

cachito/web/models.py
tkdchen: Add py38 to tox testenv as well?

cachito/web/models.py
```diff
@@ -647,8 +653,8 @@ class Request(db.Model):
     """A Cachito user request."""
 
     id = db.Column(db.Integer, primary_key=True)
-    repo = db.Column(db.String, nullable=False)
-    ref = db.Column(db.String, nullable=False)
+    repo = db.Column(db.String, nullable=False, index=True)
+    ref = db.Column(db.String, nullable=False, unique=True)
```
chmeliik, line 657: I don't think we can use a unique constraint for `ref`. You can submit a request for the same `repo` and `ref` twice, which will result in two different rows in the db. Example:
```shell
for i in 1 2; do
    -X POST \
    -d '{
      "repo": "https://github.com/cachito-testing/cachito-pip-without-deps.git",
      "ref": "fb07c8492432631d557aba15e1715af9cef9c844",
      "pkg_managers": [
        "pip"
      ]
    }'
done

docker-compose exec db psql cachito -U cachito -c 'select * from request'
```
```
 id |                              repo                               |                   ref                    | user_id | submitted_by_id | request_state_id 
----+-----------------------------------------------------------------+------------------------------------------+---------+-----------------+------------------
  1 | https://github.com/cachito-testing/cachito-pip-without-deps.git | fb07c8492432631d557aba15e1715af9cef9c844 |         |                 |                8
  2 | https://github.com/cachito-testing/cachito-pip-without-deps.git | fb07c8492432631d557aba15e1715af9cef9c844 |         |                 |               14

```

tkdchen: :sweat_smile:

cachito/web/models.py
tkdchen: Done.

cachito/web/models.py
tkdchen: Updated.

cachito/web/api_v1.py
```diff
@@ -74,6 +77,31 @@ def get_requests():
...
+            if not name:
+                # Ignore if pkg_manager= presents in the querystring
+                continue
+            pkg_manager: PackageManager = PackageManager.get_by_name(name)
+            if pkg_manager is None:
+                raise ValidationError(f"Cachito does not have package manager {name}.")
+            pkg_manager_ids.append(pkg_manager.id)
+        query = (
+            query.join(PackageManager, Request.pkg_managers)
+            .filter(PackageManager.id.in_(pkg_manager_ids))
```
chmeliik, line 101: Could we do an outer join and allow `PackageManager.id` to be NULL in the filter? That way, we could search for requests that do not use any package managers with `/api/v1/requests?pkg_manager=`

chmeliik (commented):
Looks good!

I have a suggestion about searching for requests that don't use package managers using empty `?pkg_manager=`. If used together with other package managers, e.g. `?pkg_manager=gomod&pkg_manager=pip&pkg_manager=`, then the empty value should be ignored, but if `?pkg_manager=` is used by itself, it should return requests without package managers.

cachito/web/api_v1.py
chmeliik: (or just do 2 different queries, this one if `pkg_manager_ids` is not empty, outer join and select NULL if `pkg_manager_ids` is empty)

chmeliik (approved):
LGTM :+1:

As we discussed elsewhere, `?pkg_manager=` will be ignored. To search explicitly for requests with no package managers, we may add a special argument in the future.

PR #450, opened 2021-05-13
Run black on python3.8

ejegrova:
We encountered a bug in python3.9 that is causing occasional failures
when running black.
https://bugs.python.org/issue43498


chmeliik (commented):
Could you also add a comment to tox.ini to explain why it is needed? And maybe update the comment in .travis.yml as well

chmeliik (approved):
LGTM :+1:

PR #452, opened 2021-05-14
Restrict request state transition

tkdchen:
CLOUDBLD-5455

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -6,6 +6,7 @@ import functools
 import itertools
 import os
 import re
+from typing import Dict, List
 
 import flask
 from flask_login import UserMixin, current_user
@@ -100,6 +101,35 @@ class RequestStateMapping(Enum):
         :rtype: list<str>
         """
         return ["complete", "failed"]
+
+    @staticmethod
+    def allow_transition(from_state, to_state) -> bool:
+        """Check if the state transition is allowed.
+
+        :param from_state: transition from this state.
+        :type from_state: RequestStateMapping
+        :param to_state: transition to this state.
+        :type to_state: RequestStateMapping
+        :return: True if the transition is allowed, otherwise False is returned.
+        :rtype: bool
+        """
+        allowed_from_states = ALLOWED_REQUEST_STATE_TRANSITION.get(to_state)
+        if allowed_from_states is None:
+            return False
+        return from_state in allowed_from_states
+
+
+# to_state: from_states
+ALLOWED_REQUEST_STATE_TRANSITION: Dict[RequestStateMapping, List[RequestStateMapping]] = {
+    RequestStateMapping.in_progress: [RequestStateMapping.in_progress],
+    RequestStateMapping.complete: [RequestStateMapping.in_progress],
+    RequestStateMapping.failed: [RequestStateMapping.in_progress],
+    RequestStateMapping.stale: [
+        RequestStateMapping.complete,
+        RequestStateMapping.failed,
+        RequestStateMapping.in_progress,
+    ],
+}
 
 
 class Package(db.Model):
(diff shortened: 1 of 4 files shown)


cachito/web/models.py
```diff
@@ -101,6 +102,36 @@ def get_final_states():
...
+        """Check if the state transition is allowed.
+
+        :param from_state: transition from this state.
+        :type from_state: int or None
+        :param int to_state: transition to this state.
+        :return: True if the transition is allowed, otherwise False is returned.
+        :rtype: bool
+        """
+        if from_state is None:
+            return True  # Allow to set the initial state.
```
chmeliik, line 116: IMO `from_state` should not be `None`. Instead, this method should only be called if there already is a previous state.

cachito/web/models.py
```diff
@@ -101,6 +102,36 @@ def get_final_states():
...
+        allowed_from_states: Optional[List[int]] = AllowedRequestStateTransition.get(to_state)
+        if allowed_from_states is None:
+            return False
+        return from_state in allowed_from_states
+
+
+# to_state: from_states
+AllowedRequestStateTransition: Dict[int, List[int]] = {
```
chmeliik, line 124: Why is the name in `CamelCase`? As a constant, shouldn't it be `UPPERCASE_WITH_UNDERSCORES`?

cachito/web/models.py
```diff
@@ -1015,6 +1043,14 @@ def add_state(self, state, state_reason):
                 )
             )
 
+        if self.state:
+            from_state_name: str = self.state.state_name
+            from_state: int = RequestStateMapping.__members__[from_state_name].value
```
chmeliik, line 1048:
```suggestion
            from_state: int = RequestStateMapping[from_state_name].value
```

tests/test_api_v1.py
```diff
@@ -1130,24 +1130,46 @@ def test_set_state_stale_failed_to_schedule(
...
+        ["in_progress", "failed", HTTPStatus.OK],
+        ["in_progress", "stale", HTTPStatus.OK],
+        ["complete", "stale", HTTPStatus.OK],
+        ["failed", "stale", HTTPStatus.OK],
+        # Invalid transition
+        ["complete", "in_progress", HTTPStatus.BAD_REQUEST],
+        ["failed", "in_progress", HTTPStatus.BAD_REQUEST],
+        ["stale", "in_progress", HTTPStatus.BAD_REQUEST],
+        ["stale", "complete", HTTPStatus.BAD_REQUEST],
+        ["stale", "failed", HTTPStatus.BAD_REQUEST],
```
akhmanova, line 1147: Thank you for your work! Maybe it worths adding these invalid transactions 
image

tests/test_models.py
```diff
@@ -14,3 +14,29 @@ def test_package_manager_get_by_name(name, expected, app, db, auth_env):
...
+        [RequestStateMapping.in_progress, RequestStateMapping.stale, True],
+        # Invalid transitions
+        [RequestStateMapping.complete, RequestStateMapping.failed, False],
+        [RequestStateMapping.complete, RequestStateMapping.in_progress, False],
+        [RequestStateMapping.failed, RequestStateMapping.complete, False],
+        [RequestStateMapping.failed, RequestStateMapping.in_progress, False],
+        [RequestStateMapping.stale, RequestStateMapping.complete, False],
+        [RequestStateMapping.stale, RequestStateMapping.failed, False],
+        [RequestStateMapping.stale, RequestStateMapping.in_progress, False],
+        [RequestStateMapping.stale, RequestStateMapping.stale, False],
```
akhmanova, line 36: Do we want to add this invalid transactions: 
```suggestion
        [RequestStateMapping.stale, RequestStateMapping.stale, False],
        [RequestStateMapping.failed, RequestStateMapping.failed, False],
        [RequestStateMapping.complete, RequestStateMapping.complete, False],
```
And I have another question: why do we have a different set of invalid transactions in two tests?

tests/test_api_v1.py
chmeliik: @akhmanova that's a pretty cool graphic, how did you make it?

akhmanova: @chmeliik I started to use miro.com instead of paper notes. Because
- paper notes usually are the mess
- it is easy to share something virtual

maybe it is eco friendly as well :D But I don't know 100%

chmeliik (commented):
LGTM when Elina's comments are addressed

tkdchen:
Hi @akhmanova, thanks for your comments. PTAL.

akhmanova (approved):
@tkdchen LGTM. 
I have only one question: why there were two similar tests from the start and now there is only one?

PR #455, opened 2021-05-17
Fix mock for test_fetch_source_archive_exists

tkdchen:
When test parameter gitsubmodule is True, a new SourcesDir object is
created in the scm.Git.fetch_source call, which causes the original mock
of exists method becomes useless.


chmeliik (approved):
LGTM :+1:

PR #456, opened 2021-05-18
Fix pip docs reference to build requirements configuration key

compi-migui:
requirements_build_files is the key actually used by Cachito when looking for that value.


chmeliik (approved):
D'oh! Thanks for fixing this!

PR #457, opened 2021-05-20
Include a timeout to cachito subprocess calls

tkdchen:


chmeliik (commented):
Looks good, +1 on moving `run_cmd()` :+1:

Could you document the new config option in https://github.com/release-engineering/cachito#configuring-workers?

cachito/workers/config.py
```diff
@@ -93,6 +93,8 @@ class Config(object):
     # other tasks aren't starved when processing a large archive.
     worker_prefetch_multiplier = 1
 
+    subprocess_timeout = 3600  # 1 hour
```
chmeliik, line 96:
```suggestion
    cachito_subprocess_timeout = 3600  # 1 hour
```
The cachito-specific options should have the `cachito_` prefix, the ones that don't have a prefix are celery options I think

cachito/workers/config.py
```diff
@@ -93,6 +93,8 @@ class Config(object):
     # other tasks aren't starved when processing a large archive.
     worker_prefetch_multiplier = 1
 
+    cachito_subprocess_timeout = 3600  # 1 hour
```
chmeliik, line 96: Nitpick: could you move this line to the other `cachito_` options? Ideally, they should be sorted alphabetically

README.md
```diff
@@ -513,6 +513,10 @@ Custom configuration for the Celery workers are listed below:
 * `cachito_task_log_format` - the log format that Celery displays when a task is executing. This
   defaults to
   `"[%(asctime)s #%(request_id)s %(name)s %(levelname)s %(module)s.%(funcName)s] %(message)s"`.
+* `cachito_subprocess_timeout` - a number (in seconds) to set a timeout to execute a command by
+  `subprocess` module. Default is 3600 seconds. A timeout is always required, and there is no
+  way provided by Cachito to disable. Set a larger number if it is allowed to wait the subprocess
+  execution for a longer time.
```
chmeliik, line 519:
```suggestion
* `cachito_subprocess_timeout` - a number (in seconds) to set a timeout for commands executed by
  the `subprocess` module. Default is 3600 seconds. A timeout is always required, and there is no
  way provided by Cachito to disable it. Set a larger number to give the subprocess execution more
  time.
```

chmeliik (approved):
LGTM :+1:

PR #464, opened 2021-05-26
Store packages.json as a build artifact

tkdchen:

### DIFF
--- a/cachito/paths.py
+++ b/cachito/paths.py
@@ -56,6 +56,13 @@ class RequestBundleDir(type(Path())):
 
         self.bundle_archive_file = Path(root, f"{request_id}.tar.gz")
 
+        self.packages_data = Path(root, f"{request_id}-packages.json")
+        self.gomod_packages_data = self.joinpath("gomod_packages.json")
+        self.npm_packages_data = self.joinpath("npm_packages.json")
+        self.pip_packages_data = self.joinpath("pip_packages.json")
+        self.yarn_packages_data = self.joinpath("yarn_packages.json")
+        self.git_submodule_packages_data = self.joinpath("git_submodule_packages.json")
+
         return self
 
     def app_subpath(self, subpath):
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -315,7 +315,7 @@ def create_request():
             tasks.fetch_yarn_source.si(request.id, yarn_package_configs).on_error(error_callback)
         )
 
-    chain_tasks.append(tasks.create_bundle_archive.si(request.id).on_error(error_callback))
+    chain_tasks.append(tasks.finalize_request.si(request.id).on_error(error_callback))
 
     try:
         chain(chain_tasks).delay()
@@ -472,7 +472,9 @@ def patch_request(request_id):
 
     db.session.commit()
 
-    bundle_dir = RequestBundleDir(request.id, root=flask.current_app.config["CACHITO_BUNDLES_DIR"])
+    bundle_dir: RequestBundleDir = RequestBundleDir(
+        request.id, root=flask.current_app.config["CACHITO_BUNDLES_DIR"]
+    )
 
     if delete_bundle and bundle_dir.bundle_archive_file.exists():
         flask.current_app.logger.info(
@@ -480,6 +482,7 @@ def patch_request(request_id):
         )
         try:
             bundle_dir.bundle_archive_file.unlink()
+            bundle_dir.packages_data.unlink()
         except:  # noqa E722
             flask.current_app.logger.exception(
                 "Failed to delete the bundle archive %s", bundle_dir.bundle_archive_file
--- a/cachito/workers/pkg_managers/general.py
+++ b/cachito/workers/pkg_managers/general.py
@@ -116,7 +116,7 @@ def update_request_with_package(request_id, package, env_vars=None, package_subp
     """
     Update the request with the resolved packages and corresponding metadata.
 
-    :param dict package: the package that was resolved
+    :param dict[str, str] package: the package that was resolved.
     :param dict env_vars: mapping of environment variables to record. The keys represent
         the environment variable name, and its value should be another map wth the "value" and
         "kind" attributes, e.g. {"NAME": {"value": "VALUE", "kind": "KIND"}}.
--- a/cachito/workers/tasks/general.py
+++ b/cachito/workers/tasks/general.py
@@ -4,6 +4,7 @@ import os
 import shutil
 import tarfile
 from pathlib import Path
+from typing import List
 
 import requests
 
@@ -12,14 +13,16 @@ from cachito.workers.config import get_worker_config
 from cachito.workers.scm import Git
 from cachito.workers.paths import RequestBundleDir
 from cachito.workers.tasks.celery import app
-from cachito.workers.tasks.utils import runs_if_request_in_progress, get_request
+from cachito.workers.tasks.utils import PackagesData, runs_if_request_in_progress, get_request
 
 __all__ = [
+    "aggregate_packages_data",
     "create_bundle_archive",
     "failed_request_callback",
     "fetch_app_source",
-    "set_request_state",
+    "finalize_request",
     "get_request",
+    "set_request_state",
 ]
 log = logging.getLogger(__name__)
 
@@ -140,17 +143,14 @@ def failed_request_callback(context, exc, traceback, request_id):
     set_request_state(request_id, "failed", msg)
 
 
-@app.task(priority=10)
-@runs_if_request_in_progress
-def create_bundle_archive(request_id):
+def create_bundle_archive(request_id: int, flags: List[str]) -> None:
     """
     Create the bundle archive to be downloaded by the user.
 
     :param int request_id: the request the bundle is for
+    :param list[str] flags: the list of request flags.
     """
     set_request_state(request_id, "in_progress", "Assembling the bundle archive")
-    request = get_request(request_id)
-
     bundle_dir = RequestBundleDir(request_id)
 
     log.debug("Using %s for creating the bundle for request %d", bundle_dir, request_id)
@@ -161,7 +161,7 @@ def create_bundle_archive(request_id):
         return tar_info if os.path.basename(tar_info.name) != ".git" else None
 
     tar_filter = filter_git_dir
-    if "include-git-dir" in request.get("flags", []):
+    if "include-git-dir" in flags:
         tar_filter = None
 
     with tarfile.open(bundle_dir.bundle_archive_file, mode="w:gz") as bundle_archive:
@@ -173,4 +173,30 @@ def create_bundle_archive(request_id):
         # Add the dependencies to the bundle
         bundle_archive.add(str(bundle_dir.deps_dir), "deps")
 
+
+def aggregate_packages_data(request_id: int, pkg_managers: List[str]) -> None:
+    """Aggregate packages data generated for each package manager into one unified data file.
+
+    :param int request_id: the request id.
+    """
+    set_request_state(request_id, "in_progress", "Aggregating packages data")
+    bundle_dir = RequestBundleDir(request_id)
+
+    aggregated_data = PackagesData()
+    for pkg_manager in pkg_managers:
+        # Ensure git-submodule -> git_submodule
+        data_file = getattr(bundle_dir, f"{pkg_manager.replace('-', '_')}_packages_data")
+        aggregated_data.load(data_file)
+
+    log.debug("Write request %s packages data into %s", request_id, bundle_dir.packages_data)
+    aggregated_data.write_to_file(str(bundle_dir.packages_data))
+
+
+@app.task(priority=10)
+@runs_if_request_in_progress
+def finalize_request(request_id):
+    """Execute tasks to finalize the request creation."""
+    request = get_request(request_id)
+    create_bundle_archive(request_id, request.get("flags", []))
+    aggregate_packages_data(request_id, request["pkg_managers"])
     set_request_state(request_id, "complete", "Completed successfully")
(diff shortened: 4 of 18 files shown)


cachito/workers/tasks/gomod.py
```diff
@@ -116,14 +121,31 @@ def fetch_gomod_source(request_id, dep_replacements=None, package_configs=None):
...
+
+            packages_json_data["packages"].append({
+                "dependencies": package["pkg_deps"],
+                "name": pkg_info["name"],
+                "type": pkg_info["type"],
+                "version": pkg_info["version"],
+                "path": package_subpath,
+            })
+
+    bundle_dir.gomod_packages_data.write_text(json.dumps(packages_json_data), encoding="utf-8")
```
chmeliik, line 148: I realize this is a draft, just gonna comment this so that I don't forget :D

It might be worth it to check for duplicate packages. I think packages should be unique by `name, type, version`. If there are duplicates IMO we should raise an error.

Previously, it was technically possible to have two packages with the same `name, type, version` as long as they had the same `path`. If those packages somehow had different lists of dependencies, cachito would merge the lists into one. If you had two identical packages with different paths, Cachito would raise an error.

But if two packages have the same `name, type, version, path`, then they are the same package and we should never process them more than once during a cachito request. So if two packages have the same `name, type, version`, that should always be an error regardless of the `path`.

It would probably be good to check for duplicates in each of the `{gomod,npm,pip,yarn}_package.json` files as well as the final `{request_id}-packages.json`

tkdchen: @chmeliik Is `resolve_*` function the good place to make this check?

chmeliik: I think `fetch_*_source` tasks would be the right place, because they combine the results of multiple `resolve_*` calls. We should be checking that packages are still unique even after combining all the resolve calls.

cachito/workers/tasks/utils.py
```diff
@@ -203,3 +205,72 @@ def _get_request_or_fail(request_id: int, connect_error_msg: str, status_error_m
...
+            data = json.load(f)
+            for p in data["packages"]:
+                self.add_package(p, p["path"], p["dependencies"])
+
+    def __add__(self, other):
+        new = PackagesData()
+        new._packages = self._packages[:]
+        for p in other._packages:
+            new.add_package(p, p["path"], p["dependencies"])
+        return new
```
chmeliik, line 276: Could this be changed back to `__iadd__`? And instead of creating a new PackagesData instance every time, just update `self`. The JSON files may get pretty big, copying all the data when merging the files could be expensive

cachito/workers/tasks/general.py
```diff
@@ -173,4 +175,32 @@ def filter_git_dir(tar_info):
...
+    bundle_dir: RequestBundleDir = RequestBundleDir(request_id)
+
+    def _get_packages_data():
+        request = get_request(request_id)
+        for pkg_manager in request["pkg_managers"]:
+            data = PackagesData()
+            data.load(getattr(bundle_dir, f"{pkg_manager}_packages_data"))
+            yield data
+
+    aggregated_data = functools.reduce(operator.add, _get_packages_data())
```
chmeliik, line 194: If there are no package managers, `_get_packages_data` will be an empty iterator and `functools.reduce` will fail. I think `functools.reduce` accepts an initial value for this purpose

cachito/workers/tasks/utils.py
```diff
@@ -203,3 +205,72 @@ def _get_request_or_fail(request_id: int, connect_error_msg: str, status_error_m
...
+        :param str path: the path where the package is retreived. Consult with the
+            ``fetch_*_source`` for the defailed information about a package's path.
+        :param deps: a list of depencencies the package has.
+        :type deps: list[dict[str, any]]
+        :raises CachitoError: if there is a package with same name, type and version
+            has been added already.
+        """
+        key = (pkg_info["name"], pkg_info["type"], pkg_info["version"])
+        if key in self._index:
+            raise CachitoError(f"Duplicate package is detected: {pkg_info!r}")
```
chmeliik, line 231:
```suggestion
            raise CachitoError(f"Duplicate package: {pkg_info!r}")
```

cachito/workers/tasks/utils.py
```diff
@@ -203,3 +205,72 @@ def _get_request_or_fail(request_id: int, connect_error_msg: str, status_error_m
...
+
+    def write_to_file(self, file_name: Union[str, Path]) -> None:
+        """Write the added packages to a file as JSON data.
+
+        :param file_name: an absolute or relative filename to write the added packages into.
+            When a relative path is used, it will be opened directly and depends on the
+            ``os.curdir``.
+        :type file_name: str or pathlib.Path
+        """
+        log.debug("Write packages with dependencies into file %s: %r", file_name, self._packages)
```
chmeliik, line 251: `self._packages` may be huge, I think just the file name would be good enough in the log message

cachito/workers/tasks/utils.py
```diff
@@ -203,3 +205,72 @@ def _get_request_or_fail(request_id: int, connect_error_msg: str, status_error_m
...
+        with open(file_name, "w+", encoding="utf-8") as f:
```
chmeliik, line 252: What is the difference between `w+` and `w`?

cachito/paths.py
```diff
@@ -56,6 +56,12 @@ def __new__(cls, request_id, root, app_subpath=os.curdir):
 
         self.bundle_archive_file = Path(root, f"{request_id}.tar.gz")
 
+        self.packages_data = Path(root, f"{request_id}-packages.json")
+        self.gomod_packages_data = self.joinpath("gomod_packages.json")
+        self.npm_packages_data = self.joinpath("npm_packages.json")
+        self.pip_packages_data = self.joinpath("pip_packages.json")
+        self.yarn_packages_data = self.joinpath("yarn_packages.json")
```
chmeliik, line 63: `git-submodule` can also have packages, see `tasks/gitsubmodule.py`

chmeliik (commented):
Nice, I like the `PackagesData` class!

Left a few inline comments.

cachito/workers/tasks/utils.py
tkdchen: Force to overwrite an existing file.

chmeliik: I think regular `w` already does that

cachito/web/api_v1.py
```diff
@@ -472,14 +472,17 @@ def patch_request(request_id):
 
     db.session.commit()
 
-    bundle_dir = RequestBundleDir(request.id, root=flask.current_app.config["CACHITO_BUNDLES_DIR"])
+    bundle_dir: RequestBundleDir = RequestBundleDir(
```
chmeliik, line 475: Do we need the type annotation for RequestBundleDir? I'd say the `RequestBundleDir()` call makes the type pretty clear :-)

cachito/workers/pkg_managers/general.py
```diff
@@ -116,7 +116,7 @@ def update_request_with_package(request_id, package, env_vars=None, package_subp
     """
     Update the request with the resolved packages and corresponding metadata.
 
-    :param dict package: the package that was resolved
+    :param dict[str, str] package: the package that was resolved. It must be in form
```
chmeliik, line 119: Looks like this sentence is not complete

cachito/workers/tasks/general.py
```diff
@@ -173,4 +173,35 @@ def filter_git_dir(tar_info):
...
+    :param int request_id: the request id.
+    """
+    set_request_state(request_id, "in_progress", "Aggregating packages data")
+    bundle_dir: RequestBundleDir = RequestBundleDir(request_id)
+
+    def _get_packages_data():
+        request = get_request(request_id)
+        for pkg_manager in request["pkg_managers"]:
+            data = PackagesData()
+            data.load(getattr(bundle_dir, f"{pkg_manager}_packages_data"))
```
chmeliik, line 189: This will not work for `git-submodule`. It might be good to add a test case for that

cachito/workers/tasks/general.py
```diff
@@ -173,4 +173,35 @@ def filter_git_dir(tar_info):
...
+    for data in _get_packages_data():
+        aggregated_data += data
+
+    log.debug("Write request %s packages data into %s", request_id, bundle_dir.packages_data)
+    aggregated_data.write_to_file(str(bundle_dir.packages_data))
+
+
+@app.task(priority=10)
+@runs_if_request_in_progress
+def finalize_request_creation(request_id):
```
chmeliik, line 203:
```suggestion
def finalize_request(request_id):
```

cachito/workers/tasks/general.py
```diff
@@ -173,4 +173,35 @@ def filter_git_dir(tar_info):
...
+    """Execute tasks to finalize the request creation."""
+    create_bundle_archive(request_id)
+    aggregate_packages_data(request_id)
```
chmeliik, line 206: Since both of these functions will call `get_request()`, could we instead call that here and pass the entire request dict instead of the `request_id`? The request JSON can be very large in some cases, saving one http request could be significant

cachito/workers/tasks/gomod.py
```diff
@@ -113,17 +115,26 @@ def fetch_gomod_source(request_id, dep_replacements=None, package_configs=None):
             env_vars.update(config.cachito_default_environment_variables.get("gomod", {}))
         else:
             env_vars = None
-        update_request_with_package(request_id, gomod["module"], env_vars, package_subpath=subpath)
-        update_request_with_deps(request_id, gomod["module"], gomod["module_deps"])
+
+        pkg_info = gomod["module"]
```
chmeliik, line 119:
```suggestion
        mod_info = gomod["module"]
```
See next suggestion

cachito/workers/tasks/gomod.py
```diff
@@ -113,17 +115,26 @@ def fetch_gomod_source(request_id, dep_replacements=None, package_configs=None):
...
 
         # add package deps
         for package in gomod["packages"]:
-            package_subpath = _package_subpath(
-                gomod["module"]["name"], package["pkg"]["name"], subpath
-            )
-            update_request_with_package(request_id, package["pkg"], package_subpath=package_subpath)
+            pkg_info = package["pkg"]
+            package_name = pkg_info["name"]
+            package_subpath = _package_subpath(pkg_info["name"], package_name, subpath)
```
chmeliik, line 129:
```suggestion
            package_subpath = _package_subpath(mod_info["name"], pkg_info["name"], subpath)
```
See previous suggestion. Currently, the code has a bug where `pkg_info["name"]` is always the same as `package_name`

cachito/workers/tasks/pip.py
```diff
@@ -117,13 +118,20 @@ def fetch_pip_source(request_id, package_configs=None):
     worker_config = get_worker_config()
     env_vars.update(worker_config.cachito_default_environment_variables.get("pip", {}))
 
+    # pkg key: pkg info with deps
+    packages_json_data = PackagesData()
```
chmeliik, line 122: Comment is not relevant anymore

cachito/workers/tasks/utils.py
```diff
@@ -203,3 +205,74 @@ def _get_request_or_fail(request_id: int, connect_error_msg: str, status_error_m
...
+            ``os.curdir``. If the file does not exist, nothing is changed internally.
+        :type file_name: str or pathlib.Path
+        """
+        if not os.path.exists(file_name):
+            log.debug("No data is loaded from non-existing file %s.", file_name)
+            return
+        with open(file_name, "r", encoding="utf-8") as f:
+            data = json.load(f)
+            packages = data.get("packages")
+            if not packages:
```
chmeliik, line 269:
```suggestion
            if packages is None:
```
Packages probably cannot be an empty list, but IMO this would still be nicer

cachito/workers/tasks/utils.py
```diff
@@ -203,3 +205,74 @@ def _get_request_or_fail(request_id: int, connect_error_msg: str, status_error_m
...
+                log.warning("Packages data file does not include key 'packages'.")
+                return
+            for p in packages:
+                self.add_package(p, p["path"], p["dependencies"])
+
+    def __iadd__(self, other):
+        for p in other._packages:
+            self.add_package(p, p["path"], p["dependencies"])
+        return self
```
chmeliik, line 278: Now I'm thinking, instead of adding the PackagesData objects together, could we just merge the JSON files like this?
```python
data = PackagesData()
for f in package_json_files:
    data.load(f)
```

chmeliik (commented):
Found a couple more things, sorry I didn't catch these earlier

cachito/web/api_v1.py
tkdchen: Oh, this type annotation is optional. IIRC, it just makes my editor happy to give auto-complete candidates.

chmeliik: I see, we can keep it then

cachito/workers/tasks/utils.py
```diff
@@ -203,3 +205,69 @@ def _get_request_or_fail(request_id: int, connect_error_msg: str, status_error_m
...
+        key = (pkg_info["name"], pkg_info["type"], pkg_info["version"])
+        if key in self._index:
+            raise CachitoError(f"Duplicate package: {pkg_info!r}")
+        self._index.add(key)
+        self._packages.append(
+            {
+                "name": pkg_info["name"],
+                "type": pkg_info["type"],
+                "version": pkg_info["version"],
+                "path": path,
```
chmeliik, line 238: Currently, the database will only store the `path` if `path != "."`. Doing the same thing for the JSON file would make it slightly easier to use it in the API later.

chmeliik (commented):
Looks great! I tested this locally and it works perfectly, just one minor detail about not storing paths when `path == "."`

tkdchen:
@chmeliik Thanks for all your comments. All of them are addressed. PTAL.

chmeliik (approved):
LGTM :+1:

PR #467, opened 2021-06-01
Only call go list -deps once per module 

chmeliik:
CLOUDBLD-5784

Instead of calling `go list -deps` on each package from `go list -find`,
we can just call `go list -deps -json` once. This will significantly
improve performance in cases where listing deps is the bottleneck.

### DIFF
--- a/cachito/workers/__init__.py
+++ b/cachito/workers/__init__.py
@@ -1,7 +1,10 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
+import json
 import logging
+import re
 import subprocess
+from typing import Iterator
 
 from cachito.errors import CachitoError
 from cachito.workers.config import get_worker_config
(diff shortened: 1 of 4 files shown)


chmeliik:
So far, the only case I know of where this is relevant is https://github.com/stackrox/rox. The repo is private, which makes this a bit hard to test locally - you would need the right `/root/.netrc` in both the cachito-worker and the athens containers.

If you are able to set this up, the changes can be tested with this payload:
```json
{
  "repo": "https://github.com/stackrox/rox.git",
  "ref": "e92b61484594f24eab4c93fd1cf93a0f4c5413c1",
  "pkg_managers": [
    "gomod"
  ]
}
```
If we take only the time spent locally processing packages and dependencies (after downloading but before saving in DB), the difference between master and this branch is 23 minutes vs. 30 seconds. And this time, I'm fairly confident that the effect will be similar in prod (except of course, saving the deps in the DB will take much longer there :sweat_smile: )

nirzari:
Great, LGTM :ship: :it:

nirzari (approved):

chmeliik:
Rebased on master, merging

PR #184, opened 2020-06-29
Add support for env var kind

lcarva:
* CLOUDBLD-1484

PR #471, opened 2021-06-02
Adjust unique constraint on Flag table

thrbkova:
The unique constraint for the Flag table was set for "id" and "name". That was not desired because there could be multiple flag entries with the same name. The unique constraint is now only on the name column.

### DIFF
--- a/cachito/web/migrations/versions/a655a299e967_request_flags.py
+++ b/cachito/web/migrations/versions/a655a299e967_request_flags.py
@@ -23,7 +23,7 @@ def upgrade():
         sa.Column("name", sa.String(), nullable=False),
         sa.Column("active", sa.Boolean(), nullable=False),
         sa.PrimaryKeyConstraint("id"),
-        sa.UniqueConstraint("id", "name"),
+        sa.UniqueConstraint("id", "name", name="flag_id_name_key"),
     )
 
     op.create_table(
--- a/cachito/web/migrations/versions/4d17dec0cfc3_adjust_flag_unique_constraint.py
+++ b/cachito/web/migrations/versions/4d17dec0cfc3_adjust_flag_unique_constraint.py
@@ -1,0 +1,28 @@
+"""Adjust unique constraint on Flag table
+
+Revision ID: 4d17dec0cfc3
+Revises: c6ac095d8e9f
+Create Date: 2021-05-25 13:21:48.298960
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "4d17dec0cfc3"
+down_revision = "c6ac095d8e9f"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table("flag") as batch_op:
+        batch_op.create_unique_constraint("flag_name_key", ["name"])
+        batch_op.drop_constraint("flag_id_name_key", type_="unique")
+
+
+def downgrade():
+    with op.batch_alter_table("flag") as batch_op:
+        batch_op.drop_constraint("flag_name_key", type_="unique")
+        batch_op.create_unique_constraint("flag_id_name_key", ["id", "name"])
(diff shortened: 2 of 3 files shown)


cachito/web/migrations/versions/4d17dec0cfc3_adjust_unique_constraint.py
```diff
@@ -0,0 +1,38 @@
...
+    sa.MetaData(),
+    sa.Column("id", sa.Integer(), primary_key=True),
+    sa.Column("name", sa.String(), nullable=False),
+    sa.Column("active", sa.Boolean(), nullable=False, default=True),
+)
+
+
+def upgrade():
+    with op.batch_alter_table("flag") as batch_op:
+        batch_op.create_index(batch_op.f("ix_flag_name"), ["name"], unique=True)
```
chmeliik, line 31: Could this be `create_unique_constraint` instead? If yes, I would also rename it to `flag_name_key`.

cachito/web/migrations/versions/a655a299e967_request_flags.py
```diff
@@ -23,7 +23,7 @@ def upgrade():
         sa.Column("name", sa.String(), nullable=False),
         sa.Column("active", sa.Boolean(), nullable=False),
         sa.PrimaryKeyConstraint("id"),
-        sa.UniqueConstraint("id", "name"),
+        sa.UniqueConstraint("id", "name", name="flag_id_name_key"),
```
chmeliik, line 26: I'm not 100% sure, but I think updates to old migration files will not be applied retroactively. Since `flag_id_name_key` is the default name, it should be fine to drop this change.

chmeliik (commented):
Nice job! I greatly overestimated how much of the migration script would be autogenerated :D

Just a few minor comments.

cachito/web/migrations/versions/a655a299e967_request_flags.py
nirzari: I second this.

thrbkova: Without the explicit name I get ValueError: No such constraint. How can I solve it without changing this file?

chmeliik: It works for me. Here's how I tested this:
```shell
# start cachito on master branch
git checkout master
make clean
make run

# check current indexes
docker-compose exec db psql cachito -U cachito -c "select * from pg_indexes where tablename = 'flag'"

# checkout your branch
git checkout unique_flag_tbl
# also remove the explicit name from this file (only this file)

# test the upgrade and downgrade
docker-compose exec cachito-api cachito db upgrade
docker-compose exec db psql cachito -U cachito -c "select * from pg_indexes where tablename = 'flag'"
docker-compose exec cachito-api cachito db downgrade
docker-compose exec db psql cachito -U cachito -c "select * from pg_indexes where tablename = 'flag'"
```

chmeliik: Oh I see, it's the unit tests that are failing, not the real migration. I'll take a look

chmeliik: It appears that updating old migrations for this purpose is an accepted practice, I found an old commit which does exactly that: https://github.com/release-engineering/cachito/pull/184/commits/7b09691bec6b006a3f73c3f5cd633af961111513.

Let's keep this change, but please put it in a separate commit (and explain why it is needed in the commit message)

cachito/web/migrations/versions/4d17dec0cfc3_adjust_unique_constraint.py
```diff
@@ -0,0 +1,29 @@
+"""Adjust unique constraint on Flag table
+Remove unique constraint from id and name but keep name unique
```
chmeliik, line 2: Let's use a single line description, I think we can just keep the first one.

I would also suggest to rename the file `adjust_unique_constraint` -> `adjust_flag_unique_constraint`

chmeliik (commented):
LGTM, just please swap the 2 commits, the first one would break unit tests without the second. And while you're doing that, you can add `CLOUDBLD-1158` to the commit messages :wink:

chmeliik (approved):
LGTM :+1:

PR #472, opened 2021-06-07
Unified name

dcho5:
cachito_nexus_npm_proxy_repo_url is now cachito_nexus_npm_proxy_url

### DIFF
--- a/cachito/workers/config.py
+++ b/cachito/workers/config.py
@@ -39,7 +39,7 @@ class Config(object):
     cachito_nexus_hoster_username = None
     cachito_nexus_js_hosted_repo_name = "cachito-js-hosted"
     cachito_nexus_max_search_attempts = 5
-    cachito_nexus_npm_proxy_repo_url = "http://localhost:8081/repository/cachito-js/"
+    cachito_nexus_npm_proxy_url = "http://localhost:8081/repository/cachito-js/"
     cachito_nexus_pip_raw_repo_name = "cachito-pip-raw"
     cachito_nexus_proxy_password = None
     cachito_nexus_proxy_username = None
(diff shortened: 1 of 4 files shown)


chmeliik:
Travis CI is failing, looks like a few unit tests need fixing.

To make sure CI will pass, you can run unit tests locally with `tox` or `make test-unit`.

I think the name will also have to be changed in a few other places. Unit tests may not necessarily catch those because we use [mocking](https://docs.python.org/3/library/unittest.mock.html), so you might want to run integration tests as well - `tox -e integration` or `make test-integration`. These will only work if you have cachito running locally, see https://github.com/release-engineering/cachito#integration-tests

tkdchen:
Do we expect the commits are singed-off?

cachito/workers/config.py
```diff
@@ -270,14 +270,14 @@ def validate_npm_config():
...
+    if not conf.get("cachito_nexus_npm_proxy_url"):
         raise ConfigError(
-            'The configuration "cachito_nexus_npm_proxy_repo_url" must be set for this package '
+            'The configuration "cachito_nexus_npm_proxy_url" must be set for this package '
             "manager"
         )
 
 
-# yarn and npm use the same configuration options including `cachito_npm_proxy_repo_url`;
+# yarn and npm use the same configuration options including `cachito_npm_proxy_url`;
```
chmeliik, line 280:
```suggestion
# yarn and npm use the same configuration options including `cachito_nexus_npm_proxy_url`;
```
This one was wrong before, might as well use this opportunity to fix it :-)

chmeliik (commented):
Like in #473, please squash your commits into one and update your commit message with all the necessities, see https://github.com/release-engineering/cachito/pull/473#issuecomment-856537669

Apart from that, looks good!

chmeliik (approved):
LGTM :+1:

tkdchen:
:+1:

PR #475, opened 2021-06-10
Set packages and dependencies counts for requests 

chmeliik:
CLOUDBLD-4727

All the required changes (and a bit of refactoring) for storing package and dependency counts in the database.

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -377,6 +377,8 @@ def patch_request(request_id):
         "package_subpath",
         "state",
         "state_reason",
+        "packages_count",
+        "dependencies_count",
     }
     invalid_keys = set(payload.keys()) - valid_keys
     if invalid_keys:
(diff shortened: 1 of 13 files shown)


chmeliik:
Some things to note:
* The `/api/v1/requests` API endpoint still gets the package and deps counts by counting entries in the `request_package` and `request_dependency` tables rather than relying on the new attributes. That is expected, this will be changed later.
* I tested this with integration tests. The `*_count` values stored in the DB match what is being reported by `/api/v1/requests` in all cases except one - the test case for https://github.com/cachito-testing/cachito-pip-local-path.git. This test tests an expected failure. The api reports package and dep counts as 0, the value in the DB is `null`. IMO `null` is what we want.

chmeliik:
@dcho5 feel free to take a look and give feedback if you have any :-)

tkdchen (approved):

chmeliik:
Rebased on master. Also moved the `set_packages_and_deps_counts` call to `finalize_request()` and added a test for that

chmeliik:
Added docstring to PackagesData init, not sure why flake8 didn't say anything before

tkdchen (approved):

PR #477, opened 2021-06-11
Set timeout for requests made by the /status endpoint

dcho5:
CLOUDBLD-3702

### DIFF
--- a/cachito/web/status.py
+++ b/cachito/web/status.py
@@ -26,6 +26,7 @@ PKG_MANAGER_REQUIRES = {
 }
 
 no_retry_session = requests.Session()
+SERVICE_TIMEOUT = 5
 
 
 def _service_ok(url):
@@ -38,7 +39,7 @@ def _service_ok(url):
     :return: tuple (ok: bool, reason: str or None)
     """
     try:
-        resp = no_retry_session.get(url)
+        resp = no_retry_session.get(url, timeout=SERVICE_TIMEOUT)
     except requests.RequestException:
         return False, "connection failed"
 
--- a/tests/test_status.py
+++ b/tests/test_status.py
@@ -55,7 +55,7 @@ def test_service_ok(mock_requests_get, fail_reason):
         assert ok
         assert reason is None
 
-    mock_requests_get.assert_called_once_with(url)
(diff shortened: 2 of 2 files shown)


cachito/web/status.py
```diff
@@ -25,7 +26,7 @@
     "*": {DATABASE, RABBITMQ},
 }
 
-no_retry_session = requests.Session()
+retry_session = get_requests_session()
```
chmeliik, line 29: We should keep using the `no_retry_session`. The purpose of the status endpoint is to check the status at a given moment in time, retrying on failures is not desirable.

cachito/web/status.py
```diff
@@ -37,8 +38,9 @@ def _service_ok(url):
     :param str url: url to fetch
     :return: tuple (ok: bool, reason: str or None)
     """
+    config = get_worker_config()
     try:
-        resp = no_retry_session.get(url)
+        resp = retry_session.get(url, timeout=config.cachito_api_timeout)
```
chmeliik, line 43: The `cachito_api_timeout` is too long for this purpose. I would generally expect the status endpoint to respond in less than 10 seconds, for the individual requests around 3 seconds (or slightly larger than 3 seconds as recommended here https://docs.python-requests.org/en/master/user/advanced/#timeouts) should be enough.

I don't think this needs to be configurable, IMO a constant in status.py would do. But if you do want add a config option for this I won't stop you :-)

cachito/web/status.py
```diff
@@ -26,6 +26,7 @@
 }
 
 no_retry_session = requests.Session()
+service_timeout = 5
```
chmeliik, line 29: Constants should generally be uppercase, there are some exceptions such as request sessions and logger objects but this should be uppercase IMO

cachito/web/status.py
akhmanova: @chmeliik Do you know why api timeout is located in the worker config file?

chmeliik: @akhmanova it is the timeout that workers use when communicating with the API

chmeliik (approved):
LGTM :+1:

PR #478, opened 2021-06-11
Failed requests are now stale only after 7 days

dcho5:
CLOUDBLD-3849

### DIFF
--- a/tests/test_workers/test_cleanup_job.py
+++ b/tests/test_workers/test_cleanup_job.py
@@ -92,6 +92,19 @@ mock_failed = {
             "updated": "2019-09-05T18:24:50.857861",
             "user": "tom.hanks@domain.local",
         },
+        {
+            "dependencies": 309,
+            "environment_variables": {},
+            "flags": [],
+            "id": 53,
+            "pkg_managers": ["gomod"],
+            "ref": "a7ac8d4c0b7fe90d51fb911511cbf6939655c877",
+            "repo": "https://github.com/kubernetes/kubernetes.git",
+            "state": "failed",
+            "state_reason": "The request failed",
+            "updated": "2019-08-30T18:24:50.857861",
+            "user": "tom.hanks@domain.local",
+        },
     ],
     "meta": {
         "first": "https://cachito.domain.local/api/v1/requests"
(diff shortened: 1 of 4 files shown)


dcho5:
Is it bad practice to submit another PR while another one is pending? This PR seems to show two commits including CLOUDBLD-3702 from the last PR.

chmeliik:
> Is it bad practice to submit a PR while another one is pending? This PR seems to show two commits including CLOUDBLD-3702 from the last PR.

It's not a bad practice, but it would be better to create this `failed_logs` branch from the master branch rather than the `timeout` branch. That way, this branch would only contain the commits that are needed.

You could still fix it e.g. by running `git checkout failed_logs; git rebase --interactive master` and dropping the commits that do not belong here.

tests/test_workers/test_cleanup_job.py
```diff
@@ -212,14 +212,9 @@ def test_cleanup_job_success(mock_requests, mock_auth_requests, mock_dt):
             json={"state": "stale", "state_reason": "The request has expired"},
             timeout=60,
         ),
-        mock.call(
-            "http://cachito.domain.local/api/v1/requests/52",
-            json={"state": "stale", "state_reason": "The request has expired"},
-            timeout=60,
-        ),
```
chmeliik, line 219: Can we also have a test case where a failed request is cleaned up?

tests/test_workers/test_cleanup_job.py
```diff
@@ -201,6 +201,33 @@ def test_cleanup_job_success(mock_requests, mock_auth_requests, mock_dt):
...
+    assert mock_requests.call_count == 3
+    assert mock_auth_requests.call_count == 2
+    mock_auth_requests.assert_has_calls(calls)
+
+
+@mock.patch("cachito.workers.config.Config.cachito_request_lifetime", 1)
+@mock.patch("cachito.workers.cleanup_job.datetime")
+@mock.patch("cachito.workers.cleanup_job.auth_session.patch")
+@mock.patch("cachito.workers.cleanup_job.session.get")
+def test_cleanup_job_fail(mock_requests, mock_auth_requests, mock_dt):
```
chmeliik, line 225: IMO this can be merged into the previous test. For example, you could just add a second failed request to `mock_failed['items']` with an older `'updated'` value. For this to work, you would also have to change the mocked datetime object slightly, see next suggestion.

tests/test_workers/test_cleanup_job.py
```diff
@@ -201,6 +201,33 @@ def test_cleanup_job_success(mock_requests, mock_auth_requests, mock_dt):
...
+    mock_dt.utcnow = mock.Mock(return_value=datetime(2019, 9, 13))
+    mock_dt.strptime = mock.Mock(return_value=datetime(2019, 9, 5))
```
chmeliik, line 227:
```suggestion
    mock_dt.strptime = datetime.strptime
```
I think it's better to use the real `strptime` method like this. The `'updated'` values in the testing data should already be set properly to make this work. Additionally, this change would make it easier to test the different lifetimes for different requests, see previous suggestion.

chmeliik (approved):
LGTM, nice!

PR #480, opened 2021-06-15
Let the tests remove real directory and files

tkdchen:

### DIFF
--- a/tests/test_api_v1.py
+++ b/tests/test_api_v1.py
@@ -11,6 +11,7 @@ import kombu.exceptions
 import pytest
 
 from cachito.errors import CachitoError, ValidationError
+from cachito.paths import RequestBundleDir
 from cachito.web.content_manifest import BASE_ICM, PARENT_PURL_PLACEHOLDER
 from cachito.web.models import (
     ConfigFileBase64,
@@ -23,7 +24,6 @@ from cachito.web.models import (
     _validate_package_manager_exclusivity,
 )
 from cachito.web.utils import deep_sort_icm
-from cachito.workers.paths import RequestBundleDir
 from cachito.workers.tasks import (
     add_git_submodules_as_package,
     failed_request_callback,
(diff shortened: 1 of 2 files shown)


tests/test_cachito_config.py
```diff
@@ -54,11 +54,13 @@ def test_validate_cachito_config_cli(mock_isdir, app):
...
     ],
 )
-def test_validate_mutually_exclusive_package_managers(app, value, is_valid):
+def test_validate_mutually_exclusive_package_managers(app, value, is_valid, tmpdir):
     config = app.config.copy()
     config["CACHITO_MUTUALLY_EXCLUSIVE_PACKAGE_MANAGERS"] = value
 
     if is_valid:
+        # Successful validation requires this dir exists.
+        config["CACHITO_BUNDLES_DIR"] = str(tmpdir)
```
chmeliik, line 63: It looks like the other tests handle this by mocking:
```python
@patch("os.path.isdir", return_value=True)
```

chmeliik (approved):
LGTM, one optional suggestion

tests/test_cachito_config.py
tkdchen: Yeah, I tend to make it explicit for this. I actually also think it should be better ensure the directory exists than mocking the isdir :) IMO, it makes it much clear what the test tests.

tests/test_api_v1.py
```diff
@@ -1080,10 +1071,18 @@ def test_set_state_stale(
     db.session.add(request)
     db.session.commit()
 
+    bundle_dir = RequestBundleDir(1, str(tmpdir))
+    bundle_dir.mkdir(parents=True)
```
akhmanova, line 1075: @tkdchen Could you please explain where the test actually remove the directory? 
and why did we use mocks for that before? :)

tkdchen: @akhmanova 

For the first question, https://github.com/release-engineering/cachito/blob/master/cachito/web/api_v1.py#L494

For the second one, I don't know why :)

PR #481, opened 2021-06-15
Sorts packages and dependencies on a request

brunoapimentel:
CLOUDBLD-3382
Signed-off-by: Bruno Pimentel <bpimente@redhat.com>

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -19,6 +19,7 @@ from werkzeug.exceptions import Forbidden
 from cachito.web import content_manifest
 from cachito.errors import ContentManifestError, ValidationError
 from cachito.web import db
+from cachito.workers.tasks.utils import sort_packages_and_deps_in_place
 
 
 def is_request_ref_valid(ref: str) -> bool:
@@ -908,6 +909,8 @@ class Request(db.Model):
 
                 package_to_deps.setdefault(req_dep.package.id, []).append(dep_json)
 
+            sort_packages_and_deps_in_place(rv["dependencies"])
+
             rv["packages"] = [
                 req_package.package.to_json(
                     dependencies=package_to_deps.get(req_package.package_id, []),
@@ -915,6 +918,9 @@ class Request(db.Model):
                 )
                 for req_package in self.request_packages
             ]
+
+            sort_packages_and_deps_in_place(rv["packages"])
+
             if flask.current_app.config["CACHITO_REQUEST_FILE_LOGS_DIR"]:
                 rv["logs"] = {
                     "url": flask.url_for(
--- a/cachito/workers/tasks/utils.py
+++ b/cachito/workers/tasks/utils.py
@@ -21,6 +21,7 @@ __all__ = [
     "get_request_state",
     "set_request_state",
     "set_packages_and_deps_counts",
+    "sort_packages_and_deps_in_place",
     "PackagesData",
 ]
 
@@ -233,6 +234,26 @@ def set_packages_and_deps_counts(request_id: int, packages_count: int, dependenc
     )
 
 
+def sort_packages_and_deps_in_place(packages):
+    """
+    Sorts lists of packages in place.
+
+    Sorting order: type -> dev -> name -> version.
+    If a package has a "dependencies" list, the packages inside it will be sorted as well.
+
+    :param list packages: the list of packages
+    """
+
+    def sort_key(x):
+        return x["type"], x.get("dev", False), x["name"], x["version"]
+
+    packages.sort(key=sort_key)
+
+    for package in packages:
+        if "dependencies" in package:
+            package["dependencies"].sort(key=sort_key)
+
+
 def _get_request_or_fail(request_id: int, connect_error_msg: str, status_error_msg: str) -> dict:
     """
     Try to download the JSON data for a request from the Cachito API.
@@ -344,6 +365,8 @@ class PackagesData:
             ``os.curdir``.
         :type file_name: str or pathlib.Path
         """
+        sort_packages_and_deps_in_place(self._packages)
+
         log.debug("Write packages with dependencies into file %s.", file_name)
         with open(file_name, "w", encoding="utf-8") as f:
             json.dump({"packages": self._packages}, f)
--- a/tests/integration/test_data/gomod_packages.yaml
+++ b/tests/integration/test_data/gomod_packages.yaml
@@ -14,11 +14,11 @@ without_deps:
     packages:
       - dependencies: []
         name: "github.com/cachito-testing/cachito-gomod-without-deps"
-        type: "gomod"
+        type: "go-package"
         version: "v0.0.0-20200925115855-a888f7261b9a"
       - dependencies: []
         name: "github.com/cachito-testing/cachito-gomod-without-deps"
-        type: "go-package"
+        type: "gomod"
         version: "v0.0.0-20200925115855-a888f7261b9a"
     dependencies: []
   expected_files:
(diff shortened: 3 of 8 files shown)


tests/integration/test_data/gomod_packages.yaml
```diff
@@ -14,11 +14,11 @@ without_deps:
     packages:
       - dependencies: []
         name: "github.com/cachito-testing/cachito-gomod-without-deps"
-        type: "gomod"
+        type: "go-package"
```
brunoapimentel, line 17: All changes made to the `tests/integration/test_data` files are only sorting the packages/dependencies, matching them to the output produced while running the integration tests.

This allowed us to keep the sorting function as a private member of the Request model.

cachito/web/models.py
```diff
@@ -930,6 +935,21 @@ def _state_to_json(state):
         rv.update(latest_state)
         return rv
 
+    def _sort_pkgs_in_place(self, packages):
```
akhmanova, line 938: We have almost the same function for integration tests. `sort_pkgs_and_deps_in_place()`. Maybe we can make them identical. But only if it makes sense for you.

akhmanova: Sorry, I see the change in the end, NVM :D

tests/integration/utils.py
```diff
@@ -274,27 +274,6 @@ def make_list_of_packages_hashable(data):
     return sorted([[i["name"], i["type"], i["version"]] for i in data])
 
 
-def sort_pkgs_and_deps_in_place(packages=None, dependencies=None):
```
akhmanova, line 277: I don't think it's a good idea to remove the function completely. It can be helpful for creating new tests / changing existing ones.

akhmanova: @chmeliik @ejegrova  WDYT?

cachito/web/models.py
```diff
@@ -930,6 +935,21 @@ def _state_to_json(state):
...
+        Sorting order: name -> version -> type.
+        If a package has a "dependencies" list, the packages inside it will be sorted as well.
+
+        :param list packages: the list of packages
+        """
+        packages.sort(key=lambda x: (x["name"], x["version"], x["type"]))
+
+        for package in packages:
+            if "dependencies" in package:
+                self._sort_pkgs_in_place(package["dependencies"])
```
tkdchen, line 951: How about sort deps directly like this

```
sort_key = lambda x: (x["name"], x["version"], x["type"])
package["dependencies"].sort(sort_key)
```

Just thinking there seems no extra benefit to have one level recursion to sort the deps.

Dependency also has `dev`. Should it be added to the sort key for sorting deps?

Either packages or dependecies can't have two ones with same name, version and type, remove the version from sort key?

For the form of the sort key in a tuple, does the elements order matter for the sort? IMO, it would be good to sort by the type firstly, then the name. So, `(name, type)` or `(type, name)`, which one works for it?

tests/integration/utils.py
chmeliik: I don't think we'll need it for creating new tests anymore. The Cachito API will now always return the packages and dependencies properly sorted, so it should be easy to just copy-paste the content when creating the test data

cachito/web/models.py
chmeliik: +1 for sorting directly, `dependencies` do not have their own `dependencies` so recursion is not needed
+1 for sorting also by `dev` for dependencies

I don't think we can remove `version` from the sorting key. Consider this:
```json
{
  "dependencies": [
    {"name": "requests", "version": "2.25.0", "type": "pip", "dev": false},
    {"name": "requests", "version": "2.24.0", "type": "pip", "dev": false}
  ]
}
```
^ this is a valid case, and we need to make sure the order will be deterministic.

I like the idea of sorting by type first. I'd say there is also an argument for sorting by name first, as that puts `go-package` packages closer to their corresponding `gomod` modules. But type would make more sense for requests that use multiple package managers. Ultimately we just need the order to be deterministic, I have no strong preference either way.

tkdchen:
Since Cachito stores packages and dependencies into `packages.json`, and then read them from that file directly as part of the Request JSON response. Probably `aggregate_packages_data` is the good place to sort the packages and dependencies. That is, sort them only once, not on every HTTP request to get a Request.

cachito/web/models.py
brunoapimentel: So, I went with the `(type, dev, name, version)` order. I also got rid of the recursion :)
Let me know if you're ok it.

brunoapimentel:
> Since Cachito stores packages and dependencies into `packages.json`, and then read them from that file directly as part of the Request JSON response. Probably `aggregate_packages_data` is the good place to sort the packages and dependencies. That is, sort them only once, not on every HTTP request to get a Request.

I was looking into this, and I think it makes a lot of sense, but I'm still not sure how to do it. 

For starters, we'd need to keep the sorting on both the API and the worker's side, since the requests that are currently kept in the database are not sorted. From what I've talked to @chmeliik, we'd need to keep this redudancy until we drop all the database integration and delete the old requests.

The other question I had was about the the packages.json file that is generated. From the tests I ran locally, I saw it create a JSON with a `packages` key and the `dependencies` inside it, which differs from how it is presented on the API (the dependencies are also listed within the same level of depth of packages in the response).

I think we can discuss this points and see if it still makes sense to be done in this PR, or if we should get another story for this.

chmeliik:
> The other question I had was about the packages.json file that is generated. From the tests I ran locally, I saw it creates a JSON with a `packages` key and the `dependencies` inside it, which differs from how it is presented on the API (the dependencies are also listed within the same level of depth of packages in the response).

The top-level dependencies can be derived from the packages by merging all the individual `packages.dependencies` lists and de-duplicating by `(type, dev, name, version)`. But that is a good point, we will need to sort the derived dependencies list anyway.

IMO we can leave the sorting of the JSON files for a separate story if needed. The only downside of doing it that way would be that deprecating the API sorting (at least the `packages` sorting, we would still need to sort the top-level `dependencies`) would be a two step deployment.
1. Start sorting JSON files, deploy to prod
2. Wait a week until all un-sorted files are gone, deploy change that removes API sorting

It might also be interesting to check how long the sorting takes for a large request, something like tektoncd:
```shell
    -X POST \
    -d '{
  "repo": "https://github.com/openshift/tektoncd-pipeline.git",
  "ref": "2ba62c8cfc98b74a6d02e692ce181add8b2184cd",
  "pkg_managers": [
    "gomod"
  ]
}'
```
I don't think it will be that long, but it would be good to make sure.

cachito/workers/tasks/utils.py
```diff
@@ -233,6 +234,26 @@ def set_packages_and_deps_counts(request_id: int, packages_count: int, dependenc
     )
 
 
+def sort_packages_and_deps_in_place(packages):
+    """
+    Sorts lists of packages in place.
+
+    Sorting order: name -> version -> type.
```
chmeliik, line 241: This is now `type -> dev -> name -> version`

cachito/workers/tasks/utils.py
```diff
@@ -233,6 +234,26 @@ def set_packages_and_deps_counts(request_id: int, packages_count: int, dependenc
     )
 
 
+def sort_packages_and_deps_in_place(packages):
```
chmeliik, line 237: It would be good to have a unit test for this

chmeliik (commented):
Looks good, just needs a minor docstring update and a unit test

cachito/workers/tasks/utils.py
brunoapimentel: Done!

cachito/workers/tasks/utils.py
brunoapimentel: Done!

chmeliik (approved):
LGTM :+1:

PR #484, opened 2021-06-16
Logs healthcheck requests as they arrive

brunoapimentel:
CLOUDBLD-5966

Signed-off-by: Bruno Pimentel <bpimente@redhat.com>

### DIFF
--- a/cachito/web/app.py
+++ b/cachito/web/app.py
@@ -26,6 +26,8 @@ def healthcheck():
     This returns a 200 response if the application is alive and able to serve requests. It returns
     a 500 response otherwise.
     """
+    current_app.logger.info("A healthcheck request was received")
+
     try:
         db.session.execute("SELECT 1 FROM request LIMIT 0").fetchall()
     except SQLAlchemyError:
--- a/cachito/web/app.py
+++ b/cachito/web/app.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import logging
 import os
+from timeit import default_timer as timer
 
 from flask import current_app, Flask
 from flask.logging import default_handler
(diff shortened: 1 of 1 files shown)


akhmanova (approved):
LGTM :)

cachito/web/app.py
```diff
@@ -29,7 +30,12 @@ def healthcheck():
     current_app.logger.info("A healthcheck request was received")
 
     try:
+        start_time = timer()
+
         db.session.execute("SELECT 1 FROM request LIMIT 0").fetchall()
+
+        end_time = (timer() - start_time) * 1000
+        current_app.logger.info("The healthcheck database query took %0.3f milliseconds", end_time)
```
chmeliik, line 38: Based on https://docs.python.org/3/library/timeit.html#timeit.default_timer, `default_timer` is this thing https://docs.python.org/3/library/time.html#time.perf_counter, which counts in seconds. Unless I am missing something, multiplying seconds by 1000 will not give you milliseconds  :D

brunoapimentel: Well, 1000 milliseconds == 1 second, right? I got me so unsure of it that I went to look it up online :sweat_smile: 

Anyway, the only reason I did this is because the results are quite often fractions of a second, so I thought it would look better on the logs. We could simply log in seconds, if you prefer ;)

tkdchen: Why convert to milliseconds? How about just log the time difference directly? According to the livenessProbe config, 0.xxxxx should always be the expected value.

chmeliik: Oh right, I was indeed missing something :D
Yeah, we can just keep it in seconds

brunoapimentel: Done :D

tkdchen (approved):

chmeliik (approved):
LGTM :+1:

PR #487, opened 2021-06-17
Fixed logging on cachito-cleanup script

dcho5:
CLOUDBLD-2953

### DIFF
--- a/cachito/workers/cleanup_job.py
+++ b/cachito/workers/cleanup_job.py
@@ -22,6 +22,7 @@ payload = {"state": state, "state_reason": state_reason}
 
 def main():
     """Mark all end of life requests as stale using the REST API."""
+    logging.basicConfig(level=logging.INFO)
     for state in ("complete", "in_progress", "failed"):
         stale_candidate_requests = find_all_requests_in_state(state)
         identify_and_mark_stale_requests(stale_candidate_requests)
@@ -114,5 +115,4 @@ def mark_as_stale(request_id):
 
 
 if __name__ == "__main__":
-    log.setLevel(logging.INFO)
     main()


dcho5:
Wasn't able to properly test this code locally, will need to wait for merge to test on OpenShift Web Console.

Note for @chmeliik : the method you provided me with for local testing doesn't seem to work. I tested cleanup_job.py by itself, and it printed the logs to the console normally, while running cachito-cleanup didn't appear to do anything. I tested this method further with @lcarva 's suggestion to add a raise ValueError line in the main function. Even after this, cachito-cleanup did raise any errors. Could I have done something differently?

chmeliik:
Looking into it some more, it looks like I was wrong about cachito-worker picking up the changes automatically. The cachito code is mounted at `/src`, and updating anything will reload the *main* worker process. But the `cachito-cleanup` script does not run `/src/cachito/workers/cleanup_job.py`, it runs `/usr/local/lib/python3.9/site-packages/cachito/workers/cleanup_job.py`. That one is created when the image is built and does *not* get auto-updated.

You can make sure the script is updated by rebuilding the worker image. But that will take quite long, so to help with testing, you can just make a symlink from the file in /src to the one in /usr. That way, the changes will be picked up automatically.

That by itself will not be enough to fix the logging problem. Check the `cachito-cleanup` script (find the file with `command -v cachito-cleanup`) and you may be able to see why. Hint: the cleanup_job module does not run as `__main__`.

dcho5:
> Looking into it some more, it looks like I was wrong about cachito-worker picking up the changes automatically. The cachito code is mounted at `/src`, and updating anything will reload the _main_ worker process. But the `cachito-cleanup` script does not run `/src/cachito/workers/cleanup_job.py`, it runs `/usr/local/lib/python3.9/site-packages/cachito/workers/cleanup_job.py`. That one is created when the image is built and does _not_ get auto-updated.
> 
> You can make sure the script is updated by rebuilding the worker image. But that will take quite long, so to help with testing, you can just make a symlink from the file in /src to the one in /usr. That way, the changes will be picked up automatically.
> 
> That by itself will not be enough to fix the logging problem. Check the `cachito-cleanup` script (find the file with `command -v cachito-cleanup`) and you may be able to see why. Hint: the cleanup_job module does not run as `__main__`.

- Creating a symlink between the two files proved to be useful tool in debugging, will use for future tickets; allowed me to test the logs locally
- Found an error in the initial PR - `basicConfig` should be moved into the `main` function as this is what `cachito-cleanup` calls

cachito/workers/cleanup_job.py
```diff
@@ -21,6 +21,7 @@
 
 
 def main():
+    logging.basicConfig(level=logging.INFO)
     """Mark all end of life requests as stale using the REST API."""
```
chmeliik, line 25: Let's move the basicConfig call below the docstring

chmeliik (commented):
LGTM, just a minor comment :-)

cachito/workers/cleanup_job.py
dcho5: Got it, I initially put it above the docstring because I thought it directly pertained to the lines directly below.

chmeliik (approved):
LGTM :+1:

cachito/workers/cleanup_job.py
chmeliik: Right, docstrings should always be the first thing under the name of the function (unlike comments, which can be anywhere)

PR #488, opened 2021-06-18
Read request packages and dependencies from packages.json

tkdchen:
CLBOUDBLD-4728
CLBOUDBLD-4729

Both endpoints /requests and /requests/id response the Request JSON
serialized from Request.to_json. Cachito starts to read a request's
packages and dependencies from the packages.json instead of from the
database.

### DIFF
--- a/cachito/utils.py
+++ b/cachito/utils.py
@@ -1,0 +1,173 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import json
+import logging
+import os
+from pathlib import Path
+from typing import Any, Dict, Iterator, List, Tuple, Union
+
+from cachito.errors import CachitoError
+
+log = logging.getLogger(__name__)
+
+
+def package_sort_key(package: Dict[str, Any]) -> Tuple[str, bool, str, str]:
+    """Return the sort key for sorting packages.
+
+    :param package: a mapping representing a package information. It must
+        contain four keys, type, dev, name and version.
+    :type package: dict[str, any]
+    :return: a four-elements tuple as the sort key in the order of type, dev,
+        name and version.
+    :rtype: tuple[str, bool, str, str]
+    """
+    return package["type"], package.get("dev", False), package["name"], package["version"]
+
+
+def sort_packages_and_deps_in_place(packages: List[Dict[str, Any]]) -> None:
+    """
+    Sorts lists of packages in place.
+
+    Sorting order: type -> dev -> name -> version.
+    If a package has a "dependencies" list, the packages inside it will be sorted as well.
+
+    :param list packages: the list of packages
+    """
+    packages.sort(key=package_sort_key)
+    for package in packages:
+        if "dependencies" in package:
+            package["dependencies"].sort(key=package_sort_key)
+
+
+def _package_equal(left: Union[Dict[str, Any], None], right: Dict[str, Any]) -> bool:
+    """Check if left package equals to the right.
+
+    Used by unique_packages internally.
+    """
+    if left is None:
+        return False
+    return (
+        left["name"] == right["name"]
+        and left["type"] == right["type"]
+        and left["version"] == right["version"]
+        and left.get("dev", False) == right.get("dev", False)
+    )
+
+
+def unique_packages(packages: List[Dict[str, Any]]) -> Iterator:
+    """Remove duplicate packages.
+
+    If two packages as well as dependencies have same name, type, version and
+    dev, they are duplicated. This function assumes the packages have been
+    sorted already.
+
+    :param packages: a list of sorted packages to be deduplicated. If two
+        packages have same name, type, version and dev, they will be considered
+        as duplicate package.
+    :type packages: list[dict[str, any]]
+    """
+    i = -1
+    j = 0
+    while j < len(packages):
+        left = None if i < 0 else packages[i]
+        right = packages[j]
+        if _package_equal(left, right):
+            j += 1
+        else:
+            yield right
+            i = j
+            j += 1
+
+
+class PackagesData:
+    """A collection of resolved packages."""
+
+    def __init__(self) -> None:
+        """Initialize an empty PackagesData instance."""
+        self._index = set()
+        self._packages = []
+
+    @property
+    def packages(self) -> List[Dict[str, Any]]:
+        """Get added packages."""
+        return self._packages
+
+    @property
+    def all_dependencies(self) -> List[Dict[str, Any]]:
+        """Gather dependencies together from every package.
+
+        :return: a list of sorted and deduplicated dependencies gathered from
+            every package. If no package is added, an empty list will be returned.
+        :rtype: list[dict[str, any]]
+        """
+        return list(
+            unique_packages(
+                sorted(
+                    (dep for pkg in self._packages for dep in pkg["dependencies"]),
+                    key=package_sort_key,
+                )
+            )
+        )
+
+    def add_package(self, pkg_info: Dict[str, str], path: str, deps: List[Dict[str, Any]]) -> None:
+        """Add a package with deps.
+
+        :param dict[str, str] pkg_info: a mapping containing a package information.
+            It must have ``name``, ``type`` and ``version`` key/value pairs.
+        :param str path: the path where the package is retreived. Consult with the
+            ``fetch_*_source`` for the defailed information about a package's path.
+        :param deps: a list of depencencies the package has.
+        :type deps: list[dict[str, any]]
+        :raises CachitoError: if there is a package with same name, type and version
+            has been added already.
+        """
+        key = (pkg_info["name"], pkg_info["type"], pkg_info["version"])
+        if key in self._index:
+            raise CachitoError(f"Duplicate package: {pkg_info!r}")
+        self._index.add(key)
+        package = {
+            "name": pkg_info["name"],
+            "type": pkg_info["type"],
+            "version": pkg_info["version"],
+            "dependencies": deps,
+        }
+        if path != os.curdir:
+            package["path"] = path
+        self._packages.append(package)
+
+    def write_to_file(self, file_name: Union[str, Path]) -> None:
+        """Write the added packages to a file as JSON data.
+
+        It ensures that the packages and every package's dependencies are sorted
+        by the combination in the order of type, dev, name and version.
+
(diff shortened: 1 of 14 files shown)


cachito/web/models.py
```diff
@@ -890,31 +864,22 @@ def _state_to_json(state):
...
+            )
+            if bundle_dir.packages_data.exists():
+                packages_data = json.loads(bundle_dir.packages_data.read_bytes())
+                packages = packages_data.get("packages")
+                if packages:
+                    rv["packages"] = packages
+                    deps = []
+                    for package in packages_data["packages"]:
+                        deps.extend(package.get("dependencies", []))
+                    rv["dependencies"] = deps
```
chmeliik, line 881: The dependencies will need to be de-duplicated by `(name, version, type, dev)`

It would also be good to add unit tests where 2 different packages have the same dependency

brunoapimentel: I think it's also worth mentioning that the dependencies will need to be sorted here as well (once #481 is merged, that is).

cachito/web/utils.py
```diff
@@ -1,5 +1,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 from collections import OrderedDict
+from collections.abc import Iterator
+from typing import Any, Dict, List
```
chmeliik, line 4: If the `Iterator` import is only for typing, could it be imported from the `typing` modules instead?

cachito/web/models.py
```diff
@@ -1,9 +1,11 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
+from cachito.web.utils import unique_packages
```
chmeliik, line 2: Could this be moved to the other `cachito` imports?

cachito/web/models.py
```diff
@@ -893,35 +868,25 @@ def _state_to_json(state):
...
-                )
-                for req_package in self.request_packages
-            ]
-
-            sort_packages_and_deps_in_place(rv["packages"])
+            bundle_dir = RequestBundleDir(
+                self.id, root=flask.current_app.config["CACHITO_BUNDLES_DIR"]
+            )
+            if bundle_dir.packages_data.exists():
+                packages_data = json.loads(bundle_dir.packages_data.read_bytes())
```
chmeliik, line 878: Would it make sense to use the `PackagesData` class to load this data instead? The class could even get a new method for generating the top-level dependencies list

tests/test_api_v1.py
```diff
@@ -2439,3 +2303,160 @@ def test_filter_requests(
...
+                    {"name": "async", "type": "npm", "version": "1.2.0"},
+                    {"name": "underscore", "type": "npm", "version": "1.13.0"},
+                ],
+            },
+        ],
+    }
+
+    expected_deps = [
+        {"name": "d1", "type": "go-package", "version": "1"},
+        {"name": "d2", "replaces": None, "type": "go-package", "version": "2"},
```
chmeliik, line 2431: Seeing this, I realized that Cachito currently always sets `"replaces": None` for dependencies that do not have replacements.

https://github.com/release-engineering/cachito/blob/ece43b86e31e7f858bd905cd58ed3887bd7e52d4/cachito/web/models.py#L477-L478
https://github.com/release-engineering/cachito/blob/ece43b86e31e7f858bd905cd58ed3887bd7e52d4/cachito/web/models.py#L903-L906

I'm not sure how much we need to worry about this detail of the API response, but if we want to be 100% backwards compatible, we need to do the same.

cachito/web/models.py
```diff
@@ -893,35 +868,25 @@ def _state_to_json(state):
...
-                    subpath=req_package.subpath,
-                )
-                for req_package in self.request_packages
-            ]
-
-            sort_packages_and_deps_in_place(rv["packages"])
+            bundle_dir = RequestBundleDir(
+                self.id, root=flask.current_app.config["CACHITO_BUNDLES_DIR"]
+            )
+            if bundle_dir.packages_data.exists():
```
brunoapimentel, line 877: I will need to load the packages data as well to generate the content manifest (CLOUDBLD-4730), so I was thinking that it could be a good idea to make this reusable. I have two suggestions on how it could be made:

1- Load the package data from the json file right after the object is loaded from the database. This way the data would always be available and its behavior would be similar to having the packages on the database.

2- Have this behavior be done on private function on the class, so it would be easy to just reuse this part of the code. This way, it's easier to implement and we'd have to worry less about corner cases of suggestion 1.

tkdchen: I have changes to the `PackagesData` locally, I'll push it later and please review it in advance.

tests/test_api_v1.py
tkdchen: @chmeliik 

Is this already covered in every `fetch_*_source` when writing the packages data? The resolved dependencies should have replaces, and they will be written into the package manager specific packages data file and then gathered into the `packages.json`.

chmeliik: I don't think so. The dependencies returned by the `resolve_*` functions may or may not have the "replaces" key. In fact, I think only deps from `resolve_gomod` have the key.

cachito/utils.py
```diff
@@ -0,0 +1,174 @@
...
+                return
+            for p in packages:
+                self.add_package(p, p.get("path", os.curdir), p["dependencies"])
+
+    def sort(self) -> None:
+        """Sort packages and each package's dependencies in place."""
+        sort_packages_and_deps_in_place(self._packages)
+
+    def all_dependencies(self) -> List[Dict[str, Any]]:
+        """Gather dependencies together from very package.
```
chmeliik, line 161:
```suggestion
        """Gather dependencies together from every package.
```

cachito/utils.py
```diff
@@ -0,0 +1,174 @@
...
+
+        :return: a list of sorted and deduplicated dependencies gathered from
+            every package. If no package is added, an empty list will be returned.
+        :rtype: list[dict[str, any]]
+        """
+        return list(
+            unique_packages(
+                sorted(
+                    itertools.chain(*[pkg["dependencies"] for pkg in self._packages]),
```
chmeliik, line 170:
```suggestion
                    itertools.chain.from_iterable(pkg["dependencies"] for pkg in self._packages),
```

cachito/workers/tasks/utils.py
```diff
@@ -20,9 +18,6 @@
     "get_request",
     "get_request_state",
     "set_request_state",
-    "set_packages_and_deps_counts",
```
chmeliik, line 23: I think this should stay here

tests/test_utils.py
```diff
@@ -72,3 +76,264 @@ def test_deep_sort_icm(orig_items):
...
+        {"name": "pkg6", "type": "pip", "version": "1.0.0", "dev": False},
+        {"name": "pkg5", "type": "pip", "version": "1.0.0", "dev": True},
+    ]
+
+    sort_packages_and_deps_in_place(packages)
+
+    assert packages == sorted_packages
+
+
+class TestPackagesData:
```
chmeliik, line 176: Could we have a test for the all_dependencies method?

cachito/web/models.py
```diff
@@ -893,35 +866,14 @@ def _state_to_json(state):
...
-                for req_package in self.request_packages
-            ]
-
-            sort_packages_and_deps_in_place(rv["packages"])
+            bundle_dir = RequestBundleDir(
+                self.id, root=flask.current_app.config["CACHITO_BUNDLES_DIR"]
+            )
+            packages_data = PackagesData()
+            packages_data.load(bundle_dir.packages_data)
+            packages_data.sort()
```
chmeliik, line 874: Technically we don't need to sort here anymore, when loading a single sorted file, the packages data will be sorted.

We included the sorting changes in the same release as the packages.json generation, so there should not be any unsorted files in any of the Cachito environments.

chmeliik (commented):
+1 from moving the `PackagesData` class to a shared module, I just have a suggestion about the naming of that module. I would also like that change to have its own commit, but it might be difficult to split them up now so it's not a big deal.

cachito/web/models.py
brunoapimentel: I like it, I think it looks a lot better now :)

cachito/utils.py
```diff
@@ -0,0 +1,174 @@
...
+        contain four keys, type, dev, name and version.
+    :type package: dict[str, any]
+    :return: a four-elements tuple as the sort key in the order of type, dev,
+        name and version.
+    :rtype: tuple[str, bool, str, str]
+    """
+    return package["type"], package.get("dev", False), package["name"], package["version"]
+
+
+def sort_packages_and_deps_in_place(packages: List[Dict[str, Any]]) -> None:
```
brunoapimentel, line 28: Besides some unit tests, it seems this function is now only being used inside the PackageData class.

If we could work around the tests, it would be possible to make it a private function inside the PackageData class, and this way we'd have less `utils` going around.

We don't need to do this right now, though, it's just a thought.

cachito/web/models.py
```diff
@@ -16,10 +16,11 @@
 import urllib.parse
 from werkzeug.exceptions import Forbidden
 
-from cachito.web import content_manifest
 from cachito.errors import ContentManifestError, ValidationError
+from cachito.paths import RequestBundleDir
+from cachito.web import content_manifest
 from cachito.web import db
-from cachito.workers.tasks.utils import sort_packages_and_deps_in_place
+from cachito.utils import PackagesData
```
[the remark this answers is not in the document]
tkdchen: Looks errors.py and paths.py can also be moved into the common module. I'll make this change in a separate commit.

chmeliik (commented):
I found a regression after running integration tests. All requests now have `packages_count = 0, dependencies_count = 0` set in the database. I think the reason is that in the `finalize_request()` task, `get_request()` is called before `aggregate_packages_data()`. That used to work before, but not anymore

chmeliik (commented):
LGTM :+1:

Are you planning to rename the `cachito.utils` module in this PR or a different one?

tkdchen:
> Are you planning to rename the cachito.utils module in this PR or a different one?

@chmeliik Yes. The changes are in local branch already. When this PR is good enough and merged, I'll rebase the local changes and push for review.

chmeliik (approved):

PR #490, opened 2021-06-18
Update error message for git-submodule conflict

thrbkova:
* CLOUDBLD-3865

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -674,7 +674,9 @@ def _validate_gitsubmodule_exclusivity(pkg_manager_paths, mutually_exclusive):
         if any(path != "." for path in paths):
             msg = (
                 f"Cannot process non-root packages with '{pkg_manager}' "
-                "when 'git-submodule' is also set"
+                "when 'git-submodule' is also set.\nSolution: remove the git submodules "
+                "and instead process the upstream repos (corresponding to the submodules) directly. "
+                "Each repo should be processed in a separate request."
             )
             raise ValidationError(msg)
 


cachito/web/models.py
```diff
@@ -674,7 +674,7 @@ def _validate_gitsubmodule_exclusivity(pkg_manager_paths, mutually_exclusive):
         if any(path != "." for path in paths):
             msg = (
                 f"Cannot process non-root packages with '{pkg_manager}' "
-                "when 'git-submodule' is also set"
+                "when 'git-submodule' is also set. Consider using multiple upstream repos."
```
chmeliik, line 677: Ideally, the error message should mention how to solve the problem. Here, the solution would typically be to remove the git submodules and instead process the upstream repos (corresponding to the submodules) directly.

Also, since processing multiple upstream repos is not a Cachito feature, the message may need to say that the individual repos should be processed in separate requests.

cachito/web/models.py
```diff
@@ -674,7 +674,9 @@ def _validate_gitsubmodule_exclusivity(pkg_manager_paths, mutually_exclusive):
         if any(path != "." for path in paths):
             msg = (
                 f"Cannot process non-root packages with '{pkg_manager}' "
-                "when 'git-submodule' is also set"
+                "when 'git-submodule' is also set.\nSolution: remove the git submodules"
+                "and instead process the upstream repos (corresponding to the submodules) directly."
+                "Each repo should be processed in a separate request."
```
chmeliik, line 679:
```suggestion
                "when 'git-submodule' is also set.\nSolution: remove the git submodules "
                "and instead process the upstream repos (corresponding to the submodules) directly. "
                "Each repo should be processed in a separate request."
```
Missing two spaces. You can check the error message with
```shell
    -X POST \
    -d '{
  "repo": "https://whatever.org",                                
  "ref": "b470410e50caff0447bc18ca4f011663681e7e17",
  "pkg_managers": ["git-submodule", "gomod"],
  "packages": {"gomod": [{"path": "foo"}]}
}'
```

chmeliik (approved):
LGTM :+1:

PR #493, opened 2021-06-22
Fix cached dependecy integration tests

brunoapimentel:
The gomod test repositories need to have their versions set during the cached dependency integration test. However, the replacement was only being made to the first package that appeared on the packages list (the replacement index was hard-coded).

This PR fixes that and sorts the dependencies so they match the copy on the [Gitlab repository](https://gitlab.cee.redhat.com/osbs/cachito-tests).

CLOUDBLD-3382
Signed-off-by: Bruno Pimentel bpimente@redhat.com

### DIFF
--- a/tests/integration/test_data/cached_dependencies.yaml
+++ b/tests/integration/test_data/cached_dependencies.yaml
@@ -46,20 +46,20 @@ pip_cached_deps:
   response_expectations:
     dependencies:
     - dev: false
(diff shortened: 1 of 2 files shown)


akhmanova (approved):
LGTM, thank you!

tkdchen (commented):
LGTM.

PR #499, opened 2021-06-25
Move shared code to common module

tkdchen:


cachito/common/packages_data.py
```diff
@@ -6,7 +6,7 @@
 from pathlib import Path
 from typing import Any, Dict, Iterator, List, Tuple, Union
 
-from cachito.errors import CachitoError
+from cachito.common.errors import CachitoError
```
chmeliik, line 9: Maybe `errors` could stay in `cachito.errors`, it's pretty common to keep exceptions in `<project>.errors`

chmeliik (commented):
LGTM :+1:

Just an optional suggestion about keeping errors where they are

chmeliik (approved):
LGTM :+1:

PR #500, opened 2021-06-25
Move sort_packages_and_deps_in_place to PackagesData.sort

tkdchen:
This PR requires #499 . Please review in advance. I'll rebase this PR once 499 is merged.

### DIFF
--- a/cachito/common/packages_data.py
+++ b/cachito/common/packages_data.py
@@ -11,7 +11,7 @@ from cachito.errors import CachitoError
 log = logging.getLogger(__name__)
 
 
-def package_sort_key(package: Dict[str, Any]) -> Tuple[str, bool, str, str]:
+def _package_sort_key(package: Dict[str, Any]) -> Tuple[str, bool, str, str]:
     """Return the sort key for sorting packages.
 
     :param package: a mapping representing a package information. It must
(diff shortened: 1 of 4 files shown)


chmeliik (commented):
Looks good :+1:

The `_package_sort_key` and `_package_equal` methods could be replaced by a `Package` class with `__lt__` and `__eq__` methods, but that's completely optional (and better left for a different PR anyway).

I heard that @brunoapimentel has some plans for a Package class already :-)

tkdchen:
@brunoapimentel Please review. Are you going to add that feature to Package class?

brunoapimentel:
I was planning on having a new Package class, that would be based on the packages loaded from the json file, as a replacement for the current one we have. The reason for this is that the current model class holds all the purl generation methods.

But I discussed it with Adam and we thought it is not the best place for those method. I've been working on a class that will be exclusively used for the content manifest, and it seems like it is a good idea to leave all these responsibilities on the same module. So, I guess it would not help us here. :(

Anyway, I think having a Package class to replace the current model class could indeed be a good idea, and we could implement the methods suggested by Adam. I do agree that it could be left for another PR. :)

brunoapimentel:
LGTM :+1:

PR #502, opened 2021-06-25
Generate content manifest from json file

brunoapimentel:
CLOUDBLD-4730

### DIFF
--- a/cachito/web/content_manifest.py
+++ b/cachito/web/content_manifest.py
@@ -1,10 +1,15 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 from copy import deepcopy
+from dataclasses import dataclass, field
+import os
+import re
+import urllib.parse
 
 import flask
+import pkg_resources
+from cachito.errors import ContentManifestError
 from cachito.web.utils import deep_sort_icm
 from cachito.workers.pkg_managers import gomod
-
 
 PARENT_PURL_PLACEHOLDER = "PARENT_PURL"
 
@@ -31,13 +36,15 @@ BASE_ICM = {
 class ContentManifest:
     """A content manifest associated with a Cachito request."""
 
-    def __init__(self, request):
+    def __init__(self, request, packages):
         """
         Initialize ContentManifest.
 
         :param Request request: the request to generate a ContentManifest for
+        :param list[Package] packages: the packages contained in the request
         """
         self.request = request
+        self.packages = packages
         # dict to store go package level data; uses the package id as key to identify a package
         self._gopkg_data = {}
         # dict to store go module level purl dependencies. Module names are used as keys
@@ -74,7 +81,7 @@ class ContentManifest:
         """
         if dependency.type == "go-package":
             icm_dependency = {"purl": dependency.to_purl()}
-            self._gopkg_data[package.id]["dependencies"].append(icm_dependency)
+            self._gopkg_data[package]["dependencies"].append(icm_dependency)
 
     def set_go_package_sources(self):
         """
@@ -152,9 +159,9 @@ class ContentManifest:
         pkg_type_data = getattr(self, f"_{pkg_type}_data")
 
         icm_dependency = {"purl": dependency.to_purl()}
-        pkg_type_data[package.id]["sources"].append(icm_dependency)
+        pkg_type_data[package]["sources"].append(icm_dependency)
         if not dependency.dev:
-            pkg_type_data[package.id]["dependencies"].append(icm_dependency)
+            pkg_type_data[package]["dependencies"].append(icm_dependency)
 
     def to_json(self):
         """
@@ -170,44 +177,43 @@ class ContentManifest:
         self._yarn_data = {}
         self._gitsubmodule_data = {}
 
-        # Address the possibility of packages having no dependencies
-        for request_package in self.request.request_packages:
-            package = request_package.package
+        for package in self.packages:
 
             if package.type == "go-package":
-                purl = package.to_top_level_purl(self.request, subpath=request_package.subpath)
+                purl = package.to_top_level_purl(self.request, subpath=package.path)
                 self._gopkg_data.setdefault(
-                    package.id,
+                    package,
                     {"name": package.name, "purl": purl, "dependencies": [], "sources": []},
                 )
             elif package.type == "gomod":
-                purl = package.to_top_level_purl(self.request, subpath=request_package.subpath)
+                purl = package.to_top_level_purl(self.request, subpath=package.path)
                 self._gomod_data.setdefault(package.name, {"purl": purl, "dependencies": []})
             elif package.type in ("npm", "pip", "yarn"):
-                purl = package.to_top_level_purl(self.request, subpath=request_package.subpath)
+                purl = package.to_top_level_purl(self.request, subpath=package.path)
                 data = getattr(self, f"_{package.type}_data")
-                data.setdefault(package.id, {"purl": purl, "dependencies": [], "sources": []})
+                data.setdefault(package, {"purl": purl, "dependencies": [], "sources": []})
             elif package.type == "git-submodule":
-                purl = package.to_top_level_purl(self.request, subpath=request_package.subpath)
+                purl = package.to_top_level_purl(self.request, subpath=package.path)
                 self._gitsubmodule_data.setdefault(
-                    package.id, {"purl": purl, "dependencies": [], "sources": []}
+                    package, {"purl": purl, "dependencies": [], "sources": []}
                 )
             else:
                 flask.current_app.logger.debug(
                     "No ICM implementation for '%s' packages", package.type
                 )
 
-        for req_dep in self.request.request_dependencies:
-            if req_dep.package.type == "go-package":
-                self.process_go_package(req_dep.package, req_dep.dependency)
-            elif req_dep.package.type == "gomod":
-                self.process_gomod(req_dep.package, req_dep.dependency)
-            elif req_dep.package.type == "npm":
-                self.process_npm_package(req_dep.package, req_dep.dependency)
-            elif req_dep.package.type == "pip":
-                self.process_pip_package(req_dep.package, req_dep.dependency)
-            elif req_dep.package.type == "yarn":
-                self.process_yarn_package(req_dep.package, req_dep.dependency)
+        for package in self.packages:
+            for dependency in package.dependencies:
+                if package.type == "go-package":
+                    self.process_go_package(package, dependency)
+                elif package.type == "gomod":
+                    self.process_gomod(package, dependency)
+                elif package.type == "npm":
+                    self.process_npm_package(package, dependency)
+                elif package.type == "pip":
+                    self.process_pip_package(package, dependency)
+                elif package.type == "yarn":
+                    self.process_yarn_package(package, dependency)
 
         # Adjust source level dependencies for go packages
         self.set_go_package_sources()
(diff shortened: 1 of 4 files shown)


tkdchen:
By considering CLOUDBLD-4731 together, we could have a single class `Package` eventually.

Based on current code base, the first package of a request is created by the backend by sending a PATCH request to the endpoint `/api/v1/requests/id`. That issue is about removing this behavior and relative tables from database. Since all the resolved packages and dependencies are stored in the `packages.json` already, and we don't store any such data in the database, so the `package` table will be removed eventually, that is the `Package` model will be removed. Then, the new class `Package` can be created to provide the expected behaviors.

Except the benefits we have discussed, the new class seems can also help us to avoid passing package information dict in the code base, like `{"name": "...", "type": "...", "version": "..."}`.

Just a quick idea, we could have:

```python
class BasePackageInfo:
    name: str
    type: str
    version: str

    def to_json(self) -> Dict[str, str]:
        pass

class Package(BasePackageInfo):
    dev: bool
    subpath: str

    def to_json(self) -> Dict[str, Any]:
        pass

    def to_purl(self) -> str:
        pass

    def to_vcs_purl(self) -> str:
        pass

    def to_top_level_purl(self) -> str:
        pass
    
    def content_manifest(self) -> Dict[str, Any]:
        pass
```

Maybe we could have concrete package classes for different package managers to implement the `*_purl` and `content_manifest` interface specifically. Alternatively, separate dedicated classes for individual package manager to generate purl and content manifest, like `PackageURLGenerator` and `PackageContentManifestGenerator`, and their subclasses e.g. `NPMPackageURLGenerator`.

brunoapimentel:
> Except the benefits we have discussed, the new class seems can also help us to avoid passing package information dict in the code base, like {"name": "...", "type": "...", "version": "..."}.

+1 to that. I think it makes perfect sense and would make the code more readable and maintanable.

brunoapimentel:
> By considering CLOUDBLD-4731 together, we could have a single class Package eventually.

This was my first thought, but one of the things that I made me consider putting this in a separte class is that the logic to generating the PURLs is kinda big, and it is only used in the web module while generating the content manifests (afaik). So, we could have all this logic encapsulated in this module without cluttering the more generic Package file. 

I don't know if it would make sense that the PURLs would be available for general use in Cachito, in this case, it would indeed be better to have it as you suggested (having it in the common module, within the general use Package class or classes) . Let me know what you think about it.

> Maybe we could have concrete package classes for different package managers to implement the *_purl and content_manifest interface specifically.

I like this idea. I think we could have a much more readable code and maybe it would even be easier to test.

Building on you suggestion, I was thinking we could have subclasses for the BasePackageInfo that would be specialized in generating the PURLs:

```
# cachito/common/package.py
class BasePackageInfo:
    name: str
    type: str
    version: str

    def to_json(self) -> Dict[str, str]:
        pass
 

# cachito/web/content_manifest.py
class GomodManifestPackage(BasePackageInfo):
    dev: bool
    replaces: str

    def to_vcs_purl(self) -> str:
        pass

    def to_top_level_purl(self) -> str:
        pass
    
    def content_manifest(self) -> Dict[str, Any]:
        pass
```

In any case, I think it would be best to leave these ideas as improvements and do them in a following PR (CLOUDBLD-4730 is a little late).

tests/test_content_manifest.py
```diff
@@ -267,71 +272,6 @@ def test_to_json(mock_top_level_purl, app, package, subpath):
...
-            "_gomod_data",
-            # gomod is special, it is only used to finalize go-package data
-            {"example.com/org/project": {"purl": "mock-package-purl", "dependencies": []}},
-        ),
-    ],
-)
-@mock.patch("cachito.web.models.Package.to_top_level_purl")
-# set_go_package_sources must be mocked because it is destructive towards _gopkg_data
-@mock.patch("cachito.web.content_manifest.ContentManifest.set_go_package_sources")
-def test_to_json_properly_sets_internal_data(
```
brunoapimentel, line 313: This test was removed because the `test_process_*` tests are now actually using the to_json logic to function (making it redundant).

cachito/web/content_manifest.py
```diff
@@ -232,3 +237,246 @@ def generate_icm(self, image_contents=None):
...
+        :rtype: dict
+        """
+        rv = {"name": self.name, "type": self.type, "version": self.version}
+        if dependencies is not None:
+            rv["dependencies"] = dependencies
+        if subpath:
+            rv["path"] = subpath
+        return rv
+
+    def to_purl(self):
```
brunoapimentel, line 337: All the logic to generating the purls is copied and pasted from the Package model class, which will be removed in a following PR (CLOUDBLD-4731).

tkdchen: The plan is slightly changed. :) The package and dependency model and data are not removed at this moment and will be kept for a period of time.

chmeliik: @tkdchen could we remove the models? I'd like to keep only the db data but not the corresponding python models.

tkdchen (commented):
Current code LGTM except the name `ContentManifestPackage` always confuses me. :) I think you can create the new Package class in a different module even if the Package model will be kept there for some time. There should not be confusion on the two Package classes. This idea just FYI.

cachito/web/content_manifest.py
```diff
@@ -1,7 +1,12 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 from copy import deepcopy
+import pkg_resources
+import os
+import re
+import urllib.parse
 
 import flask
+from cachito.errors import ContentManifestError, ValidationError
```
chmeliik, line 9: I'd like to keep stdlib, 3rd party and cachito imports separated by an empty line. I think `copy, os, re, urllib` would be the stdlib modules, `flask` and `pkg_resources` 3rd party.

cachito/web/content_manifest.py
tkdchen: @chmeliik Removing the models will not have corresponding database migrations to remove the table and data. Will this bring some potential issue?

cachito/web/content_manifest.py
```diff
@@ -232,3 +237,246 @@ def generate_icm(self, image_contents=None):
...
+
+
+class ContentManifestPackage:
+    """
+    A package within a content manifest.
+
+    It is used primarily to generate a package URL (purl).
+    """
+
+    def __init__(self, name, type, version, dev=False, dependencies=[], path=None, replaces=None):
```
chmeliik, line 249: Mutable defaults (the list) are usually a [bad idea](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments), typically one would use `None` instead

cachito/web/content_manifest.py
```diff
@@ -69,12 +75,12 @@ def process_go_package(self, package, dependency):
...
         Process go-package package.
 
-        :param Package package: the go-package package to process
+        :param ContentManifestPackage package: the go-package package to process
         :param Dependency dependency: the go-package package dependency to process
         """
         if dependency.type == "go-package":
             icm_dependency = {"purl": dependency.to_purl()}
-            self._gopkg_data[package.id]["dependencies"].append(icm_dependency)
+            self._gopkg_data[package.hash]["dependencies"].append(icm_dependency)
```
chmeliik, line 83: While hash collisions are quite unlikely, I would rather use a key that is guaranteed unique (see next suggestion)

cachito/web/content_manifest.py
tkdchen: One issue would be before the tables are removed from the database, every time to generate database migration , alembic will detect the existing tables and generate new migration to remove them. That could make it inconvenient and error-prone probably.

cachito/web/content_manifest.py
```diff
@@ -232,3 +237,246 @@ def generate_icm(self, image_contents=None):
...
+    def __repr__(self):
+        return (
+            f"<{self.__class__.__name__} name={self.name}, type={self.type}, "
+            f"version={self.version}>"
+        )
+
+    @property
+    def hash(self):
+        """Generate the package hash based on its name, type and version."""
+        return hash(f"{self.name}{self.type}{self.version}")
```
chmeliik, line 284: Instead of hashing by the hash, I think we should hash by a unique key, such as a tuple of `(name, version, type, dev)`. Alternatively, you could give this class `__hash__` and `__eq__` to make it properly hashable.

cachito/web/content_manifest.py
```diff
@@ -232,3 +237,246 @@ def generate_icm(self, image_contents=None):
...
+
+    @classmethod
+    def validate_json(cls, package):
```
chmeliik, line 287: Not sure if we need to do this here, if the data is invalid then something went very wrong earlier. I'm afraid we may have lost a bit of validation during the transition to JSON files, but this is probably not the place to do it.

cachito/web/content_manifest.py
```diff
@@ -232,3 +237,246 @@ def generate_icm(self, image_contents=None):
...
+        Create a Package object from JSON.
+
+        :param dict package: the dictionary representing the package
+        :return: the Package object
+        :rtype: ContentManifestPackage
+        """
+        cls.validate_json(package)
+        return cls(**package)
+
+    def to_json(self, dependencies=None, subpath=None):
```
chmeliik, line 320: Is this ever used?

cachito/web/models.py
```diff
@@ -866,11 +876,7 @@ def _state_to_json(state):
             latest_state = states[0]
             rv["state_history"] = states
 
-            bundle_dir = RequestBundleDir(
-                self.id, root=flask.current_app.config["CACHITO_BUNDLES_DIR"]
-            )
-            packages_data = PackagesData()
-            packages_data.load(bundle_dir.packages_data)
+            packages_data = self._get_packages_data()
```
chmeliik, line 879: Nice :rocket:

cachito/web/content_manifest.py
chmeliik: Hmm, that's true. I don't like the idea of keeping around dead code, but I also don't want to drop the data right away...

It seems that we might be able to get around that with some alembic magic: https://alembic.sqlalchemy.org/en/latest/cookbook.html#don-t-generate-any-drop-table-directives-with-autogenerate

cachito/web/content_manifest.py
brunoapimentel: Oh shit, I did not know that :open_mouth: 

That's what you get when you think you're smarter than your IDE :see_no_evil:

cachito/web/content_manifest.py
brunoapimentel: Done!

cachito/web/content_manifest.py
brunoapimentel: Done!

cachito/web/content_manifest.py
brunoapimentel: You're right, I removed the validation.

cachito/web/content_manifest.py
brunoapimentel: Apparently, no :grin:

cachito/web/models.py
brunoapimentel: The tests demanded it!

brunoapimentel:
> Current code LGTM except the name `ContentManifestPackage` always confuses me. :) I think you can create the new Package class in a different module even if the Package model will be kept there for some time. There should not be confusion on the two Package classes. This idea just FYI.

I agree, I did not like the name either. I think it's just Java talking through me, old habits die hard. :sweat_smile: 

I changed the name to simple `Package`, I believe the module itself can give context on what's the class purpose. The code is looking better now :slightly_smiling_face:

cachito/web/content_manifest.py
brunoapimentel: In fact, it was being used in a test :grin: 

I don't think it was enough to justify the function existence, though.

cachito/web/content_manifest.py
chmeliik: Now that package implements `__hash__` and `__eq__`, it should be used as the key directly:
```suggestion
            self._gopkg_data[package]["dependencies"].append(icm_dependency)
```

cachito/web/content_manifest.py
```diff
@@ -232,3 +237,216 @@ def generate_icm(self, image_contents=None):
...
+        )
+
+    def __hash__(self):
+        return hash((self.name, self.type, self.version, self.dev))
+
+    def __eq__(self, o: object):
+        if type(object) != self.__class__:
+            return False
+
+        return self.__hash__() == o.__hash__()
```
chmeliik, line 289: While `pkg1 == pkg2` implies `hash(pkg1) == hash(pkg2)`, the reverse is not true. The hash should not be used for the eq implementation

chmeliik: Also, objects implementing `__hash__` and `__eq__` should be, as far as Python allows, immutable. A fairly elegant way to achieve that would be with [@dataclass(frozen=True)](https://docs.python.org/3/library/dataclasses.html) (which would require a few more changes such as removing `__init__`), or just by making all the attribute names private with a leading underscore

cachito/web/content_manifest.py
```diff
@@ -232,3 +237,216 @@ def generate_icm(self, image_contents=None):
...
+            f"<{self.__class__.__name__} name={self.name}, type={self.type}, "
+            f"version={self.version}>"
+        )
+
+    def __hash__(self):
+        return hash((self.name, self.type, self.version, self.dev))
+
+    def __eq__(self, o: object):
+        if type(object) != self.__class__:
+            return False
```
chmeliik, line 287:
```suggestion
        if not isinstance(o, Package):
            return False
```
The typo of o -> object would also make this always `False`, it may be good to have a unit test for package equality

cachito/web/content_manifest.py
brunoapimentel: /me likes @dataclass :slightly_smiling_face:

cachito/web/content_manifest.py
```diff
@@ -31,13 +37,14 @@
...
 
-    def __init__(self, request):
+    def __init__(self, request, packages):
         """
         Initialize ContentManifest.
 
         :param Request request: the request to generate a ContentManifest for
         """
         self.request = request
+        self.packages = [Package.from_json(package) for package in packages]
```
tkdchen, line 47: Ideally, `ContentManifest` should not have the responsibility to convert a `package` dict to a `Package` object. It just consumes the `Package` objects to get purl and other content manifest.

It would be good to convert in the `content_manifest` property instead.

tkdchen, line 47: Furthermore, we can make refactor in the `PackagesData` to ensure every loaded package and dependency from the `packages.json` is a Package object. This is out of this PR's scope. Just share my thought.

cachito/web/content_manifest.py
```diff
@@ -31,13 +37,14 @@
 class ContentManifest:
     """A content manifest associated with a Cachito request."""
 
-    def __init__(self, request):
+    def __init__(self, request, packages):
         """
         Initialize ContentManifest.
 
         :param Request request: the request to generate a ContentManifest for
```
tkdchen, line 44: Missing description of argument `packages` in the docstring.

cachito/web/content_manifest.py
```diff
@@ -232,3 +238,205 @@ def generate_icm(self, image_contents=None):
...
+        return (
+            f"<{self.__class__.__name__} name={self.name}, type={self.type}, "
+            f"version={self.version}>"
+        )
+
+    def __hash__(self):
+        return hash((self.name, self.type, self.version, self.dev))
+
+    @classmethod
+    def from_json(cls, package):
```
chmeliik, line 277: I would like a unit test for this

cachito/web/content_manifest.py
```diff
@@ -232,3 +238,205 @@ def generate_icm(self, image_contents=None):
         icm = deepcopy(BASE_ICM)
         icm["image_contents"] = image_contents or []
         return deep_sort_icm(icm)
+
+
+@dataclass(frozen=True)
+class Package:
```
tkdchen, line 244: What do you think to have separate `Package` and `Dependency` classes? IMO, this separation could give a clear model design of the packages and dependencies used in Cachito.

chmeliik (commented):
Looking good, just a few minor things and some unit test changes left

tests/test_api_v1.py
```diff
@@ -2141,7 +2150,33 @@ def test_validate_package_manager_exclusivity(
...
+    def get_packages_data():
+        nonlocal next_index
+
+        package = packages_lists[next_index]
+        package_json = {"name": package.name, "version": package.version, "type": package.type}
+
+        packages_data = PackagesData()
+        packages_data._packages = [package_json]
+
+        next_index += next_index
```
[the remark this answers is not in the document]
chmeliik: Oh, I think I know why. All npm packages (not dependencies) will always have the same purl unless they have different subpaths. Could you change the packages in that test? Giving each a different subpath should be enough.

cachito/web/content_manifest.py
chmeliik: +1 for that, but it should not be the `content_manifest.Package` class, that one should be for content manifests only. We could have a separate `packages_data.Package` class and then the content_manifest one would just be a wrapper around it that adds purl methods.

tkdchen: How about make a general `Package` class? And rename the `content_manifest.Package` to another name to reflect the content manifest generation directly? :)

chmeliik: Yeah, in my mind the `packages_data.Package` would be the general one. What would you suggest for the new name, `ContentManifestPackage` as it was called before? Or something else?

tkdchen: Hopefully, this is helpful to express my idea.

cachito-content-manifests

tests/test_api_v1.py
```diff
@@ -1924,7 +1932,8 @@ def test_get_environment_variables(app, client, db, worker_auth_env, env_vars):
 
     env_vars_expected = env_vars or {}
 
-    EnvironmentVariable.query.count() == len(env_vars_expected)
+    assert EnvironmentVariable.query.count() == len(env_vars_expected)
```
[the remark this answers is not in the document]
brunoapimentel: :sunglasses:

cachito/web/content_manifest.py
```diff
@@ -232,3 +238,205 @@ def generate_icm(self, image_contents=None):
...
+    It is used primarily to generate a package URL (purl).
+    """
+
+    name: str
+    type: str
+    version: str
+    dev: bool = False
+    dependencies: list = field(default_factory=list)
+    path: str = None
+    replaces: str = None
```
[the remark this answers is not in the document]
brunoapimentel: I decided to take away the replaces value. :slightly_smiling_face: 

About importing the `__future__` module, although I think this is a great feature, I saw that it would be the first import to this module on the whole project, so I thought it would be best to leave it aside. If it's something we consider it's worth doing, I'm completely ok to start doing it now, anyways, so let me know.

cachito/web/content_manifest.py
brunoapimentel: Done it, take a look :smile:

cachito/web/content_manifest.py
```diff
@@ -232,3 +238,205 @@ def generate_icm(self, image_contents=None):
...
+        """
+        # Avoids changing the original JSON dependencies
+        package = package.copy()
+
+        if "dependencies" in package:
+            package["dependencies"] = [
+                Package.from_json(dependency) for dependency in package["dependencies"]
+            ]
+
+        return cls(**package)
```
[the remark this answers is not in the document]
brunoapimentel: Since we took away the `replaces` attribute, I had to rewrite the whole thing so we wouldn't have errors on the __init__ when this key was actually available in the json package.

I did avoid copying the dict, anyway :+1:

cachito/web/content_manifest.py
```diff
@@ -232,3 +238,205 @@ def generate_icm(self, image_contents=None):
...
+    """
+
+    name: str
+    type: str
+    version: str
+    dev: bool = False
+    dependencies: list = field(default_factory=list)
+    path: str = None
+    replaces: str = None
+
```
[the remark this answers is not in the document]
brunoapimentel: Done!

tests/test_api_v1.py
```diff
@@ -2172,18 +2207,13 @@ def test_get_content_manifests_by_requests(app, client, db, auth_env):
...
-    db.session.add(pkg1)
-    db.session.add(pkg2)
-    db.session.add(pkg3)
 
-    request1.add_package(pkg1)
-    request2.add_package(pkg2)
-    request3.add_package(pkg3)
-
-    db.session.commit()
+    mock_get_packages_data.side_effect = _mock_packages_json_loading([pkg1, pkg2, pkg3, pkg3])
```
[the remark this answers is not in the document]
brunoapimentel: You're right, I think it's less confusing now.

I tried to do a little refactor on this test while at it.

tests/test_api_v1.py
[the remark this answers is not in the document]
brunoapimentel: Shit, so many typos ><

cachito/web/content_manifest.py
brunoapimentel: I did just that, take a look now :wink:

cachito/web/content_manifest.py
brunoapimentel: FTFY :slightly_smiling_face:

cachito/web/content_manifest.py
brunoapimentel: I totally agree, I believe we can follow the design you posted on a previous comment.

cachito/web/content_manifest.py
```diff
@@ -232,3 +238,208 @@ def generate_icm(self, image_contents=None):
...
+        :rtype: Package
+        """
+        dependencies = [
+            Package.from_json(dependency) for dependency in package.get("dependencies", [])
+        ]
+
+        return cls(
+            name=package.get("name"),
+            type=package.get("type"),
+            version=package.get("version"),
```
tkdchen, line 294: I think the `.get` should not be used here. The `name`, `type` and `version` are required for both package and dependency. When use `dict.get`, it makes me feel the `name`, `type` and `version` are optional.

cachito/web/content_manifest.py
```diff
@@ -232,3 +238,208 @@ def generate_icm(self, image_contents=None):
...
+            dev=package.get("dev", False),
+            dependencies=dependencies,
+            path=package.get("path"),
```
tkdchen, line 297: In `packages.json`, the `path` is not present if it is `os.curdir`. For a loaded object from the file, I think we can assign `os.curdir` to the `path` rather than `None`.

cachito/web/content_manifest.py
brunoapimentel: Done!

cachito/web/content_manifest.py
brunoapimentel: I don't know if I got it right, but I tried assigning instead of none and the purls started to be generated with a trailing `.`, like this:

`pkg:golang/example.com%2Forg%2Fproject@1.1.1#.`

(which broke a bunch of tests :laughing: )

cachito/web/content_manifest.py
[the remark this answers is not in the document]
chmeliik: Good point, let's not do it. Come python 3.10, it will be available without the `__future__` import anyway

cachito/web/content_manifest.py
chmeliik: The code expects that packages without a subpath will have `path = None` rather than `path = "."`, so  using None was correct

tests/test_api_v1.py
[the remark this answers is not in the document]
chmeliik: Hmm, wait a second, if the requests for those packages have different repo urls, then the purls should *not* be the same, even for npm packages :thinking: 

So... I still don't understand why the test worked. Anyway, probably not worth worrying about it now :D

cachito/web/content_manifest.py
```diff
@@ -232,3 +238,208 @@ def generate_icm(self, image_contents=None):
...
+            f"<{self.__class__.__name__} name={self.name}, type={self.type}, "
+            f"version={self.version}>"
+        )
+
+    def __hash__(self):
+        return hash((self.name, self.type, self.version, self.dev))
+
+    def __eq__(self, other):
+        return (
+            type(other) == Package
```
chmeliik, line 269: As a minor improvement, this could be `isinstance(other, Package)` or `type(other) == type(self)`. The first if the parent class should be able to compare with subclasses, the second if not. It looks like dataclasses use the second option by default.

The current code has a minor issue in that subclasses could never compare equal, even to the same subclass.
```python
class X(Package):
    pass

X("a", "b", "c") == X("a", "b", "c")  # False
```

chmeliik (approved):
LGTM, just one minor comment. Unless we're renaming the `Package` class in this PR, I wasn't sure if that is to be done now or later.

tkdchen:
:+1: LGTM

chmeliik (approved):
LGTM :+1:

PR #503, opened 2021-06-29
Stop storing packages and dependencies into database

tkdchen:
CLOUDBLD-4731

Since the packages and dependencies are stored into the packages.json,
the API endpoint /requests/id is updated to not store packages and
dependencies into the database.

The existing data and the package and dependency table are not deleted
from database in this patch. Once we consider the data is not used
anymore after a period of time, then they can be removed.


tkdchen:
It would be good to merge #502 firstly, which should address the following tests.

```
FAILED tests/test_api_v1.py::test_fetch_request_content_manifest_go[complete] - AssertionError: assert {'image_content...
FAILED tests/test_api_v1.py::test_fetch_request_content_manifest_go[stale] - AssertionError: assert {'image_contents':...
FAILED tests/test_api_v1.py::test_fetch_request_content_manifest_npm_or_pip[complete-npm-npm] - AssertionError: assert...
FAILED tests/test_api_v1.py::test_fetch_request_content_manifest_npm_or_pip[complete-pip-pypi] - AssertionError: asser...
FAILED tests/test_api_v1.py::test_fetch_request_content_manifest_npm_or_pip[stale-npm-npm] - AssertionError: assert {'...
FAILED tests/test_api_v1.py::test_fetch_request_content_manifest_npm_or_pip[stale-pip-pypi] - AssertionError: assert {...
```

cachito/web/migrations/env.py
```diff
@@ -79,13 +81,22 @@ def process_revision_directives(context, revision, directives):
...
             **current_app.extensions["migrate"].configure_args,
         )
 
         with context.begin_transaction():
             context.run_migrations()
 
 
+# Refer to: https://alembic.sqlalchemy.org/en/latest/cookbook.html#don-t-generate-any-drop-table-directives-with-autogenerate
+def include_object(object, name, type_, reflected, compare_to):
+    if type_ == "table" and reflected and compare_to is None:
```
chmeliik, line 94: Would it make sense to explicitly ignore only the 3 affected tables - `"package", "request_package", "request_dependency"`?

cachito/workers/pkg_managers/general.py
```diff
@@ -68,96 +67,39 @@ def update_request_with_config_files(request_id, config_files):
...
-
+    payload = {"environment_variables": env_vars}
     try:
         rv = requests_auth_session.patch(
             request_url, json=payload, timeout=config.cachito_api_timeout
         )
     except requests.RequestException:
-        msg = f"The connection failed when adding a package to the request {request_id}"
+        msg = (
+            f"The connection failed when updating environment variables to the request {request_id}"
```
chmeliik, line 90:
```suggestion
            f"The connection failed when updating environment variables on the request {request_id}"
```

cachito/workers/pkg_managers/general.py
```diff
@@ -68,96 +67,39 @@ def update_request_with_config_files(request_id, config_files):
...
+        )
         log.exception(msg)
         raise CachitoError(msg)
-
     if not rv.ok:
         log.error(
-            "The worker failed to add a package to the request %d. The status was %d. "
-            "The text was:\n%s",
+            "The worker failed to update environment variables to the request %d. "
```
chmeliik, line 96:
```suggestion
            "The worker failed to update environment variables on the request %d. "
```

cachito/workers/pkg_managers/general.py
```diff
@@ -68,96 +67,39 @@ def update_request_with_config_files(request_id, config_files):
...
+            "The status was %d. The text was:\n%s",
             request_id,
             rv.status_code,
             rv.text,
         )
-        raise CachitoError(f"Setting a package on request {request_id} failed")
+        raise CachitoError(f"Update environment variables on request {request_id} failed")
```
chmeliik, line 102:
```suggestion
        raise CachitoError(f"Updating environment variables on request {request_id} failed")
```

chmeliik (commented):
1000 lines removed, this is my new favorite PR :rocket:  :smile: 

I like it, left a couple minor suggestions and a more important one (the one about `dependency_replacements`)

cachito/web/models.py
```diff
@@ -963,9 +487,6 @@ def from_json(cls, kwargs):
         if not isinstance(dependency_replacements, list):
             raise ValidationError('"dependency_replacements" must be an array')
 
-        for dependency_replacement in dependency_replacements:
-            Dependency.validate_replacement_json(dependency_replacement)
```
chmeliik, line 967: I would like to keep all the validation if possible. Since the `Dependency` class is being removed, could you just extract the `validate_replacement_json` method to a separate function?

cachito/web/validation.py
```diff
@@ -0,0 +1,48 @@
...
+        if key not in replacement and key in optional:
+            continue
+
+        if not isinstance(replacement[key], str):
+            raise ValidationError(
+                'The "{}" key of the dependency replacement must be a string'.format(key)
+            )
+
+
+def validate_dependency_replacements(replacements: List[Dict[str, Any]]) -> None:
```
chmeliik, line 37: Do we have a test for this?

tkdchen: `test_create_request_invalid_dependency_replacement` is restored.

tkdchen: And other existing test like `test_create_and_fetch_request` covers the valid replacements.

chmeliik: Gotcha, seems reasonable :+1:

chmeliik (approved):
LGTM :rocket: 

Once we implement the `Package` and `Dependency` classes as @tkdchen suggested, we should consider validating the JSON data that is passed to `PackagesData.add_package()`. Especially for dependencies, those currently miss some validation. But I don't think we need to do that right away.

cachito/workers/tasks/pip.py
```diff
@@ -118,18 +117,15 @@ def fetch_pip_source(request_id, package_configs=None):
     worker_config = get_worker_config()
     env_vars.update(worker_config.cachito_default_environment_variables.get("pip", {}))
 
-    packages_json_data = PackagesData()
+    update_request_env_vars(request_id, env_vars)
 
+    packages_json_data = PackagesData()
     # Finally, perform DB operations
```
brunoapimentel, line 123: Isn't this comment outdated?

brunoapimentel (approved):

brunoapimentel:
LGTM :+1:

cachito/workers/tasks/pip.py
tkdchen: Good catch. Fixed in a new commit. Thanks.

PR #508, opened 2021-07-07
Provide bundle archive checksum

tkdchen:


cachito/web/api_v1.py
```diff
@@ -211,15 +212,25 @@ def download_archive(request_id):
         )
         raise InternalServerError()
 
+    checksum = compute_file_checksum(get_hasher(), str(bundle_dir.bundle_archive_file))
+    store_checksum = bundle_dir.bundle_archive_checksum.read_text(encoding="utf-8")
+    if checksum != store_checksum:
+        flask.current_app.logger.error(
+            "Checksum of bundle archive %s has been changed.", bundle_dir.bundle_archive_file
+        )
+        raise InternalServerError()
```
chmeliik, line 221: Could we  have an error message here?

cachito/workers/tasks/general.py
```diff
@@ -154,12 +156,29 @@ def aggregate_packages_data(request_id: int, pkg_managers: List[str]) -> Package
     return aggregated_data
 
 
+def save_bundle_archive_checksum(request_id: int) -> None:
+    """Compute and store bundle archive's checksum.
+
+    :param int request_id: the request id.
+    """
+    set_request_state(request_id, "in_progress", "Compute and store bundle archive's checksum.")
```
chmeliik, line 164: I don't think we need a state for this

cachito/web/api_v1.py
```diff
@@ -211,15 +212,25 @@ def download_archive(request_id):
...
         "Sending the bundle at %s for request %d", bundle_dir.bundle_archive_file, request_id
     )
-    return flask.send_file(
+    resp = flask.send_file(
         str(bundle_dir.bundle_archive_file),
         mimetype="application/gzip",
         as_attachment=True,
         attachment_filename=f"cachito-{request_id}.tar.gz",
     )
+    resp.headers["checksum_sha256"] = checksum
```
chmeliik, line 232: Should we use a standardized [Digest](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Digest) header?

cachito/web/api_v1.py
```diff
@@ -211,15 +212,24 @@ def download_archive(request_id):
...
         "Sending the bundle at %s for request %d", bundle_dir.bundle_archive_file, request_id
     )
-    return flask.send_file(
+    resp = flask.send_file(
         str(bundle_dir.bundle_archive_file),
         mimetype="application/gzip",
         as_attachment=True,
         attachment_filename=f"cachito-{request_id}.tar.gz",
     )
+    resp.headers["Digest"] = f"sha-256={checksum}"
```
chmeliik, line 231: According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Digest#directives, the checksum should be base64 encoded

cachito/web/api_v1.py
```diff
@@ -211,15 +212,24 @@ def download_archive(request_id):
         )
         raise InternalServerError()
 
+    checksum = compute_file_checksum(get_hasher(), str(bundle_dir.bundle_archive_file))
+    store_checksum = bundle_dir.bundle_archive_checksum.read_text(encoding="utf-8")
+    if checksum != store_checksum:
+        msg = f"Checksum of bundle archive {bundle_dir.bundle_archive_file} has been changed."
+        flask.current_app.logger.error(msg)
+        raise InternalServerError(msg)
```
chmeliik, line 220: IMO the error message should not show the full file path, maybe just the file name or not at all (but it would be good to show the full file path in the log message).

And I would suggest to change the wording from `has been changed` to `has changed` - the first one kind of sounds like the checksum was changed intentionally

cachito/web/api_v1.py
```diff
@@ -211,15 +213,25 @@ def download_archive(request_id):
...
         "Sending the bundle at %s for request %d", bundle_dir.bundle_archive_file, request_id
     )
-    return flask.send_file(
+    resp = flask.send_file(
         str(bundle_dir.bundle_archive_file),
         mimetype="application/gzip",
         as_attachment=True,
         attachment_filename=f"cachito-{request_id}.tar.gz",
     )
+    resp.headers["Digest"] = f"sha-256={base64.b64encode(hasher.digest())}"
```
chmeliik, line 233: Minor problem, when I look at the headers locally with `curl -v localhost:8080/api/v1/requests/1/download`, I see this:
```
Digest: sha-256=b'UKVLAQIq0YRm4gY3VSV1avgwTZoAUvmeTbqDuAxnr24='
```
I think we want to `.decode()` the bytes, not just convert to str.

Optional - do you think it would make sense to have a function in the `checksum` module for converting digests to the base64 encoded value?

tkdchen: > Optional - do you think it would make sense to have a function in the checksum module for converting digests to the base64 encoded value?

Absolutely.

tkdchen: @chmeliik Both are fixed. PTAL. Thank you very much for catching this bug.

cachito/common/checksum.py
```diff
@@ -1,34 +1,48 @@
...
+        """Proxy of the hashlib hash object's digest method."""
+        return self._hasher.digest()
+
+    def hexdigest(self) -> str:
+        """Proxy of the hashlib hash object's hexdigest method."""
+        return self._hasher.hexdigest()
+
+    def base64digest(self) -> str:
+        """Return the digest encoded in base64."""
+        return base64.b64encode(self.digest()).decode("utf-8")
```
chmeliik, line 28: Nice! :rocket: 

Could we have some unit tests for it?

cachito/common/checksum.py
```diff
@@ -1,34 +1,48 @@
...
+
 
-    :param str algorithm: the algorithm name used to get a hasher from the hashlib module.
-    :return: the hasher got from the hashlib module.
+def hash_file(file_path: Union[str, Path], chunk_size: int = 10240, algorithm: str = "sha256"):
```
chmeliik, line 31:
```suggestion
def hash_file(file_path: Union[str, Path], chunk_size: int = 10240, algorithm: str = "sha256") -> Hasher:
```

tests/test_checksum.py
```diff
@@ -1,34 +1,46 @@
...
+def test_hasher_hexdigest():
+    h = hashlib.new("sha256")
+    hasher = Hasher(h)
+    assert h.hexdigest() == hasher.hexdigest()
+
+
+def test_hasher_base64digest():
+    h = hashlib.new("sha256")
+    hasher = Hasher(h)
+    assert base64.b64encode(h.digest()).decode("utf-8") == hasher.base64digest()
```
chmeliik, line 46: I'm not a big fan of tests that are essentially just a copy-paste of the implementation, maybe the expected value could be hardcoded instead?

chmeliik (commented):
LGTM, just one note on unit tests.

And as you pointed out, during deployment, we will need to generate checksums for all archives that are already present (just posting this here so that everyone is aware)

cachito/common/checksum.py
```diff
@@ -1,34 +1,51 @@
...
+
+    def base64digest(self) -> str:
+        """Return the digest encoded in base64."""
+        return base64.b64encode(self.digest()).decode("utf-8")
+
 
-    :param str algorithm: the algorithm name used to get a hasher from the hashlib module.
-    :return: the hasher got from the hashlib module.
+def hash_file(
+    file_path: Union[str, Path], chunk_size: int = 10240, algorithm: str = "sha256"
```
MartinBasti, line 32: Why `10240` ?

cachito/web/api_v1.py
```diff
@@ -211,15 +212,25 @@ def download_archive(request_id):
...
         "Sending the bundle at %s for request %d", bundle_dir.bundle_archive_file, request_id
     )
-    return flask.send_file(
+    resp = flask.send_file(
         str(bundle_dir.bundle_archive_file),
         mimetype="application/gzip",
         as_attachment=True,
         attachment_filename=f"cachito-{request_id}.tar.gz",
     )
+    resp.headers["Digest"] = f"sha-256={hasher.base64digest()}"
```
MartinBasti, line 232: Would be safer to provide `store_checksum` here instead of dynamically computed hash?

cachito/web/api_v1.py
```diff
@@ -445,6 +456,7 @@ def patch_request(request_id):
         )
         try:
             bundle_dir.bundle_archive_file.unlink()
```
MartinBasti, line 458: Should be this block more error resilient and try to remove everything instead and reporting summary error instead of exception from first failure and skipping the rest of cleanup?

cachito/web/api_v1.py
tkdchen: Not sure what's the point? :) Why would it be safer?

cachito/common/checksum.py
tkdchen: Not sure. It was there originally. By goolging it quickly, seems the consideration behind this number is to have a good performance of reading files from the file system. But, I may be wrong to understand. :)

cachito/web/api_v1.py
tkdchen: What would be the benefit of the behavior "try to remove everything and report an error summary"?

cachito/web/api_v1.py
MartinBasti: if we change implementation in comparison and we wrongly compare values, passing original value will cause client to notice.

I prefer to send the static value directly not the computed one.

cachito/common/checksum.py
MartinBasti: ok, I was just curious

cachito/web/api_v1.py
MartinBasti: if my understanding is correct, this leaves some garbage behind in case unlink of bundle_archive_file failed

cachito/web/api_v1.py
tkdchen: Does "static value" mean the stored checksum?

MartinBasti: yes

cachito/web/api_v1.py
tkdchen: Make sense. I'll implement this in a separate patch.

cachito/web/api_v1.py
MartinBasti: resolved f2f. We don't need this

MartinBasti (approved):

PR #509, opened 2021-07-08
Document more details of gomod vendoring

chmeliik:
Call out that Cachito will not provide the module cache when vendoring
is used. Explain what that means for `go mod` commands.


chmeliik:
PTAL @arewm

arewm (approved):
LGTM, thanks!

tkdchen (approved):

PR #510, opened 2021-07-08
Add PATCH and POST to allowed HTTP methods for retrying requests

serg-cymbaluk:
Separete session instances for different cases:
  - keep requests_auth_session and requests_session for internal Cachito requests
  - cachito.workers.nexus.nexus_requests_session - for Nexus requests
  - cachito.workers.pkg_managers.general.pkg_requests_session - for package managers

Add 'retry_options' parameter in get_requests_session to allow customize retrying for each case.
Specify extended set of allowed methods for retrying internal Cachito requests.
Specify limited set of allowed methonds (only safe one) for retrying requests to external services.

### DIFF
--- a/cachito/workers/requests.py
+++ b/cachito/workers/requests.py
@@ -9,12 +9,28 @@ from cachito.workers.config import get_worker_config
 
 log = logging.getLogger(__name__)
 
+# The set is extended version of constant Retry.DEFAULT_ALLOWED_METHODS
+# with PATCH and POST methods included.
+ALL_REQUEST_METHODS = frozenset(
+    {"GET", "POST", "PATCH", "PUT", "DELETE", "HEAD", "OPTIONS", "TRACE"}
+)
+# The set includes only methods which don't modify state of the service.
+SAFE_REQUEST_METHODS = frozenset({"GET", "HEAD", "OPTIONS", "TRACE"})
+DEFAULT_RETRY_OPTIONS = {
+    "total": 5,
+    "read": 5,
+    "connect": 5,
+    "backoff_factor": 1.3,
+    "status_forcelist": (500, 502, 503, 504),
+}
 
-def get_requests_session(auth=False):
+
+def get_requests_session(auth=False, retry_options={}):
     """
     Create a requests session with authentication (when enabled).
 
     :param bool auth: configure authentication on the session
+    :param dict retry_options: overwrite options for initialization of Retry instance
     :return: the configured requests session
     :rtype: requests.Session
     """
@@ -28,14 +44,15 @@ def get_requests_session(auth=False):
         elif config.cachito_auth_type == "cert":
             session.cert = config.cachito_auth_cert
 
-    retry = Retry(
-        total=5, read=5, connect=5, backoff_factor=1.3, status_forcelist=(500, 502, 503, 504)
-    )
-    adapter = requests.adapters.HTTPAdapter(max_retries=retry)
+    retry_options = {**DEFAULT_RETRY_OPTIONS, **retry_options}
+    adapter = requests.adapters.HTTPAdapter(max_retries=Retry(**retry_options))
     session.mount("http://", adapter)
     session.mount("https://", adapter)
     return session
 
 
-requests_auth_session = get_requests_session(auth=True)
-requests_session = get_requests_session()
+# These sessions are only for connecting to the internal Cachito API
+requests_auth_session = get_requests_session(
+    auth=True, retry_options={"allowed_methods": ALL_REQUEST_METHODS}
+)
+requests_session = get_requests_session(retry_options={"allowed_methods": ALL_REQUEST_METHODS})
--- a/cachito/workers/config.py
+++ b/cachito/workers/config.py
@@ -4,12 +4,14 @@ import tempfile
 import logging
 from typing import Dict, List, Optional
 
+import celery
 import kombu
 
 from cachito.errors import ConfigError
 
-
 ARCHIVES_VOLUME = os.path.join(tempfile.gettempdir(), "cachito-archives")
+
+app = celery.Celery()
 
 
 class Config(object):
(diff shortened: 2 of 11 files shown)


serg-cymbaluk:
@chmeliik 

Could you please take a quick look at the PR?

chmeliik:
Linters are complaining, please run `tox -e black,flake8` locally (or just check the Travis results) to see the issues

serg-cymbaluk:
@chmeliik 

The formatting issue is fixed. All checks are green.

cachito/workers/requests.py
```diff
@@ -9,6 +9,10 @@
 
 log = logging.getLogger(__name__)
 
+RETRY_ALLOWED_METHODS = frozenset(
+    {"GET", "POST", "PATCH", "PUT", "DELETE", "HEAD", "OPTIONS", "TRACE"}
```
chmeliik, line 13: Could you add a comment to explain that this is all the default methods + PATCH and POST? Ideally with a link to the docs where the default methods are mentioned

cachito/workers/requests.py
```diff
@@ -29,7 +33,12 @@ def get_requests_session(auth=False):
             session.cert = config.cachito_auth_cert
 
     retry = Retry(
-        total=5, read=5, connect=5, backoff_factor=1.3, status_forcelist=(500, 502, 503, 504)
+        total=5,
+        read=5,
+        connect=5,
+        backoff_factor=1.3,
+        status_forcelist=(500, 502, 503, 504),
```
chmeliik, line 40: You said you wanted to remove 500 from the status list, do you still think that should be done?

serg-cymbaluk: I don't sure for now. Because I don't have statistics about how often the 500 status is caused by internal errors and how often it is only an external outage (so it makes sense to repeat the request).

chmeliik: That's a good point, let's keep it for now

cachito/workers/requests.py
serg-cymbaluk: Updated.

chmeliik:
Reading through the description of the issue, it occurred to me that we only want to allow PATCH and POST methods for requests made from the Cachito workers to the Cachito API. The `requests_session` is used for requests to external services as well, we probably don't want those to retry PATCH and POST for those.

I think the `requests_auth_session` is used only for requests to Cachito. I also believe all the POST and PATCH endpoints require authorization. Maybe allowing the methods only for that session would work? Or we could have new request sessions for communicating with the API.

serg-cymbaluk:
> Reading through the description of the issue, it occurred to me that we only want to allow PATCH and POST methods for requests made from the Cachito workers to the Cachito API. The `requests_session` is used for requests to external services as well, we probably don't want to retry PATCH and POST for those.
> 
> I think the `requests_auth_session` is used only for requests to Cachito. I also believe all the POST and PATCH endpoints require authorization. Maybe allowing the methods only for that session would work? Or we could have new request sessions for communicating with the API.

Good point.
Let me explore all those cases.
Marking the PR as WIP...

serg-cymbaluk:
Fixed tests.

Replaced mocks like `@mock.patch("cachito.workers.requests.requests_auth_session")` with `@mock.patch.object(requests_auth_session, "patch")`.
It allows to patch objects which are already loaded, when first notation should be used before the session variable is loaded somewhere.

cachito/workers/requests.py
```diff
@@ -28,14 +44,15 @@ def get_requests_session(auth=False):
...
+    adapter = requests.adapters.HTTPAdapter(max_retries=Retry(**retry_options))
     session.mount("http://", adapter)
     session.mount("https://", adapter)
     return session
 
 
-requests_auth_session = get_requests_session(auth=True)
-requests_session = get_requests_session()
+# These sessions are only for connecting to the internal Cachito API
+requests_auth_session = get_requests_session(
```
brunoapimentel, line 55: Just a small nitpick: I saw that we have these two sessions being defined here, and we also have the `pkg_requests_session` and the `nexus_requests_session` being defined in different files.

I saw that both `pkg_requests_session` and the `nexus_requests_session` are identical now (they retry only safe methods), and that although the `requests_session` is defined to retry all methods, it is only being used on a single GET method.

So, I was thinking we could have them all on only two sessions defined on this file:

`requests_auth_session` - is authenticated and allows the retry of all methods
`requests_session` - allows the retry of safe methods (replaces `nexus_requests_session` and `pkg_requests_session` )

Anyway, it's just a suggestion, let me know what you think.

brunoapimentel (approved):

brunoapimentel:
LGTM :rocket: 

Also, great job on getting rid of the cyclic imports, the code looks cleaner now :100:

tkdchen (approved):
LGTM

cachito/workers/requests.py
serg-cymbaluk: I would distinguish sessions not based on how they are configured currently, but take to account a potential changes of each case.

It is better to instantiate session as close to a corresponding use case as possible, to not broke open-closed principle and keep possibility to tune everything.

serg-cymbaluk: I didn't want to go further with that refactoring, because here are the following dependencies:

- `fetch_npm_source depends` calls `set_request_state` and `update_request_with_config_files`
- both `set_request_state` and `update_request_with_config_files` use `requests_auth_session`
- so there is one case for the auth session, which involves different modules
   * `set_request_state` is part of `cachito.workers.tasks.utils`
   * `update_request_with_config_files` is part of `cachito.workers.pkg_managers.general`
   * common package is `cachito.workers`
So until we bring some other mechanism for session instantiation, we need to keep the session instance somewhere in common modules (the `workers.requests` is proper one from that perspective).

Only one thing I can change here it is the name of that session. It could be called `cachito_requests_auth_session` or just `cachito_auth_session`.

serg-cymbaluk: @brunoapimentel 

So, do I need to change something in the PR or we can postpone that?

brunoapimentel: We can definitely postpone it, I saw it as a minor nitpick, just wanted to leave the suggestion here.

serg-cymbaluk:
@brunoapimentel @tkdchen 

Thanks for your review!

PR #511, opened 2021-07-09
Document metadata formats

chmeliik:
CLOUDBLD-6022
    
Document the request JSON format, the Content Manifest format and the
differences between the two.

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -1,0 +1,590 @@
+# Cachito request metadata
+
+* [Request JSON](#request-json)
+* [Content Manifest](#content-manifest)
+* [Request JSON vs. Content Manifest](#request-json-vs-content-manifest)
+
+One of Cachito's secondary (but no less important) responsibilities is reporting all
+the sources that were provided for a specific request. Cachito provides two different
+metadata formats for this purpose.
+
+The first format is the JSON response at `/api/v1/requests/{id}`, which contains all
+the information about a request. In the rest of the document, this will be referred to
+as [request JSON](#request-json).
+
+The second format is the [Content Manifest](#content-manifest) at
+`/api/v1/requests/{id}/content-manifest`. This is a JSON document as well, but contains
+only information about the packages and dependencies present in the request bundle.
+
+All the sources are also present in the request bundle downloaded from
+`/api/v1/requests/{id}/download`. The packages from the requested git repository are
+under `app/`. All the dependencies downloaded by Cachito are under `deps/`. If vendoring
+is used, the dependencies will not be under `deps/` but instead somewhere under `app/`,
+depending on the package manager. For example, vendored `gomod` dependencies will be
+under `app/vendor/`, or `app/{subpath}/vendor` if the Go module is not in the repository
+root. Vendoring is optional (enabled by a [flag](../README.md#flags)) and not all
+package managers support it.
+
+## Request JSON
+
+The request JSON object has a lot of attributes. When it comes to reporting sources,
+the relevant ones are `repo`, `ref`, `packages` and `dependencies`.
+
+<details>
+<summary>Example</summary>
+
+```json
+{
+  "repo": "https://github.com/cachito-testing/cachito-yarn-lorem-ipsum",
+  "ref": "b470410e50caff0447bc18ca4f011663681e7e17",
+  "packages": [
+    {
+      "dependencies": [
+        {
+          "dev": true,
+          "name": "yarn",
+          "replaces": null,
+          "type": "npm",
+          "version": "1.22.10"
+        }
+      ],
+      "name": "bootstrap-yarn",
+      "path": "bootstrap-yarn",
+      "type": "npm",
+      "version": "1.0.0"
+    },
+    {
+      "dependencies": [
+        {
+          "name": "axios",
+          "replaces": null,
+          "type": "yarn",
+          "version": "0.21.1"
+        },
+        {
+          "name": "follow-redirects",
+          "replaces": null,
+          "type": "yarn",
+          "version": "1.14.0"
+        }
+      ],
+      "name": "lorem-ipsum",
+      "type": "yarn",
+      "version": "1.0.0"
+    }
+  ],
+  "dependencies": [
+    {
+      "dev": true,
+      "name": "yarn",
+      "replaces": null,
+      "type": "npm",
+      "version": "1.22.10"
+    },
+    {
+      "name": "axios",
+      "replaces": null,
+      "type": "yarn",
+      "version": "0.21.1"
+    },
+    {
+      "name": "follow-redirects",
+      "replaces": null,
+      "type": "yarn",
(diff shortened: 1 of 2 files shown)


chmeliik:
Should be more or less ready. I apologize in advance to anyone who is not already familiar with Cachito's metadata for having to go through this.

@lcarva @MartinBasti you may find this interesting.

chmeliik:
Rendered doc at https://github.com/chmeliik/cachito/blob/document-metadata/docs/metadata.md

docs/metadata.md
```diff
@@ -0,0 +1,582 @@
...
+pkg:golang/github.com%2Frelease-engineering%2Fretrodep%2Fv2@v2.1.1
+```
+</details>
+
+One unfortunate downside of this approach is that if your repository is a fork intended
+for use with the [replace][gomod-replace] directive, the purl will be incorrect. Take
+for example the [openshift/kubernetes-apiserver][openshift-apiserver] repository. The
+name in `go.mod` is `k8s.io/apiserver`, which is the name of the package that this fork
+is meant to replace. That is expected as per the `replace` rules, but it also causes the
+purl to incorrectly reference the original rather than the fork.
```
chmeliik, line 440: We may also have the opposite problem when somebody uses [replace](https://golang.org/ref/mod#go-mod-file-replace) for a dependency, e.g.
```
replace k8s.io/apiserver => github.com/openshift/kubernetes-apiserver
```
IIRC, Cachito takes the replacement name, so the purl ends up being `pkg:golang/github.com/openshift/kubernetes-apiserver`. Now, the purl semantics were never super clear to me, but one could make a good argument that such a purl is wrong, because `github.com/openshift/kubernetes-apiserver` is not a module that exists anywhere. The module is `k8s.io/apiserver`, the openshift fork is merely an alternative location for that module.

`go get` would seem to support that argument:
```
$ go get github.com/openshift/kubernetes-apiserver
go: github.com/openshift/kubernetes-apiserver upgrade => v0.0.0-20190820102804-33bf53f7c421
go get: github.com/openshift/kubernetes-apiserver@v0.0.0-20190820102804-33bf53f7c421: parsing go.mod:
	module declares its path as: k8s.io/apiserver
	        but was required as: github.com/openshift/kubernetes-apiserver
```
Then again, I don't know what a correct purl would be in this case. Probably a generic, non-golang one.

chmeliik:
Found another bug for go packages, fixed in #548

docs/metadata.md
```diff
@@ -0,0 +1,582 @@
+# Cachito request metadata
```
brunoapimentel, line 1: Wdyt about adding an index for all lvl 2 titles (##) right under the main title?

brunoapimentel:
As far as my knowledge goes, LGTM :+1: 

This was a great read, btw. Will need to read it over at least once more to understand some of the points here. I think this should be a mandatory read for any new Cachito dev.

Is this doc linked from anywhere in the main README? I think this must not get buried or forgotten.

docs/metadata.md
```diff
@@ -0,0 +1,582 @@
...
+
+The second format is the [Content Manifest](#content-manifest) at
+`/api/v1/requests/{id}/content-manifest`. This is a JSON document as well, but contains
+only information about the packages and dependencies present in the request bundle.
+
+All the sources are also present in the request bundle downloaded from
+`/api/v1/requests/{id}/download`. The packages from the requested git repository are
+under `app/`. All the dependencies downloaded by Cachito are under `deps/`, unless
+vendoring was used, in which case they will be in some vendor location specific to the
+given package manager.
```
lcarva, line 19: In the `app/` directory in this case, right?

How about something like this:
> All the dependencies downloaded by Cachito are usually under `deps/`. However, when vendoring is used, the source code for the dependencies will be in the vendor location specific to the given package manager under the `app/` directory. For example, vendored `gomod` dependencies will be under `app/vendor/`, not `deps/`. Vendoring is optional and not all package managers support it.

docs/metadata.md
```diff
@@ -0,0 +1,582 @@
...
+  "replaces": null
+}
+```
+</details>
+
+While the version value may not always be a valid semantic version, it *should* always
+be a valid way to specify the dependency within the rules of the given package manager.
+
+Now comes the time to explain why the part about dependencies never coming from the
+source repo was a lie. Package managers that support local files as dependencies can, in
```
lcarva, line 206: Where is the lie? Two paragraphs above it states that `local files` are allowed.

docs/metadata.md
```diff
@@ -0,0 +1,582 @@
...
+But wait, why is the purl for `k8s.io/client-go` relative to `k8s.io/kubernetes`, not
+to `k8s.io.kubernetes/cmd/kubeadm`? Dependency management is fully up to the module.
+All the dependencies, including local ones, are defined in the go.mod file in the
+module. The paths are likewise relative to the module, not to the package that ends up
+using the dependencies. That is why purls for local dependencies are relative to modules
+and not packages.
+
+## Request JSON vs. Content Manifest
+
+Both formats attempt to represent similar information about a Cachito request. Both do
```
lcarva, line 540: Is it naive to think that if we simply add `purl` attributes to each package/dependency in the `request-json` manifest format, then we have the best of both worlds?

chmeliik: That is a surprisingly simple solution. It might actually work :open_mouth: 

I was under the impression that Content Manifest was the intended winner of the unification efforts, but if that is not the case, then this might just be all we need.

docs/metadata.md
chmeliik: Oh, true. I got lost in my own doc :sweat_smile:

Will rephrase

docs/metadata.md
chmeliik: I like it :+1:

chmeliik: Added

docs/metadata.md
chmeliik: The sentences were getting a bit long, so I reworded it a little. The general idea should be captured there

docs/metadata.md
chmeliik: Rephrased

tkdchen (approved):
:+1:  Learned lots of knowledge from this document. Thanks Adam.

chmeliik:
> As far as my knowledge goes, LGTM +1
> 
> This was a great read, btw. Will need to read it over at least once more to understand some of the points here. I think this should be a mandatory read for any new Cachito dev.
> 
> Is this doc linked from anywhere in the main README? I think this must not get buried or forgotten.

@brunoapimentel great point, I was going to do that based on the discussion in Daniel's PR. Thanks for reminding me :)

Added in second commit

lcarva (approved):
Fantastic job!

nirzari (approved):
Learnt a lot going through this. Thank you.

chmeliik:
Noticed a small typo. Fixed, will merge once CI succeeds.

PR #512, opened 2021-07-09
Migrate to GH workflows

MartinBasti:


.github/workflows/gating.yaml
```diff
@@ -95,3 +95,19 @@ jobs:
...
+        dockerfile:
+          - Dockerfile-api
+          - Dockerfile-flower
+          - Dockerfile-workers
+    steps:
+    - uses: actions/checkout@v2
+    - uses: hadolint/hadolint-action@v1.5.0
+      with:
+        dockerfile: docker/${{ matrix.dockerfile }}
+        ignore: DL3041
```
akhmanova, line 112: I am not sure I know what is it, could you please add a comment?

.github/workflows/gating.yaml
```diff
@@ -0,0 +1,97 @@
...
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        COVERALLS_FLAG_NAME: python-${{ matrix.python-version }}
+        COVERALLS_PARALLEL: true
+      run: |
+        pip3 install --upgrade pip
+        pip3 install --upgrade setuptools
+        pip3 install --upgrade coveralls
+        coveralls --service=github
+
+  coveralls-finish:
```
akhmanova, line 43: We set up the coverage checking with the `.coveragerc` file. Is coveralls better? 

Could you please add a comment about a coverage percentage it uses here Maybe we even will not need the `coveragerc` file anymore ..

.github/workflows/gating.yaml
```diff
@@ -0,0 +1,97 @@
...
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install tox
+    - name: Test with tox
+      run: tox -e flake8
+
+  black:
```
akhmanova, line 80: Do we not need `basepython` with `black` anymore?

tox.ini
```diff
@@ -31,8 +36,9 @@ commands =
     black --check --diff bin --line-length=88
 
 [testenv:flake8]
+skip_install = true
 deps =
-    -rrequirements-test.txt
+    flake8
```
tkdchen, line 41: requirements-test.in includes:

```
flake8==3.9.2
flake8-docstrings==1.6.0
```

.github/workflows/gating.yaml
```diff
@@ -0,0 +1,113 @@
...
+          pip3 install --upgrade coveralls
+          coveralls --finish --service=github
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  flake8:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.8]
```
tkdchen, line 61: Either flake8 or black just needs to run with one Python version. Can this matrix be removed? Or, what requires this matrix setting?

MartinBasti: I can remove matrix, I kept it as matrix in case that we would need run flake/black on multiple versions.
With black it seems not needed at all, and with flake given that there is no py2, probably neither. I'll remove.

.github/workflows/gating.yaml
MartinBasti: coveralls provide results directly into PR and it has nice web interface for checking coverage diffs. Coveralls can measure overall decrease/increase in code coverage.

I'll check coveragerc, but in general coveragerc and coveralls are not mutually exclusive

.github/workflows/gating.yaml
MartinBasti: https://github.com/hadolint/hadolint/wiki/DL3041

.github/workflows/gating.yaml
MartinBasti: I don't follow

.github/workflows/gating.yaml
MartinBasti: I added comment with detailed explanation about what is ignored

.github/workflows/gating.yaml
MartinBasti: I'd keep  `coveragerc` as is

akhmanova (approved):
LGTM, thank you! Let's wait for @tkdchen approval.

tkdchen (approved):

PR #513, opened 2021-07-09
Run integration-tests in GH workflows

MartinBasti:
Requires: https://github.com/release-engineering/cachito/pull/512

### DIFF
--- a/.github/workflows/gating.yaml
+++ b/.github/workflows/gating.yaml
@@ -105,3 +105,29 @@ jobs:
         # * DL3041 - Specify version with dnf install -y <package>-<version>
         ignore: DL3041
         failure-threshold: warning
+
+  integration_tests:
+    runs-on: ubuntu-latest
+    needs:
+    - tests
+    - flake8
+    - black
+    - hadolint
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install required packages
+      run: |
+        sudo apt-get install -y \
+          libkrb5-dev
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install tox
+    - name: Start cachito services
+      run: make run-start UP_OPTS=-d
+    - name: Run integration tests
+      run: tox -e integration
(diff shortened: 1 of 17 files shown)


Makefile
```diff
@@ -44,7 +44,7 @@ run run-start:
 	# - nexus needs to write to ./tmp/nexus-data
 	# - integration tests need to write to (and create) ./tmp/cachito-archives
 	chmod -R 0777 ./tmp
-	$(CACHITO_COMPOSE_ENGINE) up
+	$(CACHITO_COMPOSE_ENGINE) up -d
```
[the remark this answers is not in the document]
MartinBasti: updated

akhmanova (approved):
LGTM! Thank you :)

tests/integration/conftest.py
```diff
@@ -6,7 +6,7 @@
 import pytest
 import yaml
 
-import utils
+from . import utils
```
brunoapimentel, line 9: Just curious, why is importing like this better?

.github/workflows/gating.yaml
```diff
@@ -105,3 +105,29 @@ jobs:
...
         ignore: DL3041
         failure-threshold: warning
+
+  integration_tests:
+    runs-on: ubuntu-latest
+    needs:
+    - tests
+    - flake8
+    - black
+    - hadolint
```
brunoapimentel, line 115: It seems we need to fix our Dockerfiles now :smile:

brunoapimentel (approved):

brunoapimentel:
LGTM :100: 

I'm just wondering if we can run the `cached_dependencies` tests on Github. The default test configuration just skips them.

The only problem I found running them locally is that, since my timezone is UTC-3, Athens gets some inconsistencies comparing the timestamps from the deps it downloads, or something of the like. Setting it to UTC makes all `cached_dependencies` test run fine. This should be a no brainer in the Github env.

Anyway, just putting the idea out there, I don't think this is something to be done in this PR,

akhmanova:
@brunoapimentel 
> I'm just wondering if we can run the cached_dependencies tests on Github. The default test configuration just skips them.
Hm, I think there is another problem: we need to grant writing access to (cachito-testing org)[https://github.com/cachito-testing]

tests/integration/conftest.py
MartinBasti: It didn't work for me without it. python was not able to find module without installing cachito. Why cachito should be installed for int tests?

PR #514, opened 2021-07-12
Update optionalDependencies in package.json

thrbkova:
* CLOUDBLD-3917

### DIFF
--- a/cachito/workers/pkg_managers/npm.py
+++ b/cachito/workers/pkg_managers/npm.py
@@ -229,7 +229,7 @@ def get_package_and_deps(package_json_path, package_lock_path):
 
         package_json_original = copy.deepcopy(package_json)
         for dep_name, dep_version in top_level_replacements:
-            for dep_type in ("dependencies", "devDependencies"):
+            for dep_type in ("dependencies", "devDependencies", "optionalDependencies"):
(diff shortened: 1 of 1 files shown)


brunoapimentel:
I was wondering if it might be a good idea to add a unit test to cover this new scenario (`optionalDependencies`). But then I saw that there is no test covering the `devDependencies` as well. :smile: 

If we do choose to add them, it does not seem to be a trivial task. So maybe we could do a pair programming session to help @thrbkova?

nirzari:
> I was wondering if it might be a good idea to add a unit test to cover this new scenario (`optionalDependencies`). But then I saw that there is no test covering the `devDependencies` as well. 
> 
> If we do choose to add them, it does not seem to be a trivial task. So maybe we could do a pair programming session to help @thrbkova?

+1 for implementing unit test. Ideally even if the unit test doesn't exist for the code line we are changing, we expect a unit test to be added just because we are modifying that.

tkdchen:
IIUC, `pkg_managers.yarn._replace_deps_in_package_json` should also be updated. The issue is expected the changes will work for both npm and yarn package managers.

brunoapimentel:
Closing this PR since I'll be taking over this ticket.

PR #516, opened 2021-07-13
Loads package data only when a request is complete

brunoapimentel:
CLOUDBLD-6452

The purpose of this PR is essentially to avoid looking for the file containing the packages/dependencies data when it is not supposed to be there (any state different from `complete`). This scenario is derived from the work of changing the storage method of this data from the database to JSON files.

The main issue that took us to this implementation is the failure of the integration tests running on OCP, which were somehow impacted by the many calls to the API made in a short period of time.

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -340,11 +340,20 @@ class Request(db.Model):
 
         return content_manifest.ContentManifest(self, packages)
 
-    def _get_packages_data(self):
-        bundle_dir = RequestBundleDir(self.id, root=flask.current_app.config["CACHITO_BUNDLES_DIR"])
+    def _is_complete(self):
+        if self.state:
+            return self.state.state_name == RequestStateMapping.complete.name
 
+        return False
+
+    def _get_packages_data(self):
         packages_data = PackagesData()
-        packages_data.load(bundle_dir.packages_data)
+
+        if self._is_complete():
+            bundle_dir = RequestBundleDir(
+                self.id, root=flask.current_app.config["CACHITO_BUNDLES_DIR"]
+            )
+            packages_data.load(bundle_dir.packages_data)
 
         return packages_data
 
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -2,7 +2,9 @@
 
 from typing import Optional
 import pytest
-from cachito.web.models import PackageManager
+from unittest import mock
+
+from cachito.web.models import PackageManager, Request, RequestStateMapping
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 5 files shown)


cachito/web/models.py
```diff
@@ -340,11 +340,20 @@ def content_manifest(self):
...
+        if self.state:
+            return self.state.state_name == RequestStateMapping.complete.name
+
+        return False
 
+    def _get_packages_data(self):
         packages_data = PackagesData()
-        packages_data.load(bundle_dir.packages_data)
+
+        if self._is_complete():
```
brunoapimentel, line 352: I thought it would be best to leave this check here, because it affects the loading of package data anywhere it's used (atm: request and content-manifest endpoints).

tests/test_models.py
```diff
@@ -14,3 +16,38 @@ def test_package_manager_get_by_name(name, expected, app, db, auth_env):
         assert expected == pkg_manager
     else:
         assert expected == pkg_manager.name
+
+
+class TestRequest:
```
brunoapimentel, line 21: I think putting all tests from the same class inside a single test class helps the organization, let me know what you think about it.

tox.ini
```diff
@@ -45,7 +45,7 @@ exclude = venv,.git,.tox,dist,*egg,cachito/web/migrations,.env
 ignore = D100,D104,D105,W503,E203
 per-file-ignores =
     # Ignore missing docstrings in the tests and migrations
-    tests/*:D103
+    tests/*:D101,D102,D103
```
brunoapimentel, line 48: This makes flake8 also ignore docstrings for classes and public methods in the test folder.

nirzari (approved):
LGTM

tests/test_models.py
akhmanova: Great idea! I would just move it to another commit. Not because it's too big, but because it has a different topic.

akhmanova: Hm, seems reasonable but: 
1) we already have a logical separation by files
2) there is only one test in the class (do we expect more?)

MartinBasti (approved):
LGTM

tests/test_models.py
brunoapimentel: 1. Yes, but this file, for instance, has 11 different classes. :D 
So, we could simply group the same class test functions together and keep them at top level, but we might benefit from scoping then under a test class.

2. There were no unit tests for this class at all. I believe the coverage is coming from the api tests. It might be good idea to add then.

tkdchen (approved):

PR #518, opened 2021-07-14
Improve Dockerfiles

MartinBasti:
Address hadolint findings

### DIFF
--- a/docker/Dockerfile-api
+++ b/docker/Dockerfile-api
@@ -19,9 +19,9 @@ RUN dnf -y install \
 COPY . .
 COPY ./docker/cachito-httpd.conf /etc/httpd/conf/httpd.conf
 
-RUN pip3 install -r requirements.txt --no-deps --no-cache-dir --require-hashes
-RUN pip3 install -r requirements-web.txt --no-deps --no-cache-dir --require-hashes
-RUN pip3 install . --no-deps --no-cache-dir
+RUN pip3 install -r requirements.txt --no-deps --no-cache-dir --require-hashes \
+    && pip3 install -r requirements-web.txt --no-deps --no-cache-dir --require-hashes \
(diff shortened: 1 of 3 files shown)


brunoapimentel (approved):

docker/Dockerfile-api
```diff
@@ -19,9 +19,9 @@ RUN dnf -y install \
 COPY . .
 COPY ./docker/cachito-httpd.conf /etc/httpd/conf/httpd.conf
 
-RUN pip3 install -r requirements.txt --no-deps --no-cache-dir --require-hashes
-RUN pip3 install -r requirements-web.txt --no-deps --no-cache-dir --require-hashes
-RUN pip3 install . --no-deps --no-cache-dir
+RUN pip3 install -r requirements.txt --no-deps --no-cache-dir --require-hashes \
+    && pip3 install -r requirements-web.txt --no-deps --no-cache-dir --require-hashes \
+    && pip3 install . --no-deps --no-cache-dir
```
serg-cymbaluk, line 24: There is one concern.
Many patches don't include any requirements changes.
So, to rebuild an image for such patch, it would be enough to run only the last layer. And the separated package installation is super fast.

But I don't sure that Docker tracks changes of the requirements.txt correctly or it re-runs all that steps anyway.

akhmanova (approved):

thrbkova (approved):

docker/Dockerfile-api
MartinBasti: Caching of layers is always tricky, I'd even disable that feature altogether. I already had hard times because of cached layer.

tkdchen (approved):

docker/Dockerfile-api
serg-cymbaluk: Ok, lets postpone the topic.

PR #525, opened 2021-07-21
Print tool version information in worker logs

serg-cymbaluk:
CLOUDBLD-3369

Return version for the following tools: go, pip, npm and Node.js (for npm and Yarn package managers).

### DIFF
--- a/README.md
+++ b/README.md
@@ -679,6 +679,18 @@ Offline Installations   | ✓     | x   | x   | x    |
   they could find ways to do an offline install for any package manager, but only `gomod` supports
   this out of the box (i.e. the user does not need to change their workflow).
 
+### Actual Tool Versions
+
+Tool   | Version |
+---    | ---     |
+Go     | 1.15.12 |
+npm    | 6.14.13 |
+node   | 14.17.0 |
+pip    | 20.2.2  |
+Python | 3.9.5   |
+git    | 2.31.1  |
+Yarn   | 1.x     |
+
 ### gomod
 
 The gomod package manager works by parsing the `go.mod` file present in the source repository to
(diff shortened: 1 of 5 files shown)


cachito/workers/tasks/gomod.py
```diff
@@ -54,6 +55,9 @@ def fetch_gomod_source(request_id, dep_replacements=None, package_configs=None):
     :param list package_configs: the list of optional package configurations submitted by the user
     :raises CachitoError: if the dependencies could not be retrieved
     """
+    version_output = run_cmd(["go", "version"], {})
```
akhmanova, line 58: I would be really glad if we add supported versions to documentation as well. Some users ask about it first before submitting a request. And we had a few questions about it before. We need to answer them every time and it could be better just to point to the Cachito docs. 

 Please, tell me if you need help with it.

serg-cymbaluk: Which kind of docs do you mean? README, some document in `docs/` directory or some Confluence page?

akhmanova: README will be the best. There is a part for every package manager

serg-cymbaluk: Updated.

brunoapimentel (approved):
LGTM :rocket:

README.md
```diff
@@ -679,6 +679,17 @@ Offline Installations   | ✓     | x   | x   | x    |
...
+
+Tool   | Version |
+---    | ---     |
+Go     | 1.15.12 |
+npm    | 6.14.13 |
+node   | 14.17.0 |
+pip    | 20.2.2  |
+Python | 3.9.5   |
+git    | 2.31.1  |
+
```
akhmanova, line 692: Could you please add that we support the first version of yarn?

akhmanova (approved):
Thank you so much! Looks good to me. Can be merged after you add yarn version.

README.md
serg-cymbaluk: Added.

serg-cymbaluk:
@akhmanova @brunoapimentel 

Thanks for your review!

PR #529, opened 2021-07-23
Upgrade base images to Fedora 34

brunoapimentel:
CLOUDBLD-6383

### DIFF
--- a/README.md
+++ b/README.md
@@ -683,13 +683,16 @@ Offline Installations   | ✓     | x   | x   | x    |
 
 Tool   | Version |
 ---    | ---     |
-Go     | 1.15.12 |
-npm    | 6.14.13 |
-node   | 14.17.0 |
-pip    | 20.2.2  |
-Python | 3.9.5   |
-git    | 2.31.1  |
-Yarn   | 1.x     |
+Go     | 1.16.5  |
+Npm    | 6.14.13 |
+Node   | 14.17.0 |
+Pip    | 21.0.1  |
+Python | 3.9.6   |
+Git    | 2.31.1  |
+Yarn*  | 1.x     |
+
+* Cachito does not use the Yarn runtime. The processing of yarn.lock files is handled by
+  [PYarn](https://github.com/containerbuildsystem/pyarn), which is compatible with any 1.x file.
 
 ### gomod
 
--- a/docker/Dockerfile-api
+++ b/docker/Dockerfile-api
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:33
+FROM registry.fedoraproject.org/fedora:34
 LABEL maintainer="Red Hat"
 
 WORKDIR /src
--- a/docker/Dockerfile-flower
+++ b/docker/Dockerfile-flower
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:33
+FROM registry.fedoraproject.org/fedora:34
 LABEL maintainer="Red Hat"
 
 RUN dnf -y install \
--- a/docker/Dockerfile-workers
+++ b/docker/Dockerfile-workers
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:33
+FROM registry.fedoraproject.org/fedora:34
 LABEL maintainer="Red Hat"
 
 WORKDIR /src


README.md
```diff
@@ -683,13 +683,13 @@ Offline Installations   | ✓     | x   | x   | x    |
...
+go     | 1.16.5 |
 npm    | 6.14.13 |
 node   | 14.17.0 |
-pip    | 20.2.2  |
-Python | 3.9.5   |
+pip    | 21.0.1  |
+python | 3.9.6   |
 git    | 2.31.1  |
-Yarn   | 1.x     |
+yarn   | 1.x     |
```
brunoapimentel, line 692: I saw that yarn is not really installed in the worker pods, am I right? In this case, it is not possible to write a specific version for Yarn here. 

Do you think we should add a notice that says this version is related to the ability of yarn.lock processing?

```
* The Yarn runtime is not used by Cachito. The processing of yarn.lock files is made by PYarn, which is compatible with any 1.x file.  
```

serg-cymbaluk: > Do you think we should add a notice that says this version is related to the ability of yarn.lock processing?

Yes, there could be note like (via PYarn).

Also in the first column is a tool name, but not executable. So some names could be capitalized or written in a free form.

akhmanova (approved):
LGTM! The only change I expect is a documentation one (`via PYarn`).

README.md
brunoapimentel: Yup, I was just kind of bothered that some of them were capitalized and some not (nevermind me, I can focus on unimportant things at times :sweat_smile:). I just changed them all to be capitalized, it seems to be the most correct form.

chmeliik:
Changelog for go 1.16: https://golang.org/doc/go1.16

This part is pretty interesting
> The all pattern
>
> When the main module's go.mod file declares go 1.16 or higher, the all package pattern now matches only those packages that are transitively imported by a package or test found in the main module. (Packages imported by tests of packages imported by the main module are no longer included.) This is the same set of packages retained by go mod vendor since Go 1.11.

This might mean that the code we use for listing packages will get different results for go 1.16. We need to find a module that gets different results based on whether vendoring is used or not (that can already happen in Cachito now) and check how Cachito handles it with go 1.16.

https://github.com/containernetworking/plugins might be a decent candidate, IIRC that one is where I first noticed the discrepancy between vendoring vs. not vendoring. We would have to fork it and set go version to 1.16 in go.mod.

brunoapimentel:
> Changelog for go 1.16: https://golang.org/doc/go1.16
> 
> This part is pretty interesting
> 
> > The all pattern
> > When the main module's go.mod file declares go 1.16 or higher, the all package pattern now matches only those packages that are transitively imported by a package or test found in the main module. (Packages imported by tests of packages imported by the main module are no longer included.) This is the same set of packages retained by go mod vendor since Go 1.11.
> 
> This might mean that the code we use for listing packages will get different results for go 1.16. We need to find a module that gets different results based on whether vendoring is used or not (that can already happen in Cachito now) and check how Cachito handles it with go 1.16.
> 
> https://github.com/containernetworking/plugins might be a decent candidate, IIRC that one is where I first noticed the discrepancy between vendoring vs. not vendoring. We would have to fork it and set go version to 1.16 in go.mod.

So, the results are not good D:. 

The contents were very different when using Go 1.16. I have uploaded the requests' json here so you can look at it: https://drive.google.com/drive/u/0/folders/1xtgYIkocnycMdYcm8DR3-SwnN7nqwx6E.

b-ranto:
This is only the case when you declare go version 1.16 or higher in your `go.mod` file, right? I'd assume you use golang 1.16 (or higher) when you declare it in your `go.mod` file.

I do agree that this might make some changes for some projects but IMO, the changes are "for the better", i.e. it is fixing the deps/modules to what upstream project uses?

chmeliik:
> So, the results are not good D:
> 
> The contents were very different when using Go 1.16. I have uploaded the requests' json here so you can look at it: https://drive.google.com/drive/u/0/folders/1xtgYIkocnycMdYcm8DR3-SwnN7nqwx6E.
> 
> Will keep investigating tomorrow.

@brunoapimentel this *might* actually be fine. Could you try the same repo on both f33 and f34? The difference between no flag and vendor flag is expected, I was more afraid that there would *not* be a difference.

If f33-no-flag matches f34-no-flag and f33-vendor-flag matches f34-vendor-flag (for the same repo), then we should be good.

chmeliik:
> This is only the case when you declare go version 1.16 or higher in your `go.mod` file, right? I'd assume you use golang 1.16 (or higher) when you declare it in your `go.mod` file.
> 
> I do agree that this might make some changes for some projects but IMO, the changes are "for the better", i.e. it is fixing the deps/modules to what upstream project uses?

That's an interesting point. In my opinion, Cachito should accurately report everything that was downloaded and nothing more. We normally use `go mod download` for downloading modules, or `go mod vendor` if the `gomod-vendor` flag is used.

Those do not behave the same way, `go mod download` downloads all of the transitive module dependencies, while `go mod vendor` downloads only a subset of it which is actually required. The new behavior of the `all` pattern kind of sounds like it might match the vendoring behavior. We already have custom logic in place for vendoring (so Cachito will not report more than what was downloaded), but I was worried that this change in golang would make us report dependencies as if vendoring was used even when that is not the case.

brunoapimentel:
> So, the results are not good D:
> 
> The contents were very different when using Go 1.16. I have uploaded the requests' json here so you can look at it: https://drive.google.com/drive/u/0/folders/1xtgYIkocnycMdYcm8DR3-SwnN7nqwx6E.
> 
> Will keep investigating tomorrow.

Completely ignore this. I somehow mixed up the requests I made and the Fedora 33 results were indeed from a different repo. :man_facepalming:   

Uploaded the correct results to the folder now. I ran a diff between the matching requests (flag/no-flag) and no differences showed up (besides timestamps and request number). So, I guess we're good to go? Wdyt @chmeliik?

chmeliik:
> Completely ignore this. I somehow mixed up the requests I made and the Fedora 33 results were indeed from a different repo. man_facepalming
> 
> Uploaded the correct results to the folder now. I ran a diff between the matching requests (flag/no-flag) and no differences showed up (besides timestamps and request number). So, I guess we're good to go? Wdyt @chmeliik?

Yup, should be good :tada:

I think it doesn't affect us because we are only using the `all` pattern for _modules_, not packages. But I wanted to be sure, since the words "modules" and "packages" can sometimes be used interchangeably.

chmeliik (approved):

PR #530, opened 2021-07-26
Enable cachito_gomod_strict_vendor flag by default

serg-cymbaluk:
CLOUDBLD-1710

The option means that the request will fail in case a vendor folder exists
and there is no 'gomod-vendor' set.

### DIFF
--- a/tests/test_workers/test_pkg_managers/test_gomod.py
+++ b/tests/test_workers/test_pkg_managers/test_gomod.py
@@ -3,6 +3,7 @@ import os
 import re
 import tarfile
 import textwrap
+from contextlib import nullcontext
 from tempfile import TemporaryDirectory as tempDir
 from textwrap import dedent
 from unittest import mock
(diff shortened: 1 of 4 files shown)


brunoapimentel:
LGTM :+1: 

The only thing I did not quite get is why we have two flags to handle gomod vendoring (`gomod-vendor` and `gomod-vendor-check`). I did read the README, still not sure. :sweat_smile: 

Is `gomod-vendor-check` related to enabling/disabling the check for the vendor folder? Maybe we can clear it up in the README>

chmeliik:
> The only thing I did not quite get is why we have two flags to handle gomod vendoring (`gomod-vendor` and `gomod-vendor-check`). I did read the README, still not sure. sweat_smile
> 
> Is `gomod-vendor-check` related to enabling/disabling the check for the vendor folder? Maybe we can clear it up in the README.

When you use `gomod-vendor`, cachito will run `go mod vendor` in the requested repository. That command can make changes in the repo (in its vendor/ directory, more specifically), which is usually unexpected. The `gomod-vendor-check` exists for that purpose - after Cachito runs `go mod vendor`, it will check that nothing changed in the repo. Basically, the `gomod-vendor-check` flag is used to verify that the user did the vendoring properly on their own.

tests/test_workers/test_pkg_managers/test_gomod.py
```diff
@@ -344,30 +345,51 @@ def test_resolve_gomod_vendor_dependencies(
...
+    ]
 
     archive_path = "/this/is/path/to/archive.tar.gz"
     request = {"id": 3, "ref": "c50b93a32df1c9d700e3e80996845bc2e13be848"}
     expected_error = (
         'The "gomod-vendor" or "gomod-vendor-check" flag must be set when your repository has '
         "vendored dependencies."
     )
-    with pytest.raises(CachitoError, match=expected_error):
+    with raise_error and pytest.raises(CachitoError, match=expected_error) or nullcontext():
```
chmeliik, line 392: Ooh fancy, that's the first time I've seen `nullcontext()` :rocket:

chmeliik (approved):
LGTM, we just need to be aware that this is a breaking change and notify users

brunoapimentel (approved):

serg-cymbaluk:
@brunoapimentel @chmeliik 

Thank you for the review!

> we just need to be aware that this is a breaking change and notify users

Yes, I will hold the PR until we manage to do that.

PR #531, opened 2021-07-26
Throws exception when the packages json file does not exist

brunoapimentel:


brunoapimentel:
Lots of broken unit tests, still fixing them.

cachito/workers/tasks/gomod.py
```diff
@@ -76,6 +76,7 @@ def fetch_gomod_source(request_id, dep_replacements=None, package_configs=None):
 
         # missing gomod files is supported if there is only one path referenced
         if config.cachito_gomod_ignore_missing_gomod_file and len(subpaths) == 1:
+            PackagesData().write_to_file(bundle_dir.gomod_packages_data)
```
brunoapimentel, line 79: I thought it would be best to cover this corner case and generate an empty packages.json file. Let me know what you think.

cachito/common/packages_data.py
```diff
@@ -145,14 +145,13 @@ def load(self, file_name: Union[str, Path]) -> None:
...
-            return
+            raise CachitoError(f"File {file_name} does not exist.")
 
         with open(file_name, "r", encoding="utf-8") as f:
             data = json.load(f)
             packages = data.get("packages")
             if packages is None:
-                log.warning("Packages data file does not include key 'packages'.")
+                raise CachitoError(f"Packages data file does not include key 'packages'.")
                 return
```
chmeliik, line 155: Maybe just `data["packages"]`? Under normal circumstances, if the file is valid JSON then it must have the "packages" key. So IMO, if somehow that is not the case, we can just leave that as an uncaught KeyError.

cachito/workers/tasks/gomod.py
chmeliik: Sounds good, let's just add a comment to explain why it's needed

tests/test_api_v1.py
```diff
@@ -1036,8 +1036,9 @@ def test_set_state(state, app, client, db, worker_auth_env, tmpdir):
 
     assert patch_rv.status_code == 200
 
-    get_rv = client.get(f"/api/v1/requests/{request_id}")
-    assert get_rv.status_code == 200
+    with mock.patch("cachito.common.packages_data.PackagesData.load", return_value="{}"):
```
chmeliik, line 1039: Is the `return_value` needed? The `load()` method doesn't return anything IIRC

PR #532, opened 2021-07-29
Checks packages file before finalizing a request

brunoapimentel:
This PR adds a mechanism to verify it the packages file could be properly read by an API pod before the request is moved to the `completed` state. This is done by querying the API from the worker pod that is processing the request and comparing its response with the number of packages and dependencies that were just written.

We also made so that if this check fails, it will be retried considering a configured interval and number of attempts.

image

CLOUDBLD-6599

### DIFF
--- a/cachito/workers/config.py
+++ b/cachito/workers/config.py
@@ -31,6 +31,8 @@ class Config(object):
             "SKIP_SASS_BINARY_DOWNLOAD_FOR_CI": {"value": "true", "kind": "literal"},
         },
     }
+    cachito_finalize_request_packages_check_interval = 1.0
+    cachito_finalize_request_packages_check_max_attempts = 10
     cachito_deps_patch_batch_size = 50
     cachito_gomod_ignore_missing_gomod_file = True
     cachito_gomod_strict_vendor = False
@@ -141,6 +143,8 @@ class TestingConfig(DevelopmentConfig):
             "SKIP_SASS_BINARY_DOWNLOAD_FOR_CI": {"value": "true", "kind": "literal"},
         },
     }
+    cachito_finalize_request_packages_check_interval = 0.1
+    cachito_finalize_request_packages_check_max_attempts = 3
     cachito_npm_file_deps_allowlist = {"han_solo": ["millennium-falcon"]}
     cachito_request_file_logs_dir = None
 
--- a/cachito/workers/tasks/general.py
+++ b/cachito/workers/tasks/general.py
@@ -3,6 +3,7 @@ import logging
 import os
 import shutil
 import tarfile
+import time
 from pathlib import Path
 from typing import Any, Callable, List, Optional
 
@@ -11,6 +12,7 @@ import requests
 from cachito.common.checksum import hash_file
 from cachito.common.packages_data import PackagesData
 from cachito.errors import CachitoError, ValidationError
+from cachito.workers.config import get_worker_config
 from cachito.workers.paths import RequestBundleDir
 from cachito.workers.scm import Git
 from cachito.workers.tasks.celery import app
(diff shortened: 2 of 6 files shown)


brunoapimentel:
Still working on unit tests, but that's the general idea.

tests/test_workers/test_tasks/test_general.py
```diff
@@ -259,25 +263,49 @@ def test_finalize_request(
...
+                "pkg_managers": ["pip"],
+                "packages": [],
+                "dependencies": [],
+            }
+
+    mock_get_request.side_effect = side_effect
     mock_aggregate_data.return_value = mock.Mock(packages=[pkg], all_dependencies=[pkg, pkg])
 
-    tasks.finalize_request(42)
+    with raise_error and pytest.raises(CachitoError) or nullcontext():
```
brunoapimentel, line 297: Stole this shamelessly from @serg-cymbaluk ;)

cachito/workers/config.py
```diff
@@ -64,6 +64,10 @@ class Config(object):
     cachito_task_log_format = (
         "[%(asctime)s #%(request_id)s %(name)s %(levelname)s %(module)s.%(funcName)s] %(message)s"
     )
+    # How many times a worker tries to check the packages file before finalizing a request
+    check_packages_file_max_retries = 10
+    # Interval between checks for the packages file, in seconds
+    check_packages_file_interval = 1.0
```
chmeliik, line 70: Cachito-specific options should have the `cachito_` prefix and should be documented in the README

tests/test_workers/test_tasks/test_general.py
```diff
@@ -259,25 +263,49 @@ def test_finalize_request(
...
+    # the mock will be called at least twice, once for the first loading of the request and
+    # once for the first check on the packages file
+    expected_mock_calls = retries_until_file_is_ready + 2
+
+    def side_effect(*args):
+        call_count = mock_get_request.call_count
+
+        # the mock should return a full answer only when we reach the expected amount of retries
+        if call_count == expected_mock_calls:
+            return {
```
brunoapimentel, line 280: I wish I could have declared this variable once and used [dict union](https://www.python.org/dev/peps/pep-0584/) to patch it.

* cries in python 3.8

cachito/workers/tasks/general.py
```diff
@@ -169,6 +171,36 @@ def save_bundle_archive_checksum(request_id: int) -> None:
...
+    while not check_passed and retries <= config.check_packages_file_max_retries:
+        if retries > 0:
+            time.sleep(config.check_packages_file_interval)
+
+        log.info("Checking if the packages file was properly loaded. Retried %s times.", retries)
+        check_passed = _check_packages_data_on_api(request_id, packages_count, dependencies_count)
+        retries += 1
+
+    if not check_passed:
+        raise CachitoError("Packages file could not be loaded for request %i.", request_id)
```
chmeliik, line 201: I don't think this will format the string properly, this would for a log message but not for an exception

tests/test_workers/test_tasks/test_general.py
chmeliik: You could still do something like this
```python
return {**request, "packages": [], "dependencies": []}
```
But dict union would be very fancy indeed :D

tests/test_workers/test_tasks/test_general.py
chmeliik: Maybe check the expected error message here (with `raises(..., match="some message")`)?

tests/test_workers/test_tasks/test_general.py
```diff
@@ -259,25 +263,49 @@ def test_finalize_request(
...
-    mock_get_request.assert_called_once_with(42)
+    mock_get_request.assert_called_with(42)
     mock_create_archive.assert_called_once_with(42, ["some-flag"])
     mock_save_bundle_archive_checksum.assert_called_once_with(42)
     mock_aggregate_data.assert_called_once_with(42, ["pip"])
     mock_set_counts.assert_called_once_with(42, 1, 2)
-    mock_set_state.assert_called_once_with(42, "complete", "Completed successfully")
+
+    if not raise_error:
+        assert mock_get_request.call_count == expected_mock_calls
```
chmeliik, line 307: Shouldn't this assertion be true even in the error case?

cachito/workers/tasks/general.py
```diff
@@ -169,6 +171,36 @@ def save_bundle_archive_checksum(request_id: int) -> None:
...
 
 
+def _check_packages_data_on_api(
+    request_id: int, packages_count: int, dependencies_count: int
+) -> bool:
+    request = get_request(request_id)
+    packages = request.get("packages", [])
+    dependencies = request.get("dependencies", [])
+
+    return len(packages) == packages_count and len(dependencies) == dependencies_count
```
chmeliik, line 181: Should we also log the counts when there is a mismatch? Imagine if we one day see a case where all the counts are non-zero but they still do not match o_O

chmeliik (commented):
+1 on the overall approach, couple nitpicks as usual :)

tests/test_workers/test_tasks/test_general.py
brunoapimentel: Since the test config is limited to 3 retries, it will never reach the expected retries of 5. The ideia here is just making sure the code raises an exception once the limit is over.

chmeliik: oh, right

cachito/workers/tasks/general.py
brunoapimentel: /me facepalms heavily.

cachito/workers/tasks/general.py
```diff
@@ -169,6 +171,51 @@ def save_bundle_archive_checksum(request_id: int) -> None:
...
+        return
+
+    attempts = 0
+    check_passed = False
+
+    while not check_passed and attempts < max_attempts:
+        if attempts > 0:
+            time.sleep(interval)
+
+        log.info("Checking if the packages file was properly loaded (attempt %i).", attempts)
```
chmeliik, line 211: Should the first attempt be logged as "attempt 1" rather than 0?

brunoapimentel: Sorry, this is Python, not Lua :(

Jokes apart, most definitely, will fix it right now.

chmeliik (approved):
LGTM :+1:

brunoapimentel:
The tests made with this code on OCP were not very promising: when the issue of not loading the packages/dependencies did show up, the retries were not enough to allow the API pods to be able to read the file.

We suspect this may be caused by the flush of the file from the processes memory to the NFS itself, which can be held until the process finishes. Some adjustments will be made to the approach and a new PR will be open.

PR #533, opened 2021-07-30
Add Python 3.9 as supported in setup.py

elfosardo:
Also change python_requires=">=3.8" in setup.py and add Python 3 only mention.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,6 +36,8 @@ setup(
     },
     classifiers=[
         "License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)",
+        "Programming Language :: Python :: 3 :: Only",
+        "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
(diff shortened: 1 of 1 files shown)


chmeliik:
Hi @elfosardo, good catch on this inconsistency. However, the problem is the opposite, we have failed to keep the `python_requires` value up to date - it should be `>=3.8` now.

May I ask what motivated this contribution? :-)

elfosardo:
@chmeliik hi! I will change the PR to reflect that, to be honest I was a bit undecided also looking at the tox and github actions configs, where the tests are set for python 3.8 and 3.9 only

The reason of the contribution is that I started looking at the code because I'm very interested in the pip support, I saw the inconsistency, and I just thought to spend a couple of minutes to fix it :)

chmeliik (approved):
LGTM, thanks!

MartinBasti (approved):

tkdchen:
:+1:

akhmanova (approved):
Thank you!

PR #501, opened 2021-06-25
Add integration test for gomod-vendor with no deps

chmeliik:
CLOUDBLD-6138

Test how Cachito behaves when gomod vendoring is requested but there is
nothing to vendor.

Signed-off-by: Adam Cmiel <acmiel@redhat.com>

PR #534, opened 2021-07-30
Implement integration test for gomod-vendor-check

dcho5:
CLOUDBLD-5578

Add test function which fails with inconsistent vendor or empty vendor.
Add test data for consistent vendor and no vendor.

### DIFF
--- a/tests/integration/test_data/gomod_vendor_check.yaml
+++ b/tests/integration/test_data/gomod_vendor_check.yaml
@@ -1,0 +1,188 @@
+# gomod package with gomod-vendor-check flag
+# repo: The URL for the upstream git repository
+# ref: A git reference at the given git repository
+# flags: A list of strings with Cachito flags
+# expected_files: Expected source files <relative_path>: <file_URL>
+# expected_deps_files: Expected dependencies files (empty)
+# response_expectations: Parts of the Cachito response to check
+# content_manifest: PURLs for image contents part
+wrong_vendor:
+  repo: https://github.com/cachito-testing/gomod-vendor-check-fail.git
+  ref:  8553df6498705b2b36614320ca0c65bc24a1d9e6
+  flags: ["gomod-vendor-check"]
+  pkg_managers: ["gomod"]
+empty_vendor:
+  repo: https://github.com/cachito-testing/gomod-vendor-check-empty-vendor.git
+  ref:  9989e210ac2993196e22d0a23fe18ce460012058
+  flags: ["gomod-vendor-check"]
+  pkg_managers: ["gomod"]
+correct_vendor:
+  repo: https://github.com/cachito-testing/gomod-vendor-check-pass.git
+  ref:  0543a5034b687df174c6b12b7b6b9c04770a856f
+  flags: ["gomod-vendor-check"]
+  pkg_managers: ["gomod"]
+  response_expectations:
+    dependencies:
+      - name: golang.org/x/text/internal/tag
+        replaces: null
+        type: go-package
+        version: v0.0.0-20170915032832-14c0d48ead0c
+      - name: golang.org/x/text/language
+        replaces: null
+        type: go-package
+        version: v0.0.0-20170915032832-14c0d48ead0c
+      - name: rsc.io/quote
+        replaces: null
+        type: go-package
+        version: v1.5.2
+      - name: rsc.io/sampler
+        replaces: null
+        type: go-package
+        version: v1.3.0
+      - name: golang.org/x/text
+        replaces: null
+        type: gomod
+        version: v0.0.0-20170915032832-14c0d48ead0c
+      - name: rsc.io/quote
+        replaces: null
+        type: gomod
+        version: v1.5.2
+      - name: rsc.io/sampler
+        replaces: null
+        type: gomod
+        version: v1.3.0
+    packages:
+      - dependencies:
+        - name: golang.org/x/text/internal/tag
+          replaces: null
+          type: go-package
+          version: v0.0.0-20170915032832-14c0d48ead0c
+        - name: golang.org/x/text/language
+          replaces: null
+          type: go-package
+          version: v0.0.0-20170915032832-14c0d48ead0c
+        - name: rsc.io/quote
+          replaces: null
+          type: go-package
+          version: v1.5.2
+        - name: rsc.io/sampler
+          replaces: null
+          type: go-package
+          version: v1.3.0
+        name: github.com/cachito-testing/gomod-vendor-check-pass
+        type: go-package
+        version: v0.0.0-20210802173203-0543a5034b68
+      - dependencies:
+        - name: golang.org/x/text
+          replaces: null
+          type: gomod
+          version: v0.0.0-20170915032832-14c0d48ead0c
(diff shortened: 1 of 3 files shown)


dcho5:
I haven't implemented the test case for `empty_vendor` in `test_gomod_vendor_check_pass`, yet, but how does this look so far, @akhmanova ?

tests/integration/test_gomod_packages.py
```diff
@@ -32,3 +32,73 @@ def test_gomod_vendor_without_flag(test_env):
...
+        "The content of the vendor directory is not consistent with go.mod. "
+        "Run `go mod vendor` locally to fix this problem. See the logs for more details."
+    )
+    assert error_msg in completed_response.data["state_reason"], (
+        f"#{completed_response.id}: Request failed correctly, but with unexpected message: "
+        f"{completed_response.data['state_reason']}. Expected message was: {error_msg}"
+    )
+
+
+def test_gomod_vendor_check_pass(test_env):
```
akhmanova, line 67: @chmeliik 
> Would the empty_vendor case be the same as #501? Or is that something else?

Your example shows a gomod repo without deps. This question is mostly about a gomod repo with dependencies but with an empty vendor directory. I think it can be the same result as for `test_gomod_vendor_check_fai()`, because it is just a special case of the general. But I am not 100% sure and if @dcho5 has time, I would like it to be checked as well.

akhmanova: WDYT? :)

tests/integration/test_gomod_packages.py
```diff
@@ -32,3 +32,73 @@ def test_gomod_vendor_without_flag(test_env):
         )
     else:
         utils.assert_properly_completed_response(completed_response)
+
+
+def test_gomod_vendor_check_fail(test_env):
```
akhmanova, line 37: The first test looks good to me! Thank you!

tests/integration/test_gomod_packages.py
```diff
@@ -32,3 +32,73 @@ def test_gomod_vendor_without_flag(test_env):
...
+    )
+    assert error_msg in completed_response.data["state_reason"], (
+        f"#{completed_response.id}: Request failed correctly, but with unexpected message: "
+        f"{completed_response.data['state_reason']}. Expected message was: {error_msg}"
+    )
+
+
+def test_gomod_vendor_check_pass(test_env):
+    """
+    Validate successful completion of gomod vendor request with gomod-vendor-check flag.
```
akhmanova, line 69: I think this test copies [the common one](https://github.com/release-engineering/cachito/blob/master/tests/integration/test_packages.py#L25). Could you please move the test data there or point me why you choose to have a new one?

dcho5: I'll move the first case to `test_packages.py`, and keep the `empty_vendor` and `no_vendor` test cases in this file.

dcho5:
Moved `correct_vendor` and `no_vendor` test cases to `test_packages` from `test_gomod_vendor_check_pass`. For, `no_vendor` an initial Cachito request is completed for `gomod-vendor-check-no-vendor`, which has dependencies in `go.mod` but no vendor directory. The test checks that the request is successful, and that the cached files match those of `gomod-vendor-check-pass` (which has a consistent vendor directory) instead of `gomod-vendor-check-no-vendor`.

tests/integration/test_gomod_packages.py
akhmanova: Great, thank you

dcho5:
Implemented all tests, please review @akhmanova

tests/integration/test_gomod_packages.py
chmeliik: I think it's the same case as `test_gomod_vendor_check_fail`, yeah. But doesn't hurt to have both :-)

tests/integration/test_data/gomod_vendor_check.yaml
```diff
@@ -0,0 +1,188 @@
...
+# expected_deps_files: Expected dependencies files (empty)
+# response_expectations: Parts of the Cachito response to check
+# content_manifest: PURLs for image contents part
+wrong_vendor:
+  repo: https://github.com/cachito-testing/gomod-vendor-check-fail.git
+  ref:  8553df6498705b2b36614320ca0c65bc24a1d9e6
+  flags: ["gomod-vendor-check"]
+  pkg_managers: ["gomod"]
+empty_vendor:
+  repo: https://github.com/cachito-testing/gomod-vendor-check-empty-vendor.git
```
chmeliik, line 15: I expected that https://github.com/cachito-testing/gomod-vendor-check-empty-vendor would have an empty `vendor/` directory. Would that be a more interesting test case than just having empty `modules.txt`? IMO that is very similar to https://github.com/cachito-testing/gomod-vendor-check-fail

Note: it can't be completely empty, since git doesn't recognize empty dirs. You would have to put something like an empty `.gitkeep` file in there

tests/integration/test_gomod_packages.py
```diff
@@ -32,3 +34,34 @@ def test_gomod_vendor_without_flag(test_env):
...
+            "flags": env_data["flags"],
+            "pkg_managers": env_data["pkg_managers"],
+        },
+    )
+    completed_response = client.wait_for_complete_request(initial_response)
+    assert completed_response.status == 200
+    assert completed_response.data["state"] == "failed"
+    error_msg = (
+        "The content of the vendor directory is not consistent with go.mod. "
+        "Run `go mod vendor` locally to fix this problem. See the logs for more details."
```
chmeliik, line 62: It might also be interesting to check the logs here, since we promise there will be more details :D

AFAIK it's not something we've done in integration tests yet so I'd be OK with leaving it as a future improvement, I'll leave it up to you.

chmeliik (commented):
Nice! I've got some suggestions :-)

tests/integration/test_data/gomod_vendor_check.yaml
dcho5: That's exactly what I thought initially as well, but I also knew that git doesn't recognize empty dirs. I wasn't aware of the `.gitkeep` trick. I'll give it a try!

dcho5: Done!

tests/integration/test_gomod_packages.py
dcho5: Were you thinking that we confirm the details of the log (the same way we confirm the `error_msg`) in the test function as well? The logs we are looking for are logged in `_vendor_changed` - how can I access the logs in `test_gomod_vendor_check_fail`?

chmeliik: Yeah that's what I was thinking. Request logs can be accessed at a url, e.g. localhost:8080/api/v1/requests/1/logs

tests/integration/test_data/gomod_vendor_check.yaml
chmeliik: Nice :+1:

chmeliik (approved):
LGTM :tada: 

Checks for logs can be added later as we discussed

akhmanova (approved):

PR #536, opened 2021-08-02
Checks the packages file before finalizing a request

brunoapimentel:
This PR makes three changes:
- Allows the API to read the packages file regardless of the request's state (reverting a change done in a previous PR)
- Creates a new task that will be responsible for writing all files related to the request and setting the pkgs/deps count
- Adds a mechanism to verify it the packages file can be properly read by an API pod before the request is moved to the completed state. 

The reason for the task split is because the retry did not work when the file write and the check were part of the same task. We suspect the flush of the file to the NFS is only made when the task process is finished.

The check itself is done by querying the API from the worker pod that is processing the request and comparing its response with the number of packages and dependencies that were just written. We also made so that if this check fails, it will be retried considering a configured interval and number of attempts.

Here's a simplified diagram that shows how a request processing will be once this PR is merged:

image

CLOUDBLD-6599

### DIFF
--- a/README.md
+++ b/README.md
@@ -427,6 +427,13 @@ Custom configuration for the Celery workers are listed below:
   `value` must be a string which specifies the value of the environment variable. The `kind` must
   also be a string which specifies the type of value, either `"path"` or `"literal"`. Check
   `cachito/workers/config.py::Config` for the default value of this configuration.
+* `cachito_finalize_request_packages_check_interval` - when finalizing a request, a worker will
+  query the API pods and check their response to make sure the packages and dependencies were
+  properly loaded. In case there's a failure, the worker will retry the check. This property
+  sets the interval between checks, in seconds.
+* `cachito_finalize_request_packages_check_max_attempts` - similar to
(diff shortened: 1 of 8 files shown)


brunoapimentel:
This PR replaces #532, since we're changing the overall approach here.

tkdchen (commented):
Hi Bruno,

What would be the difference between reading the data from the worker and the api? I'm not clear why calling the api to test the availability of the `packages.json`.

brunoapimentel:
> Hi Bruno,
> 
> What would be the difference between reading the data from the worker and the api? I'm not clear why calling the api to test the availability of the `packages.json`.

That is a great question. My first thought was to make the check on the worker side itself, since it was easier and we did not need an extra call to the API. But all integration tests run on OCP4 with this code passed on the first attempt, which means the bug was not triggered. We did not leave it for a very long time on the test environment, though. I left a comment about it on our investigation doc, if want to check it out (Jul 28 01:22).

We thought the issue of file sync can be happening only between different pods, and that's why we decided to query the API for the data, since it would be the closest that an actual client would be seeing.

tkdchen (approved):

PR #541, opened 2021-08-04
Adds endpoints that lists packages file contents

brunoapimentel:
This endpoint is intented to be used by the worker process while finalizing a request. The reason for this is that the current `/api/v1/requests/{id}` endpoint does not try to load the packages file unless the request is at the `completed` state.

We thought it would be best to keep it this way because there would be no impact to the current way the API is working, avoiding adding any performance issues (the NFS cache will need to be force cleaned before every loading of the file).

This PR is the first step to have a complete solution that will flush the cache before loading the packages file and check its integrity before the request is marked as `completed`. Here's a diagram showing the full solution:

image

CLOUDBLD-6599

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -13,6 +13,7 @@ from sqlalchemy import func
 from werkzeug.exceptions import BadRequest, Forbidden, Gone, InternalServerError, NotFound
 
 from cachito.common.checksum import hash_file
+from cachito.common.packages_data import PackagesData
 from cachito.common.paths import RequestBundleDir
 from cachito.common.utils import b64encode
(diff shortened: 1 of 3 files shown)


chmeliik (approved):
LGTM, I love the diagrams! :rocket: 

Do I assume correctly that the cache invalidation will be done in a different PR?

brunoapimentel:
> LGTM, I love the diagrams! 
> 
> Do I assume correctly that the cache invalidation will be done in a different PR?

Yup, next PR will cover it.

About the diagrams, take a look at this tool here: https://sequencediagram.org/. Life changer.

tkdchen (approved):

PR #542, opened 2021-08-04
Add 'Limitations' section to 'Feature Support' documentation

serg-cymbaluk:
CLOUDBLD-1486

Describe the case when a repository used for building multiple images.

### DIFF
--- a/README.md
+++ b/README.md
@@ -679,6 +679,17 @@ Offline Installations   | ✓     | x   | x   | x    |
   they could find ways to do an offline install for any package manager, but only `gomod` supports
   this out of the box (i.e. the user does not need to change their workflow).
 
+#### Limitations
+
+Cachito collects all dependencies of specified package manager types from a repository
+regardless of whether they are used by a certain application or not.
+For example, if some repository contains global `go.mod` file, dependencies from the file would be
+used for all container images built from the repository (see [gomod](#gomod) for more details).
+
+Meanwhile, Cachito still allows distinguishing between multiple applications from one source using
+**Multiple Paths** feature. So it could be used as a solution in that case (depends on the
+repository structure).
+
 ### Actual Tool Versions
 
 Tool   | Version |


README.md
```diff
@@ -679,6 +679,17 @@ Offline Installations   | ✓     | x   | x   | x    |
   they could find ways to do an offline install for any package manager, but only `gomod` supports
   this out of the box (i.e. the user does not need to change their workflow).
 
+#### Limitations
+
+Cachito collects all dependencies from repository regardless of whether they used by a certain image
```
chmeliik, line 684: This is not entirely accurate, Cachito collects dependencies for the packages that the user specifies (as you probably know, since you mention the multiple paths solution).

I'm also not sure if we should say "image" here - this is the first mention of container images in the README (apart from the containers that run Cachito itself). Cachito does not necessarily have to be used for building container images. Maybe "application," as used below, would be better.

README.md
```diff
@@ -679,6 +679,17 @@ Offline Installations   | ✓     | x   | x   | x    |
...
+or not.
+For example, if some repository contains global `go.mod` file, dependencies from the file would be
+used for all images built from the repository (see [gomod](#gomod) for more details).
+
+Meanwhile, Cachito still allows distinguishing between multiple applications from one source using
+**Multiple Paths** feature. So it could be used as a workaround in that case (depends on the
+reporitory structure).
```
chmeliik, line 691: +1, IMO this is not just a workaround but the proper solution. If you're building 2 applications from your repo, split up your repo into 2 packages.

It's not always in the users' control (e.g. building from an upstream that they do not own), but it's not something Cachito should have to do.

chmeliik (commented):
+1 on the overall solution of documenting the limitation, some suggestions on the wording

README.md
serg-cymbaluk: Replaced 'workaround' with 'solution'.

README.md
serg-cymbaluk: Used more precise wording:
```diff
- all dependencies
+ all accessible dependencies of specified package manager types
```

Replaced 'image' with 'application' in the first sentence. But in the second one it is an example, so any case specific term can be used.

chmeliik: Can you clarify what you mean by "accessible"? And can we at least use "container image" instead of "image"? The word has too many meanings in the software world these days :D

README.md
```diff
@@ -679,6 +679,17 @@ Offline Installations   | ✓     | x   | x   | x    |
   they could find ways to do an offline install for any package manager, but only `gomod` supports
   this out of the box (i.e. the user does not need to change their workflow).
 
+#### Limitations
+
+Cachito collects all accessible dependencies of specified package manager types from a repository
+regardless of whether they used by a certain application or not.
```
chmeliik, line 685:
```suggestion
regardless of whether they are used by a certain application or not.
```

README.md
serg-cymbaluk: > Can you clarify what you mean by "accessible"?

Given the nature of the dependencies search, since we go through all subpaths and check availability of corresponding files, the word could be relevant.

> And can we at least use "container image" instead of "image"?

Agree. Let me fix it.

README.md
serg-cymbaluk: Fixed.

README.md
serg-cymbaluk: Fixed 'images'.

chmeliik: > > Can you clarify what you mean by "accessible"?
> 
> Given the nature of the dependencies search, since we go through all subpaths and check availability of corresponding files, the word could be relevant.

Hmm, I see. Maybe something like "Cachito collects all the dependencies defined in packaging files" would be more accurate? Not sure if "packaging files" is the right term, I tried to find a word that that would match npm lockfiles, pip requirements files and go.mod files at the same time.

serg-cymbaluk: Removed 'accessible' to not confuse users.

chmeliik: That works :+1:

chmeliik (approved):
LGTM :+1:

brunoapimentel:
It might be just me, but I still don't understand this limitation very well (read through the ticket, related tickets and the added docs).

Does it means that if we have something like this:

```
/sideproject/requirements.txt
/sideproject/Dockerfile
/requirements.txt
```
If the request was based on the root dir, it would have all the deps from the root dir, regardless if they're used by the image at `/sideproject` or not. Is this what the limitation is about? The requester wanted that we somehow processed the Dockerfile to see what's actually being used there?

Do we have a sample request to see how this looks?

chmeliik:
> It might be just me, but I still don't understand this limitation very well (read through the ticket, related tickets and the added docs).
> 
> Does it means that if we have something like this:
> 
> ```
> /sideproject/requirements.txt
> /sideproject/Dockerfile
> /requirements.txt
> ```
> 
> If the request was based on the root dir, it would have all the deps from the root dir, regardless if they're used by the image at `/sideproject` or not. Is this what the limitation is about? The requester wanted that we somehow processed the Dockerfile to see what's actually being used there?
> 
> Do we have a sample request to see how this looks?

One such example could be kubernetes: https://github.com/kubernetes/kubernetes. There's a global go.mod file that defines the deps, but if you're only building the `kubectl`  CLI tool (https://github.com/kubernetes/kubernetes/tree/master/cmd/kubectl) then you probably only need a small subset of those. Not sure if anyone is trying to do that though.

brunoapimentel:
> One such example could be kubernetes: https://github.com/kubernetes/kubernetes. There's a global go.mod file that defines the deps, but if you're only building the `kubectl` CLI tool (https://github.com/kubernetes/kubernetes/tree/master/cmd/kubectl) then you probably only need a small subset of those. Not sure if anyone is trying to do that though.

Does this scenario happen in other pkg managers besides gomod?

chmeliik:
> > One such example could be kubernetes: https://github.com/kubernetes/kubernetes. There's a global go.mod file that defines the deps, but if you're only building the `kubectl` CLI tool (https://github.com/kubernetes/kubernetes/tree/master/cmd/kubectl) then you probably only need a small subset of those. Not sure if anyone is trying to do that though.
> 
> Does this scenario happen in other pkg managers besides gomod?

Not really AFAIK. With NPM and Yarn, you would usually split code up into packages properly. With pip, you have absolute freedom over how you split up your requirements.txt files.

README.md
```diff
@@ -679,6 +679,17 @@ Offline Installations   | ✓     | x   | x   | x    |
   they could find ways to do an offline install for any package manager, but only `gomod` supports
   this out of the box (i.e. the user does not need to change their workflow).
 
+#### Limitations
+
+Cachito collects all dependencies of specified package manager types from a repository
+regardless of whether they are used by a certain application or not.
+For example, if some repository contains global `go.mod` file, dependencies from the file would be
+used for all container images built from the repository (see [gomod](#gomod) for more details).
```
brunoapimentel, line 687:
```suggestion
For example, if a repository contains a global `go.mod` file, all dependencies from this file will be present in the resulting archive and will also be cached to allow the future reproduction of the same build (see [gomod](#gomod) for more details).
```

brunoapimentel: I think this is more in line with what Cachito actually does, since the archive could be used for whatever reason (although we know it is not). WDYT?

serg-cymbaluk:
The limitation is not relevant for now. Closing the PR.

PR #543, opened 2021-08-05
Fix empty packages data bug

brunoapimentel:
This PR introduces two mechanisms to solve the bug of Cachito not properly loading the packages file for a request:

1. Adds a check that will not allow a request to be completed unless the worker is able to query the API and successfully compare the number of packages and dependencies. This will avoid propagating any Cachito failure to Brew builds.
2. Forces the NFS to clear it's cache by listing all the packages file's parent directory contents. This is the only workaround currently found to have the cache cleared.    

Diagram showing the solution:
image

CLOUDBLD-6599

### DIFF
--- a/cachito/common/packages_data.py
+++ b/cachito/common/packages_data.py
@@ -136,6 +136,22 @@ class PackagesData:
         with open(file_name, "w", encoding="utf-8") as f:
             json.dump({"packages": self._packages}, f)
 
+    def _clear_nfs_cache(self, file_name: Union[str, Path]) -> None:
+        """
+        Force the NFS to clear its cache by listing the directory's contents.
+
+        This solution is intended for the specific use case of the application being deployed
+        to a container environment that shares a single NFS volume. The NFS cache may cause
+        issues for a second container reading the file right after it got written.
+
+        :param file_name: an absolute or relative filename that contains the packages file.
+        """
+        dirname = os.path.dirname(file_name)
+
+        if os.path.exists(dirname):
+            log.info("Forcing the clearance of cache for %s directory", dirname)
+            os.listdir(dirname)
+
     def load(self, file_name: Union[str, Path]) -> None:
         """Load data from a specified file written by write_to_file method.
 
@@ -144,6 +160,8 @@ class PackagesData:
             ``os.curdir``. If the file does not exist, nothing is changed internally.
         :type file_name: str or pathlib.Path
         """
+        self._clear_nfs_cache(file_name)
+
         if not os.path.exists(file_name):
             log.warning("No data is loaded from non-existing file %s.", file_name)
             return
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -358,7 +358,8 @@ def create_request():
             tasks.fetch_yarn_source.si(request.id, yarn_package_configs).on_error(error_callback)
         )
 
-    chain_tasks.append(tasks.finalize_request.si(request.id).on_error(error_callback))
+    chain_tasks.append(tasks.process_fetched_sources.si(request.id).on_error(error_callback))
+    chain_tasks.append(tasks.finalize_request.s(request.id).on_error(error_callback))
 
     try:
         chain(chain_tasks).delay()
--- a/cachito/workers/tasks/general.py
+++ b/cachito/workers/tasks/general.py
@@ -29,6 +29,7 @@ __all__ = [
     "finalize_request",
     "get_request",
     "save_bundle_archive_checksum",
+    "process_fetched_sources",
 ]
 log = logging.getLogger(__name__)
 
@@ -169,13 +170,25 @@ def save_bundle_archive_checksum(request_id: int) -> None:
     bundle_dir.bundle_archive_checksum.write_text(checksum, encoding="utf-8")
 
 
-@app.task(priority=10)
+@app.task
 @runs_if_request_in_progress
-def finalize_request(request_id):
-    """Execute tasks to finalize the request creation."""
+def process_fetched_sources(request_id):
+    """Generate files for request and updates the request with packages/dependencies counts."""
     request = get_request(request_id)
     create_bundle_archive(request_id, request.get("flags", []))
     save_bundle_archive_checksum(request_id)
     data = aggregate_packages_data(request_id, request["pkg_managers"])
-    set_packages_and_deps_counts(request_id, len(data.packages), len(data.all_dependencies))
+
+    packages_count = len(data.packages)
+    dependencies_count = len(data.all_dependencies)
+
+    set_packages_and_deps_counts(request_id, packages_count, dependencies_count)
+
+    return packages_count, dependencies_count
+
+
+@app.task(priority=10)
+@runs_if_request_in_progress
+def finalize_request(counts, request_id):
+    """Execute tasks to finalize the request creation."""
     set_request_state(request_id, "complete", "Completed successfully")
--- a/cachito/workers/tasks/general.py
+++ b/cachito/workers/tasks/general.py
@@ -16,6 +16,7 @@ from cachito.workers.scm import Git
 from cachito.workers.tasks.celery import app
 from cachito.workers.tasks.utils import (
     get_request,
+    get_request_packages_and_dependencies,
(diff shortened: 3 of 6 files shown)


cachito/workers/tasks/general.py
```diff
@@ -171,11 +172,26 @@ def save_bundle_archive_checksum(request_id: int) -> None:
 
 @app.task(priority=10)
 @runs_if_request_in_progress
-def finalize_request(request_id):
-    """Execute tasks to finalize the request creation."""
+def process_fetched_sources(request_id):
```
chmeliik, line 176: I'm not sure about giving this priority 10, in https://github.com/release-engineering/cachito/commit/75f0b82639383449da19f93a3bb1ae5cee8f47be, the intention seems to have been to give the final task a higher priority then all the other tasks. I don't think it was necessary though, this is probably fine :thinking:

cachito/workers/tasks/general.py
```diff
@@ -29,7 +30,7 @@
     "finalize_request",
     "get_request",
     "save_bundle_archive_checksum",
-    "process_fetched_sources"
+    "process_fetched_sources",
```
chmeliik, line 33: Should go in the prev. commit

cachito/workers/tasks/general.py
```diff
@@ -173,10 +174,7 @@ def save_bundle_archive_checksum(request_id: int) -> None:
 @app.task(priority=10)
 @runs_if_request_in_progress
 def process_fetched_sources(request_id):
-    """
-    Writes the bundle archive, packages file and checksum file.
-    It also updates the request with the number of packages and dependencies fetched.
-    """
+    """Generate files for request and updates the request with packages/dependencies counts."""
```
chmeliik, line 177: This should probably go in the prev. commit too

cachito/workers/tasks/utils.py
```diff
@@ -181,6 +181,23 @@ def get_request_state(request_id):
...
+    Get the contents of the packages file from the Cachito API.
+
+    :param request_id: the Cachito request ID this is for
+    :raises CachitoError: if the connection fails or the API returns an error response
+    """
+    log.debug("Getting packages file for request %d", request_id)
+    request = _get_request_or_fail(
+        request_id,
+        connect_error_msg=f"Packages file could not be loaded for request {request_id}: {{exc}}",
+        status_error_msg=f"Packages file could not be loaded for request {request_id}: {{exc}}",
```
chmeliik, line 195: Could be nice to have different messages so that we can differentiate between 404/500 responses from the API vs. straight up connection errors more easily

cachito/common/packages_data.py
```diff
@@ -136,6 +136,21 @@ def write_to_file(self, file_name: Union[str, Path]) -> None:
...
 
+    def _clear_nfs_cache(self, file_name: Union[str, Path]) -> None:
+        """
+        Force the NFS to clear its cache by listing the directory's contents.
+
+        This solution is a workaround found to the issue of the file occasionally not
+        appearing to the pod that tried to read it, which only happened while using a NFS volume.
+
+        :param file_name: an absolute or relative filename that contains the packages file.
+        """
```
chmeliik, line 147: Some of this (NFS, pod) is specific to the internal deployment. It is valuable info though, probably fine to keep it here

tests/test_api_v1.py
```diff
@@ -33,8 +33,8 @@
     fetch_npm_source,
     fetch_pip_source,
     fetch_yarn_source,
-    process_fetched_sources,
     finalize_request,
+    process_fetched_sources,
```
chmeliik: Should go in the prev. commit

chmeliik (commented):
Looks good, just a small mess with changes being in a different commit than they should.

About splitting the finalize task, was that also needed to make things work? It's probably a harmless change, I'm just worried we won't know which of the 2 changes between this and the listdir were truly impactful

cachito/common/packages_data.py
brunoapimentel: I will make it more clear. I just thought it would be good to have an explation of why this cache clearing is needed, and this is interal deployment related, so I guess there's no avoiding it.

brunoapimentel:
> Looks good, just a small mess with changes being in a different commit than they should.
> 
> About splitting the finalize task, was that also needed to make things work? It's probably a harmless change, I'm just worried we won't know which of the 2 changes between this and the listdir were truly impactful

We don't need to split the tasks, it was more of a groundwork to have further checks on the finalize_task. I can remove it, if you prefer, then we can add it later as needed.

cachito/web/api_v1.py
```diff
@@ -358,7 +358,8 @@ def create_request():
             tasks.fetch_yarn_source.si(request.id, yarn_package_configs).on_error(error_callback)
         )
 
-    chain_tasks.append(tasks.finalize_request.si(request.id).on_error(error_callback))
+    chain_tasks.append(tasks.process_fetched_sources.si(request.id).on_error(error_callback))
+    chain_tasks.append(tasks.finalize_request.s(request.id).on_error(error_callback))
```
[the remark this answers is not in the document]
brunoapimentel: There is. `si` stands for `immutable signature` (or something of the sort). Any tasks created with `si` will not get the previous chained tasks results. That's why we needed a simple `s` (signature) here.

cachito/workers/tasks/general.py
brunoapimentel: Should we remove the priority on the new task then? I think this is not necessary, since we are using chained tasks, and they have a strict execution order.

chmeliik:
> We don't need to split the tasks, it was more of a groundwork to have further checks on the finalize_task. I can remove it, if you prefer, then we can add it later as needed.

If we know that the `listdir()` change is the one that matters for the caching, then I'm fine with keeping it (but I would change the commit message)

cachito/workers/tasks/general.py
chmeliik: As long as the first task in the chain is lower prio than everything else then it doesn't matter AFAIK. So I don't mind either way, maybe remove the priority and let it have the default 5 for consistency

cachito/web/api_v1.py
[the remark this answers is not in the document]
chmeliik: Oh my, you would think that this difference deserves more than a single letter's worth of difference in naming, TIL, thanks :D

brunoapimentel:
> > We don't need to split the tasks, it was more of a groundwork to have further checks on the finalize_task. I can remove it, if you prefer, then we can add it later as needed.
> 
> If we know that the `listdir()` change is the one that matters for the caching, then I'm fine with keeping it (but I would change the commit message)

We did test a branch that had the task split, and it did not solve the problem. (it had the retry in it, though)

Which commit message do you think need changing?

chmeliik:
> We did test a branch that had the task split, and it did not solve the problem. (it had the retry in it, though)

Ack, good enough for me.

> Which commit message do you think need changing?

The first one, since this was not the problem
```
The intention is to have a separate task that writes all necessary
files to disk and avoid eventual file flushes being held by the
task process.
```

cachito/common/packages_data.py
chmeliik: +1, thanks for updating it

cachito/workers/tasks/general.py
```diff
@@ -169,13 +170,25 @@ def save_bundle_archive_checksum(request_id: int) -> None:
     bundle_dir.bundle_archive_checksum.write_text(checksum, encoding="utf-8")
 
 
-@app.task(priority=10)
+@app.task(priority=5)
```
chmeliik, line 174: 5 is the default: https://github.com/release-engineering/cachito/blob/9a42d2b5b31b9644cd1d5963a4ec4e359cd9d2af/cachito/workers/config.py#L84-L85

It looks like no other task defines priority=5 explicitly, it may be good to keep it that way for consistency

chmeliik (approved):
LGTM with one minor detail (and one unresolved comment above).

cachito/workers/tasks/general.py
brunoapimentel: :man_facepalming:

nirzari (approved):

PR #545, opened 2021-08-05
Add more logs to the packages files check

brunoapimentel:
CLOUDBLD-6599

- Adds log that shows the packages file content every time it gets loaded
- Adds log that shows the integrity check that is made while a request is being finalized

### DIFF
--- a/cachito/common/packages_data.py
+++ b/cachito/common/packages_data.py
@@ -168,6 +168,7 @@ class PackagesData:
 
         with open(file_name, "r", encoding="utf-8") as f:
             data = json.load(f)
+            log.info("Loaded file %s: %s", file_name, data)
             packages = data.get("packages")
             if packages is None:
                 log.warning("Packages data file does not include key 'packages'.")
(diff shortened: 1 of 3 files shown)


nirzari (approved):

lcarva (approved):

cachito/common/packages_data.py
```diff
@@ -168,6 +168,7 @@ def load(self, file_name: Union[str, Path]) -> None:
 
         with open(file_name, "r", encoding="utf-8") as f:
             data = json.load(f)
+            log.info("Loaded file %s: %s", file_name, data)
```
chmeliik, line 171: This will often be a huge log message and will appear in the /logs for a request. Not sure if we can live without it though.

brunoapimentel: I'm desperate. :sweat_smile: 
We could mark this to be removed once we believe everything is stable.

chmeliik: Sounds good to me

chmeliik (approved):

PR #546, opened 2021-08-05
Fix packages_data module logging

brunoapimentel:
This config makes the logs from the `cachito/common/packages_data.py` module appear when it is being used by Flask.

Note that other modules within this package will need the same configuration if we add logs to them in the future.

### DIFF
--- a/cachito/common/paths.py
+++ b/cachito/common/paths.py
@@ -1,12 +1,9 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
-import logging
 import os
 import shutil
 from pathlib import Path
 from typing import Any
-
-log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


tkdchen (approved):

chmeliik (approved):

brunoapimentel:
Lui commented this could be done via configuration. But I'm thinking: this should be default behavior for the application, right? The logging of this module should appear by default.

WDYT? @chmeliik @tkdchen

chmeliik:
> Lui commented this could be done via configuration. But I'm thinking: this should be default behavior for the application, right? The logging of this module should appear by default.
> 
> WDYT? @chmeliik @tkdchen

The `packages_data` module? IMO yes, maybe all the `cachito.common.*` modules

PR #547, opened 2021-08-09
Request state single source of truth in json generation method

brunoapimentel:
The `to_json` method on the Request model class currently uses two different relationships to determine a request's current state:

- state: based on a foreign key placed on the `request` table.
- states: based on a foreign key placed on the `request_state` table.

This behavior seems to cause a bug when the request is patched to the `complete` state while the `to_json` method is being processed. Each relationship attribute returns a different state, making the packages file to not be loaded and the request JSON appear with a `complete` state:

image

The fix currently only affects the "verbose" JSON generation, so we should probably remove this dual relationship later on.

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -340,8 +340,9 @@ class Request(db.Model):
         return content_manifest.ContentManifest(self, packages)
 
     def _is_complete(self):
-        if self.state:
-            return self.state.state_name == RequestStateMapping.complete.name
+        if len(self.states) > 0:
+            latest_state = self.states[-1]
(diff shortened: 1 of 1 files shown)


chmeliik (approved):
:tada: :tada: :tada: 

SQLAlchemy is evil. Or whoever added the `Request.state` attribute. Possibly both.

nirzari (approved):

nirzari:
>   
> 
> SQLAlchemy is evil. Or whoever added the `Request.state` attribute. Possibly both.

Not even going through the commit history but I bet it was Matt :smiling_imp:  :stuck_out_tongue_closed_eyes:

PR #550, opened 2021-08-12
Upgrade RabbitMQ

dcho5:
CLOUDBLD-3372

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,10 +9,9 @@ services:
       POSTGRES_INITDB_ARGS: "--auth='ident' --auth='trust'"
 
   rabbitmq:
-    image: rabbitmq:3.7-management
-    environment:
-      RABBITMQ_DEFAULT_USER: cachito
-      RABBITMQ_DEFAULT_PASS: cachito
+    image: rabbitmq:3.9-management
+    volumes:
+      - ./docker/rabbitmq.conf:/etc/rabbitmq/rabbitmq.conf:z
     ports:
       # The RabbitMQ management console
       - 8081:15672
--- a/docker/rabbitmq.conf
+++ b/docker/rabbitmq.conf
@@ -1,0 +1,3 @@
+# Only intended to be used in the local dev environment (docker-compose)
+default_user = cachito
+default_pass = cachito


dcho5:
RabbitMQ version 3.8.20 was the latest version Cachito wanted to work with. Cachito refused to run when I tried upgrading RabbitMQ to 3.9.X. What's the reason for this? According to https://www.rabbitmq.com/changelog.html, it appears we're using an old Earlang version for Cachito, is that correct? I ask because of Chen's [comment](https://issues.redhat.com/browse/CLOUDBLD-3372?focusedCommentId=17150235&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17150235).

docker-compose.yml
```diff
@@ -9,7 +9,7 @@ services:
       POSTGRES_INITDB_ARGS: "--auth='ident' --auth='trust'"
 
   rabbitmq:
-    image: rabbitmq:3.7-management
+    image: rabbitmq:3.8.20-management
```
akhmanova, line 12: > RabbitMQ version 3.8.20 was the latest version Cachito wanted to work with. Cachito refused to run when I tried upgrading RabbitMQ to 3.9.X. What's the reason for this? According to https://www.rabbitmq.com/changelog.html, it appears we're using an old Earlang version for Cachito, is that correct? I ask because of Chen's comment.

@dcho5 Team RabbitMQ produces a [package](https://github.com/rabbitmq/erlang-rpm) stripped down to only provide those components needed to run RabbitMQ. This is the recommended option for Fedora OS [https://www.rabbitmq.com/install-rpm.html].
@tkdchen @brunoapimentel @MartinBasti  WDYT?

chmeliik:
> RabbitMQ version 3.8.20 was the latest version Cachito wanted to work with. Cachito refused to run when I tried upgrading RabbitMQ to 3.9.X. What's the reason for this? According to https://www.rabbitmq.com/changelog.html, it appears we're using an old Earlang version for Cachito, is that correct? I ask because of Chen's [comment](https://issues.redhat.com/browse/CLOUDBLD-3372?focusedCommentId=17150235&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17150235).

D'oh. I though I this issue was about upgrading to 3.8.x, I forgot we agreed to go for 3.9.x. I think the problem is that we are using environment variables to configure rabbitmq. This was already deprecated in 3.8 and removed in 3.9. When you check the rabbitmq logs, you should see something like this at the beginning:
```
WARNING: 'docker-entrypoint.sh' generated/modified the RabbitMQ configuration file, which will no longer happen in 3.9 and later! (https://github.com/docker-library/rabbitmq/pull/424)

Generated end result, for reference:
------------------------------------
loopback_users.guest = false
listeners.tcp.default = 5672
default_pass = cachito
default_user = cachito
management.tcp.port = 15672
------------------------------------
```
We may need to somehow switch to using the configuration file instead.

chmeliik:
> We may need to somehow switch to using the configuration file instead.

@dcho5 there is some relevant info in https://hub.docker.com/_/rabbitmq

PR #551, opened 2021-08-18
Promethus Flask Exporter.  Group by endpoint (function).

mike-kingsbury:
This is largely the same MR as before.   WIth this change deployed, Prometheus will poll the /metrics endpoint, and get metrics from 1 random pod.  Over time, Prometheus will get the data from all the pods.  Labels in the metrics based on the hostname will allow Prometheus to sort out the data, and the sum() function in Prometheus will allow us to aggregate the counters.  

### DIFF
--- a/cachito/web/config.py
+++ b/cachito/web/config.py
@@ -26,6 +26,14 @@ class Config(object):
     CACHITO_USER_REPRESENTATIVES: List[str] = []
     CACHITO_WORKER_USERNAMES: List[str] = []
 
+    # Temp dir used by the Prometheus Flask Exporter to coalesce the metrics from the threads
+    if "PROMETHEUS_MULTIPROC_DIR" not in os.environ.keys():
+        raise ConfigError(
+            "The environment variable PROMETHEUS_MULTIPROC_DIR must be set prior to execution"
+        )
+    else:
+        PROMETHEUS_METRICS_TEMP_DIR = os.environ["PROMETHEUS_MULTIPROC_DIR"]
+
 
 class ProductionConfig(Config):
     """The production Cachito Flask configuration."""
@@ -89,6 +97,7 @@ def validate_cachito_config(config, cli=False):
         "CACHITO_LOG_FORMAT",
         "CACHITO_BUNDLES_DIR",
         "SQLALCHEMY_DATABASE_URI",
+        "PROMETHEUS_METRICS_TEMP_DIR",
     ):
         if config_var == "CACHITO_BUNDLES_DIR":
             if cli:
--- a/cachito/web/metrics.py
+++ b/cachito/web/metrics.py
@@ -1,0 +1,26 @@
+import os
+import socket
+
+from prometheus_client import multiprocess
+from prometheus_client.core import CollectorRegistry
+from prometheus_flask_exporter.multiprocess import GunicornInternalPrometheusMetrics
+
+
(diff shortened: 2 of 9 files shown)


cachito/web/config.py
```diff
@@ -26,6 +26,13 @@ class Config(object):
     CACHITO_USER_REPRESENTATIVES: List[str] = []
     CACHITO_WORKER_USERNAMES: List[str] = []
 
+    # Temp dir used by the Prometheus Flask Exporter to coalesce the metrics from the threads
+    if "PROMETHEUS_MULTIPROC_DIR" not in os.environ.keys():
+        PROMETHEUS_METRICS_TEMP_DIR = tempfile.gettempdir()
+        os.environ["PROMETHEUS_MULTIPROC_DIR"] = PROMETHEUS_METRICS_TEMP_DIR
```
chmeliik, line 32: https://github.com/prometheus/client_python has some more info about this environment variable:
> The `PROMETHEUS_MULTIPROC_DIR` environment variable must be set to a directory that the client library can use for metrics. This directory must be wiped between process/Gunicorn runs (before startup is recommended).
>
> This environment variable should be set from a start-up shell script, and not directly from Python (otherwise it may not propagate to child processes).

I'm not sure if the "not propagated to child processes" issue could affect us, but setting it using an external mechanism rather than Python does seem cleaner.

I would rather not set the variable here, instead just check that it is already set. We should probably set it in docker-compose for the local deployment, or we could have a default directly in the Dockerfile.

cachito/web/metrics.py
```diff
@@ -0,0 +1,28 @@
+import os
+import socket
+
+from prometheus_client import multiprocess
+from prometheus_client.core import CollectorRegistry
+from prometheus_flask_exporter.multiprocess import GunicornInternalPrometheusMetrics
+
+hostname = socket.gethostname()
```
chmeliik, line 8: Could this be called inside the `init_metrics()` function? It probably doesn't matter much, just feels cleaner.

tox.ini
```diff
@@ -11,16 +11,24 @@ deps =
     -rrequirements-test.txt
 whitelist_externals =
     make
+    mkdir
+    rm
 passenv = TOX_ENV_DIR
 setenv =
 	CACHITO_TESTING=true
+    PROMETHEUS_MULTIPROC_TEMP_DIR={envtmpdir}/prometheus_metrics
```
chmeliik, line 19: This should probably be `PROMETHEUS_MULTIPROC_DIR`. And the indentation does not seem to match.

chmeliik (commented):
@mike-kingsbury nice! Good job figuring this out :tada: 

Some minor details, looks good as a whole.

brunoapimentel:
LGMT :rocket:

cachito/web/metrics.py
mike-kingsbury: Moved.  I was considering it a constant that I'd never change, hence putting it up top.  On the flip side, its only used in that one place.

tox.ini
mike-kingsbury: Yes, with the other changes it does have to change.  Indents was a tabs vs spaces, fixed.

chmeliik (approved):
LGTM :+1:

ejegrova (approved):
LGTM

PR #552, opened 2021-08-18
Fixes gomod cached dependencies test to work with Go 1.16

brunoapimentel:
The test used to remove a dependency which was still being referenced in the test repo code, and this causes it to break when using Go 1.16.

See: https://blog.golang.org/go116-module-changes#TOC_3.

### DIFF
--- a/tests/integration/test_data/cached_dependencies.yaml
+++ b/tests/integration/test_data/cached_dependencies.yaml
@@ -161,6 +161,18 @@ gomod_cached_deps:
         replaces: null
         type: "gomod"
         version: "DEP_VERSION"
+      - name: "golang.org/x/text"
+        replaces: null
+        type: "gomod"
+        version: "v0.0.0-20170915032832-14c0d48ead0c"
+      - name: "rsc.io/quote"
+        replaces: null
+        type: "gomod"
+        version: "v1.5.2"
+      - name: "rsc.io/sampler"
+        replaces: null
+        type: "gomod"
+        version: "v1.3.0"
       name: "github.com/cachito-testing/cachito-gomod-with-deps"
       type: "gomod"
       version: "MAIN_VERSION"
@@ -185,6 +197,18 @@ gomod_cached_deps:
       replaces: null
       type: "gomod"
       version: "DEP_VERSION"
+    - name: "golang.org/x/text"
+      replaces: null
+      type: "gomod"
+      version: "v0.0.0-20170915032832-14c0d48ead0c"
+    - name: "rsc.io/quote"
+      replaces: null
+      type: "gomod"
+      version: "v1.5.2"
+    - name: "rsc.io/sampler"
+      replaces: null
+      type: "gomod"
+      version: "v1.3.0"
   # Expected files
   expected_files: []
   # Expected content manifest data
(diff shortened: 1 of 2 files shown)


tests/integration/test_data/cached_dependencies.yaml
```diff
@@ -185,6 +197,18 @@ gomod_cached_deps:
...
+      type: "gomod"
+      version: "v0.0.0-20170915032832-14c0d48ead0c"
+    - name: "rsc.io/quote"
+      replaces: null
+      type: "gomod"
+      version: "v1.5.2"
+    - name: "rsc.io/sampler"
+      replaces: null
+      type: "gomod"
+      version: "v1.3.0"
```
brunoapimentel, line 211: I don't really understand why these show up on both `go-package` and `gomod` parts...

brunoapimentel:
I guess it would be useful to point to the Go change that actually caused this failure, but I was not able to accurately point it. :sweat_smile:

tests/integration/test_data/cached_dependencies.yaml
chmeliik: See if this explains it: https://github.com/release-engineering/cachito/#go-package-level-dependencies-and-the-go-package-cachito-package-type

The most important part is probably this
> A directory that contains both `go.mod` and `.go` files is both a module and a package.

brunoapimentel:
> I guess it would be useful to point to the Go change that actually caused this failure, but I was not able to accurately point it.

Nevermind, I'm stupid. Updating the description.

tests/integration/test_data/cached_dependencies.yaml
chmeliik: But I guess the question is why they didn't show up before, huh?

ejegrova (approved):
LGTM

tests/integration/test_data/cached_dependencies.yaml
brunoapimentel: I'm thinking it's because the commit would remove the `rsc.io` dep, so these would not show in the resulting packages.

chmeliik: Yeah, probably. It's still a bit strange though, if go 1.15 fixed the missing `require` automatically, I would expect the modules to be properly listed. If that was not the case, we could have been reporting wrong data about the dependencies that Cachito provided. At least we would have listed the packages, only the modules would be missing.

It's good to know that go 1.16 fixes that. It may also break builds that used to work, but I wouldn't worry about that since we make the audacious assumption that users will get their package in order before trying to process it with Cachito.

tests/integration/test_using_cached_dependencies.py
```diff
@@ -458,9 +457,7 @@ def update_main_repo(env_data, repo_dir, tmpdir, new_dep_commits, dep_repo):
         dep_version = get_pseudo_version(dep_repo, new_dep_commits[0])
 
         with open(os.path.join(repo_dir, "go.mod"), "r") as f:
-            base_gomod_file = [
-                line for line in f.read().split("\n") if not line.startswith("require")
-            ]
+            base_gomod_file = [line for line in f.read().split("\n")]
         with open(os.path.join(repo_dir, "go.mod"), "w+") as f:
             for line in base_gomod_file:
                 f.write(f"{line}\n")
```
chmeliik, line 463: If you want to simplify this, we just need to open the file once for appending and write a single line to it

chmeliik (approved):
Fix LGTM, just one suggestion to simplify the code

akhmanova (approved):

tests/integration/test_using_cached_dependencies.py
brunoapimentel: FTFY

chmeliik (approved):

akhmanova (approved):

PR #553, opened 2021-08-18
Allows NPM and Yarn to replace optional dependencies

brunoapimentel:
CLOUDBLD-3917


tests/test_workers/test_pkg_managers/test_yarn.py
```diff
@@ -28,6 +28,7 @@
     "version": "1.0.0",
     "dependencies": {"chai": "^4.2.0", "fecha": HTTP_DEP_URL},
     "devDependencies": {"leftpad": GIT_DEP_URL},
+    "optionalDependencies": {"get-func-name": "^2.0.0"},
```
chmeliik, line 31: Minor thing: A dependency specified by a normal version would never be replaced in package.json. Only external dependencies get replaced - https deps, git deps etc. Doesn't really matter for unit test data, but it might be less confusing if you use a URL instead of a version.

chmeliik (commented):
I originally believed `peerDependencies` didn't matter, but based on @tkdchen's investigation of npm v7, they might matter now. I don't actually know if it is possible to specify an external peer dependency, but it shouldn't hurt to include them in the to-be-replaced list as well.

chmeliik (approved):
LGTM :+1:

tkdchen (approved):

PR #555, opened 2021-08-19
Pass download dir to download_dependencies directly

tkdchen:
Function download_dependencies is used by npm and yarn resolve tasks
only. It should make no sense to pass a request id and construct the
download directory by the function itself.

Argument pkg_manager was used for creating npm or yarn specific download
dir and validating its value to ensure function is called correctly by
npm or yarn task. In fact, the ValueError due to an invalid pkg_manager
will never be raised at the runtime, instead it should be a programming
error if happens. Therefore, the corresponding change is to use an
assertion to ensure the correctness of the passed-in value.

### DIFF
--- a/cachito/workers/pkg_managers/general_js.py
+++ b/cachito/workers/pkg_managers/general_js.py
@@ -12,13 +12,13 @@ import tarfile
 import tempfile
 import textwrap
 from dataclasses import dataclass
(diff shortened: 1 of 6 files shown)


serg-cymbaluk (approved):
LGTM

PR #557, opened 2021-08-24
Add ignore error flag for the go mod list command

brunoapimentel:
Due to some behavioral changes introduced in Go 1.16, the `go mod list -deps` command would give errors in some specific requests. 

This flag is meant to allow the use of Go 1.16 without breaking the builds that currently pass.

### DIFF
--- a/cachito/workers/pkg_managers/gomod.py
+++ b/cachito/workers/pkg_managers/gomod.py
@@ -256,7 +256,7 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
 
         log.info("Retrieving the list of package level dependencies")
         package_info = _load_list_deps(
-            run_gomod_cmd(["go", "list", "-deps", "-json", "./..."], run_params)
+            run_gomod_cmd(["go", "list", "-e", "-deps", "-json", "./..."], run_params)
         )
 
         packages = []


chmeliik (approved):
LGTM :+1:

Note that, AFAICT, adding this flag makes this command behave the same way with go 1.16 as it did with go 1.15. Here's the description of `-e` from `go help list` (identical for both versions):
> The -e flag changes the handling of erroneous packages, those that
cannot be found or are malformed. By default, the list command
prints an error to standard error for each erroneous package and
omits the packages from consideration during the usual printing.
With the -e flag, the list command never prints errors to standard
error and instead processes the erroneous packages with the usual
printing. Erroneous packages will have a non-empty ImportPath and
a non-nil Error field; other information may or may not be missing
(zeroed).

This description is wrong both versions:
* By default, the list command prints an error to standard error for each erroneous package
  * true for 1.16, false for 1.15
* omits the packages from consideration during the usual printing
  * not true for either version
* Not mentioned: if there are errors, the command will return a non-zero code
  * true for 1.16, false for 1.15

Basically, go 1.15 always behaves as if `-e` was used, 1.16 only if you actually use it. Couldn't find anything about this in the changelog, probably was a bugfix, maybe an accidental one.

ejegrova (approved):
LGTM

PR #558, opened 2021-08-26
Use exponential backoff for gomod retries

chmeliik:
The go commands download dependencies into a temporary cache directory.
The directory is reused between retries, which means that, when retried,
the command can just pick up where it left off. Three attempts with
constant backoff might not be enough to give Athens a chance to recover.
Increase the default max tries to 5 and use exponential backoff instead.

By default, Cachito will wait for 1s -> 2s -> 4s -> 8s before retrying.


ejegrova (approved):
LGTM

PR #559, opened 2021-08-30
Sort Cachito /request endpoint by date

dcho5:
CLOUDBLD-6469

Create "created" variable to represent created date of Cachito
request. Sort requests by the "created" variable, not "updated."


cachito/web/models.py
```diff
@@ -633,6 +634,7 @@ class RequestState(db.Model):
     id = db.Column(db.Integer, primary_key=True)
     state = db.Column(db.Integer, nullable=False)
     state_reason = db.Column(db.String, nullable=False)
+    created = db.Column(db.DateTime(), nullable=False, default=sqlalchemy.func.now())
     updated = db.Column(db.DateTime(), nullable=False, default=sqlalchemy.func.now())
```
chmeliik, line 638: Requests have a list of states. You don't need an extra `created` attribute, the creation time of a request is the `updated` time of the earliest state in that list.

serg-cymbaluk:
@dcho5 @chmeliik 
I would suggest here to add the `created` field to the `Request` table.
Other variants, like getting first state update, would cost us a performance reduce.
If there is needed some integrity with state `updated` field, the field could be filled in one transaction with the first state `insert`.

Then once we created a new field, we need to take care about its content. So some data migration is expected in the PR as well.

brunoapimentel:
> @dcho5 @chmeliik
> I would suggest here to add the `created` field to the `Request` table.
> Other variants, like getting first state update, would cost us a performance reduce.
> If there is needed some integrity with the state `updated` field, the `created` field could be set in one transaction with the first state `insert`.
> 
> Then once we created a new field, we need to take care about its content. So some data migration is expected in the PR as well.

@dcho5 I think it's best to follow Serhii's suggestion here. This will allow a simpler and more efficient solution. I believe it's easier to do this in separare PRs: one to add the new field and perform all the migrations, and then one to implement the actual filter and tests.

cachito/web/models.py
```diff
@@ -286,6 +286,7 @@ class Request(db.Model):
     """A Cachito user request."""
 
     id = db.Column(db.Integer, primary_key=True)
+    created = db.Column(db.DateTime(), nullable=False, default=sqlalchemy.func.now())
```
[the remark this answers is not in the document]
dcho5: Done!

cachito/web/migrations/versions/e2b2cedabc58_add_created_field.py
```diff
@@ -0,0 +1,26 @@
...
+# revision identifiers, used by Alembic.
+revision = "e2b2cedabc58"
+down_revision = "491454f79a8b"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table("request", schema=None) as batch_op:
+        batch_op.add_column(sa.Column("created", sa.Date(), nullable=True))
```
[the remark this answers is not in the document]
brunoapimentel: I'm guessing the migration was not generated again after the `index` attribute was added to the Request model.

Daniel, I believe the best way is to revert the code to the current master branch and do a `make run-rebuild`. Then, you can put back the changes you did at `models.py` and generate this migration again.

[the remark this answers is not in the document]
dcho5: Sounds good!

cachito/web/models.py
```diff
@@ -286,6 +287,7 @@ class Request(db.Model):
     """A Cachito user request."""
 
     id = db.Column(db.Integer, primary_key=True)
+    created = db.Column(db.Date(), nullable=False, index=True, default=date.today())
```
brunoapimentel, line 290: What was the reason to change this field from `datetime` to `date`? 

I believe this could be more useful as a datetime field, since we could use it for other purposes besides the date filter, and then having the exact creation timestamp would help.

dcho5: Here is the email that led me to this decision:

I believe the latter, all datetime values truncated to their date for the purpose of the comparison would be the most appropriate answer here.  Trying to make a given value more precise for comparison purposes I don't think is good behavior.  

-mike


On Wed, Sep 29, 2021 at 1:02 AM Daniel Cho <dacho@redhat.com> wrote:
- CLOUDBLD-6469
I have essentially completed this ticket. The first PR is ready to be approved and merged. The second PR will be ready to be approved and merged as soon as the first one is merged.

There is one question I have for the team, especially Michael:
For an example, the code I have implemented takes this - /api/v1/requests?start_date=2021-01-01&end_date=2021-01-01 - and treats the input not as 2021-01-01 but 2021-01-01T00:00:00.000000 instead. In other words, if the user intends to filter all Cachito requests within a single day, the end_date parameter would have to act as the exclusive upper limit (i.e. /api/v1/requests?start_date=2021-01-01&end_date=2021-01-02 for requests created on 1/1/2021).

The question is, should the code remain this way, or should the parameters be redesigned to only filter to the nearest date (i.e. all created date values would be truncated to their date for the purpose of this functionality)? The former would provide more control over the parameters, but the latter may be much easier to use.

brunoapimentel: I see. Maybe we could handle the same behavior while keeping the field as `datetime`?

For example, we could set the `start_date` parameter to always behave as if had the `00:00:00.00000` timestamp and the `end_date` if it had the `23.59:59.99999` timestamp.

Wdyt?

chmeliik: +1. Also, can the params be called `before` and `after` instead of `start_date`, `end_date`? Currently it sounds like the query should return only requests that started on day X and ended on day Y.

MartinBasti: yes please keep datetime

dcho5: I changed all of the `date` objects back to `datetime` objects. I've adjusted the `start_date` and `end_date` parameters just as you suggested, @brunoapimentel  (https://github.com/release-engineering/cachito/pull/570).

As for the name of the variables, @chmeliik, I agree with what you said. However, I want to suggest a new idea - `from_date` and `to_date`. I think this makes more sense, especially in the case where the two variables are equal to each other. I think that there is merit in keeping the word "date" in the name, since these variables represent `date`, not `datetime`. What do you think?

chmeliik: `from_date`, `to_date` works for me.

Do the variables need to represent only `date` though? Could the endpoint accept both e.g. `/requests?from_date=2021-10-04` and `/requests?from_date=2021-10-04T20:00` and behave as expected? It may not even need extra logic in code, SQLAlchemy should be pretty flexible with datetime comparisons.

dcho5: I've changed the variables to `from_date` and `to_date`. I also added the `datetime` functionality for the parameters as well. You can check it out in this PR https://github.com/release-engineering/cachito/pull/570

brunoapimentel (approved):
LGTM :rocket:

cachito/web/models.py
```diff
@@ -376,6 +377,7 @@ def to_json(self, verbose=True):
             env_vars_json[env_var.name] = env_var.value
         rv = {
             "id": self.id,
+            "created": self.created.isoformat(),
```
MartinBasti, line 380: What is timezone here? is that UTC?

cachito/web/models.py
```diff
@@ -286,6 +286,7 @@ class Request(db.Model):
     """A Cachito user request."""
 
     id = db.Column(db.Integer, primary_key=True)
+    created = db.Column(db.DateTime(), nullable=False, index=True, default=sqlalchemy.func.now())
```
MartinBasti, line 289: Can you tell me which postgresql datatype is this, pls?

brunoapimentel: Timestamp without timezone:
```
                                         Table "public.request"
       Column       |            Type             |                      Modifiers                       
--------------------+-----------------------------+------------------------------------------------------
 id                 | integer                     | not null default nextval('request_id_seq'::regclass)
 repo               | character varying           | not null
 ref                | character varying           | not null
 user_id            | integer                     | 
 submitted_by_id    | integer                     | 
 request_state_id   | integer                     | 
 packages_count     | integer                     | 
 dependencies_count | integer                     | 
 created            | timestamp without time zone | 
Indexes:
    "request_pkey" PRIMARY KEY, btree (id)
    "ix_request_request_state_id" UNIQUE, btree (request_state_id)
    "ix_request_created" btree (created)
    "ix_request_ref" btree (ref)
    "ix_request_repo" btree (repo)
```

cachito/web/models.py
MartinBasti: isn't this deployment detail, that shouldn't be even considered upstream?

What if I change pod timezone, or run this as VM?

brunoapimentel: I think I didn't get your point here :sweat_smile:. Are you suggesting we pin the timezone in the application level?

MartinBasti:
```python
>>>datetime.datetime.utcnow().isoformat()                                                                                                                                                                                                 
'2021-10-06T17:17:40.127838'
```
it provides UTC time whatever is local timezone

MartinBasti: as I was informed, cachito works with local timezone settings on other places too, so moving this to UTC here, would cause more problems than solutions

MartinBasti (approved):

PR #561, opened 2021-09-02
Define safe defaults

MartinBasti:
Keep safe default values in configuration explicitly.

### DIFF
--- a/cachito/web/config.py
+++ b/cachito/web/config.py
@@ -11,6 +11,7 @@ TEST_DB_FILE = os.path.join(os.environ.get("TOX_ENV_DIR") or tempfile.gettempdir
 class Config(object):
     """The base Cachito Flask configuration."""
 
+    DEBUG = False
     # Additional loggers to set to the level defined in CACHITO_LOG_LEVEL
     CACHITO_ADDITIONAL_LOGGERS: List[str] = ["cachito.common.packages_data"]
     CACHITO_DEFAULT_PACKAGE_MANAGERS: List[str] = ["gomod"]
@@ -25,6 +26,8 @@ class Config(object):
     # Users that are allowed to use the "user" property when creating a request
     CACHITO_USER_REPRESENTATIVES: List[str] = []
     CACHITO_WORKER_USERNAMES: List[str] = []
+    LOGIN_DISABLED = False
+    TESTING = False
 
     # Temp dir used by the Prometheus Flask Exporter to coalesce the metrics from the threads
     if "PROMETHEUS_MULTIPROC_DIR" not in os.environ.keys():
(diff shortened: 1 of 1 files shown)


cachito/web/config.py
```diff
@@ -11,6 +11,7 @@
 class Config(object):
     """The base Cachito Flask configuration."""
 
+    DEBUG = False
```
tkdchen, line 14: We may need to set `DEBUG` to True in `DevelopmentConfig` accordingly.

brunoapimentel: And we can remove this config from the `ProductionConfig class`, since it became redundant. (L44)

cachito/web/config.py
```diff
@@ -25,6 +26,8 @@ class Config(object):
     # Users that are allowed to use the "user" property when creating a request
     CACHITO_USER_REPRESENTATIVES: List[str] = []
     CACHITO_WORKER_USERNAMES: List[str] = []
+    LOGIN_DISABLED = False
```
brunoapimentel, line 29: We can remove this config from the TestConfig class, since it became redundant. (L71)

brunoapimentel:
We also have another config file for the workers, and the default authentication is "None"

MartinBasti:
Duplications removed

MartinBasti:
> We also have another config file for the workers, and the default authentication is "None". Do you think we should default it to "kerberos" (prod config)?
> 
> https://github.com/release-engineering/cachito/blob/8f9a9bbfa43245296a6cec6b65e17d848fdec9ea/cachito/workers/config.py#L24

I'd keep it as is.

brunoapimentel (approved):

cachito/web/config.py
```diff
@@ -64,7 +68,6 @@ class TestingConfig(DevelopmentConfig):
     # after the migration completes...
     #   https://github.com/miguelgrinberg/Flask-Migrate/issues/153
     SQLALCHEMY_DATABASE_URI = f"sqlite:///{TEST_DB_FILE}"
-    DEBUG = True
     LOGIN_DISABLED = False
```
ejegrova, line 71: I believe we wanted to remove line with `LOGIN_DISABLED = False` because it's in the `Config` and keep line `DEBUG = True`.

brunoapimentel: That's true, the `LOGIN_DISABLED` should be removed. The `DEBUG` was kept, though, he just moved it to the top.

MartinBasti: When I removed it, tests started failing

MartinBasti: > LOGIN_DISABLED = False because it's in the Config

But testing config inherits from `DevelopmentConfig` that requires `LOGIN_DISABLED = True`

ejegrova: > > LOGIN_DISABLED = False because it's in the Config
> 
> But testing config inherits from `DevelopmentConfig` that requires `LOGIN_DISABLED = True`

My bad, I haven't noticed that.

ejegrova (approved):

PR #562, opened 2021-09-03
Metric for requests in progress. First custom prometheus metric

mike-kingsbury:
Simple metric for requests in progress.  A sum() function in prometheus will fix the accounting issue of one API node  incrementing for a particular request and another node doing a decrementing as it moves between states.

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -19,6 +19,7 @@ from cachito.common.utils import b64encode
 from cachito.errors import CachitoError, ValidationError
 from cachito.web import db
 from cachito.web.content_manifest import BASE_ICM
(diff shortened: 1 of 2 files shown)


akhmanova:
I am not sure I know what kind of problem is insuge Gating / Unit tests. Otherwise looks good to me.

mike-kingsbury:
Its a fixturing issue.  I found some info on it, hope to solve it today.

brunoapimentel:
I had a quick discussion with Mike regarding this PR, and we talked about having a dedicated endpoint to get the number of in_progress requests in a format that is readable by Prometheus. This way, we'd avoid to have the logic to count the requests within the create/update request endpoints.

@mike-kingsbury do you think we could go ahead with this idea? :)

PR #564, opened 2021-09-07
Improve content manifest endpoint performance

tkdchen:
All the performance measurement is based on a request's packages data generated from:

* Repository: https://github.com/stackrox/rox.git
* Reference: 73de8f2f5f4490ab303e892c11e2cedf008b7ff5
* Package manager: gomod

Some basic data:

* 317 packages and 50245 dependencies included in the packages data.
* Each of them is a Python dict object.
* The file size of the packages data is 5.4M
* 1341300 deep_sort_icm function calls.

Every commit in this PR as detail information about each improvement.

As a result, based on the packages data,

* Less memory is used to generate the `content_manifest.Package` objects.
* Less memory is used by the function `deep_sort_icm`.
* Lots of function calls, e.g. `deep_sort_icm` and `isinstance`, are avoided.

Overall, about 52.6 MiB memory is saved in the deep_sort_icm function call inside generate_icm, from 133.0 MiB down to 80.4 MiB.

There is still room to improve the memory usage and even make the content manifest generation much faster.

One major thought is to remove the `deep_sort_icm` function and sort the `dependencies` and `sources` just after they are collected, and sort the final `image_contents` at once. It looks like we don't need a such general function to sort a potential huge dict object containing the content manifest, as the schema of the content manifest is known and certainty. It is probably good to implement it along with the refactor of the content manifest generation.

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -10,6 +10,7 @@ from celery import chain
 from flask import stream_with_context
 from flask_login import current_user, login_required
 from sqlalchemy import func
+from sqlalchemy.orm import joinedload, load_only
 from werkzeug.exceptions import BadRequest, Forbidden, Gone, InternalServerError, NotFound
 
 from cachito.common.checksum import hash_file
@@ -663,32 +664,34 @@ def get_content_manifest_by_requests():
     arg = flask.request.args.get("requests")
     if not arg:
         return flask.jsonify(BASE_ICM)
-    request_ids = []
+    request_ids = set()
     item: str
     for item in arg.split(","):
         if not item.strip():
             continue
         if not item.strip().isdigit():
             raise BadRequest(f"{item} is not an integer.")
-        request_ids.append(int(item))
+        request_ids.add(int(item))
 
-    requests = Request.query.filter(Request.id.in_(request_ids))
-    found_requests_ids = [request.id for request in requests]
-
-    ids_diff = list(map(str, set(request_ids) - set(found_requests_ids)))
-    if ids_diff:
-        raise BadRequest(f"Cannot find request(s) {', '.join(ids_diff)}.")
-
-    assembled_icm = deepcopy(BASE_ICM)
+    requests = Request.query.filter(Request.id.in_(request_ids)).options(
+        load_only("id"), joinedload(Request.state)
+    )
+    states = (RequestStateMapping.complete.name, RequestStateMapping.stale.name)
     request: Request
     for request in requests:
-        if request.state.state_name not in (
-            RequestStateMapping.complete.name,
-            RequestStateMapping.stale.name,
-        ):
+        if request.state.state_name not in states:
             raise BadRequest(
                 f"Request {request.id} is in state {request.state.state_name}, "
                 f"not complete or stale."
             )
-        assembled_icm["image_contents"].extend(request.content_manifest.to_json()["image_contents"])
+        request_ids.remove(request.id)
+
+    if request_ids:
+        nonexistent_ids = ",".join(map(str, request_ids))
+        raise BadRequest(f"Cannot find request(s) {nonexistent_ids}.")
+
+    assembled_icm = deepcopy(BASE_ICM)
+    for request in requests:
+        manifest = request.content_manifest.to_json()
+        assembled_icm["image_contents"].extend(manifest["image_contents"])
     return flask.jsonify(deep_sort_icm(assembled_icm))
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -706,8 +706,6 @@ def get_content_manifest_by_requests():
     for request in requests:
         manifest = request.content_manifest.to_json()
         assembled_icm["image_contents"].extend(manifest["image_contents"])
-    if len(requests) == 1:
-        final_icm = assembled_icm
-    else:
-        final_icm = deep_sort_icm(assembled_icm)
-    return create_jsonify_response(final_icm)
+    if len(requests) > 1:
+        deep_sort_icm(assembled_icm)
+    return create_jsonify_response(assembled_icm)
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -683,8 +683,10 @@ def get_content_manifest_by_requests():
             raise BadRequest(f"{item} is not an integer.")
         request_ids.add(int(item))
 
-    requests = Request.query.filter(Request.id.in_(request_ids)).options(
-        load_only("id"), joinedload(Request.state)
+    requests = (
+        Request.query.filter(Request.id.in_(request_ids))
+        .options(load_only("id"), joinedload(Request.state))
+        .all()
     )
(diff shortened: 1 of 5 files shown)


cachito/web/api_v1.py
```diff
@@ -663,32 +664,34 @@ def get_content_manifest_by_requests():
...
-    requests = Request.query.filter(Request.id.in_(request_ids))
-    found_requests_ids = [request.id for request in requests]
-
-    ids_diff = list(map(str, set(request_ids) - set(found_requests_ids)))
-    if ids_diff:
-        raise BadRequest(f"Cannot find request(s) {', '.join(ids_diff)}.")
-
-    assembled_icm = deepcopy(BASE_ICM)
+    requests = Request.query.filter(Request.id.in_(request_ids)).options(
+        load_only("id"), joinedload(Request.state)
```
akhmanova: Nice!

cachito/web/models.py
```diff
@@ -340,11 +340,7 @@ def content_manifest(self):
         return content_manifest.ContentManifest(self, packages)
 
     def _is_complete(self):
-        if len(self.states) > 0:
-            latest_state = self.states[-1]
-            return latest_state.state_name == RequestStateMapping.complete.name
-
-        return False
+        return self.state is not None and self.state.state_name == RequestStateMapping.complete.name
```
chmeliik, line 343: @brunoapimentel IIRC, wasn't this one of the causes of the empty packages bug?

cachito/web/content_manifest.py
```diff
@@ -242,20 +242,31 @@ def generate_icm(self, image_contents=None):
...
     It is used primarily to generate a package URL (purl).
     """
 
-    name: str
-    type: str
-    version: str
-    dev: bool = False
-    dependencies: list = field(default_factory=list)
-    path: Optional[str] = None
+    __slots__ = ("name", "type", "version", "dev", "dependencies", "path")
```
akhmanova, line 251: Nice! I'd never used it before

cachito/web/content_manifest.py
```diff
@@ -12,7 +12,6 @@
 from cachito.web.utils import deep_sort_icm
 from cachito.workers.pkg_managers import gomod
 
-
```
akhmanova: Can you move this change to https://github.com/release-engineering/cachito/pull/564/commits/b491af65ae02c577309ef13a7e497f1dccd2289a ?

cachito/web/utils.py
```diff
@@ -4,7 +4,6 @@
 
 from flask import request, url_for
 
-
```
akhmanova, line 6: And this one as well could be better to move into https://github.com/release-engineering/cachito/pull/564/commits/b491af65ae02c577309ef13a7e497f1dccd2289a

cachito/web/models.py
tkdchen: Yeah. Sorry. I'll revert this change.

brunoapimentel: Yeah, we need to keep using the `self.states` attribute. This might be  a good reason to try to prioritize the story that covers the refactoring of this dual relationship.

cachito/web/utils.py
```diff
@@ -21,12 +19,10 @@ def deep_sort_icm(orig_item):
...
         sorted_item = [deep_sort_icm(item) for item in orig_item]
         # If item is a list of dicts with the "purl" key, sort by the "purl" value
-        if sorted_item and isinstance(sorted_item[0], OrderedDict) and "purl" in sorted_item[0]:
+        if sorted_item and isinstance(sorted_item[0], dict) and "purl" in sorted_item[0]:
             sorted_item = sorted(sorted_item, key=lambda item: item["purl"])
     elif isinstance(orig_item, dict):
-        items = ((k, deep_sort_icm(v)) for k, v in orig_item.items())
-        # Sort items by key
-        sorted_item = OrderedDict(sorted(items, key=lambda keyval: keyval[0]))
+        sorted_item = {k: deep_sort_icm(v) for k, v in orig_item.items()}
```
chmeliik: How about changing `deep_sort_icm` to sort in place, without making copies of anything?
```python
def sort_icm_in_place(icm: dict) -> None
```
Wouldn't that reduce the memory usage of this function itself to practically 0? (Not quite, Python has loads of overhead for everything, but close).

cachito/web/api_v1.py
```diff
@@ -651,6 +653,14 @@ def get_request_logs(request_id):
     )
 
 
+def create_jsonify_response(content_manifest: Dict[str, Any]) -> flask.Response:
+    """Customize the dumped content manifest by sorting the keys."""
+    return current_app.response_class(
+        json.dumps(content_manifest, sort_keys=True),
+        mimetype=current_app.config["JSONIFY_MIMETYPE"],
+    )
```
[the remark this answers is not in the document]
chmeliik: This would also likely be slower, it's memory vs. performance. Just wondering if you thought about this.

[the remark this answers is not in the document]
tkdchen: One concern is there would be extra effort the manage the generated intermediate file. I'll rethink of it.

cachito/web/content_manifest.py
brunoapimentel: I had no idea such a thing existed o>

brunoapimentel (commented):
LGTM, awesome work Chen :rocket: 

I'm just curious on what you used to measure the memory usage of the function.

tkdchen:
@brunoapimentel I'm using memory-profiler.

chmeliik (approved):
LGTM :tada:

cachito/web/api_v1.py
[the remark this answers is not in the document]
tkdchen: @chmeliik 

When we use the `send_file`, one thing is important for us that is to cleanup the intermediate icm json files, so that such files will not occupy too much file system resources and become another kind of cost we have to monitor.

For example, based on the rox.git repo, the generated icm json file has 25M. Currently, I don't have the data of what is the percentage of requests have big icm data (>=1M or >=10M), but only get some basic data like there were around 500 HTTP GET requests in the past 24 hours and this endpoint is called frequently almost every day. Therefore, a periodic task is required to run, probably once a day, to execute the cleanup.

On the other hand, is it doable to serve the `/content-manifest` and `/download` endpoints by specific pods, separate from the api pods? If yes, we can adjust resources strategically among the pods.

Any idea, thought or comments? Thanks. :)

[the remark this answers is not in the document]
chmeliik: I was thinking the files could be temporary, something like this
```python
with tempfile.TemporaryFile() as f:
    json.dump(content_manifest, f)
    return flask.send_file(f)
```
This would hurt the performance, since writing a file is guaranteed to be slower than generating the string in memory, but might help memory usage. Probably not worth doing at the moment.

akhmanova (approved):
LGTM!

cachito/web/api_v1.py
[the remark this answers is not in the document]
tkdchen: Let's do it. Please have a look at the new commit. :)

PR #570, opened 2021-09-23
Filter Cachito /request endpoint by date

dcho5:
CLOUDBLD-6469

Introduce created_from and created_to parameters to allow
filtering requests by creation date or datetime.


cachito/web/api_v1.py
```diff
@@ -72,6 +73,28 @@ def get_requests():
...
     # pagination parameters `page` and `per_page`.
     query = Request.query.order_by(Request.id.desc())
+    from_date = flask.request.args.get("from_date")
+    if from_date:
+        try:
+            from_date = date.fromisoformat(from_date)
+            from_date = datetime.combine(from_date, datetime.min.time())
+        except ValueError:
+            try:
+                from_date = datetime.fromisoformat(from_date)
```
brunoapimentel, line 83: If you use `datetime.fromisoformat` in a date string, you have the same result as `date.fromisoformat` + `date.combine` you're using, so I guess you can simplify this code here.

```
>>> from_date = "2021-01-01"
>>> datetime.fromisoformat(from_date)
datetime.datetime(2021, 1, 1, 0, 0)
>>> datetime.combine(date.fromisoformat(from_date), datetime.min.time())
datetime.datetime(2021, 1, 1, 0, 0)
```

dcho5: Understood! My intention was to improve readability by writing extra code here to match the `to_date` code.

cachito/web/api_v1.py
```diff
@@ -72,6 +73,24 @@ def get_requests():
     # The call to `paginate` will inspect the current HTTP request for the
     # pagination parameters `page` and `per_page`.
     query = Request.query.order_by(Request.id.desc())
+    from_date = flask.request.args.get("from_date")
```
brunoapimentel, line 76: @serg-cymbaluk is using a library to validate the very same inputs. Look at #574.
I believe there's the same code could be reused between both PRs.

brunoapimentel: This is the relevante code: https://github.com/serg-cymbaluk/cachito/blob/a4bbcfbfa865f584d76c5ad37e355f2cb4427f30/cachito/web/api_v1.py#L723

It was already requested that the parameters' name were changed to match what you did here. 

Also, do mind that this lib will give out a validation error in case a `date` is passe to a `datetime` attribute, so there's the need of handling this to keep the current behavior.

dcho5: Done!

cachito/web/api_v1.py
```diff
@@ -39,6 +41,13 @@
 api_v1 = flask.Blueprint("api_v1", __name__)
 
 
+class RequestArgs(pydantic.BaseModel):
```
serg-cymbaluk, line 44: The endpoint nae has plural form:
Request -> Requests
/request -> /requests

tests/test_api_v1.py
```diff
@@ -505,6 +506,100 @@ def test_create_request_filter_state(app, auth_env, client, db):
...
+    assert fetched_requests[0]["ref"] == "c50b93a32df1c9d700e3e80996845bc2e13be848"
+    # Test created_from and created_to together
+    rv = client.get(f"/api/v1/requests?created_from={last_date}&created_to={last_date}")
+    assert rv.status_code == 200
+    fetched_requests = rv.json["items"]
+    assert len(fetched_requests) == 1
+    assert fetched_requests[0]["ref"] == "12a0692be09fc18ce82a71904562d8408fe2296a"
+
+
+def test_create_request_filter_datetime(app, auth_env, client, db):
```
serg-cymbaluk, line 556: The test looks like previous one with different input.
I would suggest replace those two tests with one, parametrize it and test more cases.
Example:
```python
@pytest.mark.parametrize("created_list,created_filter,expected_num", [
    (["2021-01-01", "2021-02-02"], {"created_from": "2021-01-01", "created_to": "2021-02-02"}, 2),
    ...
])
def test_requests_created_filter(..., created_list, created_filter, expected_num):
    ...
    for created in created_list:
        data = {
            "created": date.fromisoformat(created),
            ...
        }
        ...
    rv = client.get(f"/api/v1/requests?{urlencode(created_filter)}")
    ...
    assert len(fetched_requests) == expected_num
```

You can add as many parameters as needed.

dcho5: Done!

cachito/web/static/api_v1.yaml
```diff
@@ -146,6 +146,30 @@ paths:
...
+          schema:
+            type: string
+            example:
+              date: "2021-10-07"
+              datetime: "2021-10-07T18:52:57.577851"
+            default: null
+        - name: created_to
+          in: query
+          description: >
+            Filter requests by the given date or datetime. This
```
brunoapimentel, line 164: Do you think it's good idea to point that using date here is inclusive?

```suggestion
            Filter requests by the given date (inclusive) or datetime. This
```

brunoapimentel (approved):

PR #574, opened 2021-10-04
Add endpoints with request metrics information

serg-cymbaluk:
Add the following endpoints:
GET /requests-metrics - return a list of requests with a final state information
GET /requests-metrics/summary - return a summary about number and duration of requests

Signed-off-by: Serhii Tsymbaliuk <stsymbal@redhat.com>


serg-cymbaluk:
I didn't add unit tests for /request-metrics/summary since the corresponding DB queries incompatible with SQLite.
I going to add integration test to cover that functions.

cachito/web/api_v1.py
```diff
@@ -5,10 +5,12 @@
 import tempfile
 from collections import OrderedDict
 from copy import deepcopy
-from typing import Any, Dict
+from datetime import date, timedelta
+from typing import Any, Dict, Optional
 
 import flask
 import kombu.exceptions
+import pydantic
```
brunoapimentel, line 13: Have we settled on using pydantic for validation? LGTM, I don't have strong opinions on this, but I heard folks commeting about using JsonSchema, so I'll leave this thread open in case anyone wants to chime in. :)

cachito/web/models.py
```diff
@@ -648,6 +650,31 @@ def __repr__(self):
...
+                cls.state,
+                cls.state_reason,
+                cls.updated,
+                func.extract(
+                    "epoch",
+                    cls.updated.cast(TIMESTAMP) - func.min(initial_state.updated).cast(TIMESTAMP),
+                ).label("duration"),
+            )
+            .join(initial_state, initial_state.request_id == cls.request_id)
+            .filter(initial_state.state == RequestStateMapping.in_progress.value)
```
brunoapimentel, line 669: A request has many `in_progress` states. Does this always default to the earliest one?

cachito/web/api_v1.py
serg-cymbaluk: pydantic is not agreed, but the PR can be a good starting point for such discussion, since there is visible a real case.

cachito/web/models.py
serg-cymbaluk: We need there only func.min(initial_state.updated) and then we group by final state. Any other data from initial_state will not appear in result.

cachito/web/api_v1.py
```diff
@@ -716,3 +718,87 @@ def get_content_manifest_by_requests():
     if len(requests) > 1:
         deep_sort_icm(assembled_icm)
     return send_content_manifest_back(assembled_icm)
+
+
+class RequestMetricsArgs(pydantic.BaseModel):
+    """Query parameters for /request-metrics endpoint."""
+
+    start_date: Optional[date] = None
+    end_date: Optional[date] = None
```
chmeliik, line 727: Could the naming and behavior be consistent with #570?

cachito/web/api_v1.py
```diff
@@ -716,3 +718,87 @@ def get_content_manifest_by_requests():
...
+    pagination_query = query.paginate(max_per_page=max_per_page)
+    return flask.jsonify(
+        {
+            "items": [
+                {
+                    "id": state.request_id,
+                    "final_state": RequestStateMapping(state.state).name,
+                    "final_state_reason": state.state_reason,
+                    "finished": state.updated.isoformat(),
+                    "duration": state.duration,
```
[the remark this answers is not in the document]
chmeliik: And maybe `initial_inactive` would be useful too (time between state 1 and state 2)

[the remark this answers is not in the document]
serg-cymbaluk: As we discussed previously, the difference of time between two states would not have the same meaning.
In Cachito activity is represented by async tasks. And sometimes we set state on the start of that task, sometime in the end. Also we have couple of cases where it is set in the middle of execution.
We can not calculate the actual inactivity time as well. Inactivity in our case could mean difference between one tasks end and the following task start. And we have no such data.

However we are preparing changes that could help to fill the gap. We are going to introduce a table which will track duration of each task. Base on the data we could extend the request metrics endpoint. But currently implementing any of such metrics could just mislead users.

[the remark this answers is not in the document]
serg-cymbaluk: I would not build logic to adjust the wrong assumption.
For the certain case, you provided, it would better if Cachito could respond explicitly about a request status to avoid such guessing.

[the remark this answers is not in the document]
chmeliik: To a non-human client, the only indicator of request activity is the state updates. This metric is trying to capture apparent inactivity. But if you don't see a use for it you can drop it, just make sure to reflect that in the planned SLI work.

cachito/web/api_v1.py
```diff
@@ -716,3 +718,87 @@ def get_content_manifest_by_requests():
...
+def get_request_metrics():
+    """Return a list of completed requests with a final state information."""
+    max_per_page = flask.current_app.config["CACHITO_MAX_PER_PAGE"]
+    args = RequestMetricsArgs(**flask.request.args)
+    query = RequestState.get_final_states_query().order_by(RequestState.request_id.desc())
+
+    if args.start_date:
+        query = query.filter(RequestState.updated >= args.start_date)
+    if args.end_date:
+        query = query.filter(RequestState.updated < args.end_date + timedelta(days=1))
```
MartinBasti, line 740: why `+ timedelta(days=1)` ?

cachito/web/api_v1.py
MartinBasti: yes please, make naming consistent

cachito/web/api_v1.py
```diff
@@ -716,3 +718,87 @@ def get_content_manifest_by_requests():
     if len(requests) > 1:
         deep_sort_icm(assembled_icm)
     return send_content_manifest_back(assembled_icm)
+
+
+class RequestMetricsArgs(pydantic.BaseModel):
+    """Query parameters for /request-metrics endpoint."""
+
+    start_date: Optional[date] = None
```
MartinBasti, line 726: I'd like to use datetime instead, to support more granular queries

cachito/web/api_v1.py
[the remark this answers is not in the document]
brunoapimentel: I was thinking it might be a good idea to change the first state to something like `queued` or simply `created`. This could make it more clear, not sure if there'll be any impact though.

I do think that a `time_in_queue` metric would be very useful, I'd like to keep an eye on that. I believe that could be done in a separate PR, if @serg-cymbaluk prefers.

MartinBasti:
openapi definition must be updated

serg-cymbaluk:
- Extended unit tests
- Added integration tests
- Updated OpenAPI documentation

cachito/web/api_v1.py
```diff
@@ -716,3 +718,102 @@ def get_content_manifest_by_requests():
...
+    """Return a list of completed requests with a final state information."""
+    max_per_page = flask.current_app.config["CACHITO_MAX_PER_PAGE"]
+    args = RequestMetricsArgs(**flask.request.args)
+    query = RequestState.get_final_states_query().order_by(RequestState.request_id.desc())
+    finished_to = args.finished_to
+    if isinstance(finished_to, date):
+        finished_to = datetime.combine(finished_to, time.max)
+    if args.finished_from:
+        query = query.filter(RequestState.updated >= args.finished_from)
+    if finished_to:
```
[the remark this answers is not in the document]
serg-cymbaluk: I don't think so. It will return just empty list for such filter and the result is correct.

cachito/web/api_v1.py
```diff
@@ -735,3 +735,102 @@ def get_content_manifest_by_requests():
...
+
+
+@api_v1.route("/request-metrics", methods=["GET"])
+def get_request_metrics():
+    """Return a list of completed requests with a final state information."""
+    max_per_page = flask.current_app.config["CACHITO_MAX_PER_PAGE"]
+    args = RequestMetricsArgs(**flask.request.args)
+    query = RequestState.get_final_states_query().order_by(RequestState.request_id.desc())
+    finished_to = args.finished_to
+    if isinstance(finished_to, date):
```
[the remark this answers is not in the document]
brunoapimentel: I do think it's the same issue =/

I tried to run this request locally:
```
http://localhost:8080/api/v1/request-metrics?finished_to=2021-10-18T02:00:00
```

I added a log to see what was in the `finished_to` variable and:

```
[2021-10-18 12:43:58 cachito.web.app INFO api_v1.get_request_metrics] 2021-10-18 23:59:59.999999
```

PR #575, opened 2021-10-04
Disable celery heartbeats

brunoapimentel:
By disabling the Celery's heartbeat, we intend to avoid the connection leak issue that is happening at the Rabbitmq pod.

It seems that whenever the heartbeat fails to verify the worker's connection, it [spawns a new connection ](https://github.com/celery/kombu/blob/2aeb73248b4a0bf8ad3de81a0f40733e7cd42255/kombu/connection.py#L870)that does not get closed. This will cause the connections to accumulate, increasing the memory usage constantly and eventually blocking all new connections, essentialy crashing Cachito.

It's important to notice that this heartbeat is implemented at the application level in Celery, and duplicates the goals of the [AMQP heartbeat](https://docs.celeryproject.org/en/latest/userguide/configuration.html#broker-heartbeat) to some extent.

### DIFF
--- a/docker/Dockerfile-workers
+++ b/docker/Dockerfile-workers
@@ -34,4 +34,4 @@ ENV GIT_COMMITTER_NAME=cachito \
     GIT_AUTHOR_EMAIL=cachito@localhost
 
 EXPOSE 8080
-CMD ["celery", "-A", "cachito.workers.tasks", "worker", "--loglevel=info"]
+CMD ["celery", "-A", "cachito.workers.tasks", "worker", "--without-heartbeat", "--loglevel=info"]


chmeliik:
> duplicates the goals of the AMQP heartbeat to some extent

Does Cachito already use the AMQP heartbeat?

Do you know of any potential consequences of disabling the heartbeat?

brunoapimentel:
> > duplicates the goals of the AMQP heartbeat to some extent
> 
> Does Cachito already use the AMQP heartbeat?
> 
> Do you know of any potential consequences of disabling the heartbeat?

We have enabled the Broker heartbeats in the test enviroment while also testing the application with Celery heartbeats disabled. We have not seem any side effects for the change in almost a week.

But you have a good point here: I think we need to investigate if the Broker heartbeats are a necessity when Celery heartbeats are disabled. We don't have an answer so far.

Our idea is to keep the tests going in the Stage environment for a while.

chmeliik (approved):
> But you have a good point here: I think we need to investigate if the Broker heartbeats are a necessity when Celery heartbeats are disabled. We don't have an answer so far.

I meant more so consequences of disabling the Celery heartbeat, having at least one of the heartbeats sounds like a good idea. But stage testing basically answers my question :+1:

It would be nice to have the PR description in the commit message, apart from that LGTM

mike-kingsbury:
LGTM...

brunoapimentel:
> It would be nice to have the PR description in the commit message, apart from that LGTM

FTFY

chmeliik (approved):

serg-cymbaluk (approved):
LGTM

PR #580, opened 2021-10-08
Add checklist to pull requests

MartinBasti:
Maintaners should check checklist during reviews

### DIFF
--- a/docs/pull_request_template.md
+++ b/docs/pull_request_template.md
@@ -1,0 +1,7 @@
+# Maintainers will complete the following section
+
+- [ ] Commit messages are descriptive enough
+- [ ] Code coverage from testing does not decrease and new code is covered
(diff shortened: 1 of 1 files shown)


brunoapimentel:
We also need to update the README in case any worker configuration is changed/added. 

https://github.com/release-engineering/cachito/blob/b713e76c2d4d6e1fde3db7e720766e59eeded499/README.md#L411

Do you think we should add a reminder to the checklist? Maybe just a general reminder to check if the README needs to be updated.

serg-cymbaluk:
@MartinBasti 
One item, may be relevant here, is a DB schema review. We need to check if DB schema has been touched. If so, check the corresponding migration is provided.

akhmanova (approved):
LGTM!

serg-cymbaluk (approved):
LGTM

PR #588, opened 2021-10-15
Fix /request endpoint created_to parameter

dcho5:
CLOUDBLD-6469

Function isinstance caused created_to to treat all input as date because
datetime is a date subclass. Fix by passing datetime into isinstance
instead of date. Cover edge test cases to confirm proper functionality.


tests/test_api_v1.py
```diff
@@ -524,20 +524,20 @@ def test_create_request_filter_state(app, auth_env, client, db):
...
             {
                 "created_from": "2021-01-01T17:01:09.452358",
-                "created_to": "2021-01-01T17:01:09.452358",
+                "created_to": "2021-02-02T17:01:09.452357",
             },
             1,
         ),
         (
-            ["2021-01-01", "2021-02-02"],
+            ["2021-01-01T17:01:09.452358", "2021-02-02T17:01:09.452358"],
```
brunoapimentel, line 540: I think it would be good to keep at least one test covering `date` parameters, just to be sure they're working properly. All tests now are `datetime`. Wdyt?

brunoapimentel (approved):

MartinBasti (approved):

PR #589, opened 2021-10-15
Log api/v1/status/short endpoint errors

brunoapimentel:
Since the CachitoError is caught in this endpoint, its details are hidden from the logs, and need to be explicitly printed. 

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -61,6 +61,7 @@ def get_status_short():
         status(short=True)
         retval = {"ok": True}
     except CachitoError as e:
+        flask.current_app.logger.error(e)
         retval = {"ok": False, "reason": str(e)}
 
     return flask.jsonify(retval), 200 if retval["ok"] else 503


nirzari (approved):

cachito/web/api_v1.py
```diff
@@ -61,6 +61,7 @@ def get_status_short():
         status(short=True)
         retval = {"ok": True}
     except CachitoError as e:
+        flask.current_app.logger.error(e)
```
chmeliik, line 64: Is the goal to extract the /status/short error messages from logs? Would an easily recognizable prefix be helpful? Something like `status is not OK: %s`

MartinBasti: yes please, use a prefix

brunoapimentel: Yes, that's the intention. I omitted the prefix since we already have the exact location the error happened in the log:

```
[2021-10-18 11:48:05 cachito.web.app ERROR api_v1.get_status_short] nexus unavailable: None
```

serg-cymbaluk: I would keep as it is.

chmeliik: ack, works for me

PR #590, opened 2021-10-18
Fix date type check for /request-metrics/summary endpoint

serg-cymbaluk:
Add validator for 'finished_to' query parameter which allows to determine date values
and convert them into values that correspond to the end of the day.

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -5,7 +5,7 @@ import os
 import tempfile
 from collections import OrderedDict
 from copy import deepcopy
-from datetime import date, datetime, time
+from datetime import date, datetime
 from typing import Any, Dict, Union
 
 import flask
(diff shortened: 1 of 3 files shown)


cachito/web/api_v1.py
```diff
@@ -790,7 +790,7 @@ def get_request_metrics_summary():
     """Return a summary about completed requests for a given period of time."""
     args = RequestMetricsSummaryArgs(**flask.request.args)
     finished_to = args.finished_to
-    if isinstance(finished_to, date):
+    if type(finished_to) is date:
```
MartinBasti, line 793: I'm afraid that somebody will change it, when a linter will report this a not pythonic and suggest to use isinstance. Can we just use `if not isinstance(finished_to, datetime)` ?

PR #595, opened 2021-10-21
Add integration test for generated golang packages

dcho5:
CLOUDBLD-6925

### DIFF
--- a/tests/integration/test_data/go_generate_packages.yaml
+++ b/tests/integration/test_data/go_generate_packages.yaml
@@ -1,0 +1,142 @@
+# Repos for testing how Cachito handles packages, specifically for repos which use `go generate`.
+# See https://go.dev/blog/generate
+# Such repos can be identified with a `//go:generate ...` comment in the `main.go` file.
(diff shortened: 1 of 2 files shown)


chmeliik (approved):
LGTM, this is great :tada: 

I have to say, I'm more than a little surprised that Cachito handles these cases as well as it does, I think that's mostly by accident :smile: 

But now that we have tests for it, it's more of a formal contract :+1:

brunoapimentel:
I'm just wondering if the behavior we have now is what we expect? If you ask Cachito to process either the imported repo or the imported-generated repo, the generated package is listed as a dependency:

```
"dependencies": [
  {
    "name": "github.com/cachito-testing/go-generate-imported-generated/foobar",
    "replaces": null,
    "type": "go-package",
    "version": "v0.0.0-20211021013725-27ad22ae910c"
  }
],
```

But the deps folder in the archive is empty (just has some nested empty folders):

```/home/bpimente/Downloads/cachito-6/deps/gomod/pkg/mod/cache/download```

Wdyt, @chmeliik?

PR #606, opened 2021-11-05
Upgrades images to Fedora 35

brunoapimentel:
Whereas the primary intention was to have Go 1.17 on the worker images, we noticed that F35 still has Go 1.16 as its default version. We will still keep the image update since they're benefitial, overall.


MartinBasti (approved):

MartinBasti (approved):

akhmanova (approved):
LGTM :+1:

PR #609, opened 2021-11-08
Add integration test for Git submodules

dcho5:
CLOUDBLD-3115

### DIFF
--- a/tests/integration/test_packages.py
+++ b/tests/integration/test_packages.py
@@ -11,6 +11,7 @@ from . import utils
         ("pip_packages", "without_deps"),
         ("pip_packages", "with_deps"),
         ("pip_packages", "multiple"),
+        ("pip_packages", "git_submodule"),
         ("go_generate_packages", "go_generate"),
         ("go_generate_packages", "go_generate_generated"),
         ("go_generate_packages", "go_generate_imported"),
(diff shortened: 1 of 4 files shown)


tests/integration/test_packages.py
```diff
@@ -76,10 +79,11 @@ def test_packages(env_package, env_name, test_env, tmpdir):
     expected_response_data = env_data["response_expectations"]
     utils.assert_elements_from_response(response_data, expected_response_data)
 
-    client.download_and_extract_archive(completed_response.id, tmpdir)
-    source_path = tmpdir.join(f"download_{str(completed_response.id)}")
-    expected_files = env_data["expected_files"]
-    utils.assert_expected_files(source_path, expected_files, tmpdir)
+    if env_name != "git_submodule":
```
dcho5, line 82: It turns out that Cachito processes Git submodules better than Git does. A Git tarball does not include the contents of a submodule, leading to this test case failing every single time. I've added in this if statement to handle this for the time being. Any other ideas on how to deal with this issue?

brunoapimentel: Two ideas come to my mind:
1- Create a tarball with the correct contents and host it in yet another Github repo, just for the sake of this test
2- Delete from the `tmpdir` the submodule folders in case it's a "git_submodule" test

brunoapimentel:
Great work here! I just think we could add a readme to all new repos created... the "go generate" ones look so nice. :grin:

tests/integration/test_packages.py
dcho5: I've implemented idea #1!

dcho5:
> Great work here! I just think we could add a readme to all new repos created... the "go generate" ones look so nice. grin

Done!
i.e. https://github.com/cachito-testing/git-submodule-pip

brunoapimentel (approved):

serg-cymbaluk (approved):
Works for me.

PR #611, opened 2021-11-10
Update requests-kerberos to version 0.13.0

ejegrova:
Newer version of requests-kerberos is compatible with
changes in Python 3.10.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -101,7 +101,13 @@ cryptography==3.4.6 \
     --hash=sha256:9e98b452132963678e3ac6c73f7010fe53adf72209a32854d55690acac3f6724 \
     --hash=sha256:df186fcbf86dc1ce56305becb8434e4b6b7504bc724b71ad7a3239e0c9d14ef2 \
     --hash=sha256:fec7fb46b10da10d9e1d078d1ff8ed9e05ae14f431fdbd11145edd0550b9a964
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with python 3.9
+# This file is autogenerated by pip-compile with python 3.8
```
brunoapimentel, line 2: I think this does not change anything, but we changed the pip-compile python version here.

brunoapimentel:
Wdyt about dropping a note in the commit message to say that this upgrade is necessary because of Python 3.10 compatibility?

requirements.txt
ejegrova: yes, it's probably because my default python is 3.8. I'll generate this with python 3.9

ejegrova:
> Wdyt about dropping a note in the commit message to say that this upgrade is necessary because of Python 3.10 compatibility?

I thought it says that. Can you be more exact? Do you want me to also write there that previous versions are not compatible?

requirements.txt
ejegrova: done

brunoapimentel (approved):
LGTM :rocket:

akhmanova (approved):

PR #614, opened 2021-11-18
Identify devDependencies for Yarn

dcho5:
CLOUDBLD-3914

Use data from package.json to identify devDependencies in Yarn repos.

### DIFF
--- a/cachito/workers/pkg_managers/yarn.py
+++ b/cachito/workers/pkg_managers/yarn.py
@@ -1,6 +1,7 @@
 import copy
 import json
 import logging
+from collections import deque
 from os.path import normpath
 from pathlib import Path
 from typing import Dict, Optional
@@ -69,7 +70,41 @@ def get_yarn_proxy_repo_username(request_id):
     return f"cachito-yarn-{request_id}"
 
 
-def _get_deps(yarn_lock, file_deps_allowlist):
+def _find_reachable_deps(visited_deps, dep, yarn_lock):
+    """
+    Get a set of all dependencies reachable by a top-level dependency using BFS.
+
+    :param yarn_lock: yarn.lock file as a dictionary
+    :param dep: top-level dependency in package.json
+    :param reachable_deps: set of already visited non-dev dependencies
+    """
+    yarn_lock_parsed = _expand_yarn_lock_keys(yarn_lock)
+    visited_deps.add(dep)
+    bfs_queue = deque([dep])
+    while bfs_queue:
+        current_dep = bfs_queue.popleft()
+        package = pyarn.lockfile.Package.from_dict(current_dep, yarn_lock_parsed[current_dep])
+        bfs_queue.extend(
+            f"{name}@{version}"
+            for name, version in package.dependencies.items()
+            if f"{name}@{version}" not in visited_deps
+        )
+        visited_deps.add(current_dep)
+
+
+def _split_yarn_lock_key(dep_identifer):
+    """
+    Remove unnecessary quotes in dep_identifier and split the string into a list of dependencies.
+
+    String dep_identifer contains one or more dependencies separated by commas.
+
+    :param dep_identifier: a string which lists all of the dependencies in the identifer
+    :return: a list of all the dependencies in the identifier
+    """
+    return dep_identifer.replace('"', "").split(", ")
+
+
+def _get_deps(package_json, yarn_lock, file_deps_allowlist):
     """
     Process the dependencies in a yarn.lock file and return relevant information.
 
@@ -90,9 +125,24 @@ def _get_deps(yarn_lock, file_deps_allowlist):
     """
     deps = []
     nexus_replacements = {}
+    non_dev_deps = set()
+
+    for dep_type in ["dependencies", "peerDependencies", "optionalDependencies"]:
+        if dep_type not in package_json:
+            continue
+        for name, version in package_json[dep_type].items():
+            dep = f"{name}@{version}"
+            if dep not in non_dev_deps:
+                _find_reachable_deps(non_dev_deps, dep, yarn_lock)
 
     for dep_identifier, dep_data in yarn_lock.items():
         package = pyarn.lockfile.Package.from_dict(dep_identifier, dep_data)
+
+        dev = True
+        for dep_id in _split_yarn_lock_key(dep_identifier):
+            if dep_id in non_dev_deps:
+                dev = False
+                break
 
         if package.url:
             source = package.url
@@ -116,7 +166,7 @@ def _get_deps(yarn_lock, file_deps_allowlist):
 
         dep = {
             "bundled": False,  # yarn.lock does not seem to contain bundled deps at all
-            # "dev": <yarn.lock does not state whether a dependency is dev>
+            "dev": dev,
             "name": package.name,
             "version_in_nexus": nexus_replacement["version"] if nexus_replacement else None,
             "type": "yarn",
@@ -234,7 +284,7 @@ def _get_package_and_deps(package_json_path, yarn_lock_path):
         get_worker_config().cachito_yarn_file_deps_allowlist.get(package["name"], [])
     )
 
-    deps, nexus_replacements = _get_deps(yarn_lock, file_deps_allowlist)
+    deps, nexus_replacements = _get_deps(package_json, yarn_lock, file_deps_allowlist)
     return {
         "package": package,
         "deps": deps,
@@ -284,9 +334,9 @@ def _set_proxy_resolved_urls(yarn_lock: Dict[str, dict], proxy_repo_name: str) -
     return modified
 
 
-def _expand_replacements(nexus_replacements: Dict[str, dict]) -> Dict[str, dict]:
+def _expand_yarn_lock_keys(nexus_replacements: Dict[str, dict]) -> Dict[str, dict]:
     """
-    Expand all N:1 keys in the Nexus replacements dict into N 1:1 keys.
+    Expand all N:1 keys in the yarn.lock dict into N 1:1 keys.
 
     In the original dict, 1 key may in fact be N comma-separated keys. These N keys all have the
     same value, making them N:1 keys. In the expanded dict, these will be turned into N 1:1 keys.
@@ -297,12 +347,12 @@ def _expand_replacements(nexus_replacements: Dict[str, dict]) -> Dict[str, dict]
     :param dict nexus_replacements: a dict of nexus replacements which may contain N:1 keys
     :return: a dict of nexus replacements where all N:1 keys have been expanded to N 1:1 keys
     """
-    expanded_replacements = {
+    expanded_yarn_lock_keys = {
         key: nexus_replacements[multi_key]
         for multi_key in nexus_replacements
-        for key in map(str.strip, multi_key.split(","))
+        for key in _split_yarn_lock_key(multi_key)
     }
-    return expanded_replacements
+    return expanded_yarn_lock_keys
 
 
 def _match_to_new_version(
@@ -313,7 +363,8 @@ def _match_to_new_version(
 
     :param str dep_name: dependency name
     :param str dep_version: dependency version
-    :param dict expanded_replacements: expanded dict of Nexus replacements, see _expand_replacements
+    :param dict expanded_replacements: expanded dict of Nexus replacements,
+        see _expand_yarn_lock_keys
     :return: new version (str) or None
     """
     dep_identifier = f"{dep_name}@{dep_version}"
@@ -329,7 +380,7 @@ def _replace_deps_in_package_json(package_json, nexus_replacements):
         {<dependency identifier>: <dependency info>}
     :return: copy of package.json data with replacements applied (or None if no replacements match)
     """
-    expanded_replacements = _expand_replacements(nexus_replacements)
+    expanded_replacements = _expand_yarn_lock_keys(nexus_replacements)
 
     package_json_new = copy.deepcopy(package_json)
     modified = False
@@ -367,7 +418,7 @@ def _replace_deps_in_yarn_lock(yarn_lock, nexus_replacements):
         {<dependency identifier>: <dependency info>}
     :return: copy of yarn.lock data with replacements applied
     """
-    expanded_replacements = _expand_replacements(nexus_replacements)
+    expanded_replacements = _expand_yarn_lock_keys(nexus_replacements)
     yarn_lock_new = {}
 
     for key, value in yarn_lock.items():
--- a/README.md
+++ b/README.md
@@ -668,7 +668,7 @@ Feature                 | gomod | npm | pip | yarn |
 Baseline                | ✓     | ✓   | ✓   | ✓    |
 Content Manifest        | ✓     | ✓   | ✓   | ✓    |
 Dependency Replacements | ✓     | x   | x   | x    |
-Dev Dependencies        | ✓     | ✓   | ✓   | x    |
+Dev Dependencies        | ✓     | ✓   | ✓   | ✓    |
 External Dependencies   | N/A   | ✓   | ✓   | ✓    |
 Multiple Paths          | ✓     | ✓   | ✓   | ✓    |
 Nested Dependencies     | ✓     | ✓   | x   | ✓    |
--- a/tests/integration/test_data/cached_dependencies.yaml
+++ b/tests/integration/test_data/cached_dependencies.yaml
@@ -761,85 +761,105 @@ yarn_cached_deps:
   # Parts of the Cachito response to check
   response_expectations:
     dependencies:
-      - name: assertion-error
+      - dev: false
+        name: assertion-error
         replaces: null
         type: yarn
         version: 1.1.0
-      - name: chai
+      - dev: false
+        name: chai
         replaces: null
         type: yarn
         version: 4.2.0
-      - name: check-error
+      - dev: false
+        name: check-error
         replaces: null
         type: yarn
         version: 1.0.2
-      - name: deep-eql
+      - dev: false
+        name: deep-eql
         replaces: null
         type: yarn
         version: 3.0.1
-      - name: fecha
+      - dev: false
+        name: fecha
         replaces: null
         type: yarn
         version: https://github.com/taylorhakes/fecha/archive/91680e4db1415fea33eac878cfd889c80a7b55c7.tar.gz#f09ea0b8115b9733dddc88227086c73ba4ddc926
-      - name: get-func-name
+      - dev: false
+        name: get-func-name
         replaces: null
         type: yarn
         version: 2.0.0
-      - name: npm-test-project
+      - dev: false
+        name: npm-test-project
         replaces: null
         type: yarn
         version: git+https://gitlab.com/eakhmano-test-group/eakhmano-test-subgroup/subsubgroup/npm-test-project.git#88114c48d7d894f7e2306adf06137b096e9cdf29
-      - name: pathval
+      - dev: false
+        name: pathval
         replaces: null
         type: yarn
         version: 1.1.1
-      - name: test-npm-dep
+      - dev: false
+        name: test-npm-dep
         replaces: null
         type: yarn
         version: git+https://elinah@bitbucket.org/elinah/test-npm-dep.git#b952715c0706f30d92a7253dcdadc510a618eca5
-      - name: type-detect
+      - dev: false
+        name: type-detect
         replaces: null
         type: yarn
         version: 4.0.8
     packages:
       - dependencies:
-        - name: assertion-error
+        - dev: false
+          name: assertion-error
           replaces: null
           type: yarn
           version: 1.1.0
-        - name: chai
+        - dev: false
+          name: chai
           replaces: null
           type: yarn
           version: 4.2.0
-        - name: check-error
+        - dev: false
+          name: check-error
           replaces: null
           type: yarn
           version: 1.0.2
-        - name: deep-eql
+        - dev: false
+          name: deep-eql
           replaces: null
           type: yarn
           version: 3.0.1
-        - name: fecha
+        - dev: false
+          name: fecha
           replaces: null
           type: yarn
           version: https://github.com/taylorhakes/fecha/archive/91680e4db1415fea33eac878cfd889c80a7b55c7.tar.gz#f09ea0b8115b9733dddc88227086c73ba4ddc926
-        - name: get-func-name
+        - dev: false
+          name: get-func-name
           replaces: null
           type: yarn
           version: 2.0.0
-        - name: npm-test-project
+        - dev: false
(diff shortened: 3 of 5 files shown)


dcho5:
Please double check functionality. Only thing left to do now should be to modify unit tests.

chmeliik:
https://docs.npmjs.com/cli/v6/configuring-npm/package-lock-json#dev
> dev
>
> If true then this dependency is either a development dependency ONLY of the top level module or a transitive dependency of one. This is false for dependencies that are both a development dependency of the top level and a transitive dependency of a non-development dependency of the top level.

One fairly simple way to handle transitivity could be:
- for each direct dependency A in package.json `dependencies`:
    - traverse the dependency graph (yarn.lock) starting from A, mark all reachable deps as **non-dev**
- repeat for `optionalDependencies`? and `peerDependencies`? not sure, best check npm behavior
- once finished, everything that you haven't marked is a dev dependency

Also note that dependencies cannot be handled as `name`, but must be handled as `name@version`. There could be cases where `foo@1.0.0` is a dev dependency but `foo@2.0.0` is not.

cachito/workers/pkg_managers/yarn.py
```diff
@@ -90,6 +90,19 @@ def _get_deps(yarn_lock, file_deps_allowlist):
...
+    for dep_identifier, dep_data in yarn_lock.items():
+        package = pyarn.lockfile.Package.from_dict(dep_identifier, dep_data)
+        for dep_type in ["dependencies", "peerDependencies", "optionalDependencies"]:
+            if dep_type not in package_json:
+                continue
+            for name in package_json[dep_type]:
+                version = package_json[dep_type][name]
+                name_version = f"{name}@{version}"
+                if name_version in dep_identifier:
+                    non_dev_deps.update({dep_identifier: package.dependencies})
```
MartinBasti, line 105: I'm trying to understand why `update()`. Isn't it just:

```python
non_dev_deps[dep_identifier] = package.dependencies
```

dcho5:
Made a few significant changes:
- Create helper function to reach all non-dev dependencies using DFS (please check placement of helper function in the file as well as variable names, readability, etc.)
- Change Yarn integration test data to include the `dev` field, all set to the correct value (please review [cachito-test MR](https://gitlab.cee.redhat.com/osbs/cachito-tests/-/merge_requests/41) as well)
- Still need to work on `test_yarn.py` - temporarily made `test_get_deps` ([link](https://github.com/release-engineering/cachito/pull/614/files#diff-7360c16c071c32875c62c17eeef80b349e82f4137c9ec91f1f6b88fb7edbe0a5R330)) work by adding empty `package.json` parameters. Do we need a new unit test to test the changes, or should I simply modify `test_get_deps`?

cachito/workers/pkg_managers/yarn.py
```diff
@@ -69,7 +69,22 @@ def get_yarn_proxy_repo_username(request_id):
...
+        sub_deps = []
+        for dep_identifier, dep_data in yarn_lock.items():
+            if current_dep in dep_identifier:
+                if "dependencies" in dep_data:
+                    sub_deps = [
+                        f"{name}@{version}" for name, version in dep_data["dependencies"].items()
+                    ]
+                break
+        for sub_dep in sub_deps:
+            find_non_dev(non_dev_deps, yarn_lock, sub_dep)
```
akhmanova, line 84: Can you please elaborate on why you choose finally DFS? 

I don't think there is a high probability of hitting the recursion limit with 1000+ nested deps, but maybe it's better to avoid future problems

akhmanova: Let's use BFS with collections.deque instead of DFS

MartinBasti, line 84: I stopped reading here, replace recursion by stack and iteration.

MartinBasti (changes requested):

cachito/workers/pkg_managers/yarn.py
```diff
@@ -90,10 +105,23 @@ def _get_deps(yarn_lock, file_deps_allowlist):
     """
     deps = []
     nexus_replacements = {}
+    non_dev_deps = []
```
MartinBasti, line 108: Should be this `set` ?

cachito/workers/pkg_managers/yarn.py
MartinBasti: I've feeling that DFS may be faster and take less resources than BFS, no hard data though, I dunno how common dependencies looks like

dcho5: Apologies, I actually have implemented BFS here, @akhmanova . I accidentally wrote DFS instead of BFS in my daily standup. The reason for BFS would be I have noticed that most dependencies have a more immediate nested dependencies (many dependencies have a double-digit number of sub-dependencies) than the depth of how far those dependencies go (usually ~2-3). I will change the algorithm to use collections.deque as it should help with performance. I will also change the function from a recursive one to an iterative one, @MartinBasti

cachito/workers/pkg_managers/yarn.py
dcho5: 100%

cachito/workers/pkg_managers/yarn.py
dcho5: Got it!

cachito/workers/pkg_managers/yarn.py
dcho5: On second thought, this should be a `collections.deque`.

dcho5: Actually, would using a `set` as a stack/list be more efficient than using `collections.deque`?

MartinBasti: Isn't this result of the function?

cachito/workers/pkg_managers/yarn.py
```diff
@@ -69,7 +69,22 @@ def get_yarn_proxy_repo_username(request_id):
     return f"cachito-yarn-{request_id}"
 
 
-def _get_deps(yarn_lock, file_deps_allowlist):
+def find_non_dev(non_dev_deps, yarn_lock, current_dep):
```
MartinBasti, line 72: It misses docstring and annotations.

I don't actually understand what this function is doing, to me, it seems it only creates a list of all dependencies, where is that `non_dev` or `find` logic?

dcho5: Fixed!

cachito/workers/pkg_managers/yarn.py
dcho5: Changed! Please check.

cachito/workers/pkg_managers/yarn.py
MartinBasti: I may need explanation how this is BFS, because I see well hidden DFS

cachito/workers/pkg_managers/yarn.py
```diff
@@ -69,7 +70,37 @@ def get_yarn_proxy_repo_username(request_id):
...
 
-def _get_deps(yarn_lock, file_deps_allowlist):
+def find_reachable_deps(yarn_lock, dep):
+    """
+    Get list of all dependencies reachable by a top-level dependency using BFS.
+
+    :param yarn_lock: yarn.lock file as a dictionary
+    :param dep: top-level dependency in package.json
+    :return: list of all dependencies reachable by dep
+    :rtype: collections.deque
```
MartinBasti, line 80: set?

cachito/workers/pkg_managers/yarn.py
dcho5: I believe the old code wasn't designed as intended and caused some confusions. The new changes I have made now should be a clear BFS algorithm.

cachito/workers/pkg_managers/yarn.py
dcho5: Of course, good catch!

cachito/workers/pkg_managers/yarn.py
```diff
@@ -69,7 +70,37 @@ def get_yarn_proxy_repo_username(request_id):
...
+    :return: set of all dependencies reachable by dep
+    :rtype: set
+    """
+    reachable_deps = set()
+    bfs_queue = deque()
+    reachable_deps.add(dep)
+    bfs_queue.append(dep)
+    while bfs_queue:
+        current_dep = bfs_queue.popleft()
+        for dep_identifier, dep_data in yarn_lock.items():
```
MartinBasti, line 88: How dep_identifier looks here?

cachito/workers/pkg_managers/yarn.py
```diff
@@ -69,7 +70,37 @@ def get_yarn_proxy_repo_username(request_id):
...
+            package = pyarn.lockfile.Package.from_dict(dep_identifier, dep_data)
+            if current_dep in dep_identifier:
```
MartinBasti, line 90: Can this do false positive matching?

cachito/workers/pkg_managers/yarn.py
dcho5: Anything from `istanbul-reports@^1.0.0-alpha` to `align-text@^0.1.1, align-text@^0.1.3`, this is why I use the for-loop here to iterate through `yarn.lock` and check if `current_dep` is included in `dep_identifier`.

e.g.
```
current_dep = "align-text@^0.1.3"
dep_identifier = "align-text@^0.1.1, align-text@^0.1.3"
```

cachito/workers/pkg_managers/yarn.py
dcho5: Not as far as I can see. This may require more research. @chmeliik , what do you think?

chmeliik: Proper handling would be much nicer, you can split the multi-keys using something like this: https://github.com/release-engineering/cachito/blob/8091e35befb9026e43069faa7c11c01e20279398/cachito/workers/pkg_managers/yarn.py#L287

cachito/workers/pkg_managers/yarn.py
MartinBasti: so `dep_identifier` is a list/tuple then?

dcho5: It's a tuple.

cachito/workers/pkg_managers/yarn.py
dcho5: Done!

cachito/workers/pkg_managers/yarn.py
```diff
@@ -69,7 +70,33 @@ def get_yarn_proxy_repo_username(request_id):
...
+    Get a set of all dependencies reachable by a top-level dependency using BFS.
+
+    :param yarn_lock: yarn.lock file as a dictionary
+    :param dep: top-level dependency in package.json
+    :return: set of all dependencies reachable by dep
+    :rtype: set
+    """
+    yarn_lock_parsed = _expand_replacements(yarn_lock)
+    bfs_queue = deque([dep])
+    reachable_deps = set([dep])
```
akhmanova, line 84:
```suggestion
    bfs_queue = deque(dep)
    reachable_deps = set(dep)
```
Just nitpick

cachito/workers/pkg_managers/yarn.py
```diff
@@ -69,7 +70,33 @@ def get_yarn_proxy_repo_username(request_id):
...
+        current_dep = bfs_queue.popleft()
+        package = pyarn.lockfile.Package.from_dict(current_dep, yarn_lock_parsed[current_dep])
+        bfs_queue.extend(
+            [
+                f"{name}@{version}"
+                for name, version in package.dependencies.items()
+                if f"{name}@{version}" not in reachable_deps
+            ]
+        )
+        reachable_deps = reachable_deps.union(set(bfs_queue))
```
akhmanova, line 95: why don't add just `current_dep` to the set? We will reach that anyway

cachito/workers/pkg_managers/yarn.py
```diff
@@ -69,7 +70,33 @@ def get_yarn_proxy_repo_username(request_id):
...
+    :param yarn_lock: yarn.lock file as a dictionary
+    :param dep: top-level dependency in package.json
+    :return: set of all dependencies reachable by dep
+    :rtype: set
+    """
+    yarn_lock_parsed = _expand_replacements(yarn_lock)
+    bfs_queue = deque([dep])
+    reachable_deps = set([dep])
+    while bfs_queue:
+        current_dep = bfs_queue.popleft()
```
akhmanova, line 86:
```suggestion
        current_dep = bfs_queue.popleft()
        reachable_deps.add(current_dep)
```

dcho5: `reachable_deps.add(current_dep)` should still be in line 95 because of line 92.

cachito/workers/pkg_managers/yarn.py
dcho5: `deque()` and `set()` both require iterables as parameters. Passing strings through will treat them as an array of characters.

cachito/workers/pkg_managers/yarn.py
```diff
@@ -69,7 +70,33 @@ def get_yarn_proxy_repo_username(request_id):
     return f"cachito-yarn-{request_id}"
 
 
-def _get_deps(yarn_lock, file_deps_allowlist):
+def find_reachable_deps(yarn_lock, dep):
```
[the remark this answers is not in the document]
dcho5: Done!

cachito/workers/pkg_managers/yarn.py
akhmanova: sorry!

cachito/workers/pkg_managers/yarn.py
akhmanova: Hm, are you sure? In my opinion, they don't have any correlation. `current_dep` is a node in our "graph". If we visit this node, we need to add it to our result. 

in line 92 you check children of `current_dep`. We don't need to add them to our queue in case we already visited them.

akhmanova: But it doesn't matter actually! The time complexity is already better :)

akhmanova (approved):
LGTM. Good work!

cachito/workers/pkg_managers/yarn.py
```diff
@@ -69,7 +70,33 @@ def get_yarn_proxy_repo_username(request_id):
...
+    :rtype: set
+    """
+    yarn_lock_parsed = _expand_replacements(yarn_lock)
+    bfs_queue = deque([dep])
+    reachable_deps = set([dep])
+    while bfs_queue:
+        current_dep = bfs_queue.popleft()
+        package = pyarn.lockfile.Package.from_dict(current_dep, yarn_lock_parsed[current_dep])
+        bfs_queue.extend(
+            [
```
MartinBasti, line 89: can be generator expression

cachito/workers/pkg_managers/yarn.py
```diff
@@ -69,7 +70,33 @@ def get_yarn_proxy_repo_username(request_id):
...
+    Get a set of all dependencies reachable by a top-level dependency using BFS.
+
+    :param yarn_lock: yarn.lock file as a dictionary
+    :param dep: top-level dependency in package.json
+    :return: set of all dependencies reachable by dep
+    :rtype: set
+    """
+    yarn_lock_parsed = _expand_replacements(yarn_lock)
+    bfs_queue = deque([dep])
+    reachable_deps = set([dep])
```
MartinBasti, line 84: This looks like py2.6 way
```suggestion
    reachable_deps = {dep, }
```

cachito/workers/pkg_managers/yarn.py
```diff
@@ -90,10 +117,23 @@ def _get_deps(yarn_lock, file_deps_allowlist):
...
     deps = []
     nexus_replacements = {}
+    non_dev_deps = set()
+
+    for dep_type in ["dependencies", "peerDependencies", "optionalDependencies"]:
+        if dep_type not in package_json:
+            continue
+        for name, version in package_json[dep_type].items():
+            dep = f"{name}@{version}"
+            non_dev_deps = non_dev_deps.union(_find_reachable_deps(yarn_lock, dep))
```
MartinBasti, line 127: Why union and not just `.update()` ?  Why do you need create new instance every time and reallocate memory every time?

MartinBasti (commented):
Nitpicks

cachito/workers/pkg_managers/yarn.py
```diff
@@ -69,7 +70,33 @@ def get_yarn_proxy_repo_username(request_id):
     return f"cachito-yarn-{request_id}"
 
 
-def _get_deps(yarn_lock, file_deps_allowlist):
+def _find_reachable_deps(yarn_lock, dep):
```
brunoapimentel, line 73: Just pointing out that, since we're restarting the BFS for every non-dev top level dependency, we revisting every transitive dependency that is common to more than one top level dependency. Like:
```
A---- 
     |
     |----C   
     |
B----
``` 
C (and its dependency tree) will be visited twice in this case.
I don't think this is a big deal though, since the time for this algorithm to run is tiny fraction of all the time Cachito takes to process a request. :)

cachito/workers/pkg_managers/yarn.py
dcho5: Old habits! Yes, `update()` should be more efficient.

cachito/workers/pkg_managers/yarn.py
dcho5: I did think this as well. Would memoization save noticeable time for most Cachito requests?

cachito/workers/pkg_managers/yarn.py
dcho5: Could you elaborate? Aren't lines 90-92 already a generator expression?

cachito/workers/pkg_managers/yarn.py
dcho5: Solved!

cachito/workers/pkg_managers/yarn.py
MartinBasti: No, it's list comprehension. You are creating real list, and the that one is used. `.extend()` can consume any iterable, it should be faster with generator expression.

So you have to omit `[]`

dcho5: Got it. Learned something new!

cachito/workers/pkg_managers/yarn.py
```diff
@@ -90,10 +115,24 @@ def _get_deps(yarn_lock, file_deps_allowlist):
...
+            if dep not in non_dev_deps:
+                _find_reachable_deps(yarn_lock, dep, non_dev_deps)
 
     for dep_identifier, dep_data in yarn_lock.items():
         package = pyarn.lockfile.Package.from_dict(dep_identifier, dep_data)
 
+        dev = True
+        for non_dev_dep in non_dev_deps:
+            if non_dev_dep in dep_identifier:
+                dev = False
```
chmeliik, line 134: Is substring matching reliable enough? What about cases like this:
```python
>>> non_dev_deps = {"react@1.0.0"}
>>> "react@1.0.0" in "@types/react@1.0.0"
True
```

dcho5: I made a helper function `_split_lock_key` which splits the `dep_identifier` string into a list of dependencies (removes unnecessary " and splits string by ", "). WDYT?

i.e.
```
>>> dep_id_list = _split_lock_key('"@types/react@1.0.0, @types/react@1.2.0"')
>>> dep_id_list
["@types/react@1.0.0", "@types/react@1.2.0"]
>>> "react@1.0.0" in dep_id_list
False
>>> "@types/react@1.0.0" in dep_id_list
True
```

chmeliik: +1, that should work. How about using the helper function for `_expand_replacements` as well, to make sure they do this consistently? The current approach is to split on `","` and then strip whitespace, which might be slightly more reliable - but there is no handling for extra quotes, which would be good to have.

As a side note, `_expand_replacements` may deserve a name change, it is now used to expand the entire  yarn.lock. Maybe `_expand_yarn_lock_keys`?

chmeliik: And maybe, instead of iterating through all `non_dev_deps` and checking if they are in `dep_id_list`, would it be better to do it the other way around? The dep_id_list will usually have only one or two items, and since non_dev_deps is a set that should be constant lookup on average.

cachito/workers/pkg_managers/yarn.py
```diff
@@ -297,27 +347,27 @@ def _expand_replacements(nexus_replacements: Dict[str, dict]) -> Dict[str, dict]
...
         for multi_key in nexus_replacements
         for key in map(str.strip, multi_key.split(","))
     }
-    return expanded_replacements
+    return expanded_yarn_lock_keys
 
 
 def _match_to_new_version(
-    dep_name: str, dep_version: str, expanded_replacements: Dict[str, dict]
+    dep_name: str, dep_version: str, expanded_yarn_lock_keys: Dict[str, dict]
```
chmeliik, line 359: This rename, and many others below were probably unnecessary - this really is just a dict of Nexus replacements (a modified subset of the original yarn.lock), not the entire yarn.lock

chmeliik (commented):
LGTM, just minor nitpicks

cachito/workers/pkg_managers/yarn.py
```diff
@@ -297,27 +347,27 @@ def _expand_replacements(nexus_replacements: Dict[str, dict]) -> Dict[str, dict]
     :param dict nexus_replacements: a dict of nexus replacements which may contain N:1 keys
     :return: a dict of nexus replacements where all N:1 keys have been expanded to N 1:1 keys
     """
-    expanded_replacements = {
+    expanded_yarn_lock_keys = {
         key: nexus_replacements[multi_key]
         for multi_key in nexus_replacements
         for key in map(str.strip, multi_key.split(","))
```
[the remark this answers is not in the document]
dcho5: Must have missed it on the latest push, fixed!

chmeliik (approved):
LGTM, nice! :tada:

chmeliik:
This will also require a docs update, but it would likely be better to do that in a separate PR

MartinBasti (approved):

brunoapimentel (changes requested):
Could we at least have the README updated: https://github.com/release-engineering/cachito/blob/d429dc69d0ff4f088d7daa56223b85531726d2c3/README.md?plain=1#L671

Should be easy enough...

brunoapimentel (approved):

akhmanova:
wow! so many approvals! well done!

akhmanova:
Sorry!

PR #615, opened 2021-11-19
Change log level for failures when uploading to Nexus

ejegrova:
CLOUDBLD-7717

Failures do not have impact on the success of the request since Cachito
is able to recover from it.

### DIFF
--- a/cachito/workers/nexus.py
+++ b/cachito/workers/nexus.py
@@ -384,7 +384,7 @@ def upload_asset_only_component(repo_name, repo_type, component_path, to_nexus_h
     try:
         upload_component(params, payload, to_nexus_hoster)
     except CachitoError:
-        log.exception("Failed to upload %r to the %r Nexus repository", component_path, repo_type)
+        log.warning("Failed to upload %r to the %r Nexus repository", component_path, repo_type)
         raise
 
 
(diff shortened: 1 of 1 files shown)


serg-cymbaluk (approved):
LGTM

cachito/workers/nexus.py
```diff
@@ -384,7 +384,7 @@ def upload_asset_only_component(repo_name, repo_type, component_path, to_nexus_h
     try:
         upload_component(params, payload, to_nexus_hoster)
     except CachitoError:
-        log.exception("Failed to upload %r to the %r Nexus repository", component_path, repo_type)
+        log.warning("Failed to upload %r to the %r Nexus repository", component_path, repo_type)
```
akhmanova, line 387: @ejegrova Can you please specify a reason for this failure? As much I remember, it was because it's already in Nexus.

brunoapimentel: This exception will happen regarding any type Nexus failure: from network errors and outages to simply trying to upload a duplicate dependency. :(

That's why the log level is being changed to WARNING here... if it's an actual error, the exception will break the request's processing and be properly logged as an ERROR. Otherwise, Cachito will find its way around it and we will not see a false "error" being logged.

brunoapimentel (approved):

cachito/workers/nexus.py
ejegrova: This failure happens when `CachitoError` is raised within `upload_component` and that can happen in two cases: there's a connection error to Nexus (which will still be seen as error in log) or upload of the component failed (newly logged as warning). We can log it here as warning because when this exception reaches the top level, Cachito is logging it properly. Also Serhii will propose some changes to CachitoError so there will be a lot of changes later on.

PR #620, opened 2021-12-10
Fix a bug when processing Go repositories with submodules tags

brunoapimentel:
CLOUDBLD-3363

Cachito would fail to handle any repository that had a newer semver tag applied to a submodule. An example of a `git tag` output that would cause Cachito to fail is:

```
submodule/v1.0.1
v1.0.0
```

This change makes these tags to be ignored.

[n/a] OpenAPI schema is updated (if applicable)
[n/a] DB schema change has corresponding DB migration (if applicable)
[n/a] README updated (if worker configuration changed, or if applicable)


cachito/workers/pkg_managers/gomod.py
```diff
@@ -705,7 +705,7 @@ def _get_highest_semver_tag(repo, target_commit, major_version, all_reachable=Fa
                 "git",
                 "for-each-ref",
                 "--format",
-                "%(refname:lstrip=-1)",
+                "%(refname:lstrip=2)",
```
akhmanova, line 708: Did the code keep only the last part, right? And now you want to delete the first 2 parts.
correct me: you want to keep `module/version` from `foo/bar/module/version` instead of `version` as it was before.

brunoapimentel: You are correct. Keep in mind that we are filtering refs that have `refs/tags` in their path, so this is essentially striping `refs/tags` from the full ref path. E.g.:

```
refs/tags/submodule/version => submodule/version
refs/tags/version => version
refs/tags/some/other/stuff => some/other/stuff
```

brunoapimentel: What this does, effectively, is to make Cachito ignore refs that have anything else than the semvertag:
```
refs/tags/v0.0.0 => will be considered
refs/tags/submodule/v0.0.0 => will be ignored
```

chmeliik: I believe the correct behavior would be to consider only `refs/tags/v0.0.0` when processing the root module and consider only `refs/tags/submodule/v0.0.0` when processing `submodule`.

I'm not sure if it's OK for Cachito to always report the version of the main module. It would only produce the correct result if
a) neither the main module nor the submodule has any version tag
b) both the main module and the submodule have the same version tag

brunoapimentel: You're right. 

I was aiming only for a quick bugfix here, not to cover all the root module/submodule behavior. But it seems that this will affect submodule requests, since the version will then be ignored.

akhmanova:
What is it? ``` tests/test_workers/test_pkg_managers/golang_git_repo.tar.gz``` :)

brunoapimentel:
> What is it? ` tests/test_workers/test_pkg_managers/golang_git_repo.tar.gz` :)

It is a compressed git repository used for this unit test. It was there before, I just added a new commit, so I needed to re-add the file.

cachito/workers/pkg_managers/gomod.py
akhmanova: @brunoapimentel so we are waiting for a change for submodules, right? :)

brunoapimentel:
> I believe the correct behavior would be to consider only refs/tags/v0.0.0 when processing the root module and consider only 
> refs/tags/submodule/v0.0.0 when processing submodule.
I will close this PR and reopen it with an approach that covers this behavior.

brunoapimentel:
> I believe the correct behavior would be to consider only refs/tags/v0.0.0 when processing the root module and consider only 
> refs/tags/submodule/v0.0.0 when processing submodule.

I will close this PR and reopen it with an approach that covers this behavior.

brunoapimentel:
> I believe the correct behavior would be to consider only refs/tags/v0.0.0 when processing the root module and consider only 
> refs/tags/submodule/v0.0.0 when processing submodule.
I will close this PR and reopen it with an approach that covers this behavior.

PR #621, opened 2021-12-10
[Draft] Allow to use git+ssh protocol to fetch application sources

serg-cymbaluk:
CLOUDBLD-8038

Extend ~/.ssh/known_hosts with hosts specified by cachito_git_ssh_known_hosts config option.

### DIFF
--- a/cachito/workers/config.py
+++ b/cachito/workers/config.py
@@ -1,7 +1,9 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import logging
 import os
+import subprocess  # nosec
 import tempfile
+from pathlib import Path
 from typing import Dict, List, Optional
 
 import celery
@@ -32,6 +34,7 @@ class Config(object):
         },
     }
     cachito_deps_patch_batch_size = 50
+    cachito_git_ssh_known_hosts = ["github.com", "gitlab.com", "bitbucket.org"]
     cachito_gomod_download_max_tries = 5
     cachito_gomod_ignore_missing_gomod_file = True
     cachito_gomod_strict_vendor = False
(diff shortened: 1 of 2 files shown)


serg-cymbaluk:
Working on tests for that change.

cachito/workers/tasks/celery.py
```diff
@@ -22,6 +23,7 @@
 
 
 celeryd_init.connect(validate_celery_config)
+celeryd_init.connect(populate_ssh_known_hosts)
```
brunoapimentel, line 26: I'm wondering if this is gonna work when we deploy it to OCP. The `~/.ssh` directory is not writable in the pods.

arewm: You can set the `core.sshCommand` config to use a ssh config location of your choosing. https://stackoverflow.com/a/41219960

cachito/workers/config.py
```diff
@@ -253,6 +256,27 @@ def validate_celery_config(conf, **kwargs):
             )
 
 
+def populate_ssh_known_hosts(conf, **kwargs):
+    """
+    Populate ~/.ssh/known_hosts based on some predefined list of hosts.
+
+    The list of hosts can be provided using `cachito_git_ssh_known_hosts` configuration option.
```
arewm, line 263: Does this really protect against a MitM attack? How often does this get populated? On every deployment?

It feels like it would be "safer" to cache off the known_hosts _now_ and then just populate that. That way, we can ensure that it isn't changing without our knowledge.

serg-cymbaluk: > to cache off the known_hosts now and then just populate that

In the case, we can just mount the the ssh config and corresponding known_hosts file during the service configuration and pass them through `GIT_SSH_COMMAND` env variable (or `core.sshCommand` git option). So that patch is not needed at all.

WDYT, @brunoapimentel , @arewm ?

arewm: Yes, that works for me. It will also (presumably) allow any deployments to customize the known ssh hosts if there are also any non-public systems that they want to allow access to.

brunoapimentel: Makes sense to me. I think the known_hosts file can be added to the cachito-playbooks and mounted during the deployment, as you mentioned.

PR #622, opened 2021-12-13
Sets depth=1 to Git clone/fetch

brunoapimentel:
This is to avoid exhausting the resources while cloning very large repos, such as github.com/openshift/origin, which is consuming ~3GB memory during the cloning process.

In order to maintain compatibility, this behavior will only be applied if there is no need to keep the Git repository in the resulting tarball (i.e. `include-git-dir` flag is absent).

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -346,7 +346,11 @@ def create_request():
     error_callback = tasks.failed_request_callback.s(request.id)
     chain_tasks = [
         tasks.fetch_app_source.s(
-            request.repo, request.ref, request.id, "git-submodule" in pkg_manager_names
+            request.repo,
+            request.ref,
+            request.id,
+            "git-submodule" in pkg_manager_names,
+            "include-git-dir" in [flag.name for flag in request.flags],
         ).on_error(error_callback)
     ]
 
--- a/cachito/workers/scm.py
+++ b/cachito/workers/scm.py
@@ -143,24 +143,33 @@ class Git(SCM):
             os.unlink(self.sources_dir.archive_path)
             raise
 
-    def clone_and_archive(self, gitsubmodule=False):
+    def clone_and_archive(self, gitsubmodule=False, shallow=False):
         """
         Clone the git repository and create the compressed source archive.
 
         :param bool gitsubmodule: a bool to determine whether git submodules need to be processed.
+        :param bool shallow: determines if a shallow clone should be made (depth=1).
         :raises CachitoError: if cloning the repository fails or if the archive can't be created
         """
         with tempfile.TemporaryDirectory(prefix="cachito-") as temp_dir:
             log.debug("Cloning the Git repository from %s", self.url)
             clone_path = os.path.join(temp_dir, "repo")
+
+            kwargs = {
+                "no_checkout": True,
+                # Don't allow git to prompt for a username if we don't have access
+                "env": {"GIT_TERMINAL_PROMPT": "0"},
+            }
+
+            if shallow:
+                kwargs["depth"] = 1
+
             try:
-                repo = git.repo.Repo.clone_from(
-                    self.url,
-                    clone_path,
-                    no_checkout=True,
-                    # Don't allow git to prompt for a username if we don't have access
-                    env={"GIT_TERMINAL_PROMPT": "0"},
-                )
+                repo = git.repo.Repo.clone_from(self.url, clone_path, **kwargs)
+
+                if shallow:
+                    # Fetch the exact commit we need
+                    repo.remote().fetch(refspec=self.ref, depth=1)
             except:  # noqa E722
                 log.exception("Cloning the Git repository from %s failed", self.url)
                 raise CachitoError("Cloning the Git repository failed")
@@ -173,12 +182,13 @@ class Git(SCM):
             repo.git.gc("--prune=now")
             self._create_archive(repo.working_dir)
 
-    def update_and_archive(self, previous_archive, gitsubmodule=False):
+    def update_and_archive(self, previous_archive, gitsubmodule=False, shallow=False):
         """
         Update the existing Git history and create a source archive.
 
         :param str previous_archive: path to an archive file created before.
         :param bool gitsubmodule: a bool to determine whether git submodules need to be processed.
+        :param bool shallow: determines if only a single reference should be fetched.
         :raises CachitoError: if pulling the Git history from the remote repo or
             the checkout of the target Git ref fails.
         """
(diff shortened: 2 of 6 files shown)


chmeliik:
Will golang version resolution still work correctly?

serg-cymbaluk:
@chmeliik 
> Will golang version resolution still work correctly?
It might be correct, since the function does 'fetch' operation itself.
What kind of issue you expect there?

chmeliik:
> It might be correct, since the 'get_golang_version' function does 'fetch' operation itself. What kind of issue you expect there?

Ah, I didn't realize we pass `update_tags=True` there. Should work fine then :+1:

chmeliik:
Another potential problem is that the `include-git-dir` flag exists, and by cloning with depth 1, we change the request output when that flag is used. Maybe that's fine, I don't know if users rely on the full clone, but I would be careful

chmeliik:
> I'm thinking we could either: 1- Use shallow clones only when `include-git-dir` is absent 2- Have a new flag to force shallow clones (it is only a single repo that is causing a crash)

+1 for the first option

cachito/workers/scm.py
```diff
@@ -158,9 +158,13 @@ def clone_and_archive(self, gitsubmodule=False):
                     self.url,
                     clone_path,
                     no_checkout=True,
+                    # Clone with depth 1 to save resources on large repos
+                    depth=1,
                     # Don't allow git to prompt for a username if we don't have access
                     env={"GIT_TERMINAL_PROMPT": "0"},
                 )
+
+                repo.remote().fetch(refspec=self.ref, depth=1)
```
akhmanova, line 167: Why do we need to have this line twice?

brunoapimentel:
> I'm thinking we could either:
> 1- Use shallow clones only when `include-git-dir` is absent 
> 2- Have a new flag to force shallow clones (it is only a single repo that is causing a crash)

I started writing approach `#1`, but I felt it looked hackish, like an unexpected behavior. I'm thinking `#2` would be a better solution, since the user will need to explicitly ask for a shallow clone.

And if we go with `#2`, it would not affect how `include-git-dir` works now: it'll simply include the type of clone that was asked, deep (default) or shallow (via the new flag).

Wdyt, @serg-cymbaluk @ejegrova @akhmanova?

cachito/workers/scm.py
brunoapimentel: Because the shallow clone always defaults to the lastest master commit, so we need to fetch the commit we're actually building.

serg-cymbaluk:
> I started writing approach #1, but I felt it looked hackish, like an unexpected behavior. I'm thinking #2 would be a better solution, since the user will need to explicitly ask for a shallow clone.

It is not hackish, since the purpose of that flag is to keep the git history. And if it is not needed, we can apply `--depth=1` with no doubts.

The second one sounds more like a workaround.

brunoapimentel:
> It is not hackish, since the purpose of that flag is to keep the git history. And if it is not needed, we can apply `--depth=1` with no doubts.
> 
> The second one sounds more like a workaround.

Well, the purpose of the flag is to keep the whole git directory, although it seems the users are more interested in the refs in there.

I still feel like it is less clear to see why `include-git-dir` implicates a full clone (and it's absence a shallow clone) than having a flag that explicitly says `shallow-clone`, which is pretty self-explanatory. My overall feeling is more in the lines of "explicit is better than in implicit".

But I see your point, if we interpret the flag as `include-git-history`, it would make perfect sense. I could go with either: `#2` seems more explicit, but `#1` seems a little easier to implement (no API changes).

Maybe the feeling of hackish derived more from all the code layers I hade to "pierce" to pass a new argument... argh :D

cachito/workers/scm.py
akhmanova: Aaah, I just forgot there are some lines between 170 and 194. My bad!

akhmanova:
@brunoapimentel I am fine with the first approach too. I would not create another flag for this purpose, just renaming will be fine. Let's just update docs and say that without this flag only the last commit will be cloned.

cachito/workers/scm.py
```diff
@@ -143,7 +143,7 @@ def _create_archive(self, from_dir):
             os.unlink(self.sources_dir.archive_path)
             raise
 
-    def clone_and_archive(self, gitsubmodule=False):
+    def clone_and_archive(self, gitsubmodule=False, include_git_dir=False):
```
akhmanova, line 146: Can we add a line about `include_git_dir` param?

cachito/workers/scm.py
```diff
@@ -153,14 +153,24 @@ def clone_and_archive(self, gitsubmodule=False):
...
-                repo = git.repo.Repo.clone_from(
-                    self.url,
-                    clone_path,
-                    no_checkout=True,
-                    # Don't allow git to prompt for a username if we don't have access
-                    env={"GIT_TERMINAL_PROMPT": "0"},
-                )
+                repo = git.repo.Repo.clone_from(self.url, clone_path, **kwargs)
+
+                if include_git_dir:
```
akhmanova, line 171: And what if we need a special commit and it's not the last one (and `include_git_dir` is set to False)?

tests/test_workers/test_scm.py
```diff
@@ -89,6 +94,9 @@ def test_clone_and_archive(
         mock_ugs.assert_called_once_with(mock_clone.return_value)
     else:
         mock_ugs.assert_not_called()
+    #
```
akhmanova, line 97: should it be empty?

cachito/workers/scm.py
brunoapimentel: If `include_git_dir` is set to `False`, a deep clone will be made, which means all commits will already be there. This is the current behavior, by the way.

cachito/workers/scm.py
brunoapimentel: You mean in the docstring? Sure, I missed that one.

tests/test_workers/test_scm.py
brunoapimentel: It should not. Holiday mood was impeding my brain. :D

cachito/web/api_v1.py
```diff
@@ -346,7 +346,11 @@ def create_request():
     error_callback = tasks.failed_request_callback.s(request.id)
     chain_tasks = [
         tasks.fetch_app_source.s(
-            request.repo, request.ref, request.id, "git-submodule" in pkg_manager_names
+            request.repo,
+            request.ref,
+            request.id,
+            "git-submodule" in pkg_manager_names,
+            "include-git-dir" in request.flags,
```
serg-cymbaluk, line 353: It will not work. The request.flags is a list of objects from DB but not strings.

brunoapimentel: FTFY!

serg-cymbaluk (approved):
LGTM

ejegrova (approved):
LGTM

PR #623, opened 2021-12-15
Delete username unique index from User table

dcho5:
CLOUDBLD-7737


cachito/web/migrations/versions/02229e089b24_delete_user_username_index.py
```diff
@@ -0,0 +1,26 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table("user", schema=None) as batch_op:
+        batch_op.drop_index("ix_user_username")
+
+
+def downgrade():
```
akhmanova, line 24: to downgrade it we will not be able to get all the user names, because they were deleted, right? I think we will need to get a backup before releasing it.

akhmanova (approved):

cachito/web/migrations/versions/02229e089b24_delete_user_username_index.py
brunoapimentel: The only thing that's being deleted here is the index, no changes will be made to the actual data :)

brunoapimentel:
I did a double check and confirmed the only thing dropped is actually the Unique **index**, the Unique **constraint** is still there:

Before this PR:
```
Indexes:
    "user_pkey" PRIMARY KEY, btree (id)
    "ix_user_username" UNIQUE, btree (username)
    "user_username_key" UNIQUE CONSTRAINT, btree (username)
 ```
 
After this PR:
```
Indexes:
    "user_pkey" PRIMARY KEY, btree (id)
    "user_username_key" UNIQUE CONSTRAINT, btree (username)
 ```

brunoapimentel:
I'm thinking we could improve the commit message to make it clear why the index was dropped (it was redundant). Wdyt?

PR #625, opened 2021-12-22
Properly handle Git tags applied to Go submodules

brunoapimentel:
Currently, Cachito will ignore tags applied to submodules (e.g. submodule/v1.0.0). There is a corner case in which, if the newest version tag is applied to a submodule, the request will fail entirely. This [request](
https://cachito.engineering.redhat.com/api/v1/requests/174302) is an example of this scenario:

```
providers/v0.21.1 # newest tag is on a submodule
v0.21.0
providers/v0.21.0
``` 

The correct behavior should be to only consider tags applied specifically to the path processed, which is:

tag | path
--------|------
v0.1.0 | /
sub/v0.1.0 | /sub
sub/a/v0.1.1 | /sub/a
xpto/v0.1.1 | /xpto


tests/test_workers/test_pkg_managers/test_gomod.py
```diff
@@ -528,61 +528,118 @@ def test_go_list_cmd_failure(
...
+        ),
+        # Submodule has a semver tag applied to it
+        ("/v2", "f3ee3a4a394fb44b055ed5710b8145e6e98c0d55", "v2.5.1", "submodule"),
+        # A commit after a submodule tag
+        (
+            "/v2",
+            "cc6c9f554c0982786ff9e077c2b37c178e46828c",
+            "v2.5.2-0.20211223131312-cc6c9f554c09",
+            "submodule",
+        ),
```
akhmanova, line 631: Can we add more subpath options? Or don't we need that?

brunoapimentel: I was thinking of adding an integration test to cover this behavior. Any ideas of what we could add to the unit test, though?

akhmanova: great idea! I am not sure I got how we use submodule version tags in the test. submodule version tag is in [624 line](https://github.com/release-engineering/cachito/pull/625/files#diff-6411503dd66c67ab75b7597857d34862dd5a785a45802042b62d53df74a73c45R624)? 
And why do we have `None` instead of `submodule` in the test data sometimes? :smile_cat:

akhmanova (approved):
LGTM. I just not getting testing data separation :D

tests/test_workers/test_pkg_managers/test_gomod.py
brunoapimentel: It's easier to understand if you unpack the test repo in your computer and look at the tags there: https://github.com/release-engineering/cachito/blob/master/tests/test_workers/test_pkg_managers/golang_git_repo.tar.gz.

A normal semantic version tag looks like `v0.0.0`, which means it is applied to the root module of that repository.
A submodule tag looks like `submodule/v0.0.0`, which means it is applied to the submodule placed exaclty in the path defined in the tag (in this case, the `submodule`).

We can also have deeper nested submodules, such as `folder/anotherfolder/submodule` or even `this/is/a/very/deeply/nested/submodule`. :shrug:

brunoapimentel: The `submodule` parameter in the tests refers to which path is currently being processed by Cachito (defined in the `packages` item, when the request is created via API).

We will only consider tags that match exactly the path being processed at the moment.

cachito/workers/pkg_managers/gomod.py
```diff
@@ -723,37 +727,49 @@ def _get_highest_semver_tag(repo, target_commit, major_version, all_reachable=Fa
...
-            continue
-
         try:
-            # Exclude the 'v' prefix since this is required by Go, but it is seen as invalid by
-            # the semver Python package
-            parsed_version = semver.parse_version_info(tag_name[1:])
+            semantic_version = _get_semantic_version_from_tag(tag_name, subpath)
         except ValueError:
             log.debug(not_semver_tag_msg, tag_name)
             continue
```
chmeliik, line 734: Where are we filtering the tags that match the subpath? Is it here or did I miss something else?

brunoapimentel: Well, the `_get_semantic_version_from_tag` will try to remove the `subpath` from a tag, so in case we have:

`sub1/v0.1.1`
`sub2/v0.1.1`

And we're processin `sub1`, `sub2/v0.1.1` will not be properly parsed and will be ignored. 

But now I'm thinking... if the root module has semver tags, it will also fail because the script will not replace the `v`, although this behavior is not obvious.

I will add some more unit test cases to have this covered... might refactor things a bit.

cachito/workers/pkg_managers/gomod.py
```diff
@@ -148,6 +148,8 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
     :param list dep_replacements: dependency replacements with the keys "name" and "version"; this
         results in a series of `go mod edit -replace` commands
     :param dict git_dir_path: the full path to the application's git repository
+    :param str subpath: path to the module, relative to the root repository folder; "None" in case
+        the root module is being resolved
```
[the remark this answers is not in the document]
brunoapimentel: I'm not sure... I tried to avoid having this parameter as well, but I did not figure out how. I'll check if we can remove it.

cachito/workers/pkg_managers/gomod.py
chmeliik: +1, also if there's an incorrect root tag (like `1.0.1`, missing the `v`) and you're processing `submodule/v1.0.0`, I think the root tag would get parsed successfully (and be incorrectly applied to the submodule)

cachito/workers/pkg_managers/gomod.py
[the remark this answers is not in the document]
chmeliik: It's probably needed for all the other methods, but for resolve_gomod specifically, I think you could just get the relative path between the two paths

serg-cymbaluk (dismissed):
LGTM

cachito/workers/pkg_managers/gomod.py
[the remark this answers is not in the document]
brunoapimentel: Well, it turns out that app_source_path is a `str`. But I went with this change anyway, since it allows us to not need to check if subpath == "." (which is what Cachito does when we're processing the root module).

cachito/workers/pkg_managers/gomod.py
brunoapimentel: Great catch! That was indeed a bug. I changed the filtering logic so it is more explicit now.

chmeliik (commented):
LGTM :+1:

Just `black` has something to say about formatting, and other CI is broken

chmeliik (approved):
LGTM once CI passes

serg-cymbaluk (approved):

PR #631, opened 2022-01-06
Update py-amqp to latest version 5.0.9

ejegrova:
JIRA: CLOUDBLD-4312

Updates of py-amqp were causing cachito workers to fail to get local certificates.
Previously we pinned the version 5.0.2. With version 5.0.9 errors are not happening anymore.

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -10,9 +10,9 @@ alembic==1.5.6 \
     # via
     #   -r requirements-web.txt
     #   flask-migrate
-amqp==5.0.2 \
-    --hash=sha256:5b9062d5c0812335c75434bf17ce33d7a20ecfedaa0733faec7379868eb4068a \
-    --hash=sha256:fcd5b3baeeb7fc19b3486ff6d10543099d40ae1f5c9196eae695d1cde1b2f784
+amqp==5.0.9 \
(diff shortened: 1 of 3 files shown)


requirements-test.txt
```diff
@@ -568,9 +622,9 @@ requests==2.26.0 \
     # via
     #   -r requirements.txt
     #   requests-kerberos
-requests-kerberos==0.12.0 \
-    --hash=sha256:5733abc0b6524815f6fc72d5c0ec9f3fb89137b852adea2a461c45931f5675e0 \
-    --hash=sha256:9d21f15241c53c2ad47e813138b9aee4b9acdd04b82048c4388ade15f40a52fd
+requests-kerberos==0.13.0 \
```
serg-cymbaluk, line 625: @ejegrova 

Could you please separate those kerberos related changes and the amqp version bump into different commits?

The point is that the requests-kerberos update brings couple more dependencies. And it would be hard to recognize regressions.

ejegrova: splitted the commits, check it please

serg-cymbaluk (approved):
Thank you!
LGTM

brunoapimentel (approved):
LGTM :rocket:

PR #634, opened 2022-01-11
Revert "Uses shallow clones as default for Git operations"

brunoapimentel:
We are temporarily reverting this commit since we found that it causes Cachito to incorrectly infer Go versions in some corner cases. We're still figuring out a fix for this.

CLOUDBLD-8544


serg-cymbaluk (approved):
LGTM

MartinBasti (approved):

PR #636, opened 2022-01-12
Rollback pyamqp version bump

brunoapimentel:
CLOUDBLD-4312

The issue that happens in python3-amqp>=5.0.2 is still present, and causes Cachito to not be able to properly resolve the certificates. The issue was identified while the service was deployed to Openshift, and the error probably does not happen in a local environment.


brunoapimentel:
Converted to draft since we also needed to rollback the Celery version.

brunoapimentel:
Closed as we won't need this anymore. Issue has been fixed by #637.

PR #638, opened 2022-01-13
Add force-gomod-tidy flag to force run go mod tidy

dcho5:
CLOUDBLD-8090

The reason for adding this flag is because of a dependency pruning feature
added to Go 1.17. For compatibility reasons, the "go mod tidy" command
always runs the algorithm of the previous Go version. As a result, the
dependencies fetched with "go mod download" differ from the ones traversed
by "go mod tidy". This makes it impossible to run "go mod tidy" pointing the
GOCACHE to the tarball generated by Cachito on Go 1.17.

### DIFF
--- a/cachito/web/migrations/versions/418241dba06c_add_force_gomod_tidy_flag.py
+++ b/cachito/web/migrations/versions/418241dba06c_add_force_gomod_tidy_flag.py
@@ -1,0 +1,45 @@
+"""Add the force-gomod-tidy flag
+
+Revision ID: 418241dba06c
+Revises: 01bb0873ddcb
+Create Date: 2022-01-18 07:24:31.927867
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "418241dba06c"
+down_revision = "01bb0873ddcb"
+branch_labels = None
+depends_on = None
+
+
+flag_table = sa.Table(
+    "flag",
+    sa.MetaData(),
+    sa.Column("id", sa.Integer(), primary_key=True),
+    sa.Column("name", sa.String(), nullable=False),
+    sa.Column("active", sa.Boolean(), nullable=False, default=True),
+)
+
+
+def upgrade():
+    connection = op.get_bind()
+    res = connection.execute(
+        flag_table.select().where(flag_table.c.name == "force-gomod-tidy")
+    ).fetchone()
+    if res is None:
+        connection.execute(flag_table.insert().values(name="force-gomod-tidy", active=True))
+    else:
+        connection.execute(
+            flag_table.update().where(flag_table.c.name == "force-gomod-tidy").values(active=True)
+        )
+
+
+def downgrade():
+    connection = op.get_bind()
+    connection.execute(
+        flag_table.update().where(flag_table.c.name == "force-gomod-tidy").values(active=False)
+    )
--- a/cachito/workers/pkg_managers/gomod.py
+++ b/cachito/workers/pkg_managers/gomod.py
@@ -198,7 +198,7 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
         else:
             log.info("Downloading the gomod dependencies")
             run_download_cmd(("go", "mod", "download"), run_params)
-        if dep_replacements:
+        if "force-gomod-tidy" in flags or dep_replacements:
             run_gomod_cmd(("go", "mod", "tidy"), run_params)
 
         # main module
--- a/tests/integration/test_packages.py
+++ b/tests/integration/test_packages.py
@@ -28,6 +28,8 @@ from . import utils
         ("gomod_packages", "vendored_with_flag"),
         ("gomod_packages", "implicit_gomod"),
         ("gomod_packages", "missing_gomod"),
+        ("gomod_packages", "force_tidy_without_deps"),
+        ("gomod_packages", "force_tidy_vendored"),
         ("gomod_vendor_check", "correct_vendor"),
         ("gomod_vendor_check", "no_vendor"),
         ("npm_packages", "without_deps"),
--- a/tests/test_workers/test_pkg_managers/test_gomod.py
+++ b/tests/test_workers/test_pkg_managers/test_gomod.py
@@ -230,6 +230,7 @@ def _generate_mock_cmd_output(error_pkg="github.com/pkg/errors v1.0.0"):
     ),
 )
 @pytest.mark.parametrize("cgo_disable", [False, True])
+@pytest.mark.parametrize("force_gomod_tidy", [False, True])
 @mock.patch("cachito.workers.pkg_managers.gomod.get_golang_version")
 @mock.patch("cachito.workers.pkg_managers.gomod.GoCacheTemporaryDirectory")
 @mock.patch("cachito.workers.pkg_managers.gomod._merge_bundle_dirs")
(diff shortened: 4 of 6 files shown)


dcho5:
A couple blockers I am facing:

1. The ticket says, `Add a tidy_compat configuration option, that, when set, controls the value of the -compat flag passed to the go mod tidy invocation.` I've implemented `tidy_compat` temporarily as a local variable, but how should this new variable be passed? As an additional input parameter to `resolve_gomod`?
2. Tests for `test_resolve_gomod_vendor_dependencies` and `test_resolve_gomod_no_deps` are both failing as seen [here](https://github.com/containerbuildsystem/cachito/runs/4801327308?check_suite_focus=true). What should be done about this?

tests/test_workers/test_pkg_managers/test_gomod.py
```diff
@@ -266,8 +266,7 @@ def test_resolve_gomod(
     if dep_replacement:
         run_side_effects.append(mock.Mock(returncode=0, stdout=None))  # go mod edit -replace
     run_side_effects.append(mock.Mock(returncode=0, stdout=None))  # go mod download
-    if dep_replacement:
-        run_side_effects.append(mock.Mock(returncode=0, stdout=None))  # go mod tidy
```
akhmanova, line 270: Maybe it's failing because you don't need to run `go mod tidy` always? Only when the list `dep_replacement` is set

cachito/workers/pkg_managers/gomod.py
```diff
@@ -198,7 +198,10 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
         else:
             log.info("Downloading the gomod dependencies")
             run_download_cmd(("go", "mod", "download"), run_params)
-        if dep_replacements:
+        tidy_compat = "1.17"
```
akhmanova, line 201: > The ticket says, Add a tidy_compat configuration option, that, when set, controls the value of the -compat flag passed to the go mod tidy invocation. I've implemented tidy_compat temporarily as a local variable, but how should this new variable be passed? As an additional input parameter to resolve_gomod?

There are 2 options: 
1. use a configuration inside:
https://github.com/containerbuildsystem/cachito/blob/master/cachito/workers/config.py
2. use flags in a request

I would like to see the second option, so it will be up to our users to set. 
@brunoapimentel @serg-cymbaluk @ejegrova WDYT?

tests/test_workers/test_pkg_managers/test_gomod.py
akhmanova: Oh, I see it's a part of the story, I am sorry, just blind guess

cachito/workers/pkg_managers/gomod.py
dcho5: I believe a mix of both options would work well. We should have a default value for this variable even if the user doesn't explicitly provide the value.

dcho5: The default being the newest version of Go, of course.

dcho5: Upon reviewing, we shouldn't store a default value in `config.py`. The default behavior without an explicit `tidy_compat` value should be a simple `go mod tidy` command without `-compat=`.

tests/test_workers/test_pkg_managers/test_gomod.py
dcho5: For reference, these are the test cases:
https://github.com/containerbuildsystem/cachito/runs/4815799205?check_suite_focus=true#step:6:119
https://github.com/containerbuildsystem/cachito/runs/4815799205?check_suite_focus=true#step:6:213

cachito/workers/pkg_managers/gomod.py
brunoapimentel: It should be a new optional flag to be used during the creation of the request. That way, the user will have control on how/if it's used.

cachito/web/migrations/versions/84d67d85ed43_add_tidy_compat.py
```diff
@@ -0,0 +1,26 @@
+"""Add tidy_compat option for user requests
```
brunoapimentel, line 1: The Request table already has a NxN relation to the Flag table, so I don't think we'll need to add an extra field for that.

tests/test_workers/test_pkg_managers/test_gomod.py
brunoapimentel: As far as I understand, using `go mod tidy` will make changes to the go.mod file, and that behavior might not be wanted in all cases.  

https://go.dev/ref/mod#go-mod-tidy

So, I think we should only run `go mod tidy` in cases the new flag is present (and the current cases in which it is already run, `go-mod-vendor` if I'm not mistaken). This way, we keep backwards compatibility.

akhmanova:
@dcho5 Thank you for working on it! And thank you for your explanations. I would like to suggest separate code & integration tests changes in two commits. In addition, you can add more information about this change into a commit message (not the first line, but the commit body). 

I checked the DB migration file, unit & integration tests. Looks good to me.

MartinBasti:
> I would like to suggest separate code & integration tests changes in two commits

But why? It fixes tests so after checkout to this particular commit, tests won't be broken. Tests should be green on every commit, it shouldn't work in a way that test are fixed in followup commit

tests/integration/test_packages.py
```diff
@@ -28,6 +28,8 @@
         ("gomod_packages", "vendored_with_flag"),
         ("gomod_packages", "implicit_gomod"),
         ("gomod_packages", "missing_gomod"),
+        ("gomod_packages", "force_tidy_without_deps"),
+        ("gomod_packages", "force_tidy_vendored"),
```
akhmanova, line 32: @MartinBasti 
>But why? It fixes tests so after checkout to this particular commit, tests won't be broken. Tests should be green on every commit, it shouldn't work in a way that test are fixed in followup commit

Integration tests will not be broken because there were no tests for this flag yet. It's fine not to separate work but I hoped it will be easier to read.

cachito/web/migrations/versions/418241dba06c_add_force_gomod_tidy_flag.py
```diff
@@ -0,0 +1,43 @@
...
+        connection.execute(flag_table.insert().values(name="force-gomod-tidy", active=True))
+    else:
+        connection.execute(
+            flag_table.update().where(flag_table.c.name == "force-gomod-tidy").values(active=True)
+        )
+
+
+def downgrade():
+    connection = op.get_bind()
+    connection.execute(flag_table.update().values(name="force-gomod-tidy", active=False))
```
serg-cymbaluk, line 43: I guess the statement should be the same as the one above with `.where(flag_table.c.name == "force-gomod-tidy")` filter.

serg-cymbaluk (approved):
LGTM

brunoapimentel:
Maybe we could comment on the reason for this flag to exist on the commit message. I suggest adding something like:
```
The reason for adding this flag is because of a dependency pruning feature added to Go 1.17. For compatibility reasons, the "go mod tidy" command always runs the algorithm of the previous Go version. As a result, the dependencies fetched with "go mod download" differ from the ones traversed by "go mod tidy".

This makes it impossible to run "go mod tidy" pointing the GOCACHE to the tarball generated by Cachito on Go 1.17.
```

Besides that, LGTM :rocket:

brunoapimentel (approved):

PR #639, opened 2022-01-21
Ignore npmv7 workspace item from dependencies list

ejegrova:
Workspace is a new feature introduced by npm v7. When the workspace
is enabled, there will be an item recorded under dependencies object
within package-lock.json, and the item has version value
with prefix "file:". Such item can be skipped in order to not confuse
Cachito to treat it as a normal dependency.

CLOUDBLD-6981

### DIFF
--- a/cachito/workers/pkg_managers/npm.py
+++ b/cachito/workers/pkg_managers/npm.py
@@ -25,7 +25,7 @@ __all__ = [
 log = logging.getLogger(__name__)
 
 
-def _get_deps(package_lock_deps, file_deps_allowlist, _name_to_deps=None, workspaces=None):
+def _get_deps(package_lock_deps, file_deps_allowlist, name_to_deps=None, workspaces=None):
     """
     Get a mapping of dependencies to all versions of the dependency.
 
@@ -46,7 +46,7 @@ def _get_deps(package_lock_deps, file_deps_allowlist, _name_to_deps=None, worksp
     :param dict package_lock_deps: the value of a "dependencies" key in a package-lock.json file
     :param set file_deps_allowlist: an allow list of dependencies that are allowed to be "file"
         dependencies and should be ignored since they are implementation details
-    :param dict _name_to_deps: the current mapping of dependencies; this is not meant to be set
+    :param dict name_to_deps: the current mapping of dependencies; this is not meant to be set
         by the caller
     :param list workspaces: package workspaces defined in package-lock.json
     :return: a tuple with the first item as the mapping of dependencies where each key is a
@@ -56,8 +56,8 @@ def _get_deps(package_lock_deps, file_deps_allowlist, _name_to_deps=None, worksp
     :rtype: (dict, list)
     :raise CachitoError: if the lock file contains a dependency from an unsupported location
     """
-    if _name_to_deps is None:
-        _name_to_deps = {}
+    if name_to_deps is None:
+        name_to_deps = {}
     elif workspaces is None:
         workspaces = []
 
@@ -99,8 +99,8 @@ def _get_deps(package_lock_deps, file_deps_allowlist, _name_to_deps=None, worksp
             info.clear()
             info.update(nexus_replacement)
 
-        _name_to_deps.setdefault(name, [])
-        for d in _name_to_deps[name]:
+        name_to_deps.setdefault(name, [])
+        for d in name_to_deps[name]:
             if d["version"] == dep["version"]:
                 # If a duplicate version was found but this one isn't bundled, then mark the
                 # dependency as not bundled so it's included individually in the deps directory
@@ -112,18 +112,18 @@ def _get_deps(package_lock_deps, file_deps_allowlist, _name_to_deps=None, worksp
                     d["dev"] = False
                 break
         else:
-            _name_to_deps[name].append(dep)
+            name_to_deps[name].append(dep)
 
         if "dependencies" in info:
             _, returned_nexus_replacements = _get_deps(
-                info["dependencies"], file_deps_allowlist, _name_to_deps
+                info["dependencies"], file_deps_allowlist, name_to_deps
             )
             # If any of the dependencies were non-registry dependencies, replace the requires to be
             # the version in Nexus
             for name, version in returned_nexus_replacements:
                 info["requires"][name] = version
 
-    return _name_to_deps, nexus_replacements
+    return name_to_deps, nexus_replacements
 
 
 def convert_to_nexus_hosted(dep_name, dep_info):
--- a/cachito/workers/pkg_managers/npm.py
+++ b/cachito/workers/pkg_managers/npm.py
@@ -25,7 +25,7 @@ __all__ = [
 log = logging.getLogger(__name__)
 
 
-def _get_deps(package_lock_deps, file_deps_allowlist, _name_to_deps=None):
+def _get_deps(package_lock_deps, file_deps_allowlist, _name_to_deps=None, workspaces=None):
     """
     Get a mapping of dependencies to all versions of the dependency.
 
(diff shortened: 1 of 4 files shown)


cachito/workers/pkg_managers/npm.py
```diff
@@ -68,6 +73,11 @@ def _get_deps(package_lock_deps, file_deps_allowlist, _name_to_deps=None):
         # output of this function to download the dependencies will ignore this dependency.
         if info["version"].startswith("file:") and name in file_deps_allowlist:
             log.info("The dependency %r is an allowed exception", info)
+        #
+        elif info["version"].startswith("file:") and package_lock_lockfileversion >= 2:
```
serg-cymbaluk, line 77: If the workspaces is not empty, do we need to care about the lock file version?
Can we just check it before calling `_get_deps` whether the lockfile consistent or not and then pass only `workspaces`?

ejegrova: good point, thanks

cachito/workers/pkg_managers/npm.py
```diff
@@ -25,7 +25,7 @@
 log = logging.getLogger(__name__)
 
 
-def _get_deps(package_lock_deps, file_deps_allowlist, _name_to_deps=None):
+def _get_deps(package_lock_deps, file_deps_allowlist, package_lock_lockfileversion, workspaces, _name_to_deps=None):
```
brunoapimentel, line 28: Just be careful that this function is also clalled in line 121, and adding these new mandatory args is breaking the code when it reaches there.

cachito/workers/pkg_managers/npm.py
brunoapimentel: Or maybe you could make workspaces an optional parameter:
```
def _get_deps(package_lock_deps, file_deps_allowlist, workspaces=None, _name_to_deps=None):
```
And only fill it when lockfile_version >=2. Workspaces are a feature that should only be expected on v2, so I see no trouble reflecting this in the code. 

Or we could also go with Serhii's suggestion and ignore the lockfile_version altogether, just be sure there's no corner case we're ignoring here. I think the baseline is avoiding one additional parameter being passed down to this function.

cachito/workers/pkg_managers/npm.py
ejegrova: Sure, I just wanted your opinions first :)

cachito/workers/pkg_managers/npm.py
ejegrova: Oops, I forgot to push the changes. I changed the parameters exactly like you have it and added the check for the version in `get_package_and_deps`.

tests/test_workers/test_pkg_managers/test_npm.py
```diff
@@ -341,6 +341,17 @@ def test_get_deps_allowlisted_file_dep():
     assert replacements == []
 
 
+def test_get_deps_worspaces():
```
akhmanova, line 344: Can we get a couple more tests if you think it worth it :) For example: a test with a few elements in `workspaces` (all of them are part of `package_lock_deps`, not all of them)

cachito/workers/pkg_managers/npm.py
```diff
@@ -25,7 +25,7 @@
 log = logging.getLogger(__name__)
 
 
-def _get_deps(package_lock_deps, file_deps_allowlist, _name_to_deps=None):
+def _get_deps(package_lock_deps, file_deps_allowlist, _name_to_deps=None, workspaces=None):
```
akhmanova, line 28: Just a boring comment: maybe it's better to have a workspace before `_name_to_deps`.

tests/test_workers/test_pkg_managers/test_npm.py
```diff
@@ -462,7 +481,7 @@ def test_get_package_and_deps_dep_replacements(package_lock_deps, package_and_de
         },
     }
 
-    def _mock_get_deps(_deps, file_deps_allowlist):
+    def _mock_get_deps(_deps, file_deps_allowlist, workspaces):
```
akhmanova, line 484: do we want to test `"lockfileVersion": 2` as well?

cachito/workers/pkg_managers/npm.py
```diff
@@ -68,6 +71,10 @@ def _get_deps(package_lock_deps, file_deps_allowlist, _name_to_deps=None):
         # output of this function to download the dependencies will ignore this dependency.
         if info["version"].startswith("file:") and name in file_deps_allowlist:
             log.info("The dependency %r is an allowed exception", info)
+        # If there is npmv7 workspace entry in dependencies of package-lock.json, skip it.
```
brunoapimentel, line 74: Just a minor nitpick, but I'd leave the "v7" part out of this comment and the info log below. This is because workspaces are a permanent feature of NPM, even when v8, v9, etc. come around.

cachito/workers/pkg_managers/npm.py
brunoapimentel: I don't get why we have a parameter started with "_"... those are meant to imply private functions or attributes, and a parameter's scope is always the function :man_shrugging:

tests/test_workers/test_pkg_managers/test_npm.py
ejegrova: I don't think it's necessary, I did the parametrization for both lockfile versions in `test_get_package_and_deps` (line 413). This test `test_get_package_and_deps_dep_replacements` is testing the same method but for dependency replacements, also there's no changes regarding worspaces. But if you think it will be better, I can add it there.

tests/test_workers/test_pkg_managers/test_npm.py
ejegrova: sure, good idea :)

cachito/workers/pkg_managers/npm.py
ejegrova: right, good catch

cachito/workers/pkg_managers/npm.py
ejegrova: Changing the order will mess with the calls to this method. `_name_to_deps` is not used as parameter with `_name_to_deps=.....` on few places so the order matters. I also don't get why is the private attribute used. Maybe I can create separate commit to address this. Wdyt?

ejegrova:
> LGTM rocket after the comments left are addressed.
> 
> I do feel this story needs an integration test, wdyt about adding it to the acceptance criteria? That can be implemented in a separate PR, though.

There is one integration test ready ("npm_packages", "workspaces"), do you have some specific integration test in mind?

brunoapimentel:
> There is one integration test ready ("npm_packages", "workspaces"), do you have some specific integration test in mind?

Nevermind me, my brain was half dead when I wrote this comment. I saw the yml, but somehow thought it would feed the unit tests :facepalm:.

I do suggest improving the README on the workspaces test repo though :smile:. Maybe use this as an example: https://github.com/cachito-testing/go-generate-imported.

brunoapimentel:
I also tried to process the repository that was linked in the related Github issue locally, and it suceeded :rocket: 

Request data:
```
{
"pkg_managers": ["npm"],
"ref": "cb077cba1e881ac5ffec7828fb28cbbf14af6e0a",
"repo": "https://github.com/konveyor/forklift-ui.git",
}
```

tests/test_workers/test_pkg_managers/test_npm.py
ejegrova: added

cachito/workers/pkg_managers/npm.py
ejegrova: added second commit with changes to `name_to_deps`

ejegrova:
> > There is one integration test ready ("npm_packages", "workspaces"), do you have some specific integration test in mind?
> 
> Nevermind me, my brain was half dead when I wrote this comment. I saw the yml, but somehow thought it would feed the unit tests facepalm.
> 
> I do suggest improving the README on the workspaces test repo though smile. Maybe use this as an example: https://github.com/cachito-testing/go-generate-imported.

Okay, I made some changes to README, check it please. But we didn't really pay attention to such detail before, we don't even have readmes in other test repos :D

akhmanova:
Looks like one of the integration tests need an update of reference

```purl": "pkg:github/cachito-testing/cachito-npm-workspaces@7a83936be7177246a549acf66ab8074ea4c0d6f5",```

akhmanova (approved):
LGTM

brunoapimentel (approved):

PR #643, opened 2022-03-11
Change behavior of symlink check in the target repository

serg-cymbaluk:
CLOUDBLD-8960

The patch makes possible to process repository even if there are present
symlinks that points outside the repo.
Instead of raising exception Cachito will remove that links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -600,6 +600,10 @@ with the `mod_auth_gssapi` module.
   environment variables for the completed request. Typically, you will only want to use this if your
   package *does* use C files, and the Cachito request is failing.
 
+* `remove-unsafe-symlinks` - the flag forces Cachito to remove all symlinks that points to some
+  location outside of a cloned repository. Otherwise, if the flag isn't set, Cachito will raise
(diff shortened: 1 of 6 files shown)


serg-cymbaluk:
@brunoapimentel @ejegrova @dcho5 

Could you please take a look?

MartinBasti:
Shouldn't be this a flag to explicitly allows removing symlink, default should be just error IMO

serg-cymbaluk:
> Shouldn't be this a flag to explicitly allows removing symlink, default should be just error IMO

Ack. I will add the flag.

chmeliik (approved):
LGTM, just one detail

chmeliik:
Oh, and also the failed unit test :)

chmeliik (approved):

serg-cymbaluk:
@MartinBasti @chmeliik @brunoapimentel

The new flag is added, test are fixed.

MartinBasti (approved):

PR #644, opened 2022-03-14
Run go list with -mod=readonly when not vendoring

chmeliik:
CLOUDBLD-8966

When not vendoring, pass the -mod=readonly flag to make Go behave
accordingly - even if a vendor directory is present, we ignore it.

When we *are* vendoring, do not pass any flags and rely on Go's own
logic to decide if it should use -mod=vendor or something else.


tkdchen:
LGTM :+1:

serg-cymbaluk:
@chmeliik 

Thanks for your patch!

We need to revert the strict vendor as well: https://github.com/containerbuildsystem/cachito/pull/645
Let me test those changes.

ejegrova (approved):

chmeliik:
@serg-cymbaluk I cherry-picked your PR locally and ran integration tests, all green (except for the skipped ones, but those don't test gomod strictness)

Merging this :shipit:

PR #645, opened 2022-03-15
Revert "Enable cachito_gomod_strict_vendor flag by default"

serg-cymbaluk:
CLOUDBLD-8966

This reverts commit 818ec8a4810b0e12879380c96bb90d2f62cc633e.

### DIFF
--- a/tests/integration/test_gomod_packages.py
+++ b/tests/integration/test_gomod_packages.py
@@ -26,8 +26,7 @@ def test_gomod_vendor_without_flag(test_env):
         assert completed_response.status == 200
         assert completed_response.data["state"] == "failed"
         error_msg = (
-            'The "gomod-vendor" or "gomod-vendor-check" flag must be set when your repository has '
-            "vendored dependencies."
+            'The "gomod-vendor" flag must be set when your repository has vendored dependencies'
         )
         assert error_msg in completed_response.data["state_reason"], (
             f"#{completed_response.id}: Request failed correctly, but with unexpected message: "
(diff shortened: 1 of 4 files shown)


tests/integration/test_gomod_packages.py
```diff
@@ -26,8 +26,7 @@ def test_gomod_vendor_without_flag(test_env):
         assert completed_response.status == 200
         assert completed_response.data["state"] == "failed"
         error_msg = (
-            'The "gomod-vendor" or "gomod-vendor-check" flag must be set when your repository has '
-            "vendored dependencies."
```
chmeliik, line 30: Could we keep this and the unit test change? Otherwise LGTM

tests/test_workers/test_pkg_managers/test_gomod.py
```diff
@@ -394,30 +395,51 @@ def test_resolve_gomod_vendor_dependencies(
...
+    tmpdir,
+    strict_vendor,
+    raise_error,
 ):
     mock_isdir.return_value = True
     # Mock the get_worker_config
     mock_config = mock.Mock()
-    mock_config.cachito_gomod_strict_vendor = True
+    if strict_vendor != "default":
+        mock_config.cachito_gomod_strict_vendor = strict_vendor
```
chmeliik, line 418: I just realized that the `"default"` case doesn't actually test what is the default in our config, it tests that
```python
bool(mock.Mock().non_existing_property) is True
```
Maybe the case could be removed? But it's a minor detail, I don't mind it much

chmeliik (approved):
LGTM :+1:

tests/test_workers/test_pkg_managers/test_gomod.py
serg-cymbaluk: Removed the 'default' case. It gave a false positive result anyway.

PR #646, opened 2022-04-04
Apply blobless cloning to reduce utilization of disk space by source archives

serg-cymbaluk:
CLOUDBLD-9254

Add --filter=blob:none parameter to 'git clone'.
The parameter allows to speed up cloning and significantly save disk space for source archives.

### DIFF
--- a/cachito/workers/scm.py
+++ b/cachito/workers/scm.py
@@ -158,6 +158,7 @@ class Git(SCM):
                     self.url,
                     clone_path,
                     no_checkout=True,
+                    filter="blob:none",
                     # Don't allow git to prompt for a username if we don't have access
                     env={"GIT_TERMINAL_PROMPT": "0"},
                 )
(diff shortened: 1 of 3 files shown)


MartinBasti (approved):

chmeliik (approved):
LGTM :+1:

I have two follow-up questions:
* the `update_and_archive()` method doesn't fetch blobs, right?
* could something similar be done for git submodules, or is it not necessary?

Edit: and a third one:
* are you planning to re-clone existing repos with the filter?

serg-cymbaluk:
@chmeliik 

> the update_and_archive() method doesn't fetch blobs, right?

Let me test the case. Probably it requires the filter as well.

> could something similar be done for git submodules, or is it not necessary?

It is not critical, but could be also done.

> are you planning to re-clone existing repos with the filter?

I am not sure about that. It could be dangerous and resource consuming. I would postpone such action to the moment when it becomes really critical. In other hand, I can just provide a script for that without performing.

chmeliik:
> I am not sure about that. It could be dangerous and resource consuming. I would postpone such action to the moment when it becomes really critical. In other hand, I can just provide a script for that without performing.

Ack, yeah a script would be very nice :+1:

brunoapimentel (approved):
LGTM!

I'm linking a more thorough explanation on the blob filter for future reference (since the official docs are a little lacking in that regard).

https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/#user-content-blobless-clones

PR #651, opened 2022-04-20
Accomodating https://issues.sonatype.org/browse/NEXUS-25162, required…

mike-kingsbury:
… to support Nexux >= 3.28.0


serg-cymbaluk:
@mike-kingsbury 
Is that change backward compatible?

serg-cymbaluk:
@mike-kingsbury 
Could you please swap commit message and description?

serg-cymbaluk:
@mike-kingsbury 
> No, the change is not backwards compatible.
So if some deployment uses old Nexus the patch will break it, right?

PR #653, opened 2022-04-22
Re-attempt to delete bundle directory when state is set to 'stale'

serg-cymbaluk:
CLOUDBLD-9254

The change covers two cases:
1) Deleting of bundle dir failed due some volatile cause, so re-attempting
   would help to cleanup that eventually.
2) Some request stuck in 'in_progress' state. As result, 'complete' or 'failed'
   state is missed and deleting of the bundle dir wasn't triggered.

- [n/a] Code coverage from testing does not decrease and new code is covered


lkolacek (approved):
Looks good :+1:

PR #655, opened 2022-05-06
Fix an error condition when cloned_main_repo isn't properly set

mike-kingsbury:
When the TestCachedDependencies integration test failed before setting the [cloned_main_repo](https://github.com/containerbuildsystem/cachito/blob/03d2553e91edac9c52ecd8f3d121b9aa2414d3d0/tests/integration/test_using_cached_dependencies.py#L196) variable, the teardown method would also fail and cover up the actual root cause.

- [x ] Commit messages are descriptive enough

### DIFF
--- a/tests/integration/test_using_cached_dependencies.py
+++ b/tests/integration/test_using_cached_dependencies.py
@@ -115,7 +115,7 @@ class TestCachedDependencies:
 
     def teardown_method(self, method):
         """Delete branch with commit in the main repo."""
-        if not self.use_local:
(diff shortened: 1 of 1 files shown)


mike-kingsbury:
@brunoapimentel PTAL

brunoapimentel (approved):

lkolacek (approved):

taylormadore:
+1 LGTM

tkdchen:
:+1:

PR #658, opened 2022-05-19
Validate string length to repo field on create_request endpoint

fepas:
CLOUDBLD-7050

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -26,6 +26,11 @@ def is_request_ref_valid(ref: str) -> bool:
     return re.match(r"^[a-f0-9]{40}$", ref) is not None
 
 
+def is_request_repo_valid(repo: str) -> bool:
+    """Check if a string is a valid repo in the expected size."""
+    return len(repo) <= 200
+
+
 request_pkg_manager_table = db.Table(
     "request_pkg_manager",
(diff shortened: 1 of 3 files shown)


lkolacek:
@fepas thank you! PR looks good to me.

taylormadore (approved):
LGTM. Great job!

cachito/web/models.py
```diff
@@ -26,6 +26,11 @@ def is_request_ref_valid(ref: str) -> bool:
     return re.match(r"^[a-f0-9]{40}$", ref) is not None
 
 
+def is_request_repo_valid(repo: str) -> bool:
```
brunoapimentel, line 29: I was about to say that this function should be private... but looking at `is_request_ref_valid`, I saw it is used also to validate the `GET /api/v1/requests` endpoint.

So I'm thinking we should also use this function to validate the `repo` parameter in that endpoint.

fepas: I added the validation function in `/api/v1/requests` and changed the error message that existed in `ref`parameter validation (to keep the same message in all places). 

Also fixed the tests to work with the new error message and added a new test to check the `repo` parameter in this context.

brunoapimentel (approved):
LGTM :rocket:

PR #663, opened 2022-05-31
Catch proper exceptions and/or unconceal Cachito exceptions

fepas:
[CLOUDBLD-2833](https://issues.redhat.com/browse/CLOUDBLD-2833)

Signed-off-by: Felipe Campos <fepas.unb@gmail.com>

This issue is about fixing the blocks with the "noqa E722" comment. This was necessary because of the bare exceptions that we were using, which can cause the exceptions to be not seen. 

I found it very difficult to track down all possible exceptions that could happen in some cases, so I thought of another solution.

This is the most basic approach that I found to improve it: changing the bare except to an exception with an associated value, where we can log the exception name and its details. 

Example: 

```py
except:  # noqa E722
  flask.current_app.logger.exception(
    "Failed to delete the bundle archive %s", bundle_dir.bundle_archive_file
  )
```

It turns into:
```py
except Exception as ex:
  template = "An exception of type {0} occurred. Arguments:\n{1!r}"
  flask.current_app.logger.exception(template.format(type(ex).__name__, ex.args))
  flask.current_app.logger.exception(
    "Failed to delete the bundle archive %s", bundle_dir.bundle_archive_file
  )
```
Maybe that's not what is expected to solve this problem (also the acceptance criteria define that I should also change the API response to show the type of exception), so I'm opening this pull request to ask for your advice. :D

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -537,7 +537,7 @@ def patch_request(request_id):
             bundle_dir.bundle_archive_file.unlink()
             bundle_dir.bundle_archive_checksum.unlink()
             bundle_dir.packages_data.unlink()
-        except:  # noqa E722
+        except OSError:
             flask.current_app.logger.exception(
                 "Failed to delete the bundle archive %s", bundle_dir.bundle_archive_file
             )
@@ -548,9 +548,13 @@ def patch_request(request_id):
         )
         try:
             bundle_dir.rmtree()
-        except:  # noqa E722
+        except OSError:
             flask.current_app.logger.exception(
-                "Failed to delete the temporary files at %s", bundle_dir
+                "Failed to delete the temporary files (OSError) at %s", bundle_dir
+            )
+        except Exception as ex:
+            flask.current_app.logger.exception(
+                "Failed to delete the temporary files (%s) at %s", type(ex).__name__, bundle_dir
             )
 
     if delete_logs:
@@ -558,7 +562,7 @@ def patch_request(request_id):
         path_to_file = os.path.join(request_log_dir, f"{request_id}.log")
         try:
             os.remove(path_to_file)
-        except:  # noqa E722
+        except OSError:
             flask.current_app.logger.exception("Failed to delete the log file %s", path_to_file)
 
     for pkg_mgr in cleanup_nexus:
--- a/cachito/workers/pkg_managers/gomod.py
+++ b/cachito/workers/pkg_managers/gomod.py
@@ -813,8 +813,12 @@ def get_golang_version(module_name, git_path, commit_sha, update_tags=False, sub
     if update_tags:
         try:
             repo.remote().fetch(force=True, tags=True)
-        except:  # noqa E722
-            log.warning("Failed to fetch the tags on the Git repository for %s", module_name)
+        except Exception as ex:
+            raise CachitoError(
+                "Failed to fetch the tags on the Git repository (%s) for %s ",
+                type(ex).__name__,
+                module_name,
+            )
 
     if module_major_version:
         major_versions_to_try = (module_major_version,)
--- a/cachito/workers/scm.py
+++ b/cachito/workers/scm.py
@@ -57,10 +57,16 @@ class Git(SCM):
         try:
             repo.head.reference = repo.commit(self.ref)
             repo.head.reset(index=True, working_tree=True)
-        except:  # noqa E722
-            log.exception('Checking out the Git ref "%s" failed', self.ref)
+
+        except Exception as ex:
+            log.exception(
+                "Failed on checking out the Git ref %s, url: %s, exception: %s",
+                self.ref,
+                self.url,
+                type(ex).__name__,
+            )
             raise CachitoError(
-                "Checking out the Git repository failed. Please verify the supplied reference "
+                "Failed on checking out the Git repository. Please verify the supplied reference "
                 f'of "{self.ref}" is valid.'
             )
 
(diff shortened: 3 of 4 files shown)


cachito/web/api_v1.py
```diff
@@ -548,7 +550,9 @@ def patch_request(request_id):
         )
         try:
             bundle_dir.rmtree()
-        except:  # noqa E722
+        except Exception as ex:
```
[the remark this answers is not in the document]
brunoapimentel: But looking at the `os.remove` and the `Path.remove` functions, I think we're safe to only catch `OsError` in those blocks.

So maybe leave that extra generic exception handling only for the `rmtree` block?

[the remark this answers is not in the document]
taylormadore: Yes, this approach sounds good. It'll be nice to have this additional information captured in the logs. :slightly_smiling_face:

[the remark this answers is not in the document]
lkolacek: I agree :)

cachito/workers/pkg_managers/gomod.py
```diff
@@ -813,7 +813,9 @@ def get_golang_version(module_name, git_path, commit_sha, update_tags=False, sub
     if update_tags:
         try:
             repo.remote().fetch(force=True, tags=True)
-        except:  # noqa E722
+        except Exception as ex:
```
brunoapimentel, line 816: @chmeliik I was looking at this code block and wondering... shouldn't we fail the request if the `git fetch` fails to execute? The consequence here is that we would be potentially resolving a wrong version for the module, right? Or is there something I'm missing here?

brunoapimentel, line 816: @chmeliik I was looking at this code block and wondering... shouldn't we fail the request if the `git fetch` fails to execute? The consequence here is that we would be potentially resolving a wrong version for the module, right? Or is there something I'm missing here?

cachito/web/api_v1.py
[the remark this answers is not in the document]
fepas: Fixed it guys, thank you for the feedback :)

cachito/workers/pkg_managers/gomod.py
fepas: Now it raises a CachitoError, so I think it's better handled now. Can you guys take a look again? @brunoapimentel @chmeliik

taylormadore:
Just a couple of minor points:

- Can we update the commit message to mention something about how we're converting bare excepts to more specific ones? It might be more descriptive than E722
- Does the test coverage need to be updated? I think I noticed that it has decreased slightly

cachito/workers/scm.py
```diff
@@ -192,7 +197,10 @@ def update_and_archive(self, previous_archive, gitsubmodule=False):
                 # The reference must be specified to handle commits which are not part
                 # of a branch.
                 repo.remote().fetch(refspec=self.ref)
-            except:  # noqa E722
+
+            except Exception as ex:
+                template = "An exception of type {0} occurred. Arguments:\n{1!r}"
+                log.exception(template.format(type(ex).__name__, ex.args))
```
tkdchen, line 203: IMO, we don't need to repeat the `log.exception`. The existing `log.exception` is able to log traceback which has the exception type info.

Regarding the `ex.args`, I also don't think we need it, since what the value would be does totally depend on the concrete exception implementation. No all are valuable for developers to debug. Again, the following `log.exception` already logs the `self.url`. In addition, if we also care about the ref for the fetch operation, we should log the `self.ref` explicitly.

 Similar thought applies to other changes as well in this pr.

lkolacek: @cqi  +1 for adding template.format(type(ex).__name__  info into original `log.exception` (merge two log.exceptions here into one). I don't mind adding `ex.args` into log message. It might or might not be useful. I rather get more info. +1 fro adding `self.ref` explicitly into log.

cachito/web/api_v1.py
```diff
@@ -548,17 +548,20 @@ def patch_request(request_id):
...
         try:
             bundle_dir.rmtree()
-        except:  # noqa E722
+        except OSError:
             flask.current_app.logger.exception(
                 "Failed to delete the temporary files at %s", bundle_dir
             )
+        except Exception as ex:
+            template = "An exception of type {0} occurred. Arguments:\n{1!r}"
+            flask.current_app.logger.exception(template.format(type(ex).__name__, ex.args))
```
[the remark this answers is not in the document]
tkdchen: I just thought the original try-except wrapping `build_dir.rmtree()` is clear enough to describe the intention in this context. Feel free to keep going. I'm ok with this change. :)

cachito/workers/scm.py
lkolacek: @fepas Are we going to add also `self.ref` to log message?

lkolacek:
@fepas I added two more comments here. Otherwise, it looks great. We should merge this soon. Any more comments @taylormadore @ejegrova ??

fepas:
> > > Just a couple of minor points:
> > > 
> > > * Can we update the commit message to mention something about how we're converting bare excepts to more specific ones? It might be more descriptive than E722
> > > * Does the test coverage need to be updated? I think I noticed that it has decreased slightly
> > 
> > 
> > That makes sense, I'll do it! Thank you, Taylor!
> 
> @fepas I still see commit message `Fix noqa E722 blocks`. Can we replace it for something like: `Catch more specific exceptions` or `Replace general except for more specific exceptions`

Thank you for the idea, @lkolacek!

cachito/workers/scm.py
lkolacek: Nice, thanks!

lkolacek (approved):
LGTM :+1:

tkdchen (approved):
Looks much better 👍

ejegrova (approved):

taylormadore (approved):

sumincho22 (approved):
LGTM

PR #666, opened 2022-06-02
Parse gemlock

mtichavsky:

### DIFF
--- a/cachito/workers/pkg_managers/rubygems.py
+++ b/cachito/workers/pkg_managers/rubygems.py
@@ -1,10 +1,29 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
 import logging
+import re
+from dataclasses import dataclass
+from pathlib import Path
 
-from cachito.errors import CachitoError
+from gemlock_parser.gemfile_lock import GemfileLockParser
+
+from cachito.errors import CachitoError, ValidationError
 from cachito.workers import nexus
 from cachito.workers.errors import NexusScriptError
 
+GIT_REF_FORMAT = re.compile(r"^[a-fA-F0-9]{40}$")
+PLATFORMS_RUBY = re.compile(r"^PLATFORMS\n {2}ruby\n\n", re.MULTILINE)
+
 log = logging.getLogger(__name__)
+
+
+@dataclass
+class GemMetadata:
+    """Gem metadata."""
+
+    name: str
+    version: str
+    type: str
+    source: str
 
 
 def prepare_nexus_for_rubygems_request(rubygems_repo_name, raw_repo_name):
@@ -25,3 +44,95 @@ def prepare_nexus_for_rubygems_request(rubygems_repo_name, raw_repo_name):
     except NexusScriptError:
         log.exception("Failed to execute the script %s", script_name)
         raise CachitoError("Failed to prepare Nexus for Cachito to stage Rubygems content")
+
+
+def parse_gemlock(source_dir, gemlock_path):
+    """Parse dependencies from Gemfile.lock.
+
+    :param Path source_dir: the full path to the project directory
+    :param Path gemlock_path: the full path to Gemfile.lock
+    :return: list of Gems
+    """
+    if not gemlock_path.is_file():
+        raise ValidationError(
+            f"Gemfile.lock at path {gemlock_path} does not exist or is not a regular file."
+        )
+
+    _validate_gemlock_platforms(gemlock_path)
+
+    dependencies = []
+    all_gems = GemfileLockParser(str(gemlock_path)).all_gems
+    for gem in all_gems.values():
+        _validate_gem_metadata(gem, source_dir, gemlock_path.parent)
+        source = gem.remote if gem.type != "PATH" else gem.path
+        dependencies.append(GemMetadata(gem.name, gem.version, gem.type, source))
+
+    return dependencies
+
+
+def _validate_gemlock_platforms(gemlock_path):
+    """Make sure Gemfile.lock contains only one platform - ruby."""
+    with open(gemlock_path) as f:
+        contents = f.read()
+
+    if not PLATFORMS_RUBY.search(contents):
+        msg = "PLATFORMS section of Gemfile.lock has to contain one and only platform - ruby."
+        raise ValidationError(msg)
+
+
+def _validate_gem_metadata(gem, source_dir, gemlock_dir):
+    """Validate parsed Gem.
+
+    While individual gems may contain platform information, this function doesn't check it,
+    because it expects the Gemfile.lock to be ruby platform specific.
+    :param Gem gem: gem with information parsed from Gemfile.lock
+    :param Path source_dir: the full path to the project root
+    :param Path gemlock_dir: the root directory containing Gemfile.lock
+    :raise: ValidationError
+    """
+    if gem.name is None or gem.version is None:
+        raise ValidationError("Unspecified name or version of a RubyGem.")
+
+    if gem.type == "GEM":
+        if gem.remote != "https://rubygems.org/":
+            raise ValidationError(
+                "Cachito supports only https://rubygems.org/ as a remote for Ruby GEM dependencies."
+            )
+    elif gem.type == "GIT":
+        if not gem.remote.startswith("https://"):
+            raise ValidationError("All Ruby GIT dependencies have to use HTTPS protocol.")
+        if not GIT_REF_FORMAT.match(gem.version):
+            msg = (
+                f"No git ref for gem: {gem.name} (expected 40 hexadecimal characters, "
+                f"got: {gem.version})."
+            )
+            raise ValidationError(msg)
+    elif gem.type == "PATH":
+        _validate_path_dependency_dir(gem, source_dir, gemlock_dir)
+    else:
+        raise ValidationError("Gemfile.lock contains unsupported dependency type.")
+
+
+def _validate_path_dependency_dir(gem, project_root, gemlock_dir):
+    """Validate path of PATH dependency.
+
+    :param gem: validated gem
+    :param project_root: project root directory
+    :param gemlock_dir: absolute path to Gemfile.lock parent directory
+    """
+    dependency_dir = gemlock_dir / Path(gem.path)
+    try:
+        dependency_dir = dependency_dir.resolve(strict=True)
+        dependency_dir.relative_to(project_root.resolve())
+    except FileNotFoundError:
+        raise ValidationError(
+            f"PATH dependency {str(gem.name)} references a non-existing path: "
+            f"{str(dependency_dir)}."
+        )
+    except RuntimeError:
+        raise ValidationError(
+            f"Path of PATH dependency {str(gem.name)} contains an infinite loop: "
+            f"{str(dependency_dir)}."
+        )
+    except ValueError:
+        raise ValidationError(f"{str(dependency_dir)} is not a subpath of {str(project_root)}")
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -16,16 +16,24 @@ amqp==5.0.9 \
     # via
     #   -r requirements.txt
     #   kombu
-attrs==20.3.0 \
-    --hash=sha256:31b2eced602aa8423c2aea9c76a724617ed67cf9513173fd3a4f03e3a929c7e6 \
-    --hash=sha256:832aa3cde19744e49938b91fea06d69ecb9e649c93ba974535d08ad92164f700
+attrs==21.4.0 \
+    --hash=sha256:2d27e3784d7a565d36ab851fe94887c5eccd6a463168875832a1be79c82828b4 \
+    --hash=sha256:626ba8234211db98e869df76230a137c4c40a12d72445c45d5f5b716f076e2fd
     # via
+    #   -r requirements.txt
+    #   commoncode
     #   jsonschema
     #   pytest
 backoff==1.11.1 \
     --hash=sha256:61928f8fa48d52e4faa81875eecf308eccfb1016b018bb6bd21e05b5d90a96c5 \
     --hash=sha256:ccb962a2378418c667b3c979b504fdeb7d9e0d29c0579e3b13b86467177728cb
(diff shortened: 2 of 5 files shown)


cachito/workers/pkg_managers/ruby.py
```diff
@@ -0,0 +1,69 @@
...
+
+    :param Path gemlock_path: the full path to Gemfile.lock
+    :return: list of Gems
+    """
+    dt = gemlock_parser.GemfileLockParser(str(gemlock_path)).dependency_tree
+    dependencies = []
+
+    for key in dt:
+        gem = dt[key]
+        validate_gem_metadata(gem, gemlock_path.parents[0])
```
chmeliik, line 40: Would suggest `gemlock_path.parent` - it is equivalent but IMO clearer

cachito/workers/pkg_managers/ruby.py
```diff
@@ -0,0 +1,69 @@
...
+        raise ValidationError("Unspecified name or version of a RubyGem.")
+
+    if gem.type == "GEM":
+        if gem.remote != "https://rubygems.org/":
+            raise ValidationError(
+                "Cachito supports only https://rubygems.org/ as a remote for Ruby GEM dependencies."
+            )
+    elif gem.type == "GIT":
+        if not gem.remote.startswith("https://"):
+            raise ValidationError("All Ruby GIT dependencies have to use HTTPS protocol.")
```
chmeliik, line 63: Could you remind me why ssh:// is not allowed?

cachito/workers/pkg_managers/ruby.py
```diff
@@ -0,0 +1,69 @@
...
+
+    for key in dt:
+        gem = dt[key]
+        validate_gem_metadata(gem, gemlock_path.parents[0])
+        dependencies.append(GemMetadata(gem.name, gem.version, gem.type, gem.remote))
+
+    return dependencies
+
+
+def validate_gem_metadata(gem, gemlock_dir):
```
chmeliik, line 46: I'm missing a few validations here (based on the design doc):
> NOTE: Platform-dependent Gem references are not considered valid, since they can contain pre-compiled binaries.

> For any dependencies referenced from [other sources](https://docs.google.com/document/d/14KaPuBX4BtK3EvTzyJpV-TZ8uhCa7EM7FZARaLC-ZtQ/edit#bookmark=id.vsrcb15gznd), Cachito will check that each reference is to a pinned version or commit hash

And I would be careful about "pinned version", git tags are not immutable. I believe pip, NPM and Yarn support only commit hashes in Cachito, if acceptable, ruby should do the same

cachito/workers/pkg_managers/ruby.py
```diff
@@ -0,0 +1,69 @@
...
+            raise ValidationError(
+                "Cachito supports only https://rubygems.org/ as a remote for Ruby GEM dependencies."
+            )
+    elif gem.type == "GIT":
+        if not gem.remote.startswith("https://"):
+            raise ValidationError("All Ruby GIT dependencies have to use HTTPS protocol.")
+    elif gem.type == "PATH":
+        dependency_dir = gemlock_dir / gem.path
+        if not dependency_dir.exists():
+            raise ValidationError("PATH dependency references a non-existing path.")
```
chmeliik, line 67: We should also validate that `dependency_dir` is a subpath of `gemlock_dir` (or maybe, more permissively, a subpath of the repo directory?) in a similar way to e.g. pip's _ensure_local:
https://github.com/containerbuildsystem/cachito/blob/39d78ad0a9737c03e2d99e8f42a7c7be4ebfaec5/cachito/workers/pkg_managers/pip.py#L301-L308

tests/test_workers/test_pkg_managers/test_ruby.py
```diff
@@ -0,0 +1,189 @@
...
+
+                    GEM
+                      remote: https://rubygems.org/
+                      specs:
+                        builder (3.2.4)
+                        ci_reporter (2.0.0)
+                          builder (>= 2.1.2)
+
+                    DEPENDENCIES
+                      ci_reporter_shell!
```
chmeliik, line 51: Why do we expect only one dependency (ci_reporter_shell) in this case? I know next to nothing about Gemfile.lock, but it looks to me like there are 3 here (builder, ci_reporter, ci_reporter_shell).

chmeliik:
Interesting find on the scancode-toolkit library!

I have a few gripes with it though:
* the API is not documented
* it is huge (makes the locally built worker image 400 MB larger)
* I'm not entirely comfortable trusting that it is safe to use for Cachito's purposes (e.g. must not execute arbitrary code, should not attempt to read files we don't want it to). FWIW, I did check the gemfile_lock parser and it looks OK

Could you explain why this library is the best option?

tichavskym:
I couldn't find any other library written in Python that would be able to parse `Gemfile.lock`, so the other option is to write it from scratch. Parsing Gemfiles for dependencies doesn't make sense and using some Ruby parser would mean adding Ruby runtime dependencies, which is something we want to avoid.

To address the concerns, I could

- discuss with maintainer the possibility of creating a separate library out of it (he plans to split all of Ruby related Scancode code into a separate project in the future anyways), so we could end up with something smaller
- copy-paste it into Cachito (the source code is licensed under Apache 2.0)
- document the API

cachito/workers/pkg_managers/ruby.py
```diff
@@ -0,0 +1,69 @@
...
+    """Parse dependencies from Gemfile.lock.
+
+    :param Path gemlock_path: the full path to Gemfile.lock
+    :return: list of Gems
+    """
+    dt = gemlock_parser.GemfileLockParser(str(gemlock_path)).dependency_tree
+    dependencies = []
+
+    for key in dt:
+        gem = dt[key]
```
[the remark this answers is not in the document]
tichavskym: My bad, I should have used `.all_gems`. Dependency tree has those sub-dependencies stored in `dependencies` `Gem` field.

chmeliik:
> To address the concerns, I could
> 
> * discuss with maintainer the possibility of creating a separate library out of it (he plans to split all of Ruby related Scancode code into a separate project in the future anyways), so we could end up with something smaller
> * copy-paste it into Cachito (the source code is licensed under Apache 2.0)
> * document the API

I kinda like the first two: copy-paste it (let's call it "vendoring" :smile:) until the ruby part is separated out, then replace it with the ruby-specific library. The scancode parser is already sort of vendored: [gemfile_lock.py.ABOUT](https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/gemfile_lock.py.ABOUT), [gemfile_lock.py.LICENSE](https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/gemfile_lock.py.LICENSE)

@brunoapimentel @lkolacek @taylormadore @fepas what do you think?

Documenting it is not as important, the main thing I'm interested in is some assurance that (one of) the goals of the project are to scan packages without executing arbitrary code like Gemfile

brunoapimentel:
> I kinda like the first two: copy-paste it (let's call it "vendoring" smile) until the ruby part is separated out, then replace it with the ruby-specific library. The scancode parser is already sort of vendored: [gemfile_lock.py.ABOUT](https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/gemfile_lock.py.ABOUT), [gemfile_lock.py.LICENSE](https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/gemfile_lock.py.LICENSE)
> 
> @brunoapimentel @lkolacek @taylormadore @fepas what do you think?
> 
> Documenting it is not as important, the main thing I'm interested in is some assurance that (one of) the goals of the project are to scan packages without executing arbitrary code like Gemfile (but if we use just the gemfile_lock parser, I'd say the code itself is good enough assurance for me)

+1 for not importing the whole library inside Cachito. I think "vendoring" it would be the best approach, since we can't rely on the author to separate the project within the timeframe we need. 

I was thinking that maybe we could have a separate repository like [Pyarn](https://github.com/containerbuildsystem/pyarn). We could copy the  gemfile_lock parser and the related unit tests there. Not sure if it's overkill, though :laughing:

cachito/workers/pkg_managers/ruby.py
tichavskym: Added platform check - the way bundler works with platforms has changed over the years. See how it [works](https://prathamesh.tech/2021/04/18/bundler-2-2-3-and-deployment-of-ruby-apps/) [now](https://www.moncefbelyamani.com/understanding-the-gemfile-lock-file/#platforms) x how [it used to work](https://github.com/rubygems/rubygems/issues/4269#issuecomment-758564690). TLDR: One Gemfile.lock can contain dependencies for multiple platforms and the way they are be resolved at deploy/install time changed with bundler version 2.2.0 and 2.2.3.

I decided to implement it the simplest way - support only Gemfile.lock with only one platform specified - ruby. This way, we will be able to support all bundler versions and it will be simpler to parse the resulting file (scancode-toolkit isn't able to parse files with multiple platform versions of one package).

Added check for Git reference (40 character hash). For GEM dependencies, semantic versioning is recommended, [but not enforced](https://guides.rubygems.org/patterns/), therefore the only thing that's validated is if the version isn't None (similarly for PATH).

lkolacek:
> +1 for not importing the whole library inside Cachito. I think "vendoring" it would be the best approach, since we can't rely on the author to separate the project within the timeframe we need.
> 
> I was thinking that maybe we could have a separate repository like [Pyarn](https://github.com/containerbuildsystem/pyarn). We could copy the gemfile_lock parser and the related unit tests there. Not sure if it's overkill, though laughing

+1 to vendor only things we need in Cachito. 

I don't mind having separate repository like Pyarn. Looks good to me. :)

cachito/workers/pkg_managers/ruby.py
tichavskym: From design doc
> Maintaining credentials is a major increase to the scope of this design

and I don't know about any product we want to support that would be using SSH for git repos.

From your comment, I understand that credentials are somehow already managed, so if you really wanted to see this feature in Cachito, I could probably implement it as well, but I haven't looked into it yet, so I don't know how complex it would be.

chmeliik: AFAIK no additional changes are required, the Red Hat deployment of Cachito has an ssh key configured for git so cloning via ssh should "just work". At least for public repos - for private ones, only github is supported via the [RH-Cachito](https://github.com/RH-Cachito) user.

It's OK to disallow ssh urls until the need arises, but I think Cachito can support them (to the extent above) for free

cachito/workers/pkg_managers/ruby.py
chmeliik: Just for my understanding - semantic versioning is not required, but the lockfile _does_ pin dependencies to one version, right? Maybe the version doesn't follow semver conventions, but it's still a specific pinned version?

chmeliik:
> Also, how should we proceed about publishing the package to PyPI?

Maybe we don't need to for now, I think you could specify it in requirements.in like this
```
gemlock_parser @ https://github.com/tichavskym/gemlock-parser/archive/master.zip
```

It's slightly inconvenient in that changes to gemlock-parser will have to be followed by a manual `make pip-compile` in Cachito (proably won't be auto-detected by @dependabot), but IMO good enough to start with

cachito/workers/pkg_managers/rubygems.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+@dataclass
+class GemMetadata:
+    """Gem metadata."""
+
+    name: str
+    version: str
+    type: str
+    path: str
```
chmeliik, line 21: IMO `path` is a slightly misleading name, makes sense for PATH gems but not the other types. Maybe use `remote`, like gemlock_parser does? Or `source`, which sounds applicable to both paths and urls

cachito/workers/pkg_managers/rubygems.py
```diff
@@ -0,0 +1,110 @@
...
+
+def parse_gemlock(source_dir, gemlock_path):
+    """Parse dependencies from Gemfile.lock.
+
+    :param Path source_dir: the full path to the project directory
+    :param Path gemlock_path: the full path to Gemfile.lock
+    :return: list of Gems
+    """
+    if not gemlock_path.is_file():
+        return []
```
chmeliik, line 32: Should missing lockfile be an error? Or is it a valid case for projects that use bundler?

chmeliik (commented):
Looking good, just a few more minor comments from me :+1: 

As for your question about the gemlock_parser repo:
> Do you want to add it under `containerbuildsystem` namespace and is there anything you're missing in this repo?

Having it under `containerbuildsystem` makes sense to me. I don't think anything obvious is missing. @brunoapimentel @lkolacek agreed on putting it in our namespace? Want me to create the repo and give Milan maintainer permissions?

cachito/workers/pkg_managers/rubygems.py
tichavskym: It is required to have `Gemfile`, and therefore we could say that the project is required to have Gemfile.lock.

I noticed that implementation for Python doesn't throw any error too, so I wanted to keep it similar, but yeah, raising an exception makes sense.

cachito/workers/pkg_managers/rubygems.py
tichavskym: Hmm, my point was to unite it, because the parser actually internally copies `remote` to `path` in case of PATH dependency, but maybe it wasn't the best idea. I'll create a new field to make naming clearer.

tichavskym:
I think I addressed all your comments, so you can double-check, we're still discussing the SSH support internally, so I'd say that if we decide to implement it, I could enable it in some later PR, together with the implementation of downloading dependencies.

chmeliik (approved):
LGTM :+1:

chmeliik:
Just needs a unit test update

lkolacek:
Code looks great! Thank you. I also have insufficient permissions to fork [repository](https://github.com/tichavskym/gemlock-parser) under containerbuildsystem. @rcerven could you help us with that, please?

PR #667, opened 2022-06-10
Make sure that cachito saves UTC time to DB records

sumincho22:
CLOUDBLD-7633

**Problem**
Currently, Cachito depends on pods with UTC timezone. If there is a situation in which some pods are set in non-UTC timezones, this will produce inconsistency in the DB. Time-sensitive endpoints such as `created` and `updated` values should be based on one timezone, UTC, as the DB schema is "timestamp without timezone".

Right now, we are using the function `sqlalchemy.func.now()` to set the current DateTime for the `created` and `updated` values in a Cachito request. The problem with this function is that the timezone is based on the DB pod's timezone (which may or may not be UTC).

**Solution**
We can utilize our DB compilers, PostgreSQL and SQLite (for testing environments), to always generate the current UTC timestamp regardless of the pods' timezones.

By following this [implementation](https://docs.sqlalchemy.org/en/14/core/compiler.html#utc-timestamp-function) from the SQLAlchemy documentation, we can make a class for `utcnow` which will use compiler functions in the PostgreSQL and SQLite databases. This way, no matter which database we are using, we can generate the current UTC timestamps from the databases and properly set the `created` and `updated` endpoints for all requests.

**Doubts I have currently:**

- Where should I place the `utcnow` and its compile functions? Should I put this at the end of the file? Should I make a new file for this?
- Is the unit test reliable? Is there a better, simpler way to go about testing this?

### DIFF
--- a/tests/integration/test_valid_data_in_request.py
+++ b/tests/integration/test_valid_data_in_request.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
+import datetime
 import os
 
 import pytest
@@ -97,3 +98,27 @@ def test_various_packages(test_env):
         utils.assert_properly_completed_response(completed_response)
 
         assert len(completed_response.data["dependencies"]) == package["dependencies_count"]
+
+
+def test_utc_datetime(test_env):
+    """
+    A basic integration test to check that a request's "created" timestamp is in UTC.
+
+    Process:
+    * Send new request to the Cachito API
+    * Compare the current UTC datetime with the request's "created" datetime
+
+    Checks:
+    * Verify that the "created" datetime is within 10-second difference of the current UTC datetime
+    """
+    env_data = utils.load_test_data("pip_packages.yaml")["local_path"]
+    client = utils.Client(test_env["api_url"], test_env["api_auth_type"], test_env.get("timeout"))
+    current_utc_datetime = datetime.datetime.utcnow()
+    initial_response = client.create_new_request(
+        payload={"repo": env_data["repo"], "ref": env_data["ref"]}
+    )
+    completed_response = client.wait_for_complete_request(initial_response)
+    request_created_datetime = datetime.datetime.fromisoformat(completed_response.data["created"])
+    diff_in_secs = (current_utc_datetime - request_created_datetime).total_seconds()
+
+    assert abs(diff_in_secs) <= 10
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -9,9 +9,11 @@ from enum import Enum
 from typing import Any, Dict, List
 
 import flask
-import sqlalchemy.sql
 from flask_login import UserMixin, current_user
 from sqlalchemy import TIMESTAMP, func
+from sqlalchemy.ext.compiler import compiles
+from sqlalchemy.sql import expression
+from sqlalchemy.types import DateTime
 from werkzeug.exceptions import Forbidden
 
 from cachito.common.packages_data import PackagesData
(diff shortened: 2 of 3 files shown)


taylormadore:
I think models.py is a reasonable home for those new functions

lkolacek:
I don't have on mind any better unit test but we might want to add integration test for this specific scenario. My understanding is that our test env uses postgres and the rest ones (probably) use sqlite but we should always receive same timestamp. What do you think @sumincho22 @taylormadore

sumincho22:
I believe our test environment uses SQLite and the actual service runs on PostgreSQL. So, yes, it does make sense for me to also test the PostgreSQL compile function.

I can do a similar thing that I did with the unit test with an integration test. It will be pretty straightforward, validating the "created" timestamp.

tests/integration/test_valid_data_in_request.py
```diff
@@ -97,3 +99,15 @@ def test_various_packages(test_env):
...
+
+def test_utc_datetime(test_env):
+    env_data = utils.load_test_data("pip_packages.yaml")["local_path"]
+    client = utils.Client(test_env["api_url"], test_env["api_auth_type"], test_env.get("timeout"))
+    expected = datetime.datetime.now(timezone.utc).isoformat()[:16]
+    initial_response = client.create_new_request(
+        payload={"repo": env_data["repo"], "ref": env_data["ref"]}
+    )
+    completed_response = client.wait_for_complete_request(initial_response)
+    actual = completed_response.data["created"][:16]
```
taylormadore, line 112: Isn't it fun to test things involving time :grin: 
I wonder if we might see test failures when time is truncated to the minute and the expected/actual fall on opposite sides of it. @lkolacek what do you think?

lkolacek: We can always add some threshold :) if there's more than 10s difference it fails

sumincho22: Yeah I agree, this was something I was worried about. I will apply the threshold on both the integration and unit tests.

tests/integration/test_valid_data_in_request.py
```diff
@@ -97,3 +98,17 @@ def test_various_packages(test_env):
         utils.assert_properly_completed_response(completed_response)
 
         assert len(completed_response.data["dependencies"]) == package["dependencies_count"]
+
+
+def test_utc_datetime(test_env):
+    env_data = utils.load_test_data("pip_packages.yaml")["local_path"]
+    client = utils.Client(test_env["api_url"], test_env["api_auth_type"], test_env.get("timeout"))
+    expected = datetime.datetime.utcnow()
```
taylormadore, line 106: Minor point, but we might want to change the names of the actual/expected variables to something more meaningful. Consider something like utc_time and request_created_time. What do you think?

tests/integration/test_valid_data_in_request.py
```diff
@@ -97,3 +98,17 @@ def test_various_packages(test_env):
         utils.assert_properly_completed_response(completed_response)
 
         assert len(completed_response.data["dependencies"]) == package["dependencies_count"]
+
+
+def test_utc_datetime(test_env):
```
taylormadore, line 103: We might want to add some comments here describing what we're testing and a note about the 10 second "buffer" that we've added.

tests/integration/test_valid_data_in_request.py
sumincho22: Got it. Added the changes.

tests/integration/test_valid_data_in_request.py
sumincho22: Yep, I agree. I added documentation.

sumincho22:
@lkolacek Please let me know if there are more fixes I should do. Otherwise, this should be ready for approval. @taylormadore

taylormadore (approved):

lkolacek (approved):
Looks great

PR #669, opened 2022-06-14
Enable krb tracing

mike-kingsbury:

### DIFF
--- a/docker/Dockerfile-workers
+++ b/docker/Dockerfile-workers
@@ -39,6 +39,7 @@ ENV GIT_COMMITTER_NAME=cachito \
     GIT_COMMITTER_EMAIL=cachito@localhost \
     GIT_AUTHOR_NAME=cachito \
     GIT_AUTHOR_EMAIL=cachito@localhost
+    KRB5_TRACE=/tmp/krb5.log
 
 EXPOSE 8080
 CMD ["celery", "-A", "cachito.workers.tasks", "worker", "--loglevel=info"]


rcerven (approved):

docker/Dockerfile-workers
```diff
@@ -39,6 +39,7 @@ ENV GIT_COMMITTER_NAME=cachito \
     GIT_COMMITTER_EMAIL=cachito@localhost \
     GIT_AUTHOR_NAME=cachito \
     GIT_AUTHOR_EMAIL=cachito@localhost
```
rcerven, line 41: missing \

rcerven (changes requested):
missing \

MartinBasti (changes requested):
This shouldn't be merged into master, build it as a separate image for your debugging.

chmeliik (commented):
Python logging might also be useful, `requests_kerberos` and `spnego` in particular seem to produce nice logs. I tested with this script:
```python
#!/usr/bin/env python3
import logging

import requests
import requests_kerberos

logging.basicConfig(level=logging.INFO)

logging.getLogger("gssapi").setLevel(logging.DEBUG)
logging.getLogger("requests_kerberos").setLevel(logging.DEBUG)
logging.getLogger("spnego").setLevel(logging.DEBUG)

auth = requests_kerberos.HTTPKerberosAuth()
resp = requests.post("https://cachito.stage.engineering.redhat.com/api/v1/requests", auth=auth)
print(resp.status_code, resp.text)
assert resp.status_code == 400, f"Expected 400 Bad Request, got {resp.status_code}"
```
The most interesting part was probably this
```
DEBUG:spnego._gss:GSSAPI step input: YIGZBgkqhkiG9xIBAgICAG+BiTCBhqADAgEFoQMCAQ+iejB4oAMCARKicQRveAQ0/MsszA3H8P6dhqeQiSmcmaFh57JY8up8a+kBp64iAvNdXn2pt2V58JA5d1gwF5CLhAubBllyGNDrQDdm3CbPJxSXYZHCidchRpzaV0IPk7tgVARh2dSm6m4rj+UdOUFVT9bSUEoavSE6+3sI
```
The token can be parsed with `python3 -m spnego --token <token>` to human readable output, and may contain details about the Kerberos error. For example, I saw `KRB_AP_ERR_NOT_US` there yesterday (but `curl --verbose` produces even more readable output, it said "The ticket is not for us.")

chmeliik:
I guess the logging could be configured in [setup_task_logging](https://github.com/containerbuildsystem/cachito/blob/c69f9cf37a1dfbeec146e952bbcf7af4ba8384a2/cachito/workers/celery_logging.py#L99), the thing is I'm not sure why Cachito isn't _already_ logging at DEBUG for all imported modules (the code kinda looks like it would do that)

PR #672, opened 2022-06-27
Add containerbuildsystem/gemlock-parser as dependency

tichavskym:
Replaces repository that isn't maintained by containerbuildsystem.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,6 +10,7 @@ setup(
     install_requires=[
         "backoff",
         "celery>=5",
+        "gemlock_parser",
         "gitpython",
         "kombu>=5",  # A celery dependency but it's directly imported
         "packaging",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -10,6 +10,8 @@
     install_requires=[
         "backoff",
         "celery>=5",
+        "gemlock_parser @ https://github.com/containerbuildsystem/gemlock-parser/archive"
+        "/a2f0f4020e07b1e87813a3254eb3e40047d8e981.zip",
```
chmeliik, line 14: This can just be `gemlock_parser`. That's the difference between setup.py and requirements.txt: setup.py defines the dependencies you need (and their version ranges, optionally), requirements.txt defines the exact versions and how to get them.

tichavskym: Oh, thanks

lkolacek (approved):
Awesome, thanks! +1

PR #675, opened 2022-07-07
Configure Rubygems repositories for final usage

tichavskym:
SPMM-9899

This is basically a copy-paste of [this](https://github.com/containerbuildsystem/cachito/pull/237), but I just couldn't think of any good way to refactor it. Also note that I changed names of blob store and repositories from `ruby` to `rubygems` to match the naming conventions of the rest of the code, so you'll probably have to rebuild your environment to make it work.


chmeliik (approved):
LGTM :+1:

By my understanding, the design has changed so that the raw repositories may no longer be required, but IMO it's fine to merge this as-is and do those changes later.

ejegrova (approved):
LGTM

PR #676, opened 2022-07-11
Download Rubygems dependencies through Nexus proxy or from Git

tichavskym:


cachito/workers/config.py
```diff
@@ -45,6 +45,7 @@ class Config(object):
     cachito_nexus_max_search_attempts = 5
     cachito_nexus_npm_proxy_url = "http://localhost:8081/repository/cachito-js/"
```
tichavskym, line 46: I'm little confused about the configuration here. Why is this in the base `Config` class while the PyPI proxy URL is in the `DevelopmentConfig`? Also, I'm curious about what's the difference between `http://nexus` URL and `http://localhost`, since it looks like both of these work.

tichavskym:
Do you have any idea why are tests failing after the second refactor? They stopped working with `Refactor _download_raw_component` commit and when I'm testing it manually in my local development environment (I mean not running unit tests, but changing code a bit and creating regular curl request to Cachito), the new changes seem to work, so I'm wondering that maybe I'm patching something it those tests poorly.

cachito/workers/config.py
chmeliik: (for pip and now ruby, we do directly download stuff from the proxy url)

tests/test_workers/test_pkg_managers/test_general.py
```diff
@@ -149,3 +158,20 @@ def test_update_request_env_vars_failed(mock_patch, side_effect, expected_error)
...
+@mock.patch("cachito.workers.pkg_managers.general.nexus.upload_raw_component")
+@pytest.mark.parametrize("is_request_repo", [True, False])
+def test_upload_raw_package(mock_upload, caplog, is_request_repo):
+    """Check Nexus upload calls."""
+    name = "name"
+    path = "fakepath"
+    dest_dir = "name/varsion"
+    filename = "name.tar.gz"
+
+    upload_raw_package(name, path, dest_dir, filename, is_request_repo)
```
chmeliik, line 172: > Do you have any idea why are tests failing after the second refactor? They stopped working with `Refactor _download_raw_component` commit and when I'm testing it manually in my local development environment (I mean not running unit tests, but changing code a bit and creating regular curl request to Cachito), the new changes seem to work, so I'm wondering that maybe I'm patching something it those tests poorly.

Not sure, `mock.patch` tends to be rather fragile :disappointed: 

The only difference I can see is that the previous test used `pip.upload_raw_package()` while this one uses the imported function directly, but I'm not sure if that makes any difference

cachito/workers/pkg_managers/rubygems.py
```diff
@@ -136,3 +142,182 @@ def _validate_path_dependency_dir(gem, project_root, gemlock_dir):
...
+        "have_raw_component": have_raw_component,
+    }
+
+
+def _extract_git_info(repo_url):
+    """Extract git info from the repo url.
+
+    :param repo_url: url to the git repository
+    :return: host part of url, list with namespace parts and a repository name
+    """
```
chmeliik, line 298: Would this also be a good candidate for refactoring into the `general` module?

chmeliik (commented):
Looks good so far :+1:

cachito/workers/config.py
tichavskym: Oh, okay, I'm going to keep RubyGems config in `DevelopmentConfig` then (as pip does).

tests/test_workers/test_pkg_managers/test_general.py
tichavskym: Turns out, `download_binary_file` has to be patched twice, once called from `general` and once from `pip`.

cachito/workers/pkg_managers/rubygems.py
tichavskym: Yes. Function that pip uses does a little more than is needed for RubyGems, but to remove code duplication, it makes sense.

chmeliik (commented):
Mostly LGTM, just a minor unittest detail :+1: 

Coveralls should work now if you re-push your branch

chmeliik (approved):
LGTM :+1:

lkolacek (approved):
Looks great! :+1:

tichavskym:
@chmeliik Should I squash the commits or is it better to keep the refactors separate?

chmeliik:
> @chmeliik Should I squash the commits or is it better to keep the refactors separate?

I prefer them separate :+1: 

Merging

PR #681, opened 2022-07-21
RubyGems: Generate config file for Bundler

tichavskym:
SPMM-9901

In order to redirect Bundler to use dependencies cached by Cachito, a
configuration file will be provided.

While there already may be a `.bundle/config` file in the repository for
which dependencies were requested, it's not the case for any of the
products that we indent to support and therefore an exception is raised.

### DIFF
--- a/cachito/workers/pkg_managers/rubygems.py
+++ b/cachito/workers/pkg_managers/rubygems.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import logging
+import os
 import random
 import re
 import secrets
@@ -371,6 +372,10 @@ def resolve_rubygems(package_root, request):
         if dependency["kind"] == "GEM":
             _push_downloaded_gem(dependency, rubygems_repo_name)
 
+    for dep in dependencies:
+        if dep["kind"] == "GIT":
+            unpack_git_dependency(dep)
+
     dependencies = cleanup_metadata(dependencies)
     name, version = _get_metadata(package_root, request)
     if package_root == bundle_dir:
@@ -382,6 +387,22 @@ def resolve_rubygems(package_root, request):
         "package": {"name": name, "version": version, "type": "rubygems", "path": package_rel_path},
         "dependencies": dependencies,
     }
+
+
+def unpack_git_dependency(dep):
+    """
+    Unpack the archive with the downloaded dependency.
+
+    Only the unpacked directory is kept, the archive is deleted.
+    To get more info on local Git repos, see:
+    https://bundler.io/man/bundle-config.1.html#LOCAL-GIT-REPOS
+    :param dep: RubyGems GIT dependency
+    """
+    #
+    extracted_path = str(dep["path"]).removesuffix(".tar.gz")
+    shutil.unpack_archive(dep["path"], extracted_path)
+    os.remove(dep["path"])
+    dep["path"] = extracted_path
 
 
 def _upload_rubygems_package(repo_name, artifact_path):
--- a/cachito/workers/tasks/rubygems.py
+++ b/cachito/workers/tasks/rubygems.py
@@ -1,9 +1,16 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+from os.path import relpath
+from pathlib import Path
+from textwrap import dedent
+
+from cachito.errors import CachitoError
 from cachito.workers import nexus
 from cachito.workers.pkg_managers.rubygems import (
     get_rubygems_hosted_repo_name,
     get_rubygems_nexus_username,
 )
 from cachito.workers.tasks.celery import app
+from cachito.workers.tasks.utils import make_base64_config_file
 
 __all__ = ["cleanup_rubygems_request"]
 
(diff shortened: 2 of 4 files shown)


cachito/workers/pkg_managers/rubygems.py
```diff
@@ -376,6 +386,26 @@ def resolve_rubygems(package_root, request):
...
+    https://bundler.io/man/bundle-config.1.html#LOCAL-GIT-REPOS
+    :param dep: RubyGems GIT dependency
+    """
+    #
+    extracted_path = str(dep["path"]).removesuffix(".tar.gz")
+    shutil.unpack_archive(dep["path"], extracted_path)
+    os.remove(dep["path"])
+    dep["path"] = extracted_path
+
+    subprocess.run(f"cd {extracted_path}/app && git switch {dep['branch']}", check=True)
```
tichavskym, line 406: Can I execute this command or do I need to use something like GitPython (Bandit is complaining)?

tichavskym:
The suggested workflow is following: Cachito unpacks all RubyGems GIT dependencies in `deps/` directory, it checkouts the branch that is specified in Gemfile for the given dependency and for each top level package provides a configuration file, that should be added to `bundle_dir/app/package/.bundle/config` that redirects Bundler to use locally downloaded GIT repos and GEM dependencies to be downloaded from the proxy repo for the request.

cachito/workers/pkg_managers/rubygems.py
chmeliik: Is the checkout necessary? The archive cloned by Cachito should be checked out at the exact ref specified when creating the `Git` object

https://github.com/containerbuildsystem/cachito/blob/35fb8db4a8146422a8772f4fb5b261c2cbaef2b2/cachito/workers/pkg_managers/rubygems.py#L317

cachito/workers/tasks/rubygems.py
```diff
@@ -16,3 +23,61 @@ def cleanup_rubygems_request(request_id):
...
+        # Defaults to true when deployment is set to true
+        BUNDLE_FROZEN: "true"
+    """
+    )
+
+    config = [base_config]
+    for dependency in dependencies:
+        if dependency["kind"] == "GIT":
+            # These substitutions are required by Bundler
+            name = dependency["name"].upper().replace("-", "___")
```
chmeliik, line 69: Is this supposed to be 3 underscores? Just asking because other things seem to use 2 underscores

cachito/workers/tasks/rubygems.py
```diff
@@ -16,3 +23,61 @@ def cleanup_rubygems_request(request_id):
...
+        # Make sure gems aren’t installed system-wide but only for the given project
+        BUNDLE_PATH: "vendor/bundle"
+        # Sets mirror for all RubyGems sources
+        BUNDLE_MIRROR__ALL: "{rubygems_hosted_url}"
+        # Turn off the probing
+        BUNDLE_MIRROR__ALL__FALLBACK_TIMEOUT: "false"
+        # Install only ruby platform gems (=> gems with native extensions are compiled from source).
+        # All gems should be platform independent already, so why not keep it here.
+        BUNDLE_FORCE_RUBY_PLATFORM: "true"
+        BUNDLE_DEPLOYMENT: "true"
```
chmeliik, line 59: Will `BUNDLE_DEPLOYMENT: "true"` be universally applicable for all use cases? Or should we leave it to the user to set this?

cachito/workers/tasks/rubygems.py
```diff
@@ -16,3 +23,61 @@ def cleanup_rubygems_request(request_id):
...
+
+    Returns a Bundler config file with a mirror set for RubyGems dependencies pointing to
+    `rubygems_hosted_repo` URL. All GIT dependencies are configured to be replaced by local git
+     repos.
+
+    :param dependencies: an array of dependencies (dictionaries) with keys
+        "name": package name,
+        "path": an absolute path to a locally downloaded git repo,
+        "kind": dependency kind
+    :param bundle_dir: and absolute path to a bundle root directory
```
chmeliik, line 42: Could you mention here that `bundle_dir` is the root of the Cachito bundle? At first I thought it was a Bundler bundle and was confused what was the difference between this and `package_source_dir`

cachito/workers/pkg_managers/rubygems.py
tichavskym: I will double check this and let you know, but I tested it manually without this at first, and I was always receiving errors that there's no branch checked out in the local git repo. But looking at it, yes it's true that the right commit will be checked out based on commit hash, but it will be just a detached head and not a branch (we're not passing any information about the exact branch), which confuses Bundler.

cachito/workers/tasks/rubygems.py
akostadinov: It seems like this doesn't concern how image is built but only how cachito operated internally. So It should be transparent to the users if I'm not mistaken. The behavior of this option is described here: https://bundler.io/man/bundle-install.1.html#DEPLOYMENT-MODE

cachito/workers/tasks/rubygems.py
```diff
@@ -16,3 +23,61 @@ def cleanup_rubygems_request(request_id):
...
+    :param package_source_dir: the root directory of given package
+    :param rubygems_hosted_url: URL pointing to a request specific RubyGems hosted repo with
+     hardcoded user credentials
+    :return: dict with "content", "path" and "type" keys
+    """
+    base_config = dedent(
+        f"""
+        # Make sure gems aren’t installed system-wide but only for the given project
+        BUNDLE_PATH: "vendor/bundle"
```
akostadinov, line 51: This seems to be redundant if `BUNDLE_DEPLOYMENT` is set, according to docs.

cachito/workers/tasks/rubygems.py
tichavskym: Yes, from manual testing I can say it matches substitutions that [Bundler does for credentials](https://bundler.io/man/bundle-config.1.html#CREDENTIALS-FOR-GEM-SOURCES), where `-` is mapped to `___` (3 underscores) and `.` to `__` (2 underscores - I forgot about it)

cachito/workers/pkg_managers/rubygems.py
tichavskym: You were right, [there's an option disabling branch checks](https://bundler.io/man/bundle-config.1.html#CREDENTIALS-FOR-GEM-SOURCES): `BUNDLE_DISABLE_LOCAL_BRANCH_CHECK: "true"`

Therefore, no checkout will be necessary.

tests/test_workers/test_tasks/test_rubygems.py
```diff
@@ -12,3 +18,56 @@ def test_cleanup_rubygems_request(mock_exec_script):
...
+
+
+@pytest.mark.parametrize("exists", [True, False])
+def test_get_config_file(tmp_path, exists):
+    bundle_dir = MockBundleDir(tmp_path)
+
+    pkg_and_deps_info = {
+        "dependencies": [
+            {
+                "name": "rspec-core",
```
lkolacek, line 41: should we add name with `.` in it to test also replacement for `__`?

cachito/workers/pkg_managers/rubygems.py
```diff
@@ -378,6 +383,22 @@ def resolve_rubygems(package_root, request):
     }
 
 
+def unpack_git_dependency(dep):
```
lkolacek, line 386: Should we add a separate unit test for this method? Coverage dropped a little bit.

tichavskym: Yes, it's added

lkolacek (approved):
LGTM :+1: Thanks!

ejegrova (approved):
LGTM

PR #684, opened 2022-07-25
Add Content manifest

tichavskym:

### DIFF
--- a/cachito/web/content_manifest.py
+++ b/cachito/web/content_manifest.py
@@ -4,11 +4,13 @@ from typing import List, Optional
 
 import flask
 
+from cachito.common.utils import get_repo_name
 from cachito.web.purl import (
-    replace_parent_purl_gomod,
     replace_parent_purl_gopkg,
+    replace_parent_purl_placeholder,
     to_purl,
     to_top_level_purl,
+    to_vcs_purl,
 )
 from cachito.web.utils import deep_sort_icm
 from cachito.workers.pkg_managers import gomod
@@ -55,6 +57,8 @@ class ContentManifest:
         self._pip_data = {}
         # dict to store yarn package data; uses the package id as key to identify a package
         self._yarn_data = {}
+        # dict to store rubygems package data; uses the package id as key to identify a package
+        self._rubygems_data = {}
         # dict to store gitsubmodule package level data; uses the package id as key to identify a
         # package
         self._gitsubmodule_data = {}
@@ -69,7 +73,7 @@ class ContentManifest:
         if dependency.type == "gomod":
             parent_purl = self._gomod_data[package.name]["purl"]
             dep_purl = to_purl(dependency)
-            dep_purl = replace_parent_purl_gomod(dep_purl, parent_purl)
+            dep_purl = replace_parent_purl_placeholder(dep_purl, parent_purl)
             icm_source = {"purl": dep_purl}
             self._gomod_data[package.name]["dependencies"].append(icm_source)
 
@@ -153,6 +157,24 @@ class ContentManifest:
         if not dependency.dev:
             pkg_type_data[package]["dependencies"].append(icm_dependency)
 
+    def process_rubygems_package(self, package, dependency):
+        """
+        Process RubyGems package.
+
+        :param Package package: the RubyGems package to process
+        :param Dependency dependency: the RubyGems package dependency to process
+        """
+        if dependency.type == "rubygems":
+            parent_package_name = get_repo_name(self.request.repo).split("/")[-1]
+            parent_purl = to_vcs_purl(parent_package_name, self.request.repo, self.request.ref)
+
+            dep_purl = to_purl(dependency, package.path)
+            dep_purl = replace_parent_purl_placeholder(dep_purl, parent_purl)
+
+            icm_dependency = {"purl": dep_purl}
+            self._rubygems_data[package]["sources"].append(icm_dependency)
+            self._rubygems_data[package]["dependencies"].append(icm_dependency)
+
     def to_json(self):
         """
         Generate the JSON representation of the content manifest.
@@ -165,6 +187,7 @@ class ContentManifest:
         self._npm_data = {}
         self._pip_data = {}
         self._yarn_data = {}
+        self._rubygems_data = {}
         self._gitsubmodule_data = {}
 
         for package in self.packages:
@@ -178,7 +201,7 @@ class ContentManifest:
             elif package.type == "gomod":
                 purl = to_top_level_purl(package, self.request, subpath=package.path)
                 self._gomod_data.setdefault(package.name, {"purl": purl, "dependencies": []})
-            elif package.type in ("npm", "pip", "yarn"):
+            elif package.type in ("npm", "pip", "yarn", "rubygems"):
                 purl = to_top_level_purl(package, self.request, subpath=package.path)
                 data = getattr(self, f"_{package.type}_data")
                 data.setdefault(package, {"purl": purl, "dependencies": [], "sources": []})
@@ -204,6 +227,8 @@ class ContentManifest:
                     self.process_pip_package(package, dependency)
                 elif package.type == "yarn":
                     self.process_yarn_package(package, dependency)
+                elif package.type == "rubygems":
+                    self.process_rubygems_package(package, dependency)
 
         # Adjust source level dependencies for go packages
         self.set_go_package_sources()
@@ -213,6 +238,7 @@ class ContentManifest:
             *self._npm_data.values(),
             *self._pip_data.values(),
             *self._yarn_data.values(),
+            *self._rubygems_data.values(),
             *self._gitsubmodule_data.values(),
         ]
         return self.generate_icm(top_level_packages)
--- a/cachito/web/purl.py
+++ b/cachito/web/purl.py
@@ -10,11 +10,12 @@ from cachito.errors import ContentManifestError
 PARENT_PURL_PLACEHOLDER = "PARENT_PURL"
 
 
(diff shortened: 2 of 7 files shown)


tests/test_content_manifest.py
```diff
@@ -245,6 +242,47 @@ def test_process_pip(default_request, default_toplevel_purl):
...
+    cm = ContentManifest(default_request, packages)
+    cm.to_json()
+
+    expected_contents = {
+        package: {
+            "purl": expected_purl,
+            "dependencies": [{"purl": expected_gem_purl}, {"purl": expected_git_purl}],
+            "sources": [{"purl": expected_gem_purl}, {"purl": expected_git_purl}],
+        },
+    }
```
tichavskym, line 279: @chmeliik As I was implementing pushing of dependencies to temporary repo, I realized, it would be a good idea to think about the way returned metadata will be used for content manifest generation. What do you think I should do about PATH dependencies (dependency that is part of the top level repo itself, its source code is in some of the subdirectories of the project)? Should they be part of the generated content manifest? If so, should they be part of both "dependencies" and "sources" and do you have any preference on how to create a purl for this type, since there is nothing in the PURL spec about them?

chmeliik: PURLs do support relative paths from the package root, so you can do something like `<parent_purl>#some/path`.

Here's how Cachito does it for gomod:
https://github.com/containerbuildsystem/cachito/blob/fe5633ad782b407d99c9243270dd42bc7d91fa85/cachito/web/content_manifest.py#L322-L325
https://github.com/containerbuildsystem/cachito/blob/fe5633ad782b407d99c9243270dd42bc7d91fa85/cachito/web/content_manifest.py#L72

They should be in both "dependencies" and "sources", unless they are non-production dependencies (dev / test etc.) in which case they would go only in "sources" (but AFAIK we won't be detecting dev vs. non-dev for Ruby, so they should go in both)

chmeliik: Note that the relative paths must not contain `..` though, I don't remember if we allow PATH deps from parent directories or not but if we do, the PARENT_PURL would probably have to be a purl of the repo itself (not the package which is in some subpath of the repo)

(see https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#rules-for-each-purl-component)

tichavskym: Thank you very much

cachito/web/content_manifest.py
```diff
@@ -377,16 +403,23 @@ def to_purl(self):
...
                 # Version is a plain URL
                 fragments = urllib.parse.parse_qs(parsed_url.fragment)
                 checksum = fragments["cachito_hash"][0]
                 quoted_url = urllib.parse.quote(self.version, safe="")
                 return f"pkg:generic/{name}?download_url={quoted_url}&checksum={checksum}"
+        elif self.type == "rubygems":
+            if self.kind == "PATH":
+                # Package is relative to the parent module
+                normpath = os.path.normpath(self.path)
+                return f"{PARENT_PURL_PLACEHOLDER}#{normpath}"
```
chmeliik, line 417: Seeing this, I realized we may need more complex logic here :disappointed: 

If the parent package is not in the repo root, its purl will look something like this:
```
pkg:github/organization/repo@commit-hash#subpath
```
Appending `#{normpath}` would make the purl invalid
```
pkg:github/organization/repo@commit-hash#subpath#other-path
```
We would want
```
pkg:github/organization/repo@commit-hash#subpath/other-path
```
The only reason this works for gomod is that the parent purls never contain subpaths (top-level purls can use `pkg:golang` unlike other package managers)

chmeliik: Or maybe the more complex logic will need to go to the code which replaces the parent placeholder

tichavskym: You're right, I added a test case for this. The path that is in the dependency metadata is the path relative to the repo root, so the parent purl is now generated from the `request.repo`.

chmeliik: > The path that is in the dependency metadata is the path relative to the repo root

Hmm, I think we want it to be relative to the package

chmeliik: Or maybe it's fine, just needs to be documented

cachito/web/purl.py
```diff
@@ -101,6 +100,33 @@ def _to_purl_pip(package):
         return f"pkg:generic/{name}?download_url={quoted_url}&checksum={checksum}"
 
 
+def _to_purl_rubygems(package):
+    if package.version and package.version.startswith("./"):
+        # Package is relative to the parent module
+        normpath = os.path.normpath(package.version.removeprefix("./"))
```
chmeliik, line 106: AFAIK `normpath` removes the `./` prefix already

cachito/web/purl.py
```diff
@@ -101,6 +100,33 @@ def _to_purl_pip(package):
...
+        return f"{PARENT_PURL_PLACEHOLDER}#{normpath}"
+    parsed_url = urllib.parse.urlparse(package.version)
+    if parsed_url.scheme.startswith("git+"):
+        # refactor it out from pip
```
chmeliik, line 110: IMO this comment is not really necessary, it won't make sense out of the context of this PR

cachito/web/content_manifest.py
chmeliik: In the interest of keeping Cachito consistent, I would prefer for the `version` of PATH deps to be relative to the package dir, not to the repository root dir. In the content manifest code, you would then have to get the full path from the parent package path + the dependency path

cachito/web/content_manifest.py
```diff
@@ -153,6 +156,24 @@ def _process_standard_package(self, pkg_type, package, dependency):
...
 
+    def process_rubygems_package(self, package, dependency):
+        """
+        Process RubyGems package.
+
+        :param Package package: the RubyGems package to process
+        :param Dependency dependency: the RubyGems package dependency to process
+        """
+        if dependency.type == "rubygems":
+            parent_package_name = self.request.repo.removesuffix(".git").split("/")[-1]
```
chmeliik, line 167: It might be good to re-use the logic from scm.py for this https://github.com/containerbuildsystem/cachito/blob/35fb8db4a8146422a8772f4fb5b261c2cbaef2b2/cachito/workers/scm.py#L300

That would avoid a corner case where e.g. `https://github.com/containerbuildsystem/cachito/` resolves the package name to `''`

`repo_name` would return `containerbuildsystem/cachito`, so you might still want to split it and take only the actual repo name

Note that importing `cachito.workers.scm` in the `cachito.web.*` code probably won't work (missing git), so you would need to move the repo name logic somewhere else, probably into `cachito.common.utils`

chmeliik: Hmm, but `repo_name` has a different corner case:
```python
>>> repo_name("git@github.com:containerbuildsystem/cachito.git")
'git@github.com:containerbuildsystem/cachito'
```
But either way, it would be good to use the same repo name logic here and in #692 and address the trailing slash corner case

cachito/web/purl.py
tichavskym: Yes, sorry, this was a TODO comment for myself

cachito/web/purl.py
tichavskym: you're right

cachito/web/content_manifest.py
tichavskym: I both cases, we're getting URL from the request, so I don't think we need to care about PURL format conversion.

cachito/web/content_manifest.py
tichavskym: If we want `version` to contain relative path to parent package directory, then the function generating PURL has to get information about relative path from the parent package to the repo root. I have no better idea then to get this from package name (we could have `resolve_rubygems` function to return package path as part of the metadata it returns, if we wanted), so this PR depends on merging #692. What do you think about [New way of handling rubygems PURL](https://github.com/containerbuildsystem/cachito/pull/684/commits/8374d81b97d749d3d5c7fb3ad9e9979eeac360ef)?

tichavskym: It's not completely finished, I'll need to change the function generating PATH metadata, I'm mostly interested if I'm going in the right direction with PURL generation.

cachito/web/content_manifest.py
chmeliik: Ah, I didn't mean PURL conversion, I meant SSH urls. Cachito does support those, and some requests actually use them:
```json
{
  "ref": "f7f69d8762afe9a936d1a6a26d4d06444e8b8929",
  "repo": "git@github.com:openshift/openshift-velero-plugin.git"
  ...
}
```

chmeliik: But since you split on `/`  and take the last elem, that's not actually a problem :+1:

cachito/web/content_manifest.py
chmeliik: > the function generating PURL has to get information about relative path from the parent package to the repo root

https://github.com/containerbuildsystem/cachito/blob/8374d81b97d749d3d5c7fb3ad9e9979eeac360ef/cachito/web/content_manifest.py#L171

The `package` object should have a `path` - which is exactly the relative path from the repo root to the package (unless the package is in the repo root, in which case the path should be None)

chmeliik: The updates to the purl module make sense to me :+1:

tichavskym:
I think I addressed all the issues so far, so I'd like to ask you for additional reviews.

ejegrova (approved):
LGTM

taylormadore (approved):

PR #685, opened 2022-07-29
Add uploading of RubyGems to Nexus functionality

tichavskym:
SPMM-9902

### DIFF
--- a/cachito/workers/pkg_managers/rubygems.py
+++ b/cachito/workers/pkg_managers/rubygems.py
@@ -12,7 +12,7 @@ from gemlock_parser.gemfile_lock import GemfileLockParser
 
 from cachito.errors import NexusError, ValidationError
 from cachito.workers import get_worker_config, nexus
-from cachito.workers.errors import NexusScriptError
+from cachito.workers.errors import NexusScriptError, UploadError
 from cachito.workers.paths import RequestBundleDir
 from cachito.workers.pkg_managers.general import (
     download_binary_file,
@@ -21,6 +21,8 @@ from cachito.workers.pkg_managers.general import (
     upload_raw_package,
 )
 from cachito.workers.scm import Git
+
+GEMFILE_LOCK = "Gemfile.lock"
 
 GIT_REF_FORMAT = re.compile(r"^[a-fA-F0-9]{40}$")
 PLATFORMS_RUBY = re.compile(r"^PLATFORMS\n {2}ruby\n\n", re.MULTILINE)
@@ -181,7 +183,7 @@ def finalize_nexus_for_rubygems_request(rubygems_repo_name, raw_repo_name, usern
     return password
 
 
-def download_dependencies(request_id, dependencies):
+def download_dependencies(request_id, dependencies, package_root):
     """
     Download all dependencies from Gemfile.lock with its sources.
 
@@ -191,6 +193,7 @@ def download_dependencies(request_id, dependencies):
 
     :param int request_id: ID of the request these dependencies are being downloaded for
     :param list[GemMetadata] dependencies: List of dependencies
+    :param package_root: path to the root of the processed package
     :return: Info about downloaded packages; all items will contain "kind" and "path" keys
         (and more based on kind, see _download_*_package functions for more details)
     :rtype: list[dict]
@@ -218,16 +221,19 @@ def download_dependencies(request_id, dependencies):
             download_info = _download_git_package(
                 dep, bundle_dir.rubygems_deps_dir, rubygems_raw_repo_name, nexus_auth
             )
+        elif dep.type == "PATH":
+            download_info = _get_path_package_info(dep, bundle_dir, package_root)
         else:
             # Should not happen
             raise RuntimeError(f"Unexpected dependency type: {dep.type!r}")
 
-        log.info(
-            "Successfully downloaded gem %s (%s) to %s",
-            dep.name,
-            dep.version,
-            download_info["path"].relative_to(bundle_dir),
-        )
+        if dep.type != "PATH":
+            log.info(
+                "Successfully downloaded gem %s (%s) to %s",
+                dep.name,
+                dep.version,
+                download_info["path"].relative_to(bundle_dir),
+            )
 
         # If the raw component is not in the Nexus hoster instance, upload it there
         if dep.type == "GIT" and not download_info["have_raw_component"]:
@@ -247,6 +253,7 @@ def download_dependencies(request_id, dependencies):
             )
 
         download_info["kind"] = dep.type
+        download_info["type"] = "rubygems"
         downloads.append(download_info)
 
     return downloads
@@ -257,6 +264,7 @@ def _download_rubygems_package(gem, deps_dir, proxy_url, proxy_auth):
 
     The platform independence is ensured by downloading it from platform independent url
     (url that doesn't have any platform suffix).
+
     :param GemMetadata gem: Gem dependency from a Gemfile.lock file
     :param Path deps_dir: The deps/rubygems directory in a Cachito request bundle
     :param str proxy_url: URL of Nexus RubyGems proxy
@@ -270,7 +278,7 @@ def _download_rubygems_package(gem, deps_dir, proxy_url, proxy_auth):
     download_binary_file(proxied_url, download_path, auth=proxy_auth)
 
     return {
-        "package": gem.name,
+        "name": gem.name,
         "version": gem.version,
         "path": download_path,
     }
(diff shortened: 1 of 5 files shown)


tichavskym:
@chmeliik Ready for review

cachito/workers/pkg_managers/rubygems.py
```diff
@@ -312,11 +322,114 @@ def _download_git_package(gem, rubygems_deps_dir, rubygems_raw_repo_name, nexus_
...
+
+    rubygems_repo_name = get_rubygems_hosted_name(request["id"])
+    for dependency in dependencies:
+        if dependency["kind"] == "GEM":
+            _push_downloaded_gem(dependency, rubygems_repo_name)
+
+    return {
+        "package": {"type": "rubygems"},
+        "dependencies": dependencies,
+        "gemfile_lock": gemlock_path,
```
chmeliik, line 377: Will we need "gemfile_lock" for something? AFAIK pip only uses the paths when rewriting requirements files to point to git and https deps to the Nexus repo

cachito/workers/pkg_managers/rubygems.py
```diff
@@ -312,11 +322,114 @@ def _download_git_package(gem, rubygems_deps_dir, rubygems_raw_repo_name, nexus_
...
+    :param package_root: path to the root of the processed package
+    :return: dict with name, version and path keys
+    """
+    path = Path(package_root / dep.source).resolve().relative_to(bundle_dir.source_root_dir)
+
+    return {
+        "name": dep.name,
+        "version": dep.version,
+        "path": path,
+    }
```
chmeliik, line 352: What will be the value of `dep.version` here? The path to the dependency or a version similar to those of GEM deps? Note that, when generating the content manifest later, we will need to be able to tell the difference between a PATH dependency and a GEM dependency.

tichavskym: Even for PATH dependencies, [version is resolved](https://github.com/3scale/porta/blob/ce87a98f0f93696155dd9602220e5093bc278386/Gemfile.lock#L80) (based on .gemspec of that dependency). So here the version could be "1.0.0". Cachito doesn't really need that version later, so one way of telling the difference would be setting the version to "PATH" and then during content manifest generation considering all gems with `version == "PATH"` to be path dependencies. The second, and IMO a cleaner option is to add `kind` slot to `Package` in `content_manifest.py`, as you can see in #684.

cachito/workers/pkg_managers/rubygems.py
tichavskym: Yes, let's say there are two packages in the repository, `package1` and `package2`. We want to be able to run `bundle install` from both `repo/package1` and `repo/package2` directories. In order to redirect Bundler to use Nexus or local git repos, we have to create config files at `repo/package1/.bundle/config` and similarly for second package. And to be able to properly determine the path where to create these config files, we would do `config_file_path = gemlock_path.parent / Path(".bundle/config")`. Note that there's no support of alternative locations for `Gemfile.lock`, so we expect it always to be `package/Gemfile.lock`. Clearly, we could replace this with something like package to the package.

cachito/workers/pkg_managers/rubygems.py
lkolacek: +1 for `kind` slot to `Package` in `content_manifest.py` for better visibility

cachito/workers/pkg_managers/rubygems.py
```diff
@@ -312,11 +322,114 @@ def _download_git_package(gem, rubygems_deps_dir, rubygems_raw_repo_name, nexus_
...
+        list returned by the download_dependencies function
+    :param str rubygems_repo_name: Name of the Nexus RubyGems hosted repository to push
+        the requirement to
+    :return: dict with the cachito Dependency representation
+    :rtype: dict
+    :raises CachitoError: If Nexus upload operation fails
+    """
+    try:
+        _upload_rubygems_package(rubygems_repo_name, dependency["path"])
+    except CachitoError:
```
lkolacek, line 408: We recently changed error classification to be more detailed for debugging. Should we change it to [NetworkError](https://github.com/containerbuildsystem/cachito/blob/master/cachito/workers/nexus.py#L387)? https://github.com/containerbuildsystem/cachito/pull/680

tests/test_workers/test_pkg_managers/test_rubygems.py
```diff
@@ -633,3 +661,154 @@ def test_download_dependencies(
...
+    rubygems_repo_name = "test-rubygems-hosted"
+    name = "foo"
+    version = "1"
+    path = "some/path"
+    dep = {"name": name, "version": version, "path": path, "kind": "GEM"}
+    if uploaded:
+        rubygems._push_downloaded_gem(dep, rubygems_repo_name)
+        mock_upload.assert_called_once_with(rubygems_repo_name, path)
+    else:
+        with pytest.raises(CachitoError, match="stub"):
```
lkolacek, line 801: Also related to above comment.

cachito/workers/pkg_managers/rubygems.py
chmeliik: You should also consider how such dependencies will be presented in the request JSON https://github.com/containerbuildsystem/cachito/blob/master/docs/metadata.md#request-json

I would avoid using versions like "1.0.0" for PATH deps, as that could lead to confusion where the consumers of Cachito metadata would think that the dependency comes from the rubygems registry.

tichavskym: Oh, I see. I don't really understand the whole workflow of generating request JSON, but I assume it's produced from the same JSON file that content manifest generation is using. 

Therefore, to match the rest of the package managers, `resolve_rubygems` could return for each dependency `name`, `type`, `version` (which for PATH dependencies will contain just the path, ad gomod does, and for GIT it could be same as for pip `git+url@hash`, so that it's clear it's a different type of dependency) and maybe `dev`. `replaces` don't have to be returned, because it is filled with null automatically, if nothing is specified.

Now the question is, do we want to return `kind` as well (and possibly the other metadata that are returned in the current implementation)? I think there's no harm in doing that, and it will be better than [what `gomod` is doing](https://github.com/containerbuildsystem/cachito/blob/28a6afc130df5296945898b7d8578e5bf2157ac0/cachito/web/content_manifest.py#L322), just assuming that every version starting with `.` is a PATH dependency. What do you think?

cachito/workers/pkg_managers/rubygems.py
lkolacek: +1  feel free to add new more precise error type for your case. :)

cachito/workers/pkg_managers/rubygems.py
chmeliik: > Oh, I see. I don't really understand the whole workflow of generating request JSON, but I assume it's produced from the same JSON file that content manifest generation is using.

Yeah, it's the same file. The request JSON is generated here (the packages and dependencies sections, specifically): https://github.com/containerbuildsystem/cachito/blob/6c3de13b9b72f243218467511f04d04b96c8cbd3/cachito/web/models.py#L457-L459

AFAIK this pretty much just dumps the content of the JSON directly into the API response.

> Therefore, to match the rest of the package managers, resolve_rubygems could return for each dependency name, type, version (which for PATH dependencies will contain just the path, ad gomod does, and for GIT it could be same as for pip git+url@hash, so that it's clear it's a different type of dependency) and maybe dev. replaces don't have to be returned, because it is filled with null automatically, if nothing is specified.

+1 for the versions. Returning `dev` is not necessary if we don't have that info

> Now the question is, do we want to return kind as well (and possibly the other metadata that are returned in the current implementation)?

Ideally, the metadata returned in the API response should be common for all package managers. A `kind` key used only for ruby would not be ideal, but if it's documented well it would be OK. Another thing to consider would be that, in case we eventually want to add `kind` for other package managers, there should be a common set of values for the `kind` key.

That said, it would also be OK to  save `kind` in packages.json but clean it out of the API response

cachito/workers/pkg_managers/rubygems.py
chmeliik: Does the function need to return the path? Doesn't the caller of the `<generate bundle config>` function already know the path?

cachito/workers/pkg_managers/rubygems.py
chmeliik: > That said, it would also be OK to save kind in packages.json but clean it out of the API response

(but PATH deps should still be distinguishable from GEM deps)

cachito/workers/pkg_managers/rubygems.py
tichavskym: Yes, you're right.

tichavskym:
@chmeliik @lkolacek Added a new error type, function that performs cleanup of dependencies' metadata and documentation about `kind`.

docs/metadata.md
```diff
@@ -273,6 +274,11 @@ Note that this does not apply to replacement mechanisms supported natively by so
 package managers, such as the go.mod [replace][gomod-replace] directive. Those
 replacements will not be reflected in the request JSON in any way.
 
+#### dependency.kind
+
+Returned only for `RubyGems` dependencies, this makes it easy to distinguish between different 
+kinds of them. Possible values are `GEM`, `GIT` or `PATH`.
```
chmeliik, line 280: I feel like this sets up for more problems in the future. It's not very well justifiable why this key is only used for ruby when it's also applicable to pip, npm, yarn and maybe even gomod. But if we want to add `kind` to the other package managers in the future and the known values for it are `GEM`, `GIT`, `PATH`, it will make things awkward.

The rest of the Cachito team, what do you think?

tichavskym: I could still do it the same way as `gomod` does it with PATH dependencies starting with `./`, or I could come up with some values that are universal, such as `git`, `path`, `url`, `package-repository`.

lkolacek: I agree here with @chmeliik. In case we can't use the same set of values [GIT, GEM, PATH] for all package managers with the `kind` item, we should rather use the gomod approach and distinguish `kind/type` from parsing version string. @tichavskym do we have a set of all possible versions to be resolved here as GIT, GEM, PATH "kind" which we could add to unit tests?

tichavskym: > The RubyGems team urges gem developers to follow the [Semantic Versioning](http://semver.org/) standard for their gem’s versions. The RubyGems library itself does not enforce a strict versioning policy, but using an “irrational” policy will only be a disservice to those in the community who use your gems.

In other words, GEM dependency can have whatever it wants as a version. So, we could say GIT dependencies' version will be in format `git+<url>@hash` and PATH dependencies' version will start with `.` [as gomod is doing](https://github.com/containerbuildsystem/cachito/blob/28a6afc130df5296945898b7d8578e5bf2157ac0/cachito/web/content_manifest.py#L322) or `path+` if we wanted to come up with something new, but there's no guarantee that it won't collide with any of the GEM versions, so we would have to hope maintainers will name it so that it makes sense.

tichavskym: Meaning, for every GEM dependency with version starting with `.` (which is possible to have), we would produce the wrong purl. If we used `kind` internally and removed it from the responses, the purl would be produced correctly, but it would still be confusing in the response JSON, because users would look at version starting with `.` and be thinking that its a PATH dependency. The same would probably be true for all GEM dependencies with versions starting with `git+`. I'm not sure to which degree we want to take care of these edge cases (or what other package managers are doing about this).

chmeliik: I wouldn't worry too much about these edge cases. Maybe check that PATH deps start with `./` -- if someone uses a version like that for their GEM, I'd assume they're deliberately messing with us.

I think pip isn't all that strict about enforcing the version rules either.

And you don't even want to know what we do with NPM, the purls are not even close to valid and haven't been for 2 years :sweat_smile:

tests/test_workers/test_pkg_managers/test_rubygems.py
```diff
@@ -555,18 +579,18 @@ def test_download_dependencies(
...
-            "package": "bar",
+            "name": "bar",
             "path": git_download,
-            "url": git_url,
-            "ref": GIT_REF.lower(),
+            "version": f"git+{git_url}@{GIT_REF}",
             "raw_component_name": f"bar/bar-external-gitcommit-{GIT_REF}.tar.gz",
             "have_raw_component": have_raw_component,
         }
+        path_info = {"name": "baz", "version": "3.1.1", "path": "vendor/active-docs"}
```
chmeliik, line 593: nitpick: this is not the expected data for a path dep

docs/metadata.md
tichavskym: Ok :laughing: In that case, this pull request is ready for merge/final review.

chmeliik (approved):
LGTM, one minor nitpick

lkolacek (approved):
LGTM +1 Thank you!

PR #687, opened 2022-08-01
Update 'PYTHON_VERSION_VENV' to 3.10 and add python3.10 to tox configs

sumincho22:
CLOUDBLD-3117

Since Cachito services run on Python 3.10, the unit tests should also cover this version as well. 

Should we get rid of Python 3.8 from the unit tests now? Three runs of unit testing seems time-consuming.

Also, we will need to add Python 3.10 to GitHub Actions.

### DIFF
--- a/.github/workflows/gating.yaml
+++ b/.github/workflows/gating.yaml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.9, 3.8]
+        python-version: ["3.10", "3.9"]
 
     steps:
     - uses: actions/checkout@v1
@@ -76,7 +76,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: 3.8
+        python-version: "3.9"
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
@@ -129,7 +129,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: 3.8
+        python-version: "3.10"
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
--- a/tox.ini
+++ b/tox.ini
@@ -1,10 +1,10 @@
 [tox]
-envlist = bandit,black,flake8,mypy,py38,py39
+envlist = bandit,black,flake8,mypy,python3.9,python3.10
 
 [gh-actions]
 python =
-    3.8: py38
-    3.9: py39
+    3.9: python3.9
+    3.10: python3.10
 
 [testenv]
 deps =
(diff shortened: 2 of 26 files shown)


lkolacek:
IMO we can drop Py 3.8 form unit tests. Other thoughts?

taylormadore:
I think we'll also need to update the python-version in a few places in the gh workflows config: https://github.com/containerbuildsystem/cachito/blob/28a6afc130df5296945898b7d8578e5bf2157ac0/.github/workflows/gating.yaml#L17

sumincho22:
> IMO we can drop Py 3.8 form unit tests. Other thoughts?

- Removed Python 3.8.
- Made 3.9 the base version for unit and integration tests.
- Fixed GitHub Actions to run only 3.9 and 3.10 on unit tests

ejegrova (commented):
Please update also:
https://github.com/containerbuildsystem/cachito/blob/28a6afc130df5296945898b7d8578e5bf2157ac0/setup.py#L41
https://github.com/containerbuildsystem/cachito/blob/28a6afc130df5296945898b7d8578e5bf2157ac0/setup.py#L48

lkolacek:
And also:
```
tox.ini:35:basepython = python3.8
```

sumincho22:
When I change `target-version` to `py39`, I also have to update `black` to a version greater than 20.8b1.

ejegrova:
> When I change `target-version` to `py39`, I also have to update `black` to a version greater than 20.8b1.
> 
> Should I change `black` to the newest version? If we do this, their formatting check also changes, so we would need to reformat the code base for it to pass.

It's better to have the latest version. If it works, please do these changes.

sumincho22:
Updated `black` to version 22.6.0 (latest). Applied new formatting requirements.

.github/workflows/gating.yaml
```diff
@@ -129,7 +129,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: 3.8
+        python-version: 3.9
```
taylormadore, line 132: If python 3.10 is our target environment, should we use that for the integration tests?

tox.ini
```diff
@@ -32,12 +32,11 @@ commands_post =
 [testenv:black]
 # a bug in python3.9 is causing occasional failures: https://bugs.python.org/issue43498,
```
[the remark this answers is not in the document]
sumincho22: Yep I will remove it. I will a similar comment in `pyproject.toml`

.github/workflows/gating.yaml
sumincho22: But if the tests run successfully on python 3.9, then wouldn't python 3.10 pass as well? For this reason, I thought that the lower bound version would be better since we are testing both 3.9 and 3.10.

ejegrova: Exactly, integration tests run only on python 3.9 now. Switch it to python 3.10 please.

lkolacek: +1 to use latest greatest!

sumincho22: Gotcha, switched to python 3.10

sumincho22:
@lkolacek @taylormadore @ejegrova This should be ready for approval and merge :)

taylormadore (approved):

ejegrova (approved):
looks great!

PR #688, opened 2022-08-01
Refactor purl handling

sumincho22:
CLOUDBLD-4704

Currently, the purl generation is a part of the Package class.

However, purl handling is a complex task and deserves its own module.

The purl logic should also be organized in a way that is modular and comprehensible.

### DIFF
--- a/cachito/web/content_manifest.py
+++ b/cachito/web/content_manifest.py
@@ -1,18 +1,17 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
-import os
-import re
-import urllib.parse
 from copy import deepcopy
 from typing import List, Optional
 
 import flask
-import pkg_resources
 
-from cachito.errors import ContentManifestError
+from cachito.web.purl import (
+    replace_parent_purl_gomod,
+    replace_parent_purl_gopkg,
+    to_purl,
+    to_top_level_purl,
+)
 from cachito.web.utils import deep_sort_icm
 from cachito.workers.pkg_managers import gomod
-
-PARENT_PURL_PLACEHOLDER = "PARENT_PURL"
 
 VERSION = 1
 JSON_SCHEMA_URL = (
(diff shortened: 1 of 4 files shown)


brunoapimentel:
I think this is a step in the right direction, but there's still quite a few if/else blocks that could go away with a little more refactoring, such as https://github.com/sumincho22/cachito/blob/CLOUDBLD-4704/cachito/web/content_manifest.py#L172-L206.

It seems to me that we could try to extend the Package class to have GoPackage, PipPackage, NpmPackage, etc, and try to keep this logic more concise. Wdyt?

sumincho22:
> I think this is a step in the right direction, but there's still quite a few if/else blocks that could go away with a little more refactoring, such as https://github.com/sumincho22/cachito/blob/CLOUDBLD-4704/cachito/web/content_manifest.py#L172-L206.
> 
> It seems to me that we could try to extend the Package class to have GoPackage, PipPackage, NpmPackage, etc, and try to keep this logic more concise. Wdyt?

I suppose those sub packages can be inherited from the Package class, and each one should have its own version of the process function.

However, when we set the list of packages to pass into the ContentManifest object, we would need a way to classify each package json to a sub package when using `Package.from_json(package)`.

brunoapimentel:
> I suppose those sub packages can be inherited from the Package class, and each one should have its own version of the process function.
> 
> However, when we set the list of packages to pass into the ContentManifest object, we would need a way to classify each package json to a sub package when using `Package.from_json(package)`.

You're right, but I believe it would be the only place we'd need to employ such logic, the rest of the manifest generation should go more smoothly after that.

I do believe this would result in many more changes to the ContentManifest class itself, so feel free to ignore this idea or to pursue it in a following PR. :)

sumincho22:
> You're right, but I believe it would be the only place we'd need to employ such logic, the rest of the manifest generation should go more smoothly after that.
> 
> I do believe this would result in many more changes to the ContentManifest class itself, so feel free to ignore this idea or to pursue it in a following PR. :)

@lkolacek @taylormadore What do you guys think? I might be in a little bit of a time crunch to make these changes however. I would need to have this issue done by Monday because next week is my last week.

lkolacek:
@sumincho22 I like Bruno's idea, but we don't need to rush it in this PR. We could file a technical improvement story for it and deal with it later on. I'll review your PR early in the morning. Thanks! ;)

taylormadore:
+1 Yeah, I agree we can enhance it later if needed

lkolacek (approved):
PR looks good to me. :+1: Up to you Sumin if you want to incorporate those changes Bruno mentioned. Thank you!!

ejegrova (approved):
I like suggested changes but we can do those later. PR looks good :)

sumincho22:
@lkolacek @taylormadore @ejegrova Could someone merge this PR? Thank you!

PR #690, opened 2022-08-08
Pin git to version 2.32.0

ejegrova:
In git version 2.37.1 there is an owner check for top level directory
https://github.com/git/git/commit/8959555cee7ec045958f9b6dd62e541affb7e7d9
This causes failures for git fetch cmd in Cachito.

### DIFF
--- a/README.md
+++ b/README.md
@@ -712,13 +712,13 @@ Offline Installations   | ✓     | x   | x   | x    |
 ### Actual Tool Versions
 
 Tool   | Version |
----    | ---     |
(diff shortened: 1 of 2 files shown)


lkolacek (approved):

MartinBasti:
This is just workaround not a final fix right?

ejegrova:
> This is just workaround not a final fix right?

Yes, this should be just temporary, we have another ticket for proper investigation.

MartinBasti (approved):

PR #693, opened 2022-08-16
Remove version template

ejegrova:
We have invalid version specified, it does not comply with PEP 440.
Also 'version_config' option is depracated.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,8 @@ setup(
     license="GPLv3+",
     python_requires=">=3.9",
     setup_requires=['setuptools-git-versioning'],
-    version_config={
-        "dev_template": "{tag}.post{ccount}-git.{sha}",
+    setuptools_git_versioning={
+        "enabled": True,
+        "dev_template": "{tag}.post{ccount}+git.{sha}",
     },
 )


brunoapimentel:
I think the lines we are removing here work to create versions on the fly for commits that don't have a tag on it. For example, I created a new commit with gibberish and:
```
$ python setup.py --version
1.1.0.post0+git.6c3de13b.dirty
```
With your change, I get:
```
python setup.py --version
0.0.0
```
I'm thinking if there's a way to keep this working while being compliant to PEP 440.

chmeliik:
> I'm thinking if there's a way to keep this working while being compliant to PEP 440.

Apparently the default for [dev_template](https://setuptools-git-versioning.readthedocs.io/en/stable/options/dev_template.html#default-value) is `{tag}.post{ccount}+git.{sha}` - which is what we had but with `+` instead of `-`

Which should be a valid [local version](https://peps.python.org/pep-0440/#local-version-identifiers), but not sure if setuptools will accept it

chmeliik:
Hmm, but it already had a `+` in Bruno's test, weird :thinking:

MartinBasti:
Can we just replace `s/-/+` in current config, IIUC that default doesn't work

ejegrova:
> Can we just replace `s/-/+` in current config, IIUC that default doesn't work

This seems to be working. LMK what you think about the changes. We were also getting warning: ```UserWarning: `version_config` option is deprecated since setuptools-git-versioning v1.8.0 and will be dropped in v2.0.0```

MartinBasti:
```
/usr/lib/python3.10/site-packages/setuptools/dist.py:498: UserWarning: The version specified ('sprint-58.post20+git.083ed774') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.

```

There are still warnings, but at least it's not a hard error now :|

MartinBasti:
> ```
> /usr/lib/python3.10/site-packages/setuptools/dist.py:498: UserWarning: The version specified ('sprint-58.post20+git.083ed774') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
> ```
> 
> There are still warnings, but at least it's not a hard error now :|

`sprint-58` oh wait, I dunno why it used sprint number :D

MartinBasti (approved):

ejegrova:
> ```
> /usr/lib/python3.10/site-packages/setuptools/dist.py:498: UserWarning: The version specified ('sprint-58.post20+git.083ed774') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
> ```
> 
> There are still warnings, but at least it's not a hard error now :|

Where do you get this? I don't get any warning/error with `python3 setup.py --version`

MartinBasti:
> > ```
> > /usr/lib/python3.10/site-packages/setuptools/dist.py:498: UserWarning: The version specified ('sprint-58.post20+git.083ed774') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > There are still warnings, but at least it's not a hard error now :|
> 
> Where do you get this? I don't get any warning/error with `python3 setup.py --version`

Some unknown local magic, dunno :D  I tagged it locally and it worked for me fine

lkolacek (approved):

taylormadore (approved):

PR #694, opened 2022-08-17
Fetch rubygems

tichavskym:

### DIFF
--- a/cachito/workers/pkg_managers/rubygems.py
+++ b/cachito/workers/pkg_managers/rubygems.py
@@ -5,6 +5,7 @@ import random
 import re
 import secrets
 import shutil
+import urllib.parse
 from dataclasses import dataclass
 from pathlib import Path
 
@@ -367,7 +368,7 @@ def resolve_rubygems(package_root, request):
     dependencies = parse_gemlock(package_root, gemlock_path)
     dependencies = download_dependencies(request["id"], dependencies, package_root)
 
-    rubygems_repo_name = get_rubygems_hosted_name(request["id"])
+    rubygems_repo_name = get_rubygems_hosted_repo_name(request["id"])
     for dependency in dependencies:
         if dependency["kind"] == "GEM":
             _push_downloaded_gem(dependency, rubygems_repo_name)
@@ -376,7 +377,6 @@ def resolve_rubygems(package_root, request):
         if dep["kind"] == "GIT":
             unpack_git_dependency(dep)
 
-    dependencies = cleanup_metadata(dependencies)
     name, version = _get_metadata(package_root, request)
     if package_root == bundle_dir:
         package_rel_path = None
@@ -450,31 +450,6 @@ def _push_downloaded_gem(dependency, rubygems_repo_name):
             raise
 
 
-def get_rubygems_hosted_name(request_id):
-    """
-    Get the name of the Nexus RubyGems hosted repository for the request.
-
-    :param int request_id: the ID of the request this repository is for
-    :return: the name of the RubyGems hosted repository for the request
-    :rtype: str
-    """
-    config = get_worker_config()
-    return f"{config.cachito_nexus_request_repo_prefix}rubygems-hosted-{request_id}"
-
-
-def cleanup_metadata(dependencies):
-    """
-    Cleanup dependencies' metadata, so it can be included in request JSON.
-
-    :param dependencies: which should be cleaned up
-    :return: list[dict]
-    """
-    return [
-        {"name": dep["name"], "version": dep["version"], "type": dep["type"]}
-        for dep in dependencies
-    ]
-
-
 def get_rubygems_nexus_username(request_id):
     """
     Get the username that has read access on the RubyGems hosted repo for the request.
@@ -505,3 +480,18 @@ def _get_metadata(package_root, request):
     repo_name = get_repo_name(request["repo"]).split("/")[-1]
 
     return repo_name + relative_path, request["ref"]
+
+
+def get_rubygems_hosted_url_with_credentials(username: str, password: str, request_id: int):
+    """
+    Get URL of a RubyGems hosted repo for the request with hardcoded username and a password.
+
+    :param username: the username that has read access on the RubyGems hosted repo
+    :param password: password for the user`
+    :param int request_id: the ID of the request this repository is for
+    :return str: URL of a RubyGems hosted repo for the request with hardcoded username and password
+    """
+    config = get_worker_config()
+    url = urllib.parse.urlparse(config.cachito_nexus_url)
+    repo_name = get_rubygems_hosted_repo_name(request_id)
+    return f"{url.scheme}://{username}:{password}@{url.netloc}/repository/{repo_name}/"
--- a/cachito/workers/tasks/pip.py
+++ b/cachito/workers/tasks/pip.py
@@ -73,7 +73,7 @@ def fetch_pip_source(request_id, package_configs=None):
     packages_data = []
     requirement_file_paths = []
     for pkg_cfg in package_configs:
-        pkg_path = pkg_cfg.get("path", ".")
+        pkg_path = os.path.normpath(pkg_cfg.get("path", "."))
         source_dir = bundle_dir.app_subpath(pkg_path).source_dir
         set_request_state(
             request_id,
--- a/cachito/workers/tasks/rubygems.py
+++ b/cachito/workers/tasks/rubygems.py
@@ -1,18 +1,34 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
+import logging
+import os
 from os.path import relpath
 from pathlib import Path
 from textwrap import dedent
 
+from cachito.common.packages_data import PackagesData
 from cachito.errors import CachitoError
 from cachito.workers import nexus
+from cachito.workers.config import validate_rubygems_config
+from cachito.workers.paths import RequestBundleDir
+from cachito.workers.pkg_managers.general import update_request_with_config_files
 from cachito.workers.pkg_managers.rubygems import (
+    finalize_nexus_for_rubygems_request,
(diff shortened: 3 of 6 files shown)


cachito/workers/pkg_managers/rubygems.py
```diff
@@ -478,3 +479,18 @@ def _get_metadata(package_root, request):
...
+    Get URL of a RubyGems hosted repo for the request with hardcoded username and a password
+
+    :param username: the username that has read access on the RubyGems hosted repo
+    :param password: password for the user`
+    :param int request_id: the ID of the request this repository is for
+    :return:
+    """
+    config = get_worker_config()
+    repo_name = get_rubygems_hosted_repo_name(request_id)
+    return f"http://{username}:{password}@localhost:8082/repository/{repo_name}/"
```
tichavskym, line 495: @chmeliik @lkolacek @ejegrova  This URL is used in `.bundle/config` file to redirect Bundler to use it as a proxy repo for downloading RubyGems. I would love to get the address from some config, but I don't know what to use. There is `cachito_nexus_rubygems_proxy_url = "http://nexus:8081/repository/cachito-rubygems-proxy/"` in the worker config, but it won't work, because docker compose remaps Nexus' port 8081 to 8082 (at least in development mode). Do you have any idea what I could do about it?

cachito/workers/tasks/rubygems.py
```diff
@@ -25,6 +41,70 @@ def cleanup_rubygems_request(request_id):
...
+            f"Fetching dependencies at the {pkg_path!r} directory",
+        )
+        request = get_request(request_id)
+        pkg_and_deps_info = resolve_rubygems(
+            package_source_dir,
+            request,
+        )
+
+        # defer DB operations to use the Nexus password in the env vars
+        packages_data.append(pkg_and_deps_info)
```
tichavskym, line 79: This comment was in pip implementation and I don't really understand why. Is it pip specific or should I keep it here? Also, there was some logic about CA certificate, is that something I should implement here as well?

cachito/workers/pkg_managers/rubygems.py
chmeliik: You could do a similar thing that pip does for the `PIP_INDEX_URL`

(TL;DR - use `config.cachito_nexus_url`)

https://github.com/containerbuildsystem/cachito/blob/07f5cb1ea8307b08ac8a3b1746c28d03fda5c72c/cachito/workers/tasks/pip.py#L112-L114

https://github.com/containerbuildsystem/cachito/blob/07f5cb1ea8307b08ac8a3b1746c28d03fda5c72c/cachito/workers/pkg_managers/pip.py#L1849-L1859

tichavskym: The reason I was asking is that using `config.cachito_nexus_url` won't work with my development setup (but I don't know how Cachito is deployed internally, so it might not have the same problem), because bundler won't be able to download any dependencies from that URL (http://nexus:8081). Anyways, I implemented it this way, but in order for me to work locally, I'd probably have to change `config.cachito_nexus_url` to `http://localhost:8082`, so I'm still a bit confused.

chmeliik: Huh, yeah you're right. Now that you mention it, I think pip has the same problem, in order to install from Nexus we would have to `sed 's/nexus:8081/localhost:8082/'` in our testing.

I think the problem is that `cachito_nexus_url` has to be accessible from inside the docker-compose environment, where `localhost:8082` doesn't work. The real deployment does not have this issue because the URL is accessible both internally and externally.

To fix that perhaps we could have one internal and one external nexus url in the config, but we never bothered with it

tichavskym: Oh, in that case, I'll keep it as it is for now

cachito/workers/tasks/rubygems.py
```diff
@@ -25,8 +41,78 @@ def cleanup_rubygems_request(request_id):
...
+    log.info("Configuring Nexus for RubyGems for the request %d", request_id)
+    set_request_state(request_id, "in_progress", "Configuring Nexus for RubyGems")
+    rubygems_repo_name = get_rubygems_hosted_repo_name(request_id)
+    prepare_nexus_for_rubygems_request(rubygems_repo_name)
+
+    log.info("Fetching dependencies for request %d", request_id)
+    package_configs = package_configs or [{}]
+    packages_data = []
+    for pkg_cfg in package_configs:
+        pkg_path = pkg_cfg.get("path", ".")
```
MartinBasti, line 65: IIUC, this is submitted by user, thus shall we do some sort of validation to prevent any potential file path attack?

https://owasp.org/www-community/attacks/Path_Traversal

MartinBasti: shouldn't be `normpath` also used here?

tichavskym: I copy pasted it from pip implementation, so I didn't think much about it, but I'll look into it

chmeliik: The paths are relative
https://github.com/containerbuildsystem/cachito/blob/fcde3cbccd21a17eebb3543ec9c3a3314bd76245/cachito/web/models.py#L142
There are no symlinks that lead outside the repo
https://github.com/containerbuildsystem/cachito/blob/fcde3cbccd21a17eebb3543ec9c3a3314bd76245/cachito/workers/tasks/general.py#L80

chmeliik: +1 to use `normpath`, but IMO the extra validation is unnecessary

tichavskym: I was thinking whether there was some validation already in place. So I suppose no other validation is necessary then?

tichavskym: Ok, so as of right now, I only added normpath.

tests/test_workers/test_tasks/test_rubygems.py
```diff
@@ -70,3 +79,126 @@ def test_get_config_file(tmp_path, exists):
...
+    calls = call(request["id"], cfg)
+    assert calls == mock_update_cfg.call_args
+
+    expected = pkg_data["package"].copy()
+    del expected["path"]
+    expected_dependencies = pkg_data["dependencies"]
+    del expected_dependencies[0]["path"]
+    expected["dependencies"] = expected_dependencies
+    if package_subpath and package_subpath != os.curdir:
+        expected["path"] = package_subpath
```
chmeliik, line 182: nitpick: maybe delete `expected["path"]` only when needed? Rather than always deleting it and sometimes restoring it

cachito/workers/tasks/pip.py
```diff
@@ -73,7 +73,7 @@ def fetch_pip_source(request_id, package_configs=None):
     packages_data = []
     requirement_file_paths = []
     for pkg_cfg in package_configs:
-        pkg_path = pkg_cfg.get("path", ".")
+        pkg_path = os.path.normpath(pkg_cfg.get("path", "."))
```
chmeliik, line 76: Could the pip change be in a separate commit?

tests/test_workers/test_tasks/test_rubygems.py
```diff
@@ -57,11 +62,15 @@ def test_get_config_file(tmp_path, exists):
         )
         with pytest.raises(CachitoError, match=msg):
             rubygems._get_config_file_for_given_package(
-                pkg_and_deps_info["dependencies"], bundle_dir, package_root, rubygems_hosted_repo
+                pkg_and_deps_info["dependencies"],
+                bundle_dir,
+                package_root,
+                rubygems_hosted_repo,
+                None,
```
chmeliik, line 69: I would appreciate an extra unit test where the subpath is not None, but since one of the cases in `test_fetch_rubygems_source` already tests that it's not critical

chmeliik (commented):
LGTM, some minor details

taylormadore:
Looks good to me and pretty consistent/similar to what was done with pip :+1:

cachito/workers/pkg_managers/rubygems.py
```diff
@@ -505,3 +480,18 @@ def _get_metadata(package_root, request):
...
+
+
+def get_rubygems_hosted_url_with_credentials(username, password, request_id):
+    """
+    Get URL of a RubyGems hosted repo for the request with hardcoded username and a password.
+
+    :param username: the username that has read access on the RubyGems hosted repo
+    :param password: password for the user`
+    :param int request_id: the ID of the request this repository is for
+    :return:
```
ejegrova, line 492: can you fill in what's being returned?

cachito/workers/tasks/rubygems.py
```diff
@@ -81,3 +173,16 @@ def _get_config_file_for_given_package(
...
     final_path = config_file_path.relative_to(Path(bundle_dir))
     return make_base64_config_file(final_config, final_path)
+
+
+def cleanup_metadata(dependencies):
+    """
+    Cleanup dependencies' metadata, so it can be included in request JSON.
+
+    :param dependencies: which should be cleaned up
+    :return: list[dict]
```
ejegrova, line 183: please provide better description, you can also add :rtype:

ejegrova (commented):
look good, added few nitpicks

MartinBasti:
I don't see type annotations in new code? Is it not required anymore?

ejegrova:
> I don't see type annotations in new code? Is it not required anymore?
IIRC we never had it in Cachito. It would be nice to have though.

MartinBasti:
> > I don't see type annotations in new code? Is it not required anymore?
> 
> IIRC we never had it in Cachito. It would be nice to have though.

I see some functions with type annotations in cachito and also there is a mypy in CI

tichavskym:
This is the first time somebody brought it up here, but it makes sense to add them.

ejegrova (approved):

PR #696, opened 2022-08-22
Add Integration tests for RubyGems

tichavskym:


ejegrova:
Can you please update test data with packages from [cachito-testing repo ](https://github.com/cachito-testing?q=rubygems&type=all&language=&sort=) ? We want to keep all repos for testing under that organization. 
Please add an integration test for [cached dependencies ](https://github.com/containerbuildsystem/cachito/blob/master/tests/integration/test_data/cached_dependencies.yaml).

tests/integration/test_data/cached_dependencies.yaml
```diff
@@ -887,3 +887,84 @@ yarn_cached_deps:
     - "pkg:npm/is-positive@3.1.0"
     - "pkg:npm/pathval@1.1.1"
     - "pkg:npm/type-detect@4.0.8"
+rubygems_cached_deps:
+  # Use local version of repos.
+  # <use_local: True> is not supported in local environment and will be skipped.
+  use_local: True
```
tichavskym, line 893: @ejegrova Thanks for moving those testing repos. I started working on integration test for cached dependencies, but I'm a little confused. Those tests aren't executed neither by the pipeline nor the local dev environment, because of these lines of code (I think), so I don't know how to make sure those tests work and what's the point of having tests that aren't executed?

tichavskym:
Cached dependencies test: If I understand it correctly, [these two commits](https://github.com/containerbuildsystem/cachito/blob/f9f3db4aa9d9cb6bf51fbd99f7d93deca4a7e82f/tests/integration/test_using_cached_dependencies.py#L136) should be used as dummy dependencies for which we're testing the caching functionality. That's how I implemented this for GIT dependencies, but I'm wondering whether testing GEM dependencies is also in scope of this test (these are always downloaded from proxy repo). For example for pip, only `https + VCS` deps are updated in [update_main_repo](https://github.com/containerbuildsystem/cachito/blob/f9f3db4aa9d9cb6bf51fbd99f7d93deca4a7e82f/tests/integration/test_using_cached_dependencies.py#L408), so I suppose dependencies from PyPI are not tested here?

tests/integration/test_data/cached_dependencies.yaml
taylormadore: I believe that these tests make commits to repos under the cachito-testing organization and require a credential in order to do so. Therefore they only get executed in our internal test environment and not upstream/locally.

tests/integration/test_data/rubygems_packages.yaml
```diff
@@ -0,0 +1,137 @@
...
+# response_expectations: Parts of the Cachito response to check
+# content_manifest: PURLs for image contents part
+with_deps:
+  repo: https://github.com/cachito-testing/cachito-rubygems-with-dependencies.git
+  ref: ef6b2d7f9e4719be204531639a3ab90cfd731f26
+  pkg_managers: ["rubygems"]
+  expected_files:
+    app: https://github.com/cachito-testing/cachito-rubygems-with-dependencies/tarball/ef6b2d7f9e4719be204531639a3ab90cfd731f26
+    deps/rubygems/zeitwerk/zeitwerk-2.6.0.gem: https://rubygems.org/downloads/zeitwerk-2.6.0.gem
+    deps/rubygems/github.com/RadPad/swagger-ui_rails/swagger-ui_rails-external-gitcommit-7234e21e621b628d6b43194f9ba5cce5ca587f16/app: https://github.com/RadPad/swagger-ui_rails/tarball/7234e21e621b628d6b43194f9ba5cce5ca587f16
```
taylormadore, line 39: This might just be lack of cachito understanding on my part, but aren't deps usually archive files?

chmeliik: Usually yes, but for ruby the GIT dependencies need to be directories, bundler does not support archives.

Technically we never said what the format of `deps/` would be so we're not really breaking any promises :smile: 

We will need to check what it does with source containers though, I'll file a story for that

chmeliik: Actually nevermind, it won't hurt source containers. OSBS just dumps the entire `remote-source.tar.gz` in there, `app/` `deps/` and everything.

Denyilst exclusion should also works, OSBS makes no distinction between files and directories (as long as the pattern matches)

tests/integration/utils.py
```diff
@@ -353,15 +353,24 @@ def assert_expected_files(source_path, expected_files, tmpdir):
...
+                if archive_path.endswith(".gem"):
+                    shutil.unpack_archive(archive_path, deps_data_path, "gztar")
+                else:
+                    shutil.unpack_archive(archive_path, deps_data_path)
                 # deps_data_path is unique and contains only one expected package
-                package_root_dir = os.path.join(deps_data_path, os.listdir(deps_data_path)[0])
+                if deps_data_path.endswith(".gem"):
+                    package_root_dir = deps_data_path
+                else:
+                    package_root_dir = os.path.join(deps_data_path, os.listdir(deps_data_path)[0])
```
chmeliik, line 364: The comment about containing only one package probably belongs to the code that does `listdir()[0]`

tests/integration/test_using_cached_dependencies.py
```diff
@@ -489,3 +497,32 @@ def update_main_repo(env_data, repo_dir, tmpdir, new_dep_commits, dep_repo):
             "FIRST_DEP_COMMIT": new_dep_commits[0],
             "SECOND_DEP_COMMIT": new_dep_commits[1],
         }
+    elif env_data["pkg_managers"] == ["rubygems"]:
+        with open(os.path.join(repo_dir, "Gemfile"), "a") as f:
+            f.write(f"gem 'tilt'\ngem 'my-package', git: '{env_data['dep_archive_baseurl']}'\n")
+        with open(os.path.join(repo_dir, "Gemfile.lock"), "r") as f:
+            gemlock = f.read()
+            gemlock.replace("zeitwerk (2.6.0)", "tilt (2.0.11)\n    zeitwerk (2.6.0)")
+            gemlock.replace("swagger-ui_rails!", "swagger-ui_rails!\n  tilt")
```
chmeliik, line 506: `str.replace()` does not modify the string in place, you will have to do something like this
```suggestion
            gemlock = gemlock.replace("zeitwerk (2.6.0)", "tilt (2.0.11)\n    zeitwerk (2.6.0)")
```

tests/integration/test_using_cached_dependencies.py
```diff
@@ -489,3 +497,32 @@ def update_main_repo(env_data, repo_dir, tmpdir, new_dep_commits, dep_repo):
             "FIRST_DEP_COMMIT": new_dep_commits[0],
             "SECOND_DEP_COMMIT": new_dep_commits[1],
         }
+    elif env_data["pkg_managers"] == ["rubygems"]:
+        with open(os.path.join(repo_dir, "Gemfile"), "a") as f:
+            f.write(f"gem 'tilt'\ngem 'my-package', git: '{env_data['dep_archive_baseurl']}'\n")
+        with open(os.path.join(repo_dir, "Gemfile.lock"), "r") as f:
+            gemlock = f.read()
+            gemlock.replace("zeitwerk (2.6.0)", "tilt (2.0.11)\n    zeitwerk (2.6.0)")
```
chmeliik, line 505: I don't fully understand what this replacement does. Is it adding a new dependency?

chmeliik (commented):
@tichavskym have you been able to run the cached dependencies test? The rest LGTM

tichavskym:
> @tichavskym have you been able to run the cached dependencies test? The rest LGTM

No, I'm still not able to connect to the testing server. I removed testing GEM dependency from cached dependencies test, because I believe that's not in scope of this test.

tests/integration/test_using_cached_dependencies.py
tichavskym: This section should add a new dependency to the Gemfile and Gemfile.lock.

chmeliik (approved):
Looks like it should work :+1:

tichavskym:
Well, cached dependencies still don't work, but I have it somehow functioning locally, so hopefully I'll be able to push it soon.

tests/integration/utils.py
```diff
@@ -347,21 +347,22 @@ def assert_expected_files(source_path, expected_files, tmpdir):
...
-                archive_path = test_path
-                shutil.unpack_archive(archive_path, deps_data_path)
-                # deps_data_path is unique and contains only one expected package
-                package_root_dir = os.path.join(deps_data_path, os.listdir(deps_data_path)[0])
+
+            package_root_dir = unpack_test_archives(deps_data_path, test_path)
+
             download_archive(archive_url, expected_archive)
+            if os.path.exists(expected_data_path):
+                shutil.rmtree(expected_data_path)
```
tichavskym, line 355: Weirdly enough, without these two lines RubyGems cached dependency test would fail. But with these lines, I'm confident cached dependencies tests pass (tested with Pip and RubyGems), so we can either merge it as it is or I can look into it more.

tichavskym: The weird part is that these files should be deleted later in this function and they maybe are, but then when this function is called for the second request, it raises permission error while trying to unpack the archive, as if they still existed. And deleting them this way solves the issue.

chmeliik: Oh, I think I found the bug /facepalm
```python
            for temp_data in [deps_data_path, expected_data_path, expected_archive]:
                if os.path.exists(deps_data_path):
                                  --------------
                    shutil.rmtree(temp_data)
```
So the tests never delete `expected_data_path` or `expected_archive` :sweat_smile:

tichavskym: Well, I have no idea what I was doing when I was debugging it :sweat_smile:

tichavskym:
Ready for review/merge

chmeliik (approved):
LGTM :tada:

ejegrova (approved):
LGTM, great work :+1:

PR #697, opened 2022-08-23
Fix UploadError/NetworkError exception handling

lkolacek:
Pip package manager requests should not fail on already uploaded package

CLOUDBLD-10907

Signed-off-by: Ladislav Kolacek <lkolacek@redhat.com>

### DIFF
--- a/cachito/workers/pkg_managers/pip.py
+++ b/cachito/workers/pkg_managers/pip.py
@@ -30,7 +30,7 @@ from cachito.errors import (
 )
 from cachito.workers import nexus
 from cachito.workers.config import get_worker_config
-from cachito.workers.errors import NexusScriptError
+from cachito.workers.errors import NexusScriptError, UploadError
 from cachito.workers.paths import RequestBundleDir
 from cachito.workers.pkg_managers import general
 from cachito.workers.pkg_managers.general import (
@@ -1965,7 +1965,7 @@ def _push_downloaded_requirement(requirement, pip_repo_name, raw_repo_name):
     if requirement["kind"] == "pypi":
         try:
             upload_pypi_package(pip_repo_name, requirement["path"])
-        except (NetworkError, ValueError):
+        except UploadError:
             if nexus.get_component_info_from_nexus(
                 pip_repo_name,
                 "pypi",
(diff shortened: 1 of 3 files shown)


cachito/workers/pkg_managers/pip.py
```diff
@@ -1965,7 +1966,7 @@ def _push_downloaded_requirement(requirement, pip_repo_name, raw_repo_name):
     if requirement["kind"] == "pypi":
         try:
             upload_pypi_package(pip_repo_name, requirement["path"])
-        except (NetworkError, ValueError):
+        except (UploadError, NetworkError, ValueError):
```
lkolacek, line 1969: Checking if we can remove NetworkError from these places

cachito/workers/pkg_managers/pip.py
```diff
@@ -27,6 +27,7 @@
     NetworkError,
     NexusError,
     ValidationError,
+    UploadError,
```
chmeliik, line 30: Checked the CI failure, turns out this is in `cachito.workers.errors`, not `cachito.errors`

cachito/workers/pkg_managers/pip.py
chmeliik: I think so, +1 to remove it. The ruby code catches only `UploadError`

cachito/workers/pkg_managers/pip.py
```diff
@@ -1987,7 +1987,7 @@ def _push_downloaded_requirement(requirement, pip_repo_name, raw_repo_name):
         dest_dir, filename = requirement["raw_component_name"].rsplit("/", 1)
         try:
             upload_raw_package(raw_repo_name, requirement["path"], dest_dir, filename, True)
-        except (NetworkError, ValueError):
+        except (UploadError, NetworkError, ValueError):
```
lkolacek, line 1990: Here it looks like we can remove both `ValueError` and `NetworkError`

chmeliik: Only here, not for the other one? As far as I can tell, `ValueError` is only raised for invalid usage of the method

chmeliik:
LGTM once unit test is fixed (and optionally `ValueError` catching removed)

lkolacek:
Ready for another review pls. :)

chmeliik (approved):

taylormadore (approved):

ejegrova (approved):
LGTM

PR #700, opened 2022-08-26
Find build deps enhancements

bruno-fs:
Add a new option to `pip_find_builddeps.py` (`--only-write-on-update`) and an pre-commit hook for this script.

### DIFF
--- a/bin/pip_find_builddeps.py
+++ b/bin/pip_find_builddeps.py
@@ -117,6 +117,24 @@ def generate_file_content(builddeps, is_partial):
     return file_content
 
 
+def _parse_requirements_file(builddeps_file):
+    """Find deps requirements-build.in file."""
+    try:
+        with open(builddeps_file) as f:
+            # ignore line comments or comments added after dependency is declared
+            requirement_re = re.compile(r"^([^\s#;]+)")
+            matches = (requirement_re.match(line) for line in f)
+            return set(match.group(1) for match in matches if match)
+    except FileNotFoundError:
+        # it's ok if the file doens't exist.
+        return set()
+
+
+def _sanity_check_args(ap, args):
+    if args.only_write_on_update and not args.output_file:
+        ap.error("--only-write-on-update requires an output-file (-o/--output-file).")
+
+
 def main():
     """Run script."""
     ap = argparse.ArgumentParser(description=DESCRIPTION)
@@ -140,8 +158,17 @@ def main():
         action="store_true",
         help="generate partial output even if pip download fails",
     )
+    ap.add_argument(
+        "--only-write-on-update",
+        action="store_true",
+        help=(
+            "only write output file if dependencies will be modified - or new "
+            "dependencies will be added if used in conjunction with -a/--append."
+        ),
+    )
 
     args = ap.parse_args()
+    _sanity_check_args(ap, args)
 
     log.info(
         "Please make sure the input files meet the requirements of this script "
@@ -153,6 +180,16 @@ def main():
         no_cache=args.no_cache,
         ignore_errors=args.ignore_errors,
     )
+
+    if args.only_write_on_update:
+        original_builddeps = _parse_requirements_file(args.output_file)
+        if args.append:
+            # append only new dependencies
+            builddeps = sorted(set(builddeps) - original_builddeps)
+        if not builddeps or set(builddeps) == original_builddeps:
+            log.info("No new build dependencies found.")
+            return
+
     file_content = generate_file_content(builddeps, is_partial)
 
     log.info("Make sure to pip-compile the output before submitting a Cachito request")
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,9 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 from setuptools import find_packages, setup
 
+GEMLOCK_PARSER_REPO_URL = "https://github.com/containerbuildsystem/gemlock-parser.git"
+GEMLOCK_PARSER_PIP_REF = f"git+{GEMLOCK_PARSER_REPO_URL}@master#egg=gemlock_parser"
+
 setup(
     name="cachito",
     long_description=__doc__,
@@ -10,7 +13,7 @@ setup(
     install_requires=[
         "backoff",
         "celery>=5",
-        "gemlock_parser",
+        f"gemlock_parser @ {GEMLOCK_PARSER_PIP_REF}",
         "gitpython",
         "kombu>=5",  # A celery dependency but it's directly imported
         "packaging",
@@ -49,4 +52,5 @@ setup(
         "version_scheme": "post-release",
     },
     setup_requires=['setuptools_scm'],
+    scripts=["bin/pip_find_builddeps.py"],
 )
--- a/docs/pip.md
+++ b/docs/pip.md
@@ -168,6 +168,36 @@ dependencies you will need. Here is how you would use it:
 3. run `pip_find_builddeps.py requirements.txt -o requirements-build.in --append`
 4. run `pip-compile requirements-build.in -o requirements-build.txt --allow-unsafe`
 
+You could also use this script as [pre-commit](https://pre-commit.com/) hooks.
+To do so, copy `pip_find_builddeps.py` create a `.pre-commit-hooks.yaml` with the follwing:
+
+```
+id: update-build-requirements
+  name: update-build-requirements
+  description: find build dependencies with cachito's pip_find_builddeps.py script
+  entry: path/to/pip_find_builddeps.py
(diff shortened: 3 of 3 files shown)


bin/pip_find_builddeps.py
```diff
@@ -117,6 +117,28 @@ def generate_file_content(builddeps, is_partial):
     return file_content
 
 
+def _parse_requirements_file(builddeps_file):
+    """Find deps requirements-build.in file."""
+    try:
+        with open(builddeps_file) as f:
+            # ignore line comments or comments added after dependency is declared
+            requirement_re = re.compile(r"\s*([^\s#]+)")
```
MartinBasti, line 125: To make sure it really matches beginning not just something in comments
```diff
- requirement_re = re.compile(r"\s*([^\s#]+)")
+ requirement_re = re.compile(r"^\s*([^\s#]+)")
```

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,9 @@
+- id: update-build-requirements
```
MartinBasti, line 1: Can we mention pre-commit hook in readme instead?

MartinBasti (changes requested):
Thank you for your contribution!

.pre-commit-hooks.yaml
bruno-fs: You mean instead of creating this file?

bin/pip_find_builddeps.py
bruno-fs: [re.match](https://docs.python.org/3/library/re.html#re.match) (line 126) already takes care of matching only the beginning. With that said, if you prefer to have this regex starting with "^" anyway I can change

bin/pip_find_builddeps.py
```diff
@@ -153,6 +181,13 @@ def main():
         no_cache=args.no_cache,
         ignore_errors=args.ignore_errors,
     )
+
+    if args.only_overwrite_on_update:
+        original_builddeps = _parse_requirements_file(args.output_file)
+        if set(builddeps) == original_builddeps:
+            log.info("No new build dependencies found.")
+            exit()
```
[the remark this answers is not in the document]
bruno-fs: yes! I'll change that

bin/pip_find_builddeps.py
MartinBasti: you are right, but I'd feel safer with `^` if somebody will refactor it to `re.search` or smth.

.pre-commit-hooks.yaml
MartinBasti: yes, or maybe I need explanation how this file will help to integrate hook to other repos

bruno-fs:
@MartinBasti do you prefer I rebase before or after your next review?

setup.py
```diff
@@ -10,7 +10,7 @@
     install_requires=[
         "backoff",
         "celery>=5",
-        "gemlock_parser",
+        "gemlock_parser @ git+https://github.com/containerbuildsystem/gemlock-parser.git@master#egg=gemlock_parser",
```
MartinBasti, line 13: @tichavskym @chmeliik Shall we have this on PyPI?

tichavskym: I could definitely upload it there, I just don't know whether we have some login as an organization or how shall I proceed with this?

chmeliik: I'd vote against it, the gemlock-parser repo is supposed to be a temporary workaround until the relevant part of the scancode library (where we vendored the code from) is made standalone

https://github.com/containerbuildsystem/cachito/pull/666#issuecomment-1145756553

MartinBasti: okay, then keep it as is in this PR

MartinBasti:
@bruno-fs thank you, please fix linter issues and squash fixup commits

MartinBasti (approved):

bin/pip_find_builddeps.py
```diff
@@ -117,6 +117,26 @@ def generate_file_content(builddeps, is_partial):
     return file_content
 
 
+def _parse_requirements_file(builddeps_file):
+    """Find deps requirements-build.in file."""
+    try:
+        with open(builddeps_file) as f:
+            # ignore line comments or comments added after dependency is declared
+            requirement_re = re.compile(r"^\s*([^\s#]+)")
```
chmeliik, line 125: Why does the regex allow leading whitespace? And should the character class also exclude `;` like the one in `_filter_builddeps()`?

AFAIK semicolon would be used for environment markers:
```
poetry<1.0; python_version<"3.6"
```

bin/pip_find_builddeps.py
```diff
@@ -117,6 +117,26 @@ def generate_file_content(builddeps, is_partial):
...
+    except FileNotFoundError:
+        # it's ok if the file doens't exist.
+        return set()
+
+
+def _sanity_check_args(args):
+    if args.only_write_on_update and not args.output_file:
+        argparse.ArgumentParser.error(
+            "--only-write-on-update requires an output-file (-o/--output-file)."
+        )
```
chmeliik, line 137: I think this needs to be called on an instance
```shell
$ bin/pip_find_builddeps.py requirements.txt --only-write-on-update   
  File "/home/acmiel/RedHat/Cachito/cachito/bin/pip_find_builddeps.py", line 211, in <module>
    main()
  File "/home/acmiel/RedHat/Cachito/cachito/bin/pip_find_builddeps.py", line 173, in main
    _sanity_check_args(args)
  File "/home/acmiel/RedHat/Cachito/cachito/bin/pip_find_builddeps.py", line 135, in _sanity_check_args
    argparse.ArgumentParser.error(
TypeError: ArgumentParser.error() missing 1 required positional argument: 'message'
```

.pre-commit-hooks.yaml
chmeliik: +1 for the docs update, do we want to keep the actual pre-commit hook here? Cachito doesn't use those, nor does it declare build dependencies

bruno-fs: @chmeliik Do you know if its possible to add a hook from a third party repo?
I tried that without success...

(but maybe we could document to also copy the script - its overkill to install the whole cachito just to use a script that only relies on python standard library :thinking:)

bruno-fs: As for pre-commit dependency: one repo can contain pre-commit hooks w/o using them (and hence not relying on pre-commit).

bin/pip_find_builddeps.py
bruno-fs: whoops, sorry. I used argparser api incorrectly here

bin/pip_find_builddeps.py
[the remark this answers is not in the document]
bruno-fs: done https://github.com/bruno-fs/cachito/blob/5bb42d71577a387031c04c1da11c5455da6ad89c/bin/pip_find_builddeps.py#L190-L192

bin/pip_find_builddeps.py
bruno-fs: done https://github.com/bruno-fs/cachito/blob/5bb42d71577a387031c04c1da11c5455da6ad89c/bin/pip_find_builddeps.py#L125

bin/pip_find_builddeps.py
bruno-fs: Mmmm I didn't think thoroughly about leading whitespaces. Generated build requirements file generated by this script should not have it, so I guess skipping it is ok.

I also didn't consider environment markers. I think this match should reproduce the behavior of  `_filter_builddeps()` .

Updating this accordingly

bruno-fs: I forgot to push :facepalm: 
sorry for that :sweat_smile: 
```
% git status                                                                  ✭
On branch find-build-deps-enhancements
Your branch is ahead of 'origin/find-build-deps-enhancements' by 1 commit.
  (use "git push" to publish your local commits)
```

bruno-fs: https://github.com/bruno-fs/cachito/commit/e4c83385a79ec53d81cd06085d7b0ffce3ee289f

bruno-fs: ugh, it wasn't signed off.

Do you prefer I rebase/squash now?

chmeliik: Squashing the fixup into the first commit would be nice, otherwise the 3 commits make sense to me :+1:

chmeliik:
LGTM, there's just one LGTM alert which could be easily addressed

And the CI is failing for unrelated reasons, being fixed in #780

chmeliik (approved):
LGTM :+1:

PR #702, opened 2022-09-01
Fix RubyGems Request JSON metadata

tichavskym:
Commit 281b5217111afee923e14f901d1b15b49cd47086 moved cleanup_metatada()
to a different file but it added a `kind` key, which doesn't match
the documentation and should be deleted.

### DIFF
--- a/cachito/workers/tasks/rubygems.py
+++ b/cachito/workers/tasks/rubygems.py
@@ -181,10 +181,10 @@ def cleanup_metadata(dependencies: list[dict]):
 
     :param list dependencies: which should be processed
     :return: list of dependencies where each dependency is represented by a dictionary
-        with the following keys: name, version, type, kind
+        with the following keys: name, version, type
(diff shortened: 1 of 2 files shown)


chmeliik (approved):
LGTM, nice catch :+1:

cachito/workers/tasks/rubygems.py
```diff
@@ -185,6 +185,6 @@ def cleanup_metadata(dependencies: list[dict]):
     :rtype list[dict]:
     """
     return [
-        {"name": dep["name"], "version": dep["version"], "type": dep["type"], "kind": dep["kind"]}
+        {"name": dep["name"], "version": dep["version"], "type": dep["type"]}
```
taylormadore, line 188: Can you also drop kind from the docstring above? Other than that, LGTM :+1:

taylormadore (approved):

PR #703, opened 2022-09-06
Optimization of JavaScript Tarball Downloading - asyncio

fepas:


taylormadore:
Have you been able to successfully generate a source bundle? I tried locally (with a rather large request :sweat_smile: from rhacs) and was getting network/nexus errors while fetching the dependencies from the proxy repo. I tried dropping the number of simultaneous connections first to 10 and then to 1 but was hitting the same issues :thinking:

fepas:
@taylormadore Just posted the update using Semaphores to not kill Nexus with a lot of requests :D

PR #705, opened 2022-09-12
Fix RubyGems GIT dependencies

tichavskym:
When Cachito worker downloads GIT dependency, the downloaded Git repository is checked out at a proper revision (according to the `Gemfile.lock`), but it's not checked out at a proper branch. Until now, Cachito was using `git checkout -b <branch name>`, however, this didn't work for cases when the branch already existed (e.g. downloaded repositories usually contained `master` branch and if the `Gemfile.lock` specified `master` branch too, this command tried to create a branch that already existed).

To fix this, `-B` option should be used in `git checkout`, because in case of an existing branch, it checks out the branch and resets it to the current commit and no error is raised.

### DIFF
--- a/cachito/workers/pkg_managers/rubygems.py
+++ b/cachito/workers/pkg_managers/rubygems.py
@@ -424,7 +424,7 @@ def checkout_branch(dep: dict):
     try:
         repo = Repo(dep["path"] / "app")
         git = repo.git
-        git.checkout("HEAD", b=dep["branch"])
+        git.checkout("HEAD", B=dep["branch"])
     except CheckoutError:
         raise GitError(f"Couldn't checkout branch {dep['branch']} at {dep['path'] / 'app'}")
     except Exception:
--- a/tests/test_workers/test_pkg_managers/test_rubygems.py
+++ b/tests/test_workers/test_pkg_managers/test_rubygems.py
@@ -921,7 +921,7 @@ def test_checkout_branch(mock_repo):
     rubygems.checkout_branch({"path": Path("/yo"), "branch": "b"})
 
     mock_repo.assert_called_with(Path("/yo/app"))
-    mock_repo.return_value.git.checkout.assert_called_once_with("HEAD", b="b")
+    mock_repo.return_value.git.checkout.assert_called_once_with("HEAD", B="b")
 
 
 @mock.patch("cachito.workers.pkg_managers.rubygems.Repo")


chmeliik:
Would resetting the branch be what the user expects? Maybe Cachito should instead check that the branch already points to the proper revision?

chmeliik:
> Would resetting the branch be what the user expects? Maybe Cachito should instead check that the branch already points to the proper revision?

@tichavskym what does bundler do when the GIT dep has both a revision and a branch specified? Maybe that will help us decide what the behavior should be

tichavskym:
> > Would resetting the branch be what the user expects? Maybe Cachito should instead check that the branch already points to the proper revision?
> 
> @tichavskym what does bundler do when the GIT dep has both a revision and a branch specified? Maybe that will help us decide what the behavior should be

Good point! When user specifies only `branch:` in `Gemfile`, Bundler downloads the Git repository and finds the latest commit hash for the given branch, which it writes into Gemfile.lock. To set the working tree of the repository to this revision, Bundler executes `git reset --hard <hash>`. 

When Bundler specifies both `ref:` and `branch:`, the reference takes precedence, and in my experiments, I was able to specify a commit that wasn't on a given branch and Bundler didn't raise any error.

What you're suggesting means Cachito would have to do `git switch <branch>` - it would create a local branch tracking the remote branch (if that branch doesn't exist already, e.g. master or main) and then checkout - and then check whether it is at a proper revision, such as verifying output of `git rev-parse --verify <branch_name>` matches the revision, or something similar. There is one edge case for which this wouldn't work though.

Let's say I want to build an app. I have a dependency `my-gem` in my `Gemfile` specified using `branch: master`. I execute `bundle install` and `hash-1` is specified in my app's `Gemfile.lock`. Then I add a new commit (`hash-2`) to `my-gem` repo, but I don't update Gemfile.lock in my app repo. Now I want to build my app using OSBS + Cachito. I create a request and Cachito correctly downloads `my-gem` repository at a given commit, but `master` branch at this downloaded repo will point to `hash-2` as opposed to `hash-1` specified in my Gemfile (at least I think that's what Cachito does, I'm not sure which exact command it executes).

Problem with implementation of `git checkout -B <branch name>` is the following: if the branch don't exists anymore (it was deleted) or never existed, user will not know (though it has to exist at some point, because we don't expect anyone to create Gemfile.lock manually). 

So it's more or less about what stance should Cachito make towards changes in Git repos of the dependencies. Also note that this discusses mostly the time period between creating Gemfile.lock and caching the dependency by Cachito, because once it's cached, we don't care about what's going on in Git dependency's repository.

chmeliik:
> When Bundler specifies both ref: and branch:, the reference takes precedence, and in my experiments, I was able to specify a commit that wasn't on a given branch and Bundler didn't raise any error.

Ack, this convinced me :+1: 

> Problem with implementation of git checkout -B <branch name> is the following: if the branch don't exists anymore (it was deleted) or never existed, user will not know (though it has to exist at some point, because we don't expect anyone to create Gemfile.lock manually).

That sounds like an acceptable trade-off

chmeliik (approved):
LGTM :+1:

lkolacek (approved):
Looks good :+1:

taylormadore (approved):
Sounds reasonable to me :+1:

PR #706, opened 2022-09-12
Optimization of JavaScript Tarball Downloading - threads

fepas:
CLOUDBLD-9007

Signed-off-by: Felipe Campos <fepas.unb@gmail.com>


taylormadore:
Some additional work might need to be done here around the requests Session object, since I don't think it is guaranteed thread-safe

PR #707, opened 2022-09-14
add hostname to the labels of the metrics; pull out calls to metrics …

mike-kingsbury:
Adds hostnames to metrics; pull out metrics functionality to metrics.py file, suggested by Bruno.


chmeliik (commented):
Seems reasonable :+1: 

Are there any tests for these metrics that could be updated? If we have none, that's probably OK

cachito/web/metrics.py
```diff
@@ -27,10 +27,23 @@ def init_metrics(app):
...
     cachito_metrics["gauge_state"] = gauge_state
     cachito_metrics["request_duration"] = request_duration
+
+
+def requests_inc(state):
+    """Increase the number of requests in given state."""
+    cachito_metrics["gauge_state"].labels(state=state, host=hostname).inc()
+
+
+def requests_dec(state):
```
chmeliik, line 47: Since we have type annotations in the checklist
```suggestion
def requests_inc(state: str) -> None:
    """Increase the number of requests in given state."""
    cachito_metrics["gauge_state"].labels(state=state, host=hostname).inc()


def requests_dec(state: str) -> None:
```

brunoapimentel:
> Seems reasonable +1
> 
> Are there any tests for these metrics that could be updated? If we have none, that's probably OK

I don't think there are any :(

WIth this change, it should be easy enough to add unit tests for metrics.py. I also agree it's probably OK to have this merged without the tests, they can be added as an improvement later.

brunoapimentel:
Overall, LGTM. Only fix needed is adding type annotations, as mentioned.

MartinBasti:
What's the status of this PR? Is it pending review or pending changes?

mike-kingsbury:
Should be now pending review, changes were made to the annotations

requirements-web.txt
```diff
@@ -104,9 +106,9 @@ jinja2==3.1.2 \
     --hash=sha256:31351a702a408a9e7595a8fc6150fc3f43bb6bf7e319770cbc0db9df9437e852 \
     --hash=sha256:6088930bfe239f0e6710546ab9c19c9ef35e29792895fed6e6e31a023a182a61
     # via flask
-mako==1.2.2 \
-    --hash=sha256:3724869b363ba630a272a5f89f68c070352137b8fd1757650017b7e06fda163f \
-    --hash=sha256:8efcb8004681b5f71d09c983ad5a9e6f5c40601a6ec469148753292abc0da534
+mako==1.2.1 \
```
chmeliik, line 109: Are the `requirements*.txt` changes intentional? A lot of dependencies got downgraded, as well as the python version used to generate the files

requirements-test.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with python 3.10
```
fepas, line 2: Shouldn't we keep running pip-compile using python 3.10 since it's the newest version? (if there is any difference between using 3.9 and 3.10). Apart from that, LGTM

mike-kingsbury:
Let me revisit this, I may have done the wrong thing during a rebase...

PR #709, opened 2022-09-15
patch request should use EnvironmentVariableInfo

taylormadore:
In a PATCH request, environment_variables should
be a dict of dicts in the openapi schema


cachito/web/static/api_v1.yaml
```diff
@@ -1049,7 +1049,10 @@ components:
       type: object
       properties:
         environment_variables:
-          $ref: "#/components/schemas/EnvironmentVariable"
+          type: object
+          properties:
+            NAME:
```
chmeliik, line 1054: Out of curiosity, do you know what NAME being uppercase means?

But since the same thing is used as the /requests/id/environment-variables response, I'm going to assume this is correct :smile:

chmeliik (approved):
LGTM :+1:

cachito/web/static/api_v1.yaml
taylormadore: Good question. I had initially just replicated what was in the /requests/id/environment-variables response and functionally It appeared to work well and the redoc/swagger output looked fine.

After digging a little deeper though, I think the right approach would be to switch both to use the [additionalProperties](https://swagger.io/docs/specification/data-models/dictionaries/) keyword and drop that NAME key. I'm testing that now.

chmeliik (approved):
+1, additionalProperties makes more sense to me

cachito/web/static/api_v1.yaml
```diff
@@ -187,14 +187,23 @@ paths:
...
-                      $ref: "#/components/schemas/Request"
-                  meta:
-                    $ref: "#/components/schemas/Pagination"
+                anyOf:
+                  - type: object
+                    properties:
+                      items:
+                        type: array
+                        items:
+                          $ref: "#/components/schemas/Request"
```
chmeliik, line 196: Would it work if we `anyOf` just the items? (might make it clearer that the rest of the response is the same)
```suggestion
                      items:
                        anyOf:
                          - type: array
                            items:
                              $ref: "#/components/schemas/Request"
                          - ...
```

chmeliik (commented):
> The verbose parameter to the get requests endpoint
determines whether an array of Request or
RequestVerbose objects is returned

Wow, it really does. Today I learned :smile:

cachito/web/static/api_v1.yaml
taylormadore: I'll give it a try :+1:

brunoapimentel (approved):

cachito/web/static/api_v1.yaml
taylormadore: Thanks for mentioning this. It appears to work just fine and is more succinct :smile: 
I made one additional change: anyOf --> oneOf. I think that makes sense after this change, since the array will be of all either Request or VerboseRequest

cachito/web/static/api_v1.yaml
```diff
@@ -192,7 +192,9 @@ paths:
                   items:
                     type: array
                     items:
-                      $ref: "#/components/schemas/Request"
+                      oneOf:
+                        - $ref: "#/components/schemas/Request"
+                        - $ref: "#/components/schemas/RequestVerbose"
```
chmeliik, line 197: > the array will be of all either Request or VerboseRequest

Not sure if my intuition is correct, but doesn't the current schema essentially mean
```python
list[Request | RequestVerbose]
```
When what we want is
```python
list[Request] | list[RequestVerbose]
```
Maybe the oneOf should go around the whole array, rather than the items of this array?

taylormadore: Good point! The validation fails using oneOf when returning an empty array (since it matches both schemas), but using anyOf appears to work. I'll update

chmeliik (approved):
Re-LGTM :+1:

PR #712, opened 2022-09-19
Enhance openapi spec for connexion

taylormadore:
Enhances the cachito OpenAPI spec with some additional constraints on input parameters. Response parameters were not validated/enhanced outside of the few fixes noted in https://github.com/containerbuildsystem/cachito/pull/709

### DIFF
--- a/cachito/web/static/api_v1.yaml
+++ b/cachito/web/static/api_v1.yaml
@@ -100,6 +100,7 @@ paths:
           description: The specific page to view
           schema:
             type: integer
+            minimum: 1
             example: 1
             default: 1
         - name: per_page
@@ -107,15 +108,14 @@ paths:
           description: The number of requests to show per page
           schema:
             type: integer
+            minimum: 1
             example: 10
             default: 10
         - name: state
           in: query
           description: The state to filter requests by
           schema:
-            type: string
-            example: in_progress
-            default: null
+            $ref: "#/components/schemas/RequestState"
         - name: verbose
           in: query
           description: Shows the same view as /requests/{id}
@@ -128,15 +128,21 @@ paths:
           description: A full repository URL to filter request by
           schema:
             type: string
+            maxLength: 200
             example: https://github.com/release-engineering/retrodep.git
             default: null
+            nullable: true
         - name: ref
           in: query
           description: A git ref to filter request by
           schema:
             type: string
+            minLength: 40
+            maxLength: 40
+            pattern: '^[a-f0-9]{40}$'
             example: bc9767a71ede6e0084ae4a9e01dcd8b81c30b741
             default: null
+            nullable: true
         - name: pkg_manager
           in: query
           description: >
@@ -144,9 +150,7 @@ paths:
             can be specified multiple times, that will filter all
             requests that have the specified pkg_managers.
           schema:
-            type: string
-            example: gomod
-            default: null
+            $ref: "#/components/schemas/PackageManager"
         - name: created_from
           in: query
           description: >
@@ -157,6 +161,7 @@ paths:
             type: string
             example: "2019-09-17T18:52:57.577851"
             default: null
+            nullable: true
         - name: created_to
           in: query
           description: >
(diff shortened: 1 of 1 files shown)


cachito/web/static/api_v1.yaml
```diff
@@ -1010,19 +1073,32 @@ components:
...
+              type: array
+              items:
+                $ref: "#/components/schemas/PackageConfig"
+          additionalProperties: false
         pkg_managers:
           type: array
           items:
-            type: string
-            example: gomod
+            $ref: "#/components/schemas/PackageManager"
```
chmeliik, line 1101: PackageManager allows `null` right? I think the pkg_managers array shouldn't allow that

https://swagger.io/docs/specification/using-ref/ `$ref and Sibling Elements` makes me sad, otherwise we could make PackageManager not nullable and allow null only when needed :disappointed: 

Maybe something like `allOf` with `- nullable: false` would still work?

cachito/web/static/api_v1.yaml
```diff
@@ -1162,6 +1249,18 @@ components:
                   properties:
                     purl:
                       $ref: "#/components/schemas/Purl"
+    RequestState:
```
chmeliik, line 1252: Should `RequestUpdate.state` also be a RequestState rather than a string? (Would have the same problem with being nullable though)

cachito/web/static/api_v1.yaml
```diff
@@ -1058,14 +1139,20 @@ components:
...
+          minimum: 0
           example: 100
         error_origin:
           type: string
+          enum:
+            - client
+            - server
         error_type:
           type: string
+      additionalProperties: false
```
chmeliik, line 1155: Another thing to validate could be `minProperties: 1` (if openapi supports that)

cachito/web/static/api_v1.yaml
```diff
@@ -813,6 +828,49 @@ components:
         - type
         - version
       - $ref: "#/components/schemas/Package"
+    PackageConfig:
+      type: object
+      properties:
+        path:
+          type: string
```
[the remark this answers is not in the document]
lkolacek: +1 for checking if path is relative on Python side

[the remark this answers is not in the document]
taylormadore: Yes, definitely

cachito/web/static/api_v1.yaml
taylormadore: Yes, I think that makes sense. Good catch! (also see my comment about null in the PackageManager thread above)

cachito/web/static/api_v1.yaml
taylormadore: Will add

cachito/web/static/api_v1.yaml
taylormadore: Good point. I wonder now though if we need the null value at all here and if we're correctly using "default: null" (and the "nullable: true" I added to accommodate it) in other areas of the spec?

It appears that an empty array is still valid for pkg_managers in RequestNew even if nullable is removed from PackageManager.

cachito/web/static/api_v1.yaml
chmeliik: Hmm, yeah that's a good point. I recall thinking it made sense in some context, let me see if I can remember what context that was :smile:

cachito/web/static/api_v1.yaml
chmeliik: Hmm, yeah that's a good point. I recall thinking it made sense in some context, let me see if I can remember what context that was :smile:

chmeliik: Oh yeah, now I remember. I was about a query like

`/requests?per_page=10`

Are the `state` and `pkg_manager` params to this query considered null, or not present? I would hope the latter, in which case not making them nullable sounds like the way to go

taylormadore: I believe it's the latter:
- https://swagger.io/docs/specification/describing-parameters/#required-optional
- https://swagger.io/docs/specification/describing-parameters/#nullable

I'm running a few tests now, but I think it's safe to not have them nullable. I'll remove a few other unnecessary ones in a separate commit

chmeliik (approved):
LGTM :tada:

lkolacek (approved):
Looks great :+1:

PR #721, opened 2022-09-26
Bump Cachito base images to Fedora 36

brunoapimentel:


chmeliik (commented):
> Go 1.18 resolves additional packages that were not currently
included in the integration tests expected data.

Looks like it's just stdlib, which makes sense, that's probably going to change in every release

More interesting parts of the release notes:
> `go mod download`

> If the main module's go.mod file specifies [go 1.17](https://tip.golang.org/ref/mod#go-mod-file-go) or higher, go mod download without arguments now downloads source code for only the modules explicitly [required](https://tip.golang.org/ref/mod#go-mod-file-require) in the main module's go.mod file. (In a go 1.17 or higher module, that set already includes all dependencies needed to build the packages and tests in the main module.) To also download source code for transitive dependencies, use go mod download all.

^ I think that's what we want, but we might want to file a story to add a go 1.17 or higher integration test

> `go work`

> The go command now supports a "Workspace" mode. If a go.work file is found in the working directory or a parent directory, or one is specified using the GOWORK environment variable, it will put the go command into workspace mode. In workspace mode, the go.work file will be used to determine the set of main modules used as the roots for module resolution, instead of using the normally-found go.mod file to specify the single main module. For more information see the [go work](https://tip.golang.org/pkg/cmd/go#hdr-Workspace_maintenance) documentation.

Do we want to add a check and fail requests that use workspaces?

brunoapimentel:
> > `go mod download`
> 
> > If the main module's go.mod file specifies [go 1.17](https://tip.golang.org/ref/mod#go-mod-file-go) or higher, go mod download without arguments now downloads source code for only the modules explicitly [required](https://tip.golang.org/ref/mod#go-mod-file-require) in the main module's go.mod file. (In a go 1.17 or higher module, that set already includes all dependencies needed to build the packages and tests in the main module.) To also download source code for transitive dependencies, use go mod download all.
> 
> ^ I think that's what we want, but we might want to file a story to add a go 1.17 or higher integration test
> 
How this test would look like? Just a new cachito-testing repository that has go >=1.17?

> go work
> 
> The go command now supports a "Workspace" mode. If a go.work file is found in the working directory or a parent directory, or one is specified using the GOWORK environment variable, it will put the go command into workspace mode. In workspace mode, the go.work file will be used to determine the set of main modules used as the roots for module resolution, instead of using the normally-found go.mod file to specify the single main module. For more information see the [go work](https://tip.golang.org/pkg/cmd/go#hdr-Workspace_maintenance) documentation.
> 
> Do we want to add a check and fail requests that use workspaces?
I think this is a good idea. I filed a story so we can follow up on this feature.

chmeliik:
> I think this is a good idea. I filed a story so we can follow up on this feature.

But should we disable them right now? I think there's a decent chance that requests using workspaces will succeed -- because the support is built into various `go mod` commands -- but an equally good chance that Cachito will not report all the modules in the content manifest.

chmeliik:
> How this test would look like? Just a new cachito-testing repository that has go >=1.17?

yeah, probably just take an existing repo, bump the version to 1.18 and see what happens to the go.mod file after running `go mod tidy` with go 1.18

brunoapimentel:
> > I think this is a good idea. I filed a story so we can follow up on this feature.
> 
> But should we disable them right now? I think there's a decent chance that requests using workspaces will succeed -- because the support is built into various `go mod` commands -- but an equally good chance that Cachito will not report all the modules in the content manifest.

Let's do this right now, but in a separate PR then.

chmeliik (approved):
LGTM :+1:

docker/Dockerfile-api
```diff
@@ -27,9 +27,9 @@ RUN pip3 install -r requirements.txt --no-deps --no-cache-dir --require-hashes \
     && pip3 install . --no-deps --no-cache-dir
 
 # Use the system CA bundle for the requests library
-ENV REQUESTS_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.crt
+ENV REQUESTS_CA_BUNDLE=/etc/ssl/certs/ca-bundle.crt
 # Use the system CA bundle for native SSL calls from celery (python)
-ENV SSL_CERT_FILE=/etc/pki/tls/certs/ca-bundle.crt
+ENV SSL_CERT_FILE=/etc/ssl/certs/ca-bundle.crt
```
chmeliik, line 32: Maybe this change would deserve an explanation in the commit message

chmeliik (approved):
Re-ACK

fepas (approved):
LGTM!

fepas (approved):
LGTM!

PR #726, opened 2022-09-30
Fail a request if a Go workspace exists in the repository

brunoapimentel:
Since Cachito is not fully adapted to process Go workspaces, this check aims to avoid incorrect results in case this feature is used accidentaly.

### DIFF
--- a/cachito/workers/tasks/gomod.py
+++ b/cachito/workers/tasks/gomod.py
@@ -1,9 +1,10 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import logging
 import os
+from pathlib import Path
 
 from cachito.common.packages_data import PackagesData
-from cachito.errors import FileAccessError, GoModError, UnsupportedFeature
+from cachito.errors import FileAccessError, GoModError, InvalidRepoStructure, UnsupportedFeature
 from cachito.workers import run_cmd
 from cachito.workers.config import get_worker_config
 from cachito.workers.paths import RequestBundleDir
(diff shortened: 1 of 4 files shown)


cachito/workers/tasks/gomod.py
```diff
@@ -39,6 +40,27 @@ def _find_missing_gomod_files(bundle_dir, subpaths):
...
+    if (repo_root / "go.work").exists():
+        return True
+
+    return False
+
+
+def _fail_if_bundle_dir_has_workspaces(bundle_dir: RequestBundleDir, subpaths: list[str]):
+    for subpath in subpaths:
+        if _is_workspace(bundle_dir.source_root_dir, subpath):
+            log.warning("go.work file found at %s", bundle_dir / subpath)
```
chmeliik, line 60: go.work is not necessarily at `bundle_dir / subpath` - do we want to log the actual path or mention that it might be in  a parent dir?

brunoapimentel: My bad, I realized my mistake afterwards and forgot to update the logs.

I think it's enough to log `bundle_dir.source_root_dir / subpath`, right? Subpath might only be a `.`, in which case, pathlib.Path will resolve the operation.

chmeliik: Oh, I didn't notice the missing `source_root_dir`

My comment was about the message being misleading
```
go.work file found at .../repo/workspace/mod_a
```
But there's no go.work file there, it is at `.../repo/workspace`

chmeliik: maybe just add `go.work file found at ... (or a parent directory)`

brunoapimentel: You're right, I completely missed that.

chmeliik (approved):
LGTM :+1:

tests/integration/test_gomod_packages.py
```diff
@@ -66,3 +66,30 @@ def test_gomod_vendor_check_fail(env_name, test_env):
         f"#{completed_response.id}: Request failed correctly, but with unexpected message: "
         f"{completed_response.data['state_reason']}. Expected message was: {error_msg}"
     )
+
+
+def test_gomod_workspace_check(test_env):
+    """
+    Validate failing of gomod requests that contain worksapces.
```
lkolacek, line 73: typo workspaces

lkolacek (approved):
Looks good. Found just one nitpick. Thanks Bruno :+1:

tests/integration/test_gomod_packages.py
```diff
@@ -66,3 +66,30 @@ def test_gomod_vendor_check_fail(env_name, test_env):
...
+    initial_response = client.create_new_request(
+        payload={
+            "repo": env_data["repo"],
+            "ref": env_data["ref"],
+            "pkg_managers": env_data["pkg_managers"],
+        },
+    )
+    completed_response = client.wait_for_complete_request(initial_response)
+    assert completed_response.status == 200
+    assert completed_response.data["state"] == "failed"
```
ejegrova, line 89: nitpick: If the state is different, I don't think we'll have request ID to examine the logs.

brunoapimentel: Do you think we should log the Request.id during the test execution? I think the other tests in this package have the same flaw.

PR #728, opened 2022-09-30
Implement CLI using Typer

chmeliik:
Very WIP, just check if you're :+1: for [Typer](https://typer.tiangolo.com/) or if you'd rather stick to argparse

TODO (regardless of Typer/argparse)
- [x] Validate package input (by creating a pydantic model)
- [x] Define what `fetch_gomod_source` is supposed to return (an output model)
- [x] Provide env. vars in the output dir (as JSON?) - add them to the output model
- [x] Document the CLI better (especially what --package accepts)


chmeliik:
More or less ready, including unit tests. Left some TODOs here and there for myself, I'll try to get to those eventually

cachi2/interface/cli.py
```diff
@@ -0,0 +1,96 @@
...
+        metavar="PKG",
+    ),
+    source: Path = Option(DEFAULT_SOURCE, help="Process the git repository at this path."),
+    output: Path = Option(DEFAULT_OUTPUT, help="Write output files to this directory."),
+    # TODO: let's have actual flags like --gomod-vendor instead?
+    flags: str = Option(
+        "",
+        help="Pass additional flags as a comma-separated list.",
+        metavar="FLAGS",
+    ),
```
chmeliik, line 87: WDYT? Having `--flags=flag1,flag2,flag3` as opposed to `--flag1 --flag2 --flag3` feels very anti-CLI

cachi2/core/models/output.py
```diff
@@ -0,0 +1,113 @@
...
+        """Sort and de-duplicate dependencies."""
+        return unique_sorted(dependencies, by=lambda dep: (dep.type, dep.name, dep.version))
+
+
+class EnvironmentVariable(pydantic.BaseModel):
+    """An environment variable."""
+
+    name: str
+    value: str
+    kind: Literal["literal", "path"]
```
chmeliik, line 57: How are we gonna do environment variables? Previously, OSBS had to take `path` env. vars and make their values absolute.

Do we want to do that right away and get rid of the `kind`? Do we want a subcommand that generates fully resolved env. vars in the specified format (json, .env, ...)? Do we want to decide later?

chmeliik:
PTAL

Ignore the first commit, that's separate in #731 

And without the 2 unit-test commits, it's not _that_ huge :sweat_smile:

chmeliik:
IMO this is in review-able state now, I do want to address some of the TODOs but in a later PR

cachi2/interface/cli.py
```diff
@@ -0,0 +1,153 @@
...
+def cachi2(  # noqa: D103; docstring becomes part of --help message
+    version: bool = Option(
+        False,
+        "--version",
+        callback=version_callback,
+        is_eager=True,
+        help="Show version and exit.",
+    ),
+    log_level: LogLevel = Option(
+        "DEBUG",
```
brunoapimentel, line 40: I think that, for a better user experience, we should leave the default log leve as INFO, since it creates a more concise output.

brunoapimentel:
So, this is a not a valid command:
```
$ cachi2 fetch-deps --log-level INFO --package gomod

Usage: cachi2 fetch-deps [OPTIONS]
Try 'cachi2 fetch-deps --help' for help.

Error: No such option: --log-level
```
But this is valid:
```
$ cachi2 --log-level INFO fetch-deps --package gomod
```
I'm thinking it is a little confusing to have options that are top level and others specific to the subcommand. Also, you would only notice the `--log-level` option if you used `--help` in the top level. I would rather have the `cachi2` command only show the subcommands available, and have all options that affect the subcommand output tied to it directly.

Maybe it's only my impression, though. Wdyt?

brunoapimentel:
Also, we should aim to improve the logs, as they are the primary indicator of progress in a request now. Look at an example of INFO level logs:
```
$ cachi2 --log-level INFO fetch-deps --package gomod
2022-10-05 19:20:16,834 INFO Go version: go version go1.18.6 linux/amd64
2022-10-05 19:20:16,834 INFO Fetching the gomod dependencies for request in subpath .
2022-10-05 19:20:16,834 INFO Fetching the gomod dependencies at the "." directory
2022-10-05 19:20:16,835 INFO Downloading the gomod dependencies
2022-10-05 19:20:47,602 INFO Retrieving the list of packages
2022-10-05 19:20:47,640 INFO Retrieving the list of package level dependencies
```
I feel like it lacks in the very least a line that shows the command completed successfully. I would improve the output related to the CLI in this PR, though. We can tackle the other parts later.

cachi2/interface/cli.py
brunoapimentel: +1 for a single `--flags=str[]`  flag :)

cachi2/core/models/output.py
brunoapimentel: The fully resolved paths would be based on the `output_dir`? I'm trying to imagine how this would be used in a Tekton pipeline, and I think the build step can change the `output_dir` path when it mounts it.

On the other hand, maybe we can see this from the perspective of: you said the output should be "here", so we're delivering env variables considering that output is "here", if you change where output is, you should handle the env variables. Sounds more simple, and can be enough for an initial approach.

brunoapimentel:
I left some comments with my initial perceptions, specially based on user experience. I'll dive deeper into the code tomorrow.

cachi2/interface/cli.py
chmeliik: > Also, we should aim to improve the logs, as they are the primary indicator of progress in a request now.

+1, I'll file a story for that and change the default to INFO

chmeliik:
> I'm thinking it is a little confusing to have options that are top level and others specific to the subcommand. Also, you would only notice the --log-level option if you used --help in the top level. I would rather have the cachi2 command only show the subcommands available, and have all options that affect the subcommand output tied to it directly.

As far as I know, all the big python CLI libraries do it like that, including argparse
```python
#!/usr/bin/env python
import argparse

ap = argparse.ArgumentParser()
ap.add_argument("--log-level")

subcommands = ap.add_subparsers()
cmd = subcommands.add_parser("cmd")

ap.parse_args()
```
```shell
/tmp ·························································································
❯ ./test.py --log-level=debug cmd

/tmp ·························································································
❯ ./test.py cmd --log-level=debug
usage: test.py [-h] [--log-level LOG_LEVEL] {cmd} ...
test.py: error: unrecognized arguments: --log-level=debug
```
You can get around that by not having any top-level options and instead adding them to each subcommand separately.

But some options like `--log-level` and later `--config` should apply to every subcommand, no?

cachi2/core/models/output.py
chmeliik: Yeah, same issue with mismatched paths that I mentioned in the pip design. I'll hack together what the subcommand could look like and we can discuss it later

chmeliik:
Moved --log-level to the fetch-deps subcommand

Also changed the model validation errorfrom using `log.error()` to simply printing to stderr (to make sure users can't disable the error message accidentally)

cachi2/interface/cli.py
lkolacek: +1 for `cachi2 fetch-deps --gomod-vendor --cgo-disable`

chmeliik: Added a WIP commit that allows both forms, LMK what you think

lkolacek:
> Moved --log-level to the fetch-deps subcommand
> 
> Also changed the model validation error from using `log.error()` to simply printing to stderr (to make sure users can't disable the error message accidentally)

```
(venv) lkolacek@lkolacek-ThinkPad-T590 cachito (cli-typer=) $ cachi2 fetch-deps --log-level INFO --package gomod
2022-10-10 13:29:58,378 INFO Go version: go version go1.18.1 linux/amd64

  File "/tmp/cachito/venv/bin/cachi2", line 8, in <module>
    sys.exit(app())

  File "/tmp/cachito/cachi2/interface/cli.py", line 194, in fetch_deps
    request_output = gomod.fetch_gomod_source(request)

  File "/tmp/cachito/cachi2/core/package_managers/gomod.py", line 59, in fetch_gomod_source
    invalid_files_print = "; ".join(invalid_gomod_files)

TypeError: sequence item 0: expected str instance, PosixPath found
```
+1 for moving  --log-level to the fetch-deps subcommand
https://github.com/chmeliik/cachito/blob/429c9761fca932d19b93486210620aac990c33a9/cachi2/core/package_managers/gomod.py#L134 should this contain str type?

chmeliik:
> https://github.com/chmeliik/cachito/blob/429c9761fca932d19b93486210620aac990c33a9/cachi2/core/package_managers/gomod.py#L134 should this contain str type?

Good catch! Kept it as `list[Path]` but fixed the code that creates the error message

lkolacek (approved):
Looks great to me Adam. Thanks! I vote for merging this soon and create separate PRs for future changes. :+1: :)

cachi2/core/models/output.py
```diff
@@ -0,0 +1,81 @@
...
+
+    packages: list[Package]
+    environment_variables: list[EnvironmentVariable]
+
+    @pydantic.validator("packages")
+    def unique_packages(cls, packages: list[Package]) -> list[Package]:
+        """Sort packages and check that there are no duplicates."""
+        return unique_sorted(
+            packages,
+            by=lambda pkg: (pkg.type, pkg.name, pkg.version),
```
brunoapimentel, line 69: Should we add this key to a @property in the Package and Depedency classes?

cachi2/core/models/input.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class Request(pydantic.BaseModel):
+    """Holds all data needed for the processing of a single request."""
+
+    source_dir: Path
+    output_dir: Path
+    packages: list[PackageInput]
+    flags: frozenset[Flag] = frozenset()
+    dep_replacements: tuple[dict, ...] = ()  # TODO: do we want dep replacements at all?
```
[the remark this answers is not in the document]
chmeliik: I think the idea was that, if you do need to replace some dependency, you can quickly iterate by submitting Cachito requests with `dependency_replacements`. Once you're sure that your replacements work, you commit those to go.mod and stop using `dependency_replacements`.

But OSBS only allows setting those through `brew container-build --replace-dependency=...`, not in container.yaml. And only allows them for scratch builds. https://osbs.readthedocs.io/en/osbs_ocp3/users.html#replacing-project-dependencies-with-cachito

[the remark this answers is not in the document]
brunoapimentel: I think we should file a story to get rid of this feature in a following PR, wdyt?

cachi2/core/models/output.py
chmeliik: I think I had it like that at some point, but then I moved the key to the place where the unique-ification and sorting actually happens. Dependencies are unique in the context of a Package and Packages are unique in the context of the whole Request, so I figured I would let those decide "how unique". Plus the key is used for sorting as well, which felt less obvious when it was defined on the classes.

cachi2/core/models/input.py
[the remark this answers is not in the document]
chmeliik: We have one to investigate if we can get rid of it :-)

PR #738, opened 2022-10-10
FIX: Nexus authentication on JS tarballs download and remove npmrc_file

fepas:
CLOUDBLD-9007


cachito/workers/pkg_managers/general_js.py
```diff
@@ -165,96 +169,85 @@ def download_dependencies(
         skip_deps = set()
 
     conf = get_worker_config()
-    with tempfile.TemporaryDirectory(prefix="cachito-") as temp_dir:
-        npm_rc_file = os.path.join(temp_dir, ".npmrc")
-        if conf.cachito_nexus_ca_cert and os.path.exists(conf.cachito_nexus_ca_cert):
```
brunoapimentel, line 170: Don't we need to use the ca_cert anymore? I'm thinking that the requests in the Openshift envs could fail without the cert.

fepas: Good question. I don´t know, what do you think @chmeliik?

chmeliik: Hmm, I don't think the other package managers that make direct requests to Nexus use the CA

https://github.com/containerbuildsystem/cachito/blob/47fb9c6056f7b701f8ba51e877f82be271821a52/cachito/workers/pkg_managers/pip.py#L1573
https://github.com/containerbuildsystem/cachito/blob/47fb9c6056f7b701f8ba51e877f82be271821a52/cachito/workers/pkg_managers/rubygems.py#L300

I'm guessing the CA that `requests` is configured to use includes the Nexus CA

brunoapimentel (approved):
I would add a brief commit description that explains why the npmrc file was dropped, but otherwise, LGTM

chmeliik (approved):
LGTM :+1:

(+1 to explain why .npmrc was dropped in the commit message)

fepas:
Done!

PR #741, opened 2022-10-17
CVE-2007-4559 Patch

fepas:
CVE-2007-4559 is a 15 year old bug in the Python tarfile package. By using extract() or extractall() on a tarfile object without sanitizing input, a maliciously crafted .tar file could perform a directory path traversal attack.

This issue was initially described in [this](https://github.com/containerbuildsystem/cachito/pull/740) pull request.

CLOUDBLD-11296

### DIFF
--- a/cachito/workers/__init__.py
+++ b/cachito/workers/__init__.py
@@ -4,6 +4,8 @@ import json
 import logging
 import re
 import subprocess  # nosec
+from pathlib import Path
+from tarfile import ExtractError, TarFile
 from typing import Iterator
 
 from cachito.errors import SubprocessCallError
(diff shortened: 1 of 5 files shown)


cachito/workers/utils.py
```diff
@@ -0,0 +1,35 @@
...
+    :raise Exception: if there is a Traversal Path Attempt in the Tar File.
+    """
+    for member in tar.getmembers():
+
+        member_path = os.path.join(path, member.name)
+        abs_directory = os.path.abspath(path)
+        abs_target = os.path.abspath(member_path)
+        prefix = os.path.commonprefix([abs_directory, abs_target])
+
+        if not (prefix == abs_directory):
```
chmeliik, line 32: Could we do the relative path check the same way we do in other places? https://github.com/containerbuildsystem/cachito/blob/7662e45d43f74e96a1b745112688c0b4ed7b9e62/cachito/workers/tasks/general.py#L103-L105

Or https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_relative_to, since we no longer care about python < 3.9

cachito/workers/utils.py
```diff
@@ -0,0 +1,35 @@
...
+            raise Exception("Attempted Path Traversal in Tar File")
```
chmeliik, line 33: A more specific exception type would be good

tkdchen:
:+1:

cachito/workers/__init__.py
```diff
@@ -59,3 +61,27 @@ def load_json_stream(s: str) -> Iterator:
...
+    extracted safely and throws an exception otherwise.
+
+    :param tarfile tar: the tarfile to be extracted.
+    :param str path: specifies a different directory to extract to.
+    :param numeric_owner: if True, only the numbers for user/group names are used and not the names.
+    :raise ExtractError: if there is a Traversal Path Attempt in the Tar File.
+    """
+    for member in tar.getmembers():
+
+        member_path = PurePath(path).joinpath(member.name)
```
chmeliik, line 82: The original patch makes both `path` and `member_path` absolute before checking them. I think we need to do the same, otherwise
```python
In [8]: path = PurePath(".")

In [9]: path.joinpath("../foo").is_relative_to(path)
Out[9]: True
```

chmeliik: Ideally, we might want to use [Path.resolve](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve), which also resolves symlinks (something the original patch misses)

chmeliik (commented):
Trying to decide if we need to test this or not :thinking: 

But the original patch from Trellix didn't include tests either and I don't know how to construct malicous tarballs, so :shrug:

chmeliik:
Hmm, actually, constructing malicious tarballs is not very difficult. The `arcname` param to [TarFile.add()](https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.add) lets you use any path at all, including malicious ones.

That unit test might be worth it after all?

brunoapimentel (approved):
Besides the linter issues in CI (which are being fixed in a separte PR), this LGTM :rocket:

chmeliik (approved):

taylormadore (approved):

PR #742, opened 2022-10-17
switch to setuptools-scm for versioning

taylormadore:
A change was introduced in setuptools-git-versioning 1.12 where leading, non-numeric characters (except v) are no longer allowed in the version string. This is causing issues with our tag-based versioning.

Switch to use setuptools-scm instead.

CLOUDBLD-11280

### DIFF
--- a/docker/Dockerfile-api
+++ b/docker/Dockerfile-api
@@ -8,6 +8,7 @@ RUN dnf -y install \
     --setopt=tsflags=nodocs \
     httpd \
     gcc \
+    git-core \
     libffi-devel \
     libpq-devel \
     mod_auth_gssapi \
--- a/docker/Dockerfile-api
+++ b/docker/Dockerfile-api
@@ -24,7 +24,8 @@ COPY ./docker/cachito-httpd.conf /etc/httpd/conf/httpd.conf
 
 RUN pip3 install -r requirements.txt --no-deps --no-cache-dir --require-hashes \
     && pip3 install -r requirements-web.txt --no-deps --no-cache-dir --require-hashes \
-    && pip3 install . --no-deps --no-cache-dir
+    && pip3 install . --no-deps --no-cache-dir \
+    && rm -rf .git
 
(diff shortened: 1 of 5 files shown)


docker/Dockerfile-api
```diff
@@ -22,7 +23,8 @@ RUN dnf -y install \
 COPY . .
 COPY ./docker/cachito-httpd.conf /etc/httpd/conf/httpd.conf
 
-RUN pip3 install -r requirements.txt --no-deps --no-cache-dir --require-hashes \
+RUN --mount=type=bind,source=.git,target=/src/.git,Z \
```
taylormadore, line 26: Hadolint is going to have the same parsing error here on the "Z" as over in [Bruno's cachi2 Containerfile PR](https://github.com/containerbuildsystem/cachi2/pull/3#discussion_r994842242). Since this is a parsing error instead of a rule check, I don't think it can be ignored.

Should we just copy the .git directory in?

chmeliik: How did these builds ever work with setuptools-git-versioning :confused: 

Yeah, I'd say we should copy it and then `rm -rf` after installing. Maybe also uninstall `git-core`

chmeliik: Oh that's how
```shell
$ podman run --rm -ti quay.io/containerbuildsystem/cachito-workers:latest pip list | grep cachito
cachito            0.0.1
```

taylormadore: Yes, I chuckled a little bit when I saw that yesterday :sweat_smile:

One issue I can see with removing git-core is that in the absence of a default "fallback" version, [the local development install](https://github.com/containerbuildsystem/cachito/blob/7662e45d43f74e96a1b745112688c0b4ed7b9e62/docker-compose.yml#L57) will hit an error when trying to install the cachito package: `OSError: 'git' was not found`

chmeliik (approved):
LGTM :taco:

lkolacek (approved):

PR #744, opened 2022-10-18
workflow: use checkout action version 3

MartinBasti:
Older versions run with unsupported nodejs

### DIFF
--- a/.github/workflows/gating.yaml
+++ b/.github/workflows/gating.yaml
@@ -17,7 +17,7 @@ jobs:
         python-version: ["3.10"]
 
     steps:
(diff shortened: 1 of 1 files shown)


taylormadore (approved):

chmeliik:
You mentioned something about doing a full clone - I think we'll need that in order for setuptools-scm to work

MartinBasti:
> You mentioned something about doing a full clone - I think we'll need that in order for setuptools-scm to work

Added

lkolacek (approved):
Looks good :+1:

PR #745, opened 2022-10-18
Ensure go.sum is generated before calling `go mod vendor`

arewm:
As of go 1.18, `go.sum` is no longer updated when `go mod vendor` is called [1]. In order to maintain backwards compatibility, we should ensure that the file is updated before vendoring.

[1]: https://tip.golang.org/doc/go1.18

Signed-off-by: arewm <arewm@users.noreply.github.com>

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,4 @@
 .env/
-.git/
 .idea/
 .pytest_cache/
 .tox/
--- a/cachito/workers/pkg_managers/gomod.py
+++ b/cachito/workers/pkg_managers/gomod.py
@@ -200,11 +200,10 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
         should_vendor, can_make_changes = _should_vendor_deps(
             flags, app_source_path, worker_config.cachito_gomod_strict_vendor
         )
+        log.info("Downloading the gomod dependencies")
+        run_download_cmd(("go", "mod", "download"), run_params)
         if should_vendor:
             _vendor_deps(run_params, can_make_changes, git_dir_path)
-        else:
-            log.info("Downloading the gomod dependencies")
-            run_download_cmd(("go", "mod", "download"), run_params)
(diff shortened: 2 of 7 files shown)


arewm:
rebased on top of https://github.com/containerbuildsystem/cachito/pull/742 to see whether CI will pass.

chmeliik:
I can see some issues with this approach.

Normally, when `gomod-vendor` is enabled, we only download modules to app/vendor/. If we call `go mod download` unconditionally, the dependencies will be both in app/vendor/ and deps/gomod/. Doubling the size of the final tarball and possibly the time taken to process the request.

> ensure that the file is updated before vendoring.

By updating the checksums, Cachito bypasses checksum verification. Or more accurately, Cachito decides what the checksums are when this should be up to the user. This opens up the user to dependency confusion and such

chmeliik:
Following the gChat discussion, I don't think we want any Cachito changes for this.

The user's repository already had a correct go.sum at https://github.com/ManageIQ/manageiq-v2v-conversion_host/tree/master/vm-import-provider

The issue was that the Cachito request was being submitted for the root path, not for the `vm-import-provider` subpath. The root has a [symlink to go.mod](https://github.com/ManageIQ/manageiq-v2v-conversion_host/blob/master/go.mod) but doesn't have one for go.sum

arewm:
Thanks for the feedback, @chmeliik!

I had thought about the size doubling and was prepared to add one more step to the vendoring process to remove the cache.

Regarding the checksum changes, is this something that Cachito had been doing and we were just not checking it? We current have the ability to check to see if the vendor directory is modified after vendoring, but we don't expand that check to any other place that I am aware of. Would this functionality of `go mod download` or `go mod vendor` be considered a bug which is now only partially resolved with the patter in go 1.18?

arewm:
Closing PR. This is fixing a backwards compatibility "bug" as we shouldn't be modifying `go.sum` (or creating it) as this would prevent reproducible builds.

PR #749, opened 2022-10-27
Fix asyncio behavior when a request fails

fepas:
When there is a failure while downloading dependencies, Cachito should not continue creating download tasks for the remaining dependencies, but cancel the Request as a whole.

Log of a failed request -> https://pastebin.com/mTqpPtiL
Repo used to create a failed request -> https://github.com/fepas/cachito-yarn-with-not-found-deps.git

CLOUDBLD-9007

### DIFF
--- a/cachito/workers/pkg_managers/general.py
+++ b/cachito/workers/pkg_managers/general.py
@@ -187,8 +187,17 @@ async def async_download_binary_file(
                     if not chunk:
                         break
                     f.write(chunk)
-    except Exception as e:
-        raise NetworkError(f"Could not download {url}: {e}")
+
+    except Exception as exception:
+        log.error(f"Unsuccessful download: {tarball_name}")
+        # "from None" since we have the exception context in the logs
+        raise NetworkError(
+            (
+                f"Could not download {tarball_name} from {url}. "
+                f"exception_name: {exception.__class__.__name__}, "
+                f"details: {exception}"
+            )
+        ) from None
 
     log.debug(f"Download completed - {tarball_name}")
 
--- a/cachito/workers/pkg_managers/general_js.py
+++ b/cachito/workers/pkg_managers/general_js.py
@@ -22,6 +22,7 @@ from cachito.errors import (
     FileAccessError,
     InvalidChecksum,
     InvalidFileFormat,
+    NetworkError,
     NexusError,
     UnsupportedFeature,
 )
@@ -106,16 +107,25 @@ async def get_dependencies(
     async with aiohttp.ClientSession() as session:
 
         tasks: Set[asyncio.Task] = set()
+
         results = []
 
         for dep_identifier in deps_to_download:
-            proxied_url, tarball_name = parse_dependency(proxy_repo_url, dep_identifier)
-            results.append(tarball_name)
 
             if len(tasks) >= concurrency_limit:
                 # Wait for some download to finish before adding a new one
-                _, tasks = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
+                done, tasks = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
+                # Check for exceptions
+                try:
+                    await asyncio.gather(*done)
+                except NetworkError:
+                    for t in tasks:
+                        t.cancel()
+                    raise
 
+            proxied_url, tarball_name = parse_dependency(proxy_repo_url, dep_identifier)
+
+            results.append(tarball_name)
             tasks.add(
                 asyncio.create_task(
                     async_download_binary_file(
--- a/tests/test_workers/test_pkg_managers/test_general_js.py
+++ b/tests/test_workers/test_pkg_managers/test_general_js.py
@@ -276,7 +276,7 @@ async def test_get_dependecies(mock_async_download_binary_file):
         "http://nexus:8081/repository/cachito-yarn-53/",
         "/tmp/cachito-archives/bundles/temp/53/deps/yarn",
         deps_list,
-        5,
+        conf.cachito_js_concurrency_limit,
         conf.cachito_nexus_username,
         conf.cachito_nexus_password,
     )


chmeliik:
Looking at https://stackoverflow.com/questions/30361824/asynchronous-exception-handling-in-python again, I may have found an easier way to do things
```diff
diff --git a/cachito/workers/pkg_managers/general_js.py b/cachito/workers/pkg_managers/general_js.py
index 7242c8a..f343d83 100644
--- a/cachito/workers/pkg_managers/general_js.py
+++ b/cachito/workers/pkg_managers/general_js.py
@@ -114,7 +114,14 @@ async def get_dependencies(
 
             if len(tasks) >= concurrency_limit:
                 # Wait for some download to finish before adding a new one
-                _, tasks = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
+                done, tasks = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
+                # Check for exceptions
+                try:
+                    await asyncio.gather(*done)
+                except:
+                    for t in tasks:
+                        t.cancel()
+                    raise
 
             tasks.add(
                 asyncio.create_task(

```

cachito/workers/pkg_managers/general.py
```diff
@@ -179,18 +179,18 @@ async def async_download_binary_file(
...
             with open(os.path.join(download_dir, tarball_name), "wb") as f:
                 while True:
                     chunk = await resp.content.read(chunk_size)
                     if not chunk:
                         break
                     f.write(chunk)
-    except Exception as e:
-        raise NetworkError(f"Could not download {url}: {e}")
 
-    log.debug(f"Download completed - {tarball_name}")
```
chmeliik, line 193: Maybe we could keep the logs about starting and completing the downloads? I think it's good to have them as a progress indicator

chmeliik (commented):
LGTM, just a suggestion not to drop the log messages

cachito/workers/pkg_managers/general.py
fepas: Yeah, but it seemed like a very big spam in big requests. Don't you think that would be a problem?

cachito/workers/pkg_managers/general.py
```diff
@@ -179,18 +179,18 @@ async def async_download_binary_file(
...
+    except Exception as e:
+        log.error(f"Unsuccessful download: {tarball_name}")
+        message = f"Could not download. exception_name: {e.__class__.__name__}, details: {str(e)}"
```
MartinBasti, line 192: `str` is implicit in F-string

cachito/workers/pkg_managers/general.py
```diff
@@ -179,18 +179,18 @@ async def async_download_binary_file(
...
+        raise NetworkError(message) from None
```
MartinBasti, line 193: Can we comment here reason why we want to suppress exc context?

fepas: What do you mean by suppressing exc context?

brunoapimentel: I believe Martin is asking about the `from None` part.

cachito/workers/pkg_managers/general.py
MartinBasti, line 192: Shouldn't we keep URL/tarball name in message?

cachito/workers/pkg_managers/general.py
chmeliik: IMO this level of spam is acceptable for debug logs. If I'm not mistaken, we did log all the downloads before converting to asyncio no? We also log all downloads for pip I believe.

Maybe keep just only the completed message?

fepas: If this is a behavior that already happens in pip, I don't see a problem in adding it.

cachito/workers/pkg_managers/general.py
chmeliik: I'm wondering if we're doing this backwards. Maybe the error message doesn't need to be that verbose (keep just `f"Could not download {tarball_name} from {url}"`) and we can _not_ suppress the context instead?

On the other hand, having everything in the same message might make it easier to search in Splunk logs etc.

cachito/workers/pkg_managers/general.py
```diff
@@ -187,8 +187,17 @@ async def async_download_binary_file(
...
+    except Exception as exception:
+        log.error(f"Unsuccessful download: {tarball_name}")
+        message = (
+            f"Could not download {tarball_name} from {url}. "
+            f"exception_name: {exception.__class__.__name__}, "
+            f"details: {exception}"
+        )
+
+        raise NetworkError(message) from None
+        # Using "from None" since we have the exception context in the logs
```
chmeliik, line 200: If we keep the more verbose err. msg and suppress the context, could the comment go above the `raise` line? I completely missed it at first when looking at the diff and was wondering why the conversation was resolved :smile:

chmeliik (commented):
LGTM :+1:

One open question about not suppressing exception context vs. more verbose error messages, but I'm OK with the current state of things (if the comment is moved above the line it refers to)

cachito/workers/pkg_managers/general.py
MartinBasti: don't we usually write comments on line before?

https://peps.python.org/pep-0008/#block-comments

cachito/workers/pkg_managers/general.py
fepas: The problem, raising our exception inside the other, makes us receive in the logs that Python warning:    

    During handling of the above exception, another exception occurred 

That way, cachito logs are much more comfortable to read using `from None`.

cachito/workers/pkg_managers/general.py
MartinBasti: what adam said :D

cachito/workers/pkg_managers/general.py
chmeliik: The alternative is `raise ... from exception`, which is intended exactly for the use case of transforming errors into our custom error types
https://docs.python.org/3/tutorial/errors.html#exception-chaining

Although admittedly, it's not a whole lot better, AFAIK it just replaces
```
During handling of the above exception, another exception occurred:
```
with
```
The above exception was the direct cause of the following exception:
```

cachito/workers/pkg_managers/general.py
fepas: Sorry, I put the cart before the horse :facepalm: 

image

chmeliik (approved):
LGTM when CI passes :+1:

MartinBasti (approved):

taylormadore:
Coverage dropped a tiny bit, otherwise LGTM :+1:

fepas:
> Coverage dropped a tiny bit, otherwise LGTM +1

Yes, there are some difficulties to test the asyncio downloading tarball function, so I think it's not worth it to add new dependencies just to it.

taylormadore (approved):

PR #754, opened 2022-11-08
fix typing errors flagged by mypy 0.990

taylormadore:

### DIFF
--- a/cachito/web/migrations/env.py
+++ b/cachito/web/migrations/env.py
@@ -14,7 +14,8 @@ config = context.config
 
 # Interpret the config file for Python logging.
 # This line sets up loggers basically.
-fileConfig(config.config_file_name)
+if config.config_file_name is not None:
+    fileConfig(config.config_file_name)
 logger = logging.getLogger("alembic.env")
 
 # add your model's MetaData object here
@@ -23,7 +24,7 @@ logger = logging.getLogger("alembic.env")
 # target_metadata = mymodel.Base.metadata
 from flask import current_app
 
-config.set_main_option("sqlalchemy.url", current_app.config.get("SQLALCHEMY_DATABASE_URI"))
+config.set_main_option("sqlalchemy.url", current_app.config["SQLALCHEMY_DATABASE_URI"])
 target_metadata = current_app.extensions["migrate"].db.metadata
 
 # other values from the config, defined by the needs of env.py,
--- a/cachito/workers/pkg_managers/general.py
+++ b/cachito/workers/pkg_managers/general.py
@@ -4,7 +4,7 @@ import logging
 import os
 import urllib
 from pathlib import Path
-from typing import Dict
+from typing import Dict, Optional
 
 import aiohttp
 import requests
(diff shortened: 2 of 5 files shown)


cachito/web/migrations/env.py
```diff
@@ -23,7 +24,7 @@
 # target_metadata = mymodel.Base.metadata
 from flask import current_app
 
-config.set_main_option("sqlalchemy.url", current_app.config.get("SQLALCHEMY_DATABASE_URI"))
+config.set_main_option("sqlalchemy.url", str(current_app.config.get("SQLALCHEMY_DATABASE_URI")))
```
MartinBasti, line 27: this is suspicious `str(None)`

taylormadore: Oops, forgot to fix this in the morning. Let me know if the change looks sane

cachito/web/migrations/env.py
```diff
@@ -23,7 +24,12 @@
 # target_metadata = mymodel.Base.metadata
 from flask import current_app
 
-config.set_main_option("sqlalchemy.url", current_app.config.get("SQLALCHEMY_DATABASE_URI"))
+database_uri = (
+    ""
+    if current_app.config.get("SQLALCHEMY_DATABASE_URI") is None
+    else str(current_app.config.get("SQLALCHEMY_DATABASE_URI"))
```
MartinBasti, line 30: It could be just this
```suggestion
    current_app.config.get("SQLALCHEMY_DATABASE_URI", "")
```

but will this code even work if URL is undefined? Shouldn't we fail instead?

taylormadore: Good point. What I was going to suggest next was:
1. Try to access SQLALCHEMY_DATABASE_URI. Don't handle a potential KeyError and let it fail
2. If the value of SQLALCHEMY_DATABASE_URI is None, raise a ValueError
3. Do the config.set_main_option, while casting the value of SQLALCHEMY_DATABASE_URI to string

FYI, the checks for 1 and 2 appear to be handled ahead of time by: https://github.com/containerbuildsystem/cachito/blob/f59adc16eba2166d43345a4c4da71b80fee261ce/cachito/web/config.py#L128

MartinBasti: I vote for 2.

brunoapimentel (approved):

cachito/web/migrations/env.py
MartinBasti: why do we need `str()` there?

taylormadore: @MartinBasti I originally added str() to fix the mypy error:
`cachito/web/migrations/env.py:26: error: Argument 2 to "set_main_option" of "Config" has incompatible type "Optional[Any]"; expected "str"  [arg-type]`
But direct indexing appears to resolve it without str() :+1: Thanks @chmeliik

cachito/web/migrations/env.py
```diff
@@ -23,7 +24,10 @@
 # target_metadata = mymodel.Base.metadata
 from flask import current_app
 
-config.set_main_option("sqlalchemy.url", current_app.config.get("SQLALCHEMY_DATABASE_URI"))
+if current_app.config["SQLALCHEMY_DATABASE_URI"] is None:
```
taylormadore, line 27: Keep or drop this check? It was originally added while using get() instead of direct indexing, and it looks like it already might be covered in: https://github.com/containerbuildsystem/cachito/blob/f59adc16eba2166d43345a4c4da71b80fee261ce/cachito/web/config.py#L128

chmeliik (approved):
LGTM :+1:

PR #762, opened 2022-11-17
Validate supported package managers before creating a request

brunoapimentel:
This validation is currently being done after the actual request creation in the database, which causes any request with an invalid package manager to never reach a worker and get stuck (as we're currently seeing with the additon of rubygems tests).


taylormadore (approved):
LGTM :+1:

chmeliik (approved):

PR #773, opened 2022-11-30
use github purls for local dependencies in parent directories

taylormadore:
CLOUDBLD-10559 and CLOUDBLD-12584

Currently we don't allow gomod local dependency replacements when the dependency is located in a parent directory. When the dependency is a local path to the parent directory, the purl in the content manifest isn't valid. See https://github.com/containerbuildsystem/cachito/pull/373#discussion_r575195006 for the original discussion on this.

The idea behind allowing this case is that we can instead provide a github purl for the dependency in the content manifest. So that:
`pkg:golang/github.com%2Fsubmariner-io%2Flighthouse%2Fcoredns@v0.0.0-20220711135356-c454961ea91a#../pkg/constants`
would become:
`pkg:golang/github.com%2Fsubmariner-io%2Flighthouse%2Fpkg%2Fconstants@v0.13.0-rc2`

Diff of the content-manifest for lighthouse at https://github.com/submariner-io/lighthouse/commit/c454961ea91af3bc45a1f40e157083f32269049c 
http://pastebin.test.redhat.com/1088699

### DIFF
--- a/cachito/web/content_manifest.py
+++ b/cachito/web/content_manifest.py
@@ -1,12 +1,14 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
+import os
 from copy import deepcopy
+from functools import cached_property
+from pathlib import Path
 from typing import List, Optional
 
 import flask
 
 from cachito.common.utils import get_repo_name
 from cachito.web.purl import (
-    replace_parent_purl_gopkg,
     replace_parent_purl_placeholder,
     to_purl,
     to_top_level_purl,
@@ -63,6 +65,11 @@ class ContentManifest:
         # package
         self._gitsubmodule_data = {}
 
+    @cached_property
+    def go_modules_by_name(self) -> dict[str, "Package"]:
+        """Get a mapping of go module names to their respective package object."""
+        return {module.name: module for module in self.packages if module.type == "gomod"}
+
     def process_gomod(self, package, dependency):
         """
         Process gomod package.
@@ -71,8 +78,25 @@ class ContentManifest:
         :param Dependency dependency: the gomod package dependency to process
         """
         if dependency.type == "gomod":
-            parent_purl = self._gomod_data[package.name]["purl"]
-            dep_purl = to_purl(dependency)
+
+            parent_module_name = package.name
+            relpath_from_parent_module_to_dep = None
+
+            # if we're processing a local dependency
+            if dependency.version and dependency.version.startswith("."):
+                dep_normpath = os.path.normpath(os.path.join(package.name, dependency.version))
+                parent_module_name = gomod.match_parent_module(
+                    dep_normpath, self._gomod_data.keys()
+                )
+                # if the dependency is not in this repo, use a relative path from the root module
+                # otherwise just use the parent purl directly
+                if dependency.name not in self._gomod_data:
+                    relpath_from_parent_module_to_dep = Path(dep_normpath).relative_to(
+                        Path(parent_module_name)
+                    )
+
+            parent_purl = self._gomod_data[parent_module_name]["purl"]
+            dep_purl = to_purl(dependency, relpath_from_parent_module_to_dep)
             dep_purl = replace_parent_purl_placeholder(dep_purl, parent_purl)
             icm_source = {"purl": dep_purl}
             self._gomod_data[package.name]["dependencies"].append(icm_source)
@@ -85,8 +109,49 @@ class ContentManifest:
         :param Dependency dependency: the go-package package dependency to process
         """
         if dependency.type == "go-package":
-            icm_dependency = {"purl": to_purl(dependency)}
+            if dependency.version and dependency.version.startswith("."):
+                dep_purl = self._get_local_go_package_dep_purl(package, dependency)
+            else:
+                dep_purl = to_purl(dependency)
+
+            icm_dependency = {"purl": dep_purl}
             self._gopkg_data[package]["dependencies"].append(icm_dependency)
+
+    def _get_local_go_package_dep_purl(self, package: "Package", dependency: "Package") -> str:
+        """
+        Get the purl for a local go-package dependency.
+
+        :param Package package: the go-package package to process
+        :param Dependency dependency: the go-package package dependency to process
+        """
+        if not dependency.version or not dependency.version.startswith("."):
+            raise ValueError(f"{dependency} has an invalid version for a local dependency")
+
+        modules = self.go_modules_by_name
+        dep_module_name = gomod.match_parent_module(dependency.name, modules.keys())
+
+        # if the dep_module is in this repo, replace the dependency version with the module version
+        if dep_module_name is not None:
+            dependency.version = modules[dep_module_name].version
+            return to_purl(dependency)
+
+        # dep_module is not in this repo, so use a purl with a relative path from the root module
+        package_module_name = gomod.match_parent_module(package.name, modules.keys())
+        if package_module_name is None:
+            # This should be impossible. A top-level go-package should match a module
+            raise RuntimeError(f"Could not find parent Go module for package: {package.name}")
+
+        dep_normpath = os.path.normpath(os.path.join(package_module_name, dependency.version))
+        dep_module_name = gomod.match_parent_module(dep_normpath, modules.keys())
+        if dep_module_name is None:
+            # This should be impossible. The dep module should at least match the root module
+            raise RuntimeError(f"Could not find parent Go module for package: {dep_normpath}")
+
+        relpath_from_parent_module_to_dep = Path(dep_normpath).relative_to(Path(dep_module_name))
+        parent_purl = self._gomod_data[dep_module_name]["purl"]
+        dep_purl = to_purl(dependency, relpath_from_parent_module_to_dep)
+
+        return replace_parent_purl_placeholder(dep_purl, parent_purl)
 
     def set_go_package_sources(self):
         """
@@ -108,7 +173,6 @@ class ContentManifest:
             if module_name is not None:
                 module = self._gomod_data[module_name]
                 self._gopkg_data[package_id]["sources"] = module["dependencies"]
-                replace_parent_purl_gopkg(self._gopkg_data[package_id], module["purl"])
             else:
                 flask.current_app.logger.warning(
                     "Could not find a Go module for %s", pkg_data["purl"]
--- a/cachito/workers/tasks/gomod.py
+++ b/cachito/workers/tasks/gomod.py
@@ -4,12 +4,22 @@ import os
 from pathlib import Path
 
 from cachito.common.packages_data import PackagesData
-from cachito.errors import FileAccessError, GoModError, InvalidRepoStructure, UnsupportedFeature
+from cachito.errors import (
+    FileAccessError,
+    GoModError,
+    InvalidRepoStructure,
+    InvalidRequestData,
+    UnsupportedFeature,
+)
 from cachito.workers import run_cmd
 from cachito.workers.config import get_worker_config
 from cachito.workers.paths import RequestBundleDir
 from cachito.workers.pkg_managers.general import update_request_env_vars
-from cachito.workers.pkg_managers.gomod import path_to_subpackage, resolve_gomod
+from cachito.workers.pkg_managers.gomod import (
+    match_parent_module,
+    path_to_subpackage,
+    resolve_gomod,
+)
 from cachito.workers.tasks.celery import app
 from cachito.workers.tasks.utils import get_request, runs_if_request_in_progress, set_request_state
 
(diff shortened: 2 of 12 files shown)


chmeliik:
The submariner/lighthouse manifest looks perfect to me :+1: 

Ideally, we could do some validation to make sure the content manifest will be generate-able

cachito/web/content_manifest.py
```diff
@@ -175,6 +179,38 @@ def process_rubygems_package(self, package, dependency):
...
+            }
+            module_name = gomod.match_parent_module(normpath, modules.keys())
+
+            if module_name is None:
+                flask.current_app.logger.error(
+                    (
+                        f"Could not find a Go module containing {normpath} "
+                        f"while processing dependency {dependency} of package {package}"
+                    )
+                )
```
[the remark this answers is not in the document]
taylormadore: Yes, definitely. Will add :+1:

taylormadore:
Rebased on main, added some validation to fetch_gomod_source, and added partial unit test coverage

cachito/workers/tasks/gomod.py
```diff
@@ -62,6 +72,34 @@ def _fail_if_bundle_dir_has_workspaces(bundle_dir: RequestBundleDir, subpaths: l
...
+        package["name"] for package in packages_json_data.packages if package["type"] == "gomod"
+    ]
+    log.debug(packages_json_data.packages)
+
+    for package in packages_json_data.packages:
+        for dependency in package.get("dependencies", []):
+            if dependency["version"] and ".." in Path(dependency["version"]).parts:
+                dep_normpath = os.path.normpath(
+                    os.path.join(package["name"], dependency["version"])
+                )
```
chmeliik, line 92: I'm not 100% sure, but isn't the `dependency["version"]` relative to the **module** rather than the package?

To find the dep_normpath, I think you would first have to find the parent module of the *package*, the join the *module* name with the dependency version

cachito/workers/tasks/gomod.py
```diff
@@ -62,6 +72,34 @@ def _fail_if_bundle_dir_has_workspaces(bundle_dir: RequestBundleDir, subpaths: l
...
+                matching_module_name = match_parent_module(dep_normpath, modules)
+                if matching_module_name is None:
+                    raise InvalidRequestData(
+                        (
+                            f"Could not find a Go module in this request containing {dep_normpath} "
+                            f"while processing dependency {dependency} of {package['name']}"
+                        )
```
chmeliik, line 99: Could we try to point the user towards a solution here?

Maybe something like

> Please tell Cachito to process the module which contains the dependency. Maybe the parent module of {parent_module_of_package}?

cachito/workers/tasks/gomod.py
```diff
@@ -62,6 +72,34 @@ def _fail_if_bundle_dir_has_workspaces(bundle_dir: RequestBundleDir, subpaths: l
             raise InvalidRepoStructure("Go workspaces are not supported by Cachito.")
 
 
+def _fail_if_parent_replacement_not_included(packages_json_data: PackagesData) -> None:
+    """
+    Fail if any dependency replacement refers to a parent dir that isn't included in this request.
```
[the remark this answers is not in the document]
taylormadore: > I'm too lost in Go modules and our gomod code to suggest any other solution

:sweat_smile: Same

cachito/workers/tasks/gomod.py
```diff
@@ -62,6 +72,34 @@ def _fail_if_bundle_dir_has_workspaces(bundle_dir: RequestBundleDir, subpaths: l
...
+
+    :param PackagesData packages_json_data: the collection of resolved packages for the request
+    :raises InvalidRequestData: if the module being replaced is not part of this request
+    """
+    modules = [
+        package["name"] for package in packages_json_data.packages if package["type"] == "gomod"
+    ]
+    log.debug(packages_json_data.packages)
```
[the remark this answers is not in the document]
taylormadore: Nope, this was some debug that slipped in. Thanks for catching :+1:

cachito/workers/tasks/gomod.py
taylormadore: Yeah, I think you're right. When I was testing, the module and package were often the same, so I think I didn't notice

chmeliik:
I decided to test a particularly nasty example (sorry):  https://github.com/cachito-testing/cachito-gomod-local-parent-deps/commit/584bdfdb70279d2ada6e68028a684b6c8afa0767

Request:

```json
{
  "repo": "https://github.com/cachito-testing/cachito-gomod-local-parent-deps",
  "ref": "584bdfdb70279d2ada6e68028a684b6c8afa0767",
  "pkg_managers": ["gomod"],
  "packages": {
    "gomod": [{"path": "."}, {"path": "foo-module"}, {"path": "foo-module/bar-module"}]
  }
}
```

Doesn't including the staging/... module and the request still passes (as it should, IMO)

But when you look at the content manifest:
```shell
curl localhost:8080/api/v1/requests/2/content-manifest | jq 'del(.image_contents[].dependencies[] | select(.purl | test("@") | not))'
```

Relevant part:

```json
    {
      "dependencies": [
        {
          "purl": "pkg:golang/github.com%2Fcachito-testing%2Fcachito-gomod-local-parent-deps%2Fbaz-package@v0.0.1-0.20230124152505-584bdfdb7027"
        },
        {
          "purl": "pkg:golang/github.com%2Fcachito-testing%2Fcachito-gomod-local-parent-deps%2Ffoo-module%2Fbar-module%2Fbar-package@v0.0.0-20230124152505-584bdfdb7027"
        },
        {
          "purl": "pkg:golang/github.com%2Fcachito-testing%2Fcachito-gomod-local-parent-deps%2Ffoo-module%2Ffoo-package@v0.0.0-20230124152505-584bdfdb7027"
        },
        {
          // Not sure if we can consider this correct. I would expect:
          // pkg:golang/github.com/cachito-test/cachito-gomod-local-parent-deps@v0.0.x#staging/src/spam-module/spam
          "purl": "pkg:golang/github.com%2Fcachito-testing%2Fcachito-gomod-local-parent-deps%2Fstaging%2Fsrc%2Fspam-module%2Fspam@v0.0.1-0.20230124152505-584bdfdb7027"
        }
      ],
      "purl": "pkg:golang/github.com%2Fcachito-testing%2Fcachito-gomod-local-parent-deps%2Ffoo-module@v0.0.0-20230124152505-584bdfdb7027",
      "sources": [
        {
          "purl": "pkg:golang/github.com%2Fcachito-testing%2Fcachito-gomod-local-parent-deps%2Ffoo-module%2Fbar-module@v0.0.0-20230124152505-584bdfdb7027"
        },
        {
          "purl": "pkg:golang/github.com%2Fcachito-testing%2Fcachito-gomod-local-parent-deps@v0.0.1-0.20230124152505-584bdfdb7027"
        },
        {
          // I think this one was supposed to be spam-module (either this or the one above)
          "purl": "pkg:golang/github.com%2Fcachito-testing%2Fcachito-gomod-local-parent-deps@v0.0.1-0.20230124152505-584bdfdb7027"
        }
      ]
    }
```

cachito/web/content_manifest.py
```diff
@@ -175,6 +180,44 @@ def process_rubygems_package(self, package, dependency):
...
+                    (f"There is no Go module containing {dep_normpath} in this request")
+                )
+
+            module = modules[dep_module_name]
+            if dependency.type == "gomod":
+                dependency.name = module.name
+                dependency.version = module.version
+            elif dependency.type == "go-package":
+                dependency.name = dep_normpath
+                dependency.version = module.version
```
chmeliik, line 219: I think this will have to work differently if we want to generate correct purls for `../staging/src/some-module` type of use cases

One idea would be to do roughly this

```python
def process_gomod(self, module, dependency):
    dep_normpath = normpath(join(module.name, dependency.name))
    parent_module = get_parent_module(dep_normpath)
    relpath_from_parent_module_to_dep = relpath(parent_module, dep_normpath)

    dep_purl_with_placeholder = to_purl(dependency, parent_package_relpath=relpath_from...)
    parent_purl = self.get_purl(parent_module)

    dep_purl = replace_parent_purl_placeholder(dep_purl_with_placeholder, parent_purl)
```

But there might be an issue around making sure that you already have/can generate the parent module purl when you need it (or maybe not, haven't thought about it much)

chmeliik: Or maybe even simpler than that (I didn't check if `to_purl` does what is needed to achieve this)
```python
dep_purl = f"{parent_purl}#{relpath_from_parent_module_to_dep}"
```

cachito/web/content_manifest.py
```diff
@@ -71,8 +73,25 @@ def process_gomod(self, package, dependency):
...
+                dep_normpath = os.path.normpath(os.path.join(package.name, dependency.version))
+                parent_module_name = gomod.match_parent_module(
+                    dep_normpath, self._gomod_data.keys()
+                )
+                # if the dependency is not in this repo, use a relative path from the root module
+                # otherwise just use the parent purl directly
+                if dependency.name not in self._gomod_data:
+                    relpath_from_parent_module_to_dep = Path(dep_normpath).relative_to(
+                        Path(parent_module_name)
+                    )
```
chmeliik, line 91: It took me a while to understand why `dependency.name in self._gomod_data` is a special case. For others: it would mean that this dependency is one of the modules that the user specified in the Request. We processed the module => we have a regular purl for it, we don't need to use a subpath-based purl.

~This seems correct, but I wonder if we need to special-case it. If the dependency is in gomod_data, then `gomod.match_parent_module(dep_normpath, gomod_data) == dep_normpath`. The dependency is its own parent module, the path from parent to dependency is an empty `Path()`.~

~(we would instead have logic for empty paths in `to_purl()`)~

Edit: scratch this, we need this special case for go-packages

cachito/web/content_manifest.py
```diff
@@ -85,9 +104,50 @@ def process_go_package(self, package, dependency):
...
+        if not dependency.version or not dependency.version.startswith("."):
+            raise ValueError("{dependency} has an invalid version for a local dependency")
+
+        modules = {module.name: module for module in self.packages if module.type == "gomod"}
+        dep_module_name = gomod.match_parent_module(dependency.name, modules.keys())
+
+        # if the dep_module is in this repo, replace the dependency version with the module version
+        if dep_module_name is not None:
+            dependency.version = modules[dep_module_name].version
+            return to_purl(dependency)
```
chmeliik, line 131: Ah, but we won't be able to un-special-case it here. The relative path from the dep_module_name to the dep_normpath won't be empty unless the package is at the module root.

In that case, it's probably better to keep the special case for `gomod` as well

cachito/web/content_manifest.py
```diff
@@ -85,9 +104,50 @@ def process_go_package(self, package, dependency):
...
+        """
+        Get the purl for a local go-package dependency.
+
+        :param Package package: the go-package package to process
+        :param Dependency dependency: the go-package package dependency to process
+        """
+        if not dependency.version or not dependency.version.startswith("."):
+            raise ValueError("{dependency} has an invalid version for a local dependency")
+
+        modules = {module.name: module for module in self.packages if module.type == "gomod"}
```
chmeliik, line 125: Minor nitpick: I would prefer if we didn't have to re-create this dict every time. Maybe something like a `go_modules_by_name` [cached_property](https://docs.python.org/3/library/functools.html#functools.cached_property) could work?

cachito/web/content_manifest.py
```diff
@@ -108,7 +168,9 @@ def set_go_package_sources(self):
             if module_name is not None:
                 module = self._gomod_data[module_name]
                 self._gopkg_data[package_id]["sources"] = module["dependencies"]
-                replace_parent_purl_gopkg(self._gopkg_data[package_id], module["purl"])
+                # I believe this is fine to remove, since placeholder replacements
+                # should have already happened by this point
+                # replace_parent_purl_gopkg(self._gopkg_data[package_id], module["purl"])
```
chmeliik, line 173: +1

chmeliik (commented):
* find the parent module of the _package_
* join  the _dependency_ version (path) to the package module name
* match the result against the requested modules to find the parent module of the _dependency_
* find the relative path from the dependency module to the dependency

Insane, but correct :smiling_face_with_tear: 

Nice work, I'm about as sure as I can be that it's correct now :taco:

cachito/web/content_manifest.py
```diff
@@ -85,9 +104,50 @@ def process_go_package(self, package, dependency):
...
             self._gopkg_data[package]["dependencies"].append(icm_dependency)
 
+    def _get_local_go_package_dep_purl(self, package: "Package", dependency: "Package") -> str:
+        """
+        Get the purl for a local go-package dependency.
+
+        :param Package package: the go-package package to process
+        :param Dependency dependency: the go-package package dependency to process
+        """
+        if not dependency.version or not dependency.version.startswith("."):
```
lkolacek, line 122: Why do we check this twice? https://github.com/containerbuildsystem/cachito/pull/773/commits/b292ecdff5200276c0202524ec38f62bb1d211ae#diff-191e36300d9d76bf9bd1c4a7329c01e1c11d9c8b12b018821d7399663a23cddeR107

taylormadore: mypy wasn't pleased that version is optional in the Package class :smile:

cachito/web/content_manifest.py
taylormadore: Added :+1:

chmeliik (approved):
LGTM, great work :tada:

brunoapimentel (approved):
Besides leaving some nitpick comments, this review was more informative to me than anything. As far as I can tell, it LGTM :rocket::rocket::rocket:

Awesome work!

cachito/web/content_manifest.py
lkolacek: OK :)

lkolacek:
LGTM :+1:

lkolacek (approved):

PR #776, opened 2022-12-01
Remove dependency on `mapfile`

gsaslis:
`mapfile` doesn't exist on the standard bash version found on macOS

This should hopefully achieve the same without sacrificing too much in terms of readability...

### DIFF
--- a/bin/cachito-download.sh
+++ b/bin/cachito-download.sh
@@ -80,10 +80,7 @@ inject_config_files () {
     curl -fsS "$request_url/configuration-files" > "$config_json"
 
     echo "Injecting configuration files to remote-source/"
-    # According to shellcheck, this is the proper way to save lines in an array
-    mapfile -t paths < <(jq '.[].path' -r < "$config_json")
-
-    for path in "${paths[@]}"; do
+    jq '.[].path' -r < "$config_json" | while IFS= read -r path; do
         # Show the path indented by 4 spaces
         echo "    $path"
         mkdir -p "$(dirname "$output_dir/remote-source/$path")"


bin/cachito-download.sh
```diff
@@ -80,17 +80,15 @@ inject_config_files () {
...
-    for path in "${paths[@]}"; do
+    while IFS= read -r path; do
         # Show the path indented by 4 spaces
         echo "    $path"
         mkdir -p "$(dirname "$output_dir/remote-source/$path")"
 
         jq '.[] | select(.path == "'"$path"'") | .content' -r < "$config_json" |
             base64 --decode > "$output_dir/remote-source/$path"
-    done
+    done < <(jq '.[].path' -r < "$config_json")
```
chmeliik, line 90: Personal preference: could this be `jq ... | while IFS= read ...` ?

I feel like it might be clearer than having the input data at the end of the while loop

chmeliik (commented):
LGTM, thanks for the contribution :+1: 

The commits just [need a Signed-off-by](https://github.com/containerbuildsystem/cachito/pull/776/checks?check_run_id=9919577798) (and, IMO, to be squashed)

bin/cachito-download.sh
```diff
@@ -80,17 +80,15 @@ inject_config_files () {
...
-    for path in "${paths[@]}"; do
+    jq '.[].path' -r < "$config_json" | while IFS= read -r path; do
         # Show the path indented by 4 spaces
         echo "    $path"
         mkdir -p "$(dirname "$output_dir/remote-source/$path")"
 
         jq '.[] | select(.path == "'"$path"'") | .content' -r < "$config_json" |
             base64 --decode > "$output_dir/remote-source/$path"
     done
+
```
lkolacek, line 91: I believe we can remove this extra line.

chmeliik (approved):
LGTM :+1:

PR #777, opened 2022-12-01
decrease cachito_js_concurrency_limit to 1

taylormadore:
Is this something we would rather change in our playbooks?


chmeliik:
> Is this something we would rather change in our playbooks?

+1 (if Mike's idea doesn't fix the issues and we still want to test this after)

PR #779, opened 2022-12-07
add integration tests for private repos

taylormadore:
CLOUDBLD-12547

In environments where cachito workers are configured with the credentials necessary to access private
repositories, test requests for a private repo using both https and ssh.

Tested here, with a workaround for the test_env_vars changes not being present in downstream: https://container-build-jenkins.engineering.redhat.com/job/tmadore-TEST-cachito-stage-integration-tests/5/console

And where the env is skipped, as local/cachito-test would be: https://container-build-jenkins.engineering.redhat.com/job/tmadore-TEST-cachito-stage-integration-tests/6/console

- n/a Draft release notes are updated before merging

### DIFF
--- a/tests/integration/test_data/private_repo_packages.yaml
+++ b/tests/integration/test_data/private_repo_packages.yaml
@@ -1,0 +1,11 @@
+# Test data for the private repos test
+# The test will only be performed if $JOB_NAME includes one of the following
+private_repo_test_envs:
+  - cachito-prod
+  - cachito-stage
+private_repo_https:
+  repo: https://github.com/cachito-testing/cachito-no-package-manager-private.git
+  ref: 31e242befe367f940f878f27158d9cd036ba3b23
+private_repo_ssh:
+  repo: git@github.com:cachito-testing/cachito-no-package-manager-private.git
+  ref: 31e242befe367f940f878f27158d9cd036ba3b23
--- a/tests/integration/test_private_repos.py
+++ b/tests/integration/test_private_repos.py
@@ -1,0 +1,66 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import os
+from pathlib import Path
+from typing import Any
+
+import pytest
+from git import Repo
+
+from . import utils
+
+
+@pytest.mark.parametrize("env_package", ["private_repo_https", "private_repo_ssh"])
+def test_private_repos(env_package: str, test_env: dict[str, Any], tmp_path: Path) -> None:
+    """
+    Validate a cachito request with no package managers to a private repo.
+
+    Process:
+    Send new request to the Cachito API
+    Send request to check status of existing request
+
+    Checks:
+    * Check that the request completes successfully
+    * Check that no packages are identified in response
+    * Check that no dependencies are identified in response
+    * Check that the source tarball includes the application source code. Verify the expected files
+      by checking both the ref and diff because we don't have a clone of the private source repo.
+    * Check that the source tarball includes empty deps directory
+    * Check that the content manifest is successfully generated and empty
+    """
+    test_data = utils.load_test_data("private_repo_packages.yaml")
+    private_repo_test_envs = test_data["private_repo_test_envs"]
+    env_data = test_data[env_package]
+    is_supported_env = any(x in str(os.environ.get("JOB_NAME")) for x in private_repo_test_envs)
+    if not is_supported_env:
+        pytest.skip(
+            (
+                "This test is only executed in environments that "
+                "have been configured with the credentials needed "
+                "to access private repositories."
+            )
+        )
+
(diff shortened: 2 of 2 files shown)


tests/integration/test_private_repos.py
```diff
@@ -0,0 +1,48 @@
...
+    completed_response = client.wait_for_complete_request(initial_response)
+
+    utils.assert_properly_completed_response(completed_response)
+    assert completed_response.data["packages"] == []
+    assert completed_response.data["dependencies"] == []
+
+    client.download_and_extract_archive(completed_response.id, tmpdir)
+    source_path = tmpdir.join(f"download_{str(completed_response.id)}")
+    repo = Repo(os.path.join(source_path, "app"))
+    assert repo.head.commit.hexsha == env_data["ref"]
```
chmeliik, line 45: additional idea: check that `repo.git.diff()` is empty after checking the expected ref => we're basically checking all the expected files

taylormadore: Nice! Will add

test_env_vars.yaml
```diff
@@ -244,3 +244,8 @@ various_packages:
     repo: https://github.com/cachito-testing/cachito-gomod-test
     ref: 1827221d787cbd1e979b339cfbbf59728eddf0d4
     dependencies_count: 48
+# Test data for the private repos test
+# The test will only be performed if $JOB_NAME includes one of the following
+private_repo_test_envs:
+  - cachito-prod
+  - cachito-stage
```
ejegrova, line 251: Can we have this data elsewhere please? I believe [tests/integration/test_data/private_repo_packages.yaml](https://github.com/containerbuildsystem/cachito/pull/779/files#diff-c0f35f02e13615fcf7b7925d912b94e8299aaf78c70b071d0591f12438f5ee8d) should be fine for that. We wanted to move all specific test data from this file because we have to change it also downstream every time we make a change here.

tests/integration/test_data/private_repo_packages.yaml
```diff
@@ -0,0 +1,6 @@
+https:
```
ejegrova, line 1: nitpick: It might be nice to keep the same structure as in the other test_data files and add name of the test as the first key.

tests/integration/test_private_repos.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Process:
+    Send new request to the Cachito API
+    Send request to check status of existing request
+
+    Checks:
+    * Check that the request completes successfully
+    * Check that no packages are identified in response
+    * Check that no dependencies are identified in response
+    * Check that the source tarball includes the application source code
```
chmeliik, line 24: Thanks for adding the docstring :tada: 

Could we also mention how we check the application source code here, and why we have to do it differently than the other tests?

test_env_vars.yaml
taylormadore: Moved to the private repo test data file :+1:

tests/integration/test_data/private_repo_packages.yaml
taylormadore: Just to clarify, do you think I should change the https one to just "private_repo"?

ejegrova: no, sorry, that's not what I meant. Check out for example: https://github.com/containerbuildsystem/cachito/blob/master/tests/integration/test_data/cached_dependencies.yaml#L2
We have one key with name of the test `cached_package` and all the appropriate variables for that test are nested under that key. That way when there's another test we need to add for caching, we just use the same file for test data.

chmeliik: But `private_repo_https` and `private_repo_ssh` are two different test cases, right? The structure makes sense to me

chmeliik:
Do we care about type annotations in test code? Probably not that much, right?

chmeliik (approved):
LGTM :+1:

tests/integration/test_data/private_repo_packages.yaml
ejegrova: oh right, it's fine then

taylormadore:
> Do we care about type annotations in test code? Probably not that much, right?

It looks like we don't run mypy on the tests, but I added the annotations anyway :+1: 
In the process of doing so, I saw that the pytest tmpdir fixture [is deprecated-ish](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures), so I switched it to use tmp_path instead.

chmeliik (approved):
> In the process of doing so, I saw that the pytest tmpdir fixture [is deprecated-ish](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures), so I switched it to use tmp_path instead.

Yay for pathlib! :tada:

PR #785, opened 2022-12-20
Fix usage of vendor deps for gomod-vendor-check

lkolacek:
Make a fix for --gomod-vendor-check usage to create an empty gomod cache in the bundle directory so that any Cachi2 user does not have to guard against this directory not existing.

This change prevents from downloading dependencies to deps/ folder in output dir.

Signed-off-by: Ladislav Kolacek <lkolacek@redhat.com>

### DIFF
--- a/cachito/workers/pkg_managers/gomod.py
+++ b/cachito/workers/pkg_managers/gomod.py
@@ -281,7 +281,7 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
 
         bundle_dir = RequestBundleDir(request["id"])
 
-        if "gomod-vendor" in flags:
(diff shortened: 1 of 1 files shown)


taylormadore (approved):

chmeliik:
Minor nitpick for the commit message:

> so that any Cachi2 user does not have to guard against this directory not existing.

They don't have to guard against it either way, the directory is either empty or full but never missing. I'd say the fix makes sure that the build will use the vendored dependencies as intended, rather than the ones downloaded to deps/gomod/

chmeliik (approved):
LGTM, thanks for backporting the fix to Cachito :+1:

PR #786, opened 2022-12-21
update submodules with git instead of GitPython

taylormadore:
GitPython is unable to update submodules when the submodule repository doesn't have a master branch. Use the "git submodule update --init" command directly instead.

CLOUDBLD-12480

-  ~OpenAPI schema is updated (if applicable)~
-  ~DB schema change has corresponding DB migration (if applicable)~
- ~README updated (if worker configuration changed, or if applicable)~

### DIFF
--- a/cachito/workers/scm.py
+++ b/cachito/workers/scm.py
@@ -286,15 +286,26 @@ class Git(SCM):
         `go-github` submodule, the dir structure would look like,
         retrodep/go-github/<content_of_go-github_repo>
 
+        The git command is called directly because of an issue with GitPython's handling of
+        submodules: https://github.com/gitpython-developers/GitPython/issues/1058
+
         :param git.Repo repo: the repository object.
-        :raises RepositoryAccessError: if updating the git submodules fail.
+        :raises SubprocessCallError: if the 'git submodule update --init' command fails to
+                                     update the git submodules.
         """
+        log.debug(f"Git submodules for the requested repo are: {repo.submodules}")
+        cmd = ["git", "submodule", "update", "--init"]
         try:
-            log.debug(f"Git submodules for the requested repo are: {repo.submodules}")
-            repo.submodule_update(recursive=False)
-        except Exception as e:
-            log.exception("Updating the Git submodule(s) from '%s' failed %s", self.url, e)
-            raise RepositoryAccessError("Updating the Git submodule(s) failed")
+            run_cmd(cmd, {"cwd": repo.working_dir, "check": True})
+        except subprocess.CalledProcessError as exc:
+            log.error(
+                (
+                    f"Updating the Git submodule(s) from {self.url} failed. Cachito attempted "
+                    f"to execute the \"{' '.join(cmd)}\" command but encountered an error. "
+                    f"STDERR: {exc.stderr}"
+                )
+            )
+            raise SubprocessCallError("Updating the Git submodule(s) failed")
 
     @property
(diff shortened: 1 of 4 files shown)


cachito/workers/scm.py
```diff
@@ -286,15 +286,25 @@ def update_git_submodules(self, repo):
...
-            raise RepositoryAccessError("Updating the Git submodule(s) failed")
+            run_cmd(cmd, {"cwd": repo.working_dir, "check": True})
+        except subprocess.CalledProcessError:
+            log.exception(
+                (
+                    f"Updating the Git submodule(s) from {self.url} failed. Cachito attempted "
+                    f"to execute the \"{' '.join(cmd)}\" command but encountered an error."
+                )
+            )
+            raise SubprocessCallError("Updating the Git submodule(s) failed")
```
chmeliik, line 307: I think run_cmd already wraps CalledProcessErorr in something custom

taylormadore: I thought that when check=True that CalledProcessError would be raised for non-zero exit codes :thinking:

chmeliik: Oh, you're right, I missed the check=True param

But should we use it? I would almost consider that a misuse of the run_cmd function, since it's normally supposed to raise this for failing commands https://github.com/containerbuildsystem/cachito/blob/1221b0849785165202d8c38ce3558933b5b28a65/cachito/workers/__init__.py#L41-L45

taylormadore: Yeah, I agree that it's a bit odd. IIRC the reason that I took that approach was because I saw that run_cmd was being used in a similar way in the same file where git fsck was being called. Looking back through the history, it looked like it was done that way to take advantage of the [command timeout setting](https://github.com/containerbuildsystem/cachito/pull/457) https://github.com/containerbuildsystem/cachito/blob/f189008d440b7c001170d911da5ff4c5fd618a61/cachito/workers/scm.py#L107-L112

chmeliik: Ah, fair enough, I didn't realize it was already being used this way. Could we just make sure to log the stderr? I think the CalledProcessError exception does not include the stderr by default, we would have to log it explicitly like with the git fsck call

brunoapimentel (approved):
Besides the comment to address the exception logging, it LGTM.

tests/test_workers/test_scm.py
```diff
@@ -465,20 +465,22 @@ def test_fetch_source_invalid_archive_exists(mock_clone, mock_verify, caplog, gi
 
 
 @mock.patch("git.Repo")
-def test_update_git_submodules(mock_repo):
+@mock.patch("cachito.workers.scm.run_cmd")
+def test_update_git_submodules(mock_git_submodule_update, mock_repo):
```
brunoapimentel, line 469: Should we use this opportunity and add type annotations to this function signature here? If so, also look at the `test_update_git_submodules_failed` function.

cachito/workers/scm.py
taylormadore: Added :+1:

tests/test_workers/test_scm.py
taylormadore: Added :+1:

taylormadore:
Rebased on master

chmeliik (approved):
LGTM

Much simpler fix than I expected, kudos :tada:

PR #811, opened 2023-02-10
[STONEBLD-85] fix: Wrong handling of NPM workspaces

yaacov:
Issue: currently only paths that match exactly a workspace path are ignoread, if npm workspaces include a glob pattern, packages matching this pattern will not be ignored.

This commit test each version path for a matching workspace pattern and does not assume string matching

Signed-off-by: Yaacov <kobi.zamir@gmail.com>


yaacov:
@chmeliik please review

yaacov:
cc://@sjd78 @rszwajko

yaacov:
note:
workspaces glob pattern should not include "./" prefix, 
```python
workspaces=["./packages/mypackage"]
```
will fail to match
```python
version="file:package/mypackage"
```
but
```python
workspaces=["packages/mypackage"]
```
will work fine, it's a change we will have to do in our npm package

cachito/workers/pkg_managers/npm.py
```diff
@@ -24,6 +25,20 @@
 
 log = logging.getLogger(__name__)
 
+def _is_worksapce_version(version, workspaces):
```
chmeliik, line 28: typo: workspace

we also try to use type hints when possible:
```suggestion
def _is_workspace_version(version: str, workspaces: list[str]) -> bool:
```

cachito/workers/pkg_managers/npm.py
```diff
@@ -24,6 +25,20 @@
...
+
+    :param version a file: version string to match with a workspace glob pattern
+    :param list workspaces: package workspaces defined in package-lock.json
+    :return: true if path matches one of the workspaces
+    :rtype: boolean
+    """
+    path = version[5:]
+    for workspace in workspaces:
+        if fnmatch.fnmatch(path, workspace):
+            return True
```
chmeliik, line 40: > note:
> workspaces glob pattern should not include "./" prefix,
> ```
> workspaces=["./packages/mypackage"]
> ```
> will fail to match
> ```
> version="file:package/mypackage"
> ```

Should we normalize both the paths first? Maybe with something like this?
```python
Path(workspace).as_posix()
```

cachito/workers/pkg_managers/npm.py
```diff
@@ -24,6 +25,20 @@
...
+def _is_worksapce_version(version, workspaces):
+    """
+    Test if a version file path match one of the workspaces pattern string
+
+    :param version a file: version string to match with a workspace glob pattern
+    :param list workspaces: package workspaces defined in package-lock.json
+    :return: true if path matches one of the workspaces
+    :rtype: boolean
+    """
+    path = version[5:]
```
chmeliik, line 37: nitpick, this might be clearer:
```suggestion
    path = version.removeprefix("file:")
```

cachito/workers/pkg_managers/npm.py
yaacov: :+1:

cachito/workers/pkg_managers/npm.py
yaacov: done :+1:

cachito/workers/pkg_managers/npm.py
yaacov: :+1:

yaacov:
@chmeliik hi, I addressed the comments and added some tests, please re-review

yaacov:
@chmeliik sorry, had a typo, please re-run tests

yaacov:
@chmeliik, last time I think :crossed_fingers:

chmeliik:
> @chmeliik, last time I think crossed_fingers

Good demonstration of why we shouldn't fail-fast the linters, I'll fix that :smile:

chmeliik:
Tested locally with 
```json
{                                              
  "repo": "https://github.com/kubev2v/forklift-console-plugin",
  "ref": "c6f8c82beef917596ebec8eb888cc179531a7114",
  "pkg_managers": ["npm"]
}
```
(before the commit than remove `./` from the workspace path)

and worked :tada:

taylormadore (approved):
LGTM :+1:

PR #815, opened 2023-02-13
PoC: Use cachi2 as the worker backend for gomod

chmeliik:


chmeliik:
This is now in an almost backwards-compatible state. Integration tests are passing, the only backwards-incompatible change that I've made is, **as far as I know**, that I "accidentally" dropped the allowlist for local dependencies.

Gomod was the happy path, pip will be much more complex.

Some thoughts on the pros and cons in the final commit.

chmeliik:
Closing. After discussing with the team, we decided that cachi2 in Cachito is not worth it at the moment

PR #821, opened 2023-02-20
add -e flag to go-list command

taylormadore:
STONEBLD-631

Adding this flag to the go-list command prevents
erroneous packages from being printed to stderr
and causing the cachito request to fail needlessly

Tested locally with: https://github.com/openshift/sdn/commit/f2ff9949731708da53f8eb90dc5ff7ba7430a158

### DIFF
--- a/cachito/workers/pkg_managers/gomod.py
+++ b/cachito/workers/pkg_managers/gomod.py
@@ -208,8 +208,13 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
         if "force-gomod-tidy" in flags or dep_replacements:
             run_gomod_cmd(("go", "mod", "tidy"), run_params)
 
+        go_list = ["go", "list", "-e"]
+        if not should_vendor:
+            # Make Go ignore the vendor dir even if there is one
+            go_list.extend(["-mod", "readonly"])
+
         # main module
-        module_name = run_gomod_cmd(["go", "list", "-m"], run_params).rstrip()
+        module_name = run_gomod_cmd([*go_list, "-m"], run_params).rstrip()
 
         # module level dependencies
         if should_vendor:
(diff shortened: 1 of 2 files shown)


cachito/workers/pkg_managers/gomod.py
```diff
@@ -302,16 +303,16 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
 
         if not should_vendor:
             # Make Go ignore the vendor dir even if there is one
-            go_list = ["go", "list", "-mod", "readonly"]
+            go_list = ["go", "list", "-e", "-mod", "readonly"]
         else:
-            go_list = ["go", "list"]
+            go_list = ["go", "list", "-e"]
```
chmeliik, line 308: now that I'm looking at it, maybe we should define the `go_list` variable above the first usage of `go list` and use it consistently everywhere

We'll avoid more inconsistencies with passing -mod=readonly only sometimes

taylormadore: Yeah, that makes sense :+1:

cachito/workers/pkg_managers/gomod.py
```diff
@@ -208,8 +208,10 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
         if "force-gomod-tidy" in flags or dep_replacements:
             run_gomod_cmd(("go", "mod", "tidy"), run_params)
 
+        go_list = ["go", "list", "-e"]
```
chmeliik, line 211: Could we also decide `-mod=vendor` up here already? It looks like the only command which is currently missing -mod=vendor is `go list -m`, and I think I've seen failures because of it

taylormadore: Just checking--did you mean -mod=readonly here?

chmeliik: Ah, right, yes I did :+1:

chmeliik (approved):
LGTM, thanks for incorporating the `-mod=readonly` fix as well :grin:

cachito/workers/pkg_managers/gomod.py
lkolacek: Should we mention in commit message why we added "-mod", "readonly" flags?

lkolacek:
LGTM @taylormadore! Thank you, I left only 1 nitpick.

lkolacek (approved):

cachito/workers/pkg_managers/gomod.py
taylormadore: > Should we mention in commit message why we added "-mod", "readonly" flags?
Updated commit message :+1:

PR #827, opened 2023-02-28
Allow Yarn workspace dependencies by default (and many other fixes)

chmeliik:

- [ ] ~OpenAPI schema is updated (if applicable)~
- [ ] ~DB schema change has corresponding DB migration (if applicable)~
- [ ] ~README updated (if worker configuration changed, or if applicable)~

### DIFF
--- a/tests/test_workers/test_pkg_managers/test_general_js.py
+++ b/tests/test_workers/test_pkg_managers/test_general_js.py
@@ -6,11 +6,18 @@ import os.path
 import tarfile
 import textwrap
 from pathlib import Path
+from typing import Any
 from unittest import mock
 
 import pytest
 
-from cachito.errors import FileAccessError, InvalidFileFormat, NexusError, UnsupportedFeature
+from cachito.errors import (
+    FileAccessError,
+    InvalidFileFormat,
+    InvalidRepoStructure,
+    NexusError,
+    UnsupportedFeature,
+)
 from cachito.workers import nexus
 from cachito.workers.config import get_worker_config
 from cachito.workers.errors import NexusScriptError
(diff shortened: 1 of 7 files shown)


chmeliik:
This fixes one of the many problems that yarn workspaces introduce. https://github.com/cachito-testing/cachito-yarn-workspaces-happy-path explains some of them.

Now, users *can*  at least get correct results by adding every workspace as a `file:` dependency to package.json

chmeliik:
PTAL

This is the first part of Yarn workspace fixes - it allows `file:` dependencies in yarn.lock if they are workspaces.

The second part (**not part of this PR**) will be detecting workspaces even if they don't appear in yarn.lock

This PR also includes two related fixes:

1. Report `file:` dependencies for npm regardless of whether they are workspaces or not (previosuly, npm reported allowlisted dependencies but not workspaces)
2. Make the yarn allowlist behave the same as the npm allowlist (previously, yarn dependencies were allowlisted based on their path, not their name like npm)

brunoapimentel:
> This is the first part of Yarn workspace fixes - it allows `file:` dependencies in yarn.lock if they are workspaces.

Just curious, but does this scenario happen at all? I'm asking because the sample repo we got from this ticket had no mention of workspaces in the yarn.lock file.

brunoapimentel:
> > This is the first part of Yarn workspace fixes - it allows `file:` dependencies in yarn.lock if they are workspaces.
> 
> Just curious, but does this scenario happen at all? I'm asking because the sample repo we got from this ticket had no mention of workspaces in the yarn.lock file.

Nevermind, I just realized there are several instances of `file:` in the sample repository. I remember now the original state of the `package.json` when the user reached us: https://github.com/xperimental/kibana/commit/edcfd82cbdd2d00c65745d6aad3a9c80ea43cf1b

When you declare a dependency that is a workspace, but only use its version instead of `file:path`, this dep does not show up in the lockfile. Just a heads up that this is apparently a valid use case, so we will need to treat it in the following PR.

chmeliik:
> When you declare a dependency that is a workspace, but only use its version instead of `file:path`, this dep does not show up in the lockfile. Just a heads up that this is apparently a valid use case, so we will need to treat it in the following PR.

Yup,  I have an integration test ready for that in https://github.com/cachito-testing/cachito-yarn-workspaces

Just trying to make the code less ugly now :smile:

tests/test_workers/test_pkg_managers/test_general_js.py
```diff
@@ -794,6 +801,88 @@ def test_process_non_registry_dependency_github_not_in_nexus(mock_unrd, mock_gnc
...
+            "is_allowed": False,
+        },
+        {
+            "name": "not_a_file_dependency",
+            "version": "git+https://...",
+            "is_allowed": True,
+        },
+    ],
+)
+def test_vet_file_dependency_allowlist(test_input: dict[str, Any]):
```
brunoapimentel, line 872: Do we need to annotate the return value here?

tests/test_workers/test_pkg_managers/test_yarn.py
```diff
@@ -422,18 +480,16 @@ def mock_nexus_replacement_getitem(key):
     )
 
 
-@mock.patch("cachito.workers.pkg_managers.yarn._convert_to_nexus_hosted")
-def test_get_deps_disallowed_file_dep(mock_convert_hosted):
+def test_get_deps_disallowed_file_dep():
```
brunoapimentel, line 483: Do we need to annotate the return value here?

tests/test_workers/test_pkg_managers/test_general_js.py
chmeliik: Yup, good catch. And that's why we need stricter mypy /shakes_fist

brunoapimentel (approved):
LGTM :rocket: 
And thanks for all the fi~sh~xes

lkolacek (approved):
Looks amazing! Thank you Adam. Nice logic commits as always. :taco:

PR #842, opened 2023-03-16
Add leftpad to yarn cached dependencies test

chmeliik:
The repository used for the test was updated in
https://github.com/cachito-testing/cachito-yarn-with-deps/commit/dd651df74599872e38bd719a554bb4638631e135

The cached dependencies test uses the latest commit.


chmeliik:
Porting Taylor's fix (from the downstream gitlab) to upstream.

This test doesn't actually run upstream, but let's try to keep the data correct for the alternate reality where we get rid of the downstream copy.

taylormadore (approved):

ejegrova (approved):

PR #862, opened 2023-06-06
allow npm registry deps from the yarn registry

taylormadore:
In an npm request, consider dependencies from the yarn registry to be "registry" deps for the purpose of determining whether or not to upload to nexus.

During testing of https://github.com/containerbuildsystem/cachito/pull/857 I found a product with an npm package-lock.json file that has registry dependencies with "resolved" urls pointing to registry.yarnpkg.com. https://github.com/quay/quay/blob/1a96bd2f299f1199436a264f0d3afb8ff6f1ab60/package-lock.json

- [ ] ~New code has type annotations~ not new code
- n/a Draft release notes are updated before merging

### DIFF
--- a/cachito/workers/pkg_managers/general_js.py
+++ b/cachito/workers/pkg_managers/general_js.py
@@ -42,6 +42,7 @@ from cachito.workers.pkg_managers.general import (
 __all__ = [
     "download_dependencies",
     "get_dependencies",
+    "is_from_npm_registry",
     "parse_dependency",
     "finalize_nexus_for_js_request",
     "find_package_json",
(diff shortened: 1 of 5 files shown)


taylormadore:
@chmeliik @brunoapimentel WDYT?

chmeliik:
What would we save as `version` in the packages.json in this case? Should be the semver so that we get correct purls

brunoapimentel:
We should probably also port this to Cachi2 in case we decide to go with it.

chmeliik:
> We should probably also port this to Cachi2 in case we decide to go with it.

[on it](https://github.com/containerbuildsystem/cachi2/pull/196#issuecomment-1580225970)

chmeliik (approved):
LGTM :+1:

taylormadore:
> > > What would we save as `version` in the packages.json in this case? Should be the semver so that we get correct purls
> > 
> > 
> > I might not be sufficiently caffeinated yet this morning, because I'm not sure I follow sweat_smile
> 
> just asking if the purls in content manifest and versions in request JSON will be reported properly - I guess they will if the logic is based on the `is_registry_dep` property

Ah, I understand now. Yes, the semver is taken from the "version" field of the `package` or `dependency` in package-lock.json when _is_registry_dep is True.

brunoapimentel (approved):

PR #864, opened 2023-06-08
fix(gomod): parent replacement not always supported if major version > 1

ben-alkov:

When a Go module has local replacements to a parent directory, and the replaced module has v${major_version} in the name, Cachito may break.

Fix: when we look for a parent module, and the package name matches *something* but the dependency's normalized path doesn't match anything, we consider it found.

- https://issues.redhat.com/browse/STONEBLD-1266

Signed-off-by: Ben Alkov <ben.alkov@redhat.com>

### DIFF
--- a/cachito/web/content_manifest.py
+++ b/cachito/web/content_manifest.py
@@ -97,16 +97,23 @@ class ContentManifest:
 
             # if we're processing a local dependency
             if dependency.version and dependency.version.startswith("."):
+                dep = None
                 dep_normpath = os.path.normpath(os.path.join(package.name, dependency.version))
-                parent_module_name = gomod.match_parent_module(
+                if parent_module_name := gomod.match_parent_module(
+                    dependency.name, self._gomod_data.keys()
+                ):
+                    dep = dependency.name
+                elif parent_module_name := gomod.match_parent_module(
                     dep_normpath, self._gomod_data.keys()
-                )
+                ):
+                    dep = dep_normpath
+
+                relpath_from_parent_module_to_dep = Path(dep).relative_to(Path(parent_module_name))
+
                 # if the dependency is not in this repo, use a relative path from the root module
                 # otherwise just use the parent purl directly
-                if dependency.name not in self._gomod_data:
-                    relpath_from_parent_module_to_dep = Path(dep_normpath).relative_to(
-                        Path(parent_module_name)
-                    )
+                if relpath_from_parent_module_to_dep == Path():
+                    relpath_from_parent_module_to_dep = None
 
             parent_purl = self._gomod_data[parent_module_name]["purl"]
             dep_purl = to_purl(dependency, relpath_from_parent_module_to_dep)
--- a/cachito/workers/tasks/gomod.py
+++ b/cachito/workers/tasks/gomod.py
@@ -94,18 +94,22 @@ def _fail_if_parent_replacement_not_included(packages_json_data: PackagesData) -
                         f"Could not find parent Go module for package: {package['name']}"
                     )
 
+                dep_name = dependency["name"]
                 dep_normpath = os.path.normpath(
                     os.path.join(pkg_module_name, dependency["version"])
                 )
-                dep_module_name = match_parent_module(dep_normpath, modules)
-                if dep_module_name is None:
+                dep_module_name = match_parent_module(dep_name, modules) or match_parent_module(
+                    dep_normpath, modules
+                )
+
+                if not dep_module_name:
                     raise InvalidRequestData(
                         (
-                            f"Could not find a Go module in this request containing {dep_normpath} "
-                            f"while processing dependency {dependency} of package "
-                            f"{package['name']}. Please tell Cachito to process the module which "
-                            f"contains the dependency. Perhaps the parent module of "
-                            f"{pkg_module_name}?"
+                            f"Could not find a Go module in this request containing "
+                            f"{dependency['name']} while processing dependency {dependency} "
+                            f"of package {package['name']}. Please tell Cachito to process "
+                            f"the module which contains the dependency. Perhaps the parent "
+                            f"module of {pkg_module_name}?"
                         )
                     )
 
--- a/tests/test_content_manifest.py
+++ b/tests/test_content_manifest.py
@@ -82,39 +82,111 @@ def test_process_go(app, default_request):
     assert cm._gomod_data == expected_gomod_contents
 
 
-def test_process_gomod_with_local_dep(default_request: Request):
-    packages_json = [
-        {
-            "name": "example.com/org/project",
-            "type": "gomod",
-            "version": "1.1.1",
-            "dependencies": [],
-        },
-        {
-            "name": "example.com/org/project/foo",
-            "type": "gomod",
-            "version": "1.1.1",
-            "dependencies": [
-                {"name": "example.com/org/project", "type": "gomod", "version": "../"}
+@pytest.mark.parametrize(
+    "packages_json, expected_gomod_contents",
+    [
+        pytest.param(
+            [
+                {
+                    "name": "k8s.io/kubernetes",
+                    "type": "gomod",
+                    "version": "1.1.1",
+                    "dependencies": [],
+                },
+                {
+                    "name": "k8s.io/kubernetes/some-module",
+                    "type": "gomod",
+                    "version": "1.1.1",
+                    "dependencies": [
+                        {
+                            "name": "k8s.io/api",
+                            "type": "gomod",
+                            "version": "../staging/src/k8s.io/api",
+                        }
+                    ],
+                },
             ],
-        },
-    ]
-
-    expected_parent_purl = "pkg:golang/example.com%2Forg%2Fproject@1.1.1"
-    expected_foo_purl = "pkg:golang/example.com%2Forg%2Fproject%2Ffoo@1.1.1"
-    expected_src_purl = "pkg:golang/example.com%2Forg%2Fproject@1.1.1"
-
+            {
+                "k8s.io/kubernetes": {
+                    "purl": "pkg:golang/k8s.io%2Fkubernetes@1.1.1",
+                    "dependencies": [],
+                },
+                "k8s.io/kubernetes/some-module": {
+                    "purl": "pkg:golang/k8s.io%2Fkubernetes%2Fsome-module@1.1.1",
+                    "dependencies": [
+                        {"purl": "pkg:golang/k8s.io%2Fkubernetes@1.1.1#staging/src/k8s.io/api"},
+                    ],
+                },
+            },
+            id="normpath_matches",
+        ),
+        pytest.param(
+            [
+                {
+                    "name": "spam/v3",
+                    "type": "gomod",
+                    "version": "3.1.1",
+                    "dependencies": [],
+                },
+                {
+                    "name": "spam/ham/v3",
+                    "type": "gomod",
+                    "version": "3.1.1",
+                    "dependencies": [{"name": "spam/v3", "type": "gomod", "version": "../"}],
+                },
+            ],
+            {
+                "spam/v3": {
+                    "purl": "pkg:golang/spam%2Fv3@3.1.1",
+                    "dependencies": [],
+                },
+                "spam/ham/v3": {
+                    "purl": "pkg:golang/spam%2Fham%2Fv3@3.1.1",
+                    "dependencies": [{"purl": "pkg:golang/spam%2Fv3@3.1.1"}],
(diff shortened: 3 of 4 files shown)


chmeliik:
Tried to come up with a reasonable way to inspect the results

```python
import json
import sys
from collections import defaultdict

sbom = json.load(sys.stdin)

components_by_name = defaultdict(list)
for component in sbom["components"]:
    components_by_name[component["name"]].append(component)

etcd_components = {
    name: components
    for name, components in components_by_name.items()
    if name.startswith("go.etcd.io")
}

json.dump(etcd_components, sys.stdout, indent=2, sort_keys=True)
print()
```

```shell
request_id=1  # id of the etcd request
curl "localhost:8080/api/v1/sbom?requests=$request_id" | python the_script_above.py
```

And it's kind of an unholy mess :smile:

Here's a highlight
```json
  "go.etcd.io/etcd/api/v3": [
    {
      "name": "go.etcd.io/etcd/api/v3",
      "purl": "pkg:golang/go.etcd.io%2Fetcd%2Fapi%2Fv3@v3.5.1-0.20230511082319-da125abcfa3f",
      "type": "library",
      "version": "v3.5.1-0.20230511082319-da125abcfa3f"
    },
    {
      "name": "go.etcd.io/etcd/api/v3",
      "purl": "pkg:golang/go.etcd.io%2Fetcd%2Fclient%2Fv2@v2.0.0-20230511082319-da125abcfa3f",
      "type": "library",
      "version": "../../api"
    },
    {
      "name": "go.etcd.io/etcd/api/v3",
      "purl": "pkg:golang/go.etcd.io%2Fetcd%2Fclient%2Fv3@v3.0.0-20230511082319-da125abcfa3f",
      "type": "library",
      "version": "../../api"
    },
    {
      "name": "go.etcd.io/etcd/api/v3",
      "purl": "pkg:golang/go.etcd.io%2Fetcd%2Fetcdctl%2Fv3@v3.5.1-0.20230511082319-da125abcfa3f",
      "type": "library",
      "version": "../api"
    },
    {
      "name": "go.etcd.io/etcd/api/v3",
      "purl": "pkg:golang/go.etcd.io%2Fetcd%2Fetcdutl%2Fv3@v3.5.1-0.20230511082319-da125abcfa3f",
      "type": "library",
      "version": "../api"
    },
    {
      "name": "go.etcd.io/etcd/api/v3",
      "purl": "pkg:golang/go.etcd.io%2Fetcd%2Fserver%2Fv3@v3.5.1-0.20230511082319-da125abcfa3f",
      "type": "library",
      "version": "../api"
    },
    {
      "name": "go.etcd.io/etcd/api/v3",
      "purl": "pkg:golang/go.etcd.io%2Fetcd%2Ftests%2Fv3@v3.5.1-0.20230511082319-da125abcfa3f",
      "type": "library",
      "version": "../api"
    },
    {
      "name": "go.etcd.io/etcd/api/v3",
      "purl": "pkg:golang/go.etcd.io%2Fetcd%2Fv3@v3.5.1-0.20230511082319-da125abcfa3f",
      "type": "library",
      "version": "./api"
    }
  ],
```
The duplication is a pre-existing issue, but the purls should be the same for all the duplicates
(same as this one)
```json
    {
      "name": "go.etcd.io/etcd/api/v3",
      "purl": "pkg:golang/go.etcd.io%2Fetcd%2Fapi%2Fv3@v3.5.1-0.20230511082319-da125abcfa3f",
      "type": "library",
      "version": "v3.5.1-0.20230511082319-da125abcfa3f"
    },
```
And then all packages which are relative to `go.etcd.io/etcd/api/v3` should have the same version as `go.etcd.io/etcd/api/v3` and the purls should be based on name@verison

That part seems correct at first glance
```json
  "go.etcd.io/etcd/api/v3/authpb": [
    {
      "name": "go.etcd.io/etcd/api/v3/authpb",
      "purl": "pkg:golang/go.etcd.io%2Fetcd%2Fapi%2Fv3%2Fauthpb@v3.5.1-0.20230511082319-da125abcfa3f",
      "type": "library",
      "version": "v3.5.1-0.20230511082319-da125abcfa3f"
    }
  ],
```

chmeliik:
Not sure about this one to be honest, that may be a separate bug from what we're fixing here (since the path is not through a parent directory)

```json
    {
      "name": "go.etcd.io/etcd/api/v3",
      "purl": "pkg:golang/go.etcd.io%2Fetcd%2Fv3@v3.5.1-0.20230511082319-da125abcfa3f",
      "type": "library",
      "version": "./api"
    }
```

ben-alkov:
> And it's kind of an unholy mess :smile:

Yes, the generated content manifests have 10s of thousands of items... :roll_eyes:

chmeliik:
I think what might work is this:

1. find the parent module by dependency.name
2. if parent module not found, fall back to dep_normpath

(same change needed in content manifest code and the code in tasks/gomod.py which checks if we will be able to generate the content manifest)

ben-alkov:
>     1. find the parent module by dependency.name
>     2. if parent module not found, fall back to dep_normpath
>     same change needed in content manifest code

@chmeliik; PTAL, pushed changes for above

cachito/workers/tasks/gomod.py
```diff
@@ -94,18 +94,22 @@ def _fail_if_parent_replacement_not_included(packages_json_data: PackagesData) -
...
-                if dep_module_name is None:
+                dep_module_name = match_parent_module(dependency["name"], modules)
+                check_module_path = dep_module_name
+                if not dep_module_name:
+                    dep_normpath = os.path.normpath(
+                        os.path.join(pkg_module_name, dependency["version"])
+                    )
+                    check_module_path = dep_normpath
+
+                if not check_module_path and dependency["name"] not in modules:
```
chmeliik, line 105: I think this can be simpler:
```suggestion
                dep_name = dependency["name"]
                dep_normpath = os.path.normpath(os.path.join(pkg_module_name, dependency["version"]))
                dep_module_name = match_parent_module(dep_name, modules) or match_parent_module(dep_normpath, modules)

                if not dep_module_name:
```

cachito/web/content_manifest.py
```diff
@@ -97,7 +97,9 @@ def process_gomod(self, package, dependency, type="icm"):
 
             # if we're processing a local dependency
             if dependency.version and dependency.version.startswith("."):
-                dep_normpath = os.path.normpath(os.path.join(package.name, dependency.version))
+                dep_normpath = os.path.normpath(
+                    os.path.join(package.name, dependency.name)
+                ) or os.path.normpath(os.path.join(package.name, dependency.version))
                 parent_module_name = gomod.match_parent_module(
                     dep_normpath, self._gomod_data.keys()
                 )
```
chmeliik, line 105: And this should be equivalent to the comment above

chmeliik:
> I'm not sure what the code in tasks/gomod.py which checks if we will be able to generate the content manifest is referring to...

The `_fail_if_parent_replacement_not_included` which we're updating here exists only to verify that we will be able to generate a content manifest later.

If this function doesn't find a parent module, then the content manifest code also won't find one and will fail to generate a content manifest. But failing while generating the content manifest is too late, the request has already succeeded and now we can't return a manifest for it from the `/content-manifest` or `/sbom` endpoints.

So the function duplicates some of the logic to do the check while the request is still in progress.

Anyway, the changes here should be what we need :slightly_smiling_face:

cachito/web/content_manifest.py
```diff
@@ -99,8 +99,8 @@ def process_gomod(self, package, dependency, type="icm"):
             if dependency.version and dependency.version.startswith("."):
                 dep_normpath = os.path.normpath(os.path.join(package.name, dependency.version))
                 parent_module_name = gomod.match_parent_module(
-                    dep_normpath, self._gomod_data.keys()
-                )
+                    dependency.name, self._gomod_data.keys()
+                ) or gomod.match_parent_module(dep_normpath, self._gomod_data.keys())
                 # if the dependency is not in this repo, use a relative path from the root module
                 # otherwise just use the parent purl directly
                 if dependency.name not in self._gomod_data:
```
chmeliik, line 106: I think there's one more issue here with the relative path https://github.com/containerbuildsystem/cachito/blob/5dedcd3f5f14c18877a8924e67c802ef87ec543b/cachito/web/content_manifest.py#L106-L109

If we find the parent module based on `match_parent_module(dependency.name)`
and then try to take the relpath from the parent_module_name to `dep_normpath`
we can get an uncaught exception if the paths are not relative

If we find the parent module by dependency.name, we should take the relative path from parent_module_name to dependency.name

If we find the parent module by dep_normpath, take the relative path to dep_normpath

chmeliik:
It would also be good to unit-test the changes :pray:

cachito/web/content_manifest.py
ben-alkov: I actually understood that :hushed:

chmeliik:
@ben-alkov how is it going with this one? Need some help with the unit tests?

ben-alkov:
> @ben-alkov how is it going with this one? Need some help with the unit tests?

I pushed UT for 'cachito/cachito/workers/tasks/gomod.py'; still working on UT for content manifest

cachito/web/content_manifest.py
```diff
@@ -97,13 +97,23 @@ def process_gomod(self, package, dependency, type="icm"):
...
+                    match_type = "name"
+                elif parent_module_name := gomod.match_parent_module(
                     dep_normpath, self._gomod_data.keys()
-                )
+                ):
+                    match_type = "version"
                 # if the dependency is not in this repo, use a relative path from the root module
                 # otherwise just use the parent purl directly
-                if dependency.name not in self._gomod_data:
+                if dependency.name not in self._gomod_data and match_type == "name":
```
chmeliik, line 112: And we can replace the `if dependency.name not in self._gomod_data` logic with

```python
if relpath_from_parent_module_to_dep == Path():
    relpath_from_parent_module_to_dep = None
```
That should mean the same thing - if the dependency name is in _gomod_data, then it will match itself => the relpath will be empty

chmeliik:
If needed, I think I have serviceable unit tests for the content manifest code ready locally :slightly_smiling_face:

tests/test_content_manifest.py
```diff
@@ -206,48 +206,96 @@ def test_process_go_package_with_local_dep_unprocessed_module(default_request: R
...
             ],
-        },
-    ]
-
-    root_module_purl = "pkg:golang/example.com%2Forg%2Fproject@1.1.1"
-    foo_module_purl = "pkg:golang/example.com%2Forg%2Fproject%2Ffoo@2.2.2"
-    foo_package_purl = "pkg:golang/example.com%2Forg%2Fproject%2Ffoo%2Fpackage@2.2.2"
+            "pkg:golang/go.etcd.io%2Fetcd%2Fv3@v3.5.1-0.20230511082319-da125abcfa3f",
+            "pkg:golang/go.etcd.io%2Fetcd%2Fv3%2Fcontrib%2Flock%2Fclient@v3.5.1-0.20230511082319-da125abcfa3f",
+            "pkg:golang/go.etcd.io%2Fetcd%2Fclient%2Fpkg%2Fv3%2Fsrv@v3.5.1-0.20230511082319-da125abcfa3f",
```
chmeliik, line 285: These need fixing

One comment on the `root_module_purl`, `dep_module_purl`, `dep_package_purl` setup: I think that the "root" will have to be the purl for `go.etcd.io/etcd/client`, dep will be the one for `go.etcd.io/etcd` and dep_package the one for `go.etcd.io/etcd/api` (if you go with that)

The naming is a bit off since the "root" is not the root anymore, but basically: the root is the one that has dependencies, and the two dep_ ones are its dependencies

chmeliik:
> If needed, I think I have serviceable unit tests for the content manifest code ready locally slightly_smiling_face

Pushed it in case you want to use it as reference: #871

I used as a slightly different approach, but the three cases (both match, name matches, path matches) are there for both gomod and go-package

ben-alkov:
Ready for review; incorporates changes from #871

chmeliik (approved):
Tested locally with the etcd request, results look good :tada: 

LGTM :+1:

lkolacek (approved):
Code looks good :+1: Can we add more detailed description to commit message based on Long explanation in story -> Adam's Gitlab MR#309

ejegrova (approved):
LGTM

chmeliik:
Updated release draft

* Cachito can now process Go modules replaced by parent paths even if they have a major version in the name (e.g. `replace my-parent-module/v3 => ../`)
  * However, this only works if the replaced module is one of the `packages` that Cachito is told to process

PR #880, opened 2023-07-30
Add docker.io as the default registry name

thenets:
# Why

I'm using the `podman-compose` and, whenever I use `podman`, there's no default container registry.

Since all the default images expects `docker.io` as the default registry, I believe that makes sense to write it on the `docker-compose.yml` file.

- [ ] Draft release notes are updated before merging (I don't know what to do about this one here)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3'
 services:
   db:
-    image: postgres:9.6
+    image: docker.io/postgres:9.6
     environment:
       POSTGRES_USER: cachito
       POSTGRES_PASSWORD: cachito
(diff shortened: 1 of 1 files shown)


chmeliik (commented):
Can't really verify this, my podman resolves those to docker.io automatically

But if it works for you then LGTM :+1: 

Just missing a sign-off on the commit (`git commit --amend -s`)

thenets:
@chmeliik oh, sorry for the delay. I missed the message.
Just pushed the change :)

thenets:
I also rebased it now

chmeliik (approved):

brunoapimentel (approved):

PR #886, opened 2023-08-07
Bumps api and worker images to Fedora 38

brunoapimentel:
This change is primarily meant to get Go 1.20:
https://tip.golang.org/doc/go1.20


docker/Dockerfile-workers
```diff
@@ -12,7 +12,7 @@ RUN dnf -y install \
     krb5-devel \
     libffi-devel \
     mercurial \
-    npm \
+    nodejs-npm \
```
brunoapimentel, line 15: Dnf was acting weird and installing 2 versions of nodejs unless I specified the package as nodejs-npm :man_shrugging:

chmeliik (approved):
LGTM :+1:

chmeliik:
If we still have that version table in the README, it may need an update

brunoapimentel:
> If we still have that version table in the README, it may need an update

I was halfway doing it and forgot about it :facepalm: 

It's updated now.

taylormadore (approved):

brunoapimentel:
> 

That is a good point. Minor versions are important for Python and Golang, though. We could easily drop the patch versions.

PR #918, opened 2023-10-02
OpenTelemetry tracing.   

mike-kingsbury:

A combination of auto-instrumentation and manual instrumentation in the worker where external actions are occurring.  Traces are sent to OTel collectors based on an environment variable, or in the case of development environment, a jaeger instance brought up with docker-compose. 


cachito/web/content_manifest.py
```diff
@@ -91,7 +91,6 @@ def process_gomod(self, package, dependency, type="icm"):
         :param type: icm or sbom component
         """
         if dependency.type == "gomod":
-
```
brunoapimentel, line 94: Are these changes caused by some automatic linter you're using in the IDE? I'd say it's probably better to avoid reformating files we're not actually changing in this PR.

cachito/workers/pkg_managers/gomod.py
```diff
@@ -239,9 +243,11 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
             new_name = dep_replacement.get("new_name", name)
             version = dep_replacement["version"]
             log.info("Applying the gomod replacement %s => %s@%s", name, new_name, version)
-            run_gomod_cmd(
-                ("go", "mod", "edit", "-replace", f"{name}={new_name}@{version}"), run_params
-            )
+            with tracer.start_as_current_span("run gomod " + new_name):
```
brunoapimentel, line 246: Probably not worth having a specific trace for this piece here. This is for dependency replacements defined in the Cachito request itself, and AFAIK, nobody is using it.

cachito/workers/tasks/gomod.py
```diff
@@ -180,30 +184,33 @@ def fetch_gomod_source(request_id, dep_replacements=None, package_configs=None):
...
-            request_id,
-            "in_progress",
-            f'Fetching the gomod dependencies at the "{subpath}" directory',
-        )
-        request = get_request(request_id)
-        gomod_source_path = str(bundle_dir.app_subpath(subpath).source_dir)
-        try:
-            gomod = resolve_gomod(
-                gomod_source_path, request, dep_replacements, bundle_dir.source_dir
+        with tracer.start_as_current_span("fetching gomod deps in " + subpath):
```
brunoapimentel, line 187: Maybe it's enough to have a span in the `resolve_gomod` file, as it is the function that is doing the bulk of the work.

brunoapimentel (commented):
Is there an easy way to find the trace corresponding to a specific request id? I'm thinking it could be useful if someone wants to use the traces for troubleshooting.

Maybe we could log the trace id in the `POST /requests` endpoint, or maybe change the span title itself for something like `cachito-api: /api/v1/requests/<actual_request_id>`.

chmeliik:
What is the logic that determines what gets custom traces?

Is it every IO operation (network, file IO, subprocess)? Or bigger pieces of code that eventually end up doing some IO inside?

mike-kingsbury:
I was largely aiming for the top level functions for each type of packager,
as well as functions running network or long running processes.  One
example where this would have shown a problem in the past would have been
during the migration to MPP+.  When requests were taking longer than
expected, the traces would have shown the heavy disk I/O processes taking
considerably longer than Cachito on the ocp-c1 cluster.   Its trivial to
add or remove spans with the function decorators, but I was trying to err
on the side of too much versus too little.

cachito/web/api_v1.py
```diff
@@ -153,6 +156,8 @@ def get_requests():
     return flask.jsonify(response)
 
 
+@tracer.start_as_current_span("get_request")
+@api_v1.route("/requests/<int:request_id>", methods=["GET"])
```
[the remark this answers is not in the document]
mike-kingsbury: Removed; I must have accidently brought them back during a rebase.

cachito/web/app.py
```diff
@@ -134,6 +146,47 @@ def create_app(config_obj=None):
...
+    # Undefined; send data to the console.
+    else:
+        print("Configuring ConsoleSpanExporter")
+        processor = BatchSpanProcessor(ConsoleSpanExporter())
+
+    # Toggle between sending to jaeger and displaying span info on console
+    provider.add_span_processor(processor)
+    trace.set_tracer_provider(provider)
+
+    FlaskInstrumentor().instrument_app(app)
```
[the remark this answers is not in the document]
mike-kingsbury: It should be fairly trivial, sure, we can exclude them.

cachito/workers/pkg_managers/pip.py
```diff
@@ -78,41 +80,43 @@ def get_pip_metadata(package_dir):
     name = None
     version = None
 
-    setup_py = SetupPY(package_dir)
-    setup_cfg = SetupCFG(package_dir)
+    with tracer.start_as_current_span("get_pip_metadata"):
```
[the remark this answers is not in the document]
mike-kingsbury: Yes!

brunoapimentel:
> > Is there an easy way to find the trace corresponding to a specific request id? I'm thinking it could be useful if someone wants to use the traces for troubleshooting.
> > Maybe we could log the trace id in the `POST /requests` endpoint, or maybe change the span title itself for something like `cachito-api: /api/v1/requests/<actual_request_id>`.
> 
> Logging the trace ID in the create_request endpoint looks like this; searching for it in Jaeger pops it right up: cachito-api_1 | [2023-10-04 14:46:51 cachito.web.app INFO api_v1.create_request] An anonymous user submitted request 99; trace_id: 185d4b4d96b5d000f452d0e295f624a0
> 
> We don't want to post the actual ID in the span title as it will prevent possible aggregate functions from working properly. There are tags, more or less a custom label, that can be applied, it would probably be best to drop the request ID in there....

Yeah, I was testing it with tags, but I felt that the info can be easily lost if there's more than a few of those. It would also be only in one specific span, right? Unless we add it to all spans, which is overwork.

I think the log is a good enough solution for now. We can improve later if needed.

cachito/web/api_v1.py
```diff
@@ -342,12 +348,19 @@ def create_request():
     cachito_metrics["gauge_state"].labels(state="total").inc()
     cachito_metrics["gauge_state"].labels(state=request.state.state_name).inc()
 
+    ctx = trace.get_current_span().get_span_context()
+    trace_id = '{trace:032x}'.format(trace=ctx.trace_id)
```
chmeliik, line 352: For my understanding: trace IDs are (up to) 128-bit numbers, so we use 32 hexadecimals to represent them? With 0-padding to make sure the string is always exactly 32 characters?

(could you add a comment to explain it in the code?)

cachito/web/app.py
```diff
@@ -134,6 +145,49 @@ def create_app(config_obj=None):
 
     init_metrics(app)
 
+    # Some of the following has already been executed due to the manner in which
+    # the tasks config is included....
+
+    service_name = "cachito-api"
+    print("CACHITO-API SERVICE NAME")
```
chmeliik, line 152: We will want the prints removed or replaced with logging before this is merged

cachito/workers/__init__.py
```diff
@@ -29,22 +32,22 @@ def run_cmd(cmd, params, exc_msg=None):
     params.setdefault("capture_output", True)
     params.setdefault("universal_newlines", True)
     params.setdefault("encoding", "utf-8")
+    with tracer.start_as_current_span("running cmd " + "".join(cmd)):
```
chmeliik, line 35: Is the command intentionally whitespace-less?

> running cmd gomoddownload-json

cachito/workers/config.py
```diff
@@ -2,15 +2,53 @@
...
 
 from cachito.errors import ConfigError
 
 ARCHIVES_VOLUME = os.path.join(tempfile.gettempdir(), "cachito-archives")
 
+RequestsInstrumentor().instrument()
+
+
+@worker_process_init.connect(weak=False)
+def init_celery_tracing(*args, **kwargs):
```
chmeliik, line 27: We generally do these sort of celery hooks in https://github.com/containerbuildsystem/cachito/blob/13876fe66c0a53562bb4bbe8af22bb22290ec4f7/cachito/workers/tasks/celery.py#L24-L28

Maybe that would also solve

>     # Some of the following has already been executed due to the manner in which
>     # the tasks config is included....

?

cachito/workers/config.py
```diff
@@ -38,6 +76,7 @@ class Config(object):
     cachito_log_level = "INFO"
     cachito_js_concurrency_limit = 5
     cachito_js_download_max_tries = 5
+    cachito_js_download_batch_size = 30
```
chmeliik, line 79: Is this change intentional?

cachito/workers/config.py
```diff
@@ -102,11 +141,17 @@ class Config(object):
     # other tasks aren't starved when processing a large archive.
     worker_prefetch_multiplier = 1
 
+    cachito_jaeger_exporter_endpoint: Optional[str] = ""
+    cachito_jaeger_exporter_port: Optional[int]
+    cachito_otlp_exporter_endpoint: Optional[str] = ""
```
chmeliik, line 146: We try to keep cachito-specific configuration above the Celery configuration in the Config class (the options that don't start with cachito_ are Celery opts)

cachito/workers/config.py
```diff
@@ -2,15 +2,53 @@
...
 
+RequestsInstrumentor().instrument()
+
+
+@worker_process_init.connect(weak=False)
+def init_celery_tracing(*args, **kwargs):
+    """Initialize OTLP tracing, set the processor & endpoint."""
+    CeleryInstrumentor().instrument()
+    config = get_config()
+    if config.cachito_jaeger_exporter_endpoint:
```
chmeliik, line 31: With the way how the Config classes currently work, this will fail with an AttributeError if:
- the config is not DevelopmentConfig
- the config is DevelopmentConfig but CACHITO_JAEGER_EXPORTER_ENDPOINT is not set
- the config is DevelopmentConfig, CACHITO_JAEGER_EXPORTER_ENDPOINT is set, but CACHITO_OTLP_EXPORTER_ENDPOINT is also set

(I think the fix needs to be done in the Config class, just illustrating the problem here)

cachito/workers/config.py
```diff
@@ -140,6 +185,16 @@ class DevelopmentConfig(Config):
...
+    if not os.environ.get("CACHITO_OTLP_EXPORTER_ENDPOINT"):
+        if os.environ.get("CACHITO_JAEGER_EXPORTER_ENDPOINT"):
+            print("Jaeger endpoint defined")
+            cachito_jaeger_exporter_endpoint = os.environ.get("CACHITO_JAEGER_EXPORTER_ENDPOINT")
+            cachito_jaeger_exporter_port = cast(int, os.environ.get("CACHITO_JAEGER_EXPORTER_PORT"))
+    else:
+        print(
+            "OTLP endpoint defined: " + cast(str, os.environ.get("CACHITO_OTLP_EXPORTER_ENDPOINT"))
+        )
+        cachito_otlp_exporter_endpoint = os.environ.get("CACHITO_OTLP_EXPORTER_ENDPOINT")
```
chmeliik, line 197: Setting the attributes conditionally will cause the issue ilustrated in the comment above

I think all three of these need to be set in the base Config class and need to be set always

```python
cachito_jaeger_exporter_endpoint = os.environ.get("CACHITO_JAEGER_EXPORTER_ENDPOINT")
cachito_jaeger_exporter_port = os.environ.get("CACHITO_JAEGER_EXPORTER_PORT")
cachito_otlp_exporter_endpoint = os.environ.get("CACHITO_OTLP_EXPORTER_ENDPOINT")
```

I'm also a little nervous about reading environment variables at _import time_ (the code gets executed as soon as some part of Cachito imports cachito.workers.config). Is there a reason why these should be environment variables rather than being set directly in the code or in the config file (`prod_config_file_path`)? That's how configuration is usually handled.

cachito/workers/config.py
```diff
@@ -140,6 +185,16 @@ class DevelopmentConfig(Config):
     }
     cachito_request_file_logs_dir: Optional[str] = "/var/log/cachito/requests"
     cachito_sources_dir = os.path.join(ARCHIVES_VOLUME, "sources")
+    if not os.environ.get("CACHITO_OTLP_EXPORTER_ENDPOINT"):
+        if os.environ.get("CACHITO_JAEGER_EXPORTER_ENDPOINT"):
+            print("Jaeger endpoint defined")
+            cachito_jaeger_exporter_endpoint = os.environ.get("CACHITO_JAEGER_EXPORTER_ENDPOINT")
+            cachito_jaeger_exporter_port = cast(int, os.environ.get("CACHITO_JAEGER_EXPORTER_PORT"))
```
chmeliik, line 192: `cast()` is a type-checker assertion, it doesn't do anything at runtime

if the port really does need to be an int, you'll need to convert it with `int()` (but only if the environment variable is set, otherwise that's a `TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'`

chmeliik: Ah, it gets converted later in

```python
        jaeger_exporter = JaegerExporter(
            agent_host_name=config.cachito_jaeger_exporter_endpoint,
            agent_port=int(config.cachito_jaeger_exporter_port),
        )
```

IMO that's the right place to do the conversion, and it should be a string in the Config class. (If we're using env vars to set the config. If we set the variable directly in the code, it can be an int)

chmeliik:
Are the three commits meant to be logically separate? If yes, could you add commit messages to explain what each one does? If not, could you squash them (and also explain more in the commit message)>

mike-kingsbury:
> Are the three commits meant to be logically separate? If yes, could you add commit messages to explain what each one does? If not, could you squash them (and also explain more in the commit message)?

Squashed!

docker-compose.yml
```diff
@@ -114,6 +119,21 @@ services:
...
+    ports:
+      - "4317:4317"
+      - "5775:5775/udp"
+      - "6831:6831/udp"
+      - "6832:6832/udp"
+      - "5778:5778"
+      - "16686:16686"
+      - "14268:14268"
+      - "14250:14250"
+      - "9411:9411"
```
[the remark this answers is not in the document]
chmeliik: Removing the unused ones works too, yeah :smile: 

But I think integration tests were using 4317

docker-compose.yml
```diff
@@ -96,8 +96,6 @@ services:
       python3 -c 'from cachito.workers.nexus import create_or_update_scripts; create_or_update_scripts()' &&
       watchmedo auto-restart -d ./cachito/workers -p '*.py' --recursive \
         -- celery -A cachito.workers.tasks worker --loglevel=info
-    environment:
-      CACHITO_DEV: 'true'
```
chmeliik, line 100: I think we still need this one

tox.ini
```diff
@@ -86,7 +86,12 @@ basepython = python3
...
+    pytest -rA -vvvv --export-traces \
         --confcutdir=tests/integration \
         {posargs:tests/integration}
-passenv = KRB5CCNAME, REQUESTS_CA_BUNDLE, KRB5_CLIENT_KTNAME, CACHITO_TEST_CERT, CACHITO_TEST_KEY, JOB_NAME
+passenv = KRB5CCNAME,REQUESTS_CA_BUNDLE,KRB5_CLIENT_KTNAME,CACHITO_TEST_CERT\
+          CACHITO_TEST_KEY,JOB_NAME
+setenv =
+    OTEL_RESOURCE_ATTRIBUTES=service.name=cachito-integration-tests
+    OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317
+
```
brunoapimentel, line 97: I think this file now has two trailing whitelines.

cachito/workers/__init__.py
```diff
@@ -29,22 +32,22 @@ def run_cmd(cmd, params, exc_msg=None):
     params.setdefault("capture_output", True)
     params.setdefault("universal_newlines", True)
     params.setdefault("encoding", "utf-8")
+    with tracer.start_as_current_span("running cmd " + " ".join(cmd)):
```
brunoapimentel, line 35: The reason we're doing this inline instead of using the decorator is so that we can also track the actual `cmd`? I think it'd be better to avoid doing inline operations as much as we can, but if there's no way to capture this info, I'm fine with this.

cachito/workers/tasks/celery.py
```diff
@@ -2,15 +2,45 @@
...
     cleanup_task_logging,
     cleanup_task_logging_customization,
     setup_task_logging,
     setup_task_logging_customization,
 )
-from cachito.workers.config import app, validate_celery_config  # noqa: F401
+from cachito.workers.config import app, get_worker_config, validate_celery_config  # noqa: F401
+
+
+def init_celery_tracing(*args, **kwargs):
```
brunoapimentel, line 23: Nitpick: since this will only be used in the same module, we could make it private.

cachito/workers/tasks/gomod.py
```diff
@@ -26,6 +28,8 @@
 __all__ = ["fetch_gomod_source"]
 log = logging.getLogger(__name__)
 
+tracer = trace.get_tracer(__name__)
+
```
brunoapimentel, line 32: I think this import here is a leftover.

brunoapimentel:
Did we intentionally skip the `pkg_managers.yarn`? I think it's worth tracing at least `resolve_yarn`, maybe a couple of other functions.

tox.ini
chmeliik: And we probably need to re-expose port 4317 https://github.com/containerbuildsystem/cachito/pull/918#discussion_r1358350953

mike-kingsbury:
> Did we intentionally skip the `pkg_managers.yarn`? I think it's worth tracing at least `resolve_yarn`, maybe a couple of other functions.

🤦  Done.

PR #951, opened 2023-12-05
Add new API endpoint requests/latest

taylormadore:
The requests/latest endpoint will return the most recent request for a given repo_name and git ref

### DIFF
--- a/cachito/web/api_v1.py
+++ b/cachito/web/api_v1.py
@@ -6,7 +6,7 @@ import tempfile
 from collections import OrderedDict
 from copy import deepcopy
 from datetime import date, datetime
-from typing import Any, Dict, List, Set, Union
+from typing import Any, Dict, List, Optional, Set, Union, cast
 
 import flask
 import kombu.exceptions
@@ -15,14 +15,14 @@ from celery import chain
 from flask import stream_with_context
 from flask_login import current_user, login_required
 from opentelemetry import trace
-from sqlalchemy import and_, func
+from sqlalchemy import and_, desc, func, or_
 from sqlalchemy.orm import joinedload, load_only
 from werkzeug.exceptions import BadRequest, Forbidden, Gone, InternalServerError, NotFound
 
 from cachito.common.checksum import hash_file
 from cachito.common.packages_data import PackagesData
 from cachito.common.paths import RequestBundleDir
-from cachito.common.utils import b64encode
+from cachito.common.utils import b64encode, get_repo_name
 from cachito.errors import MessageBrokerError, NoWorkers, RequestErrorOrigin, ValidationError
 from cachito.web import db
 from cachito.web.content_manifest import BASE_ICM, BASE_SBOM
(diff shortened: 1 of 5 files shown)


cachito/web/api_v1.py
```diff
@@ -184,6 +184,70 @@ def get_request(request_id):
...
+    # Check the first result for a matching repo_name
+    first_request = query.first()
+    if first_request and get_repo_name(first_request.repo) == repo_name:
+        return first_request
+
+    # Fall back to iterating over the full result set, loading it in chunks,
+    # looking for a matching repo_name
+    for request in query.yield_per(10):
+        if get_repo_name(request.repo) == repo_name:
+            return request
```
eskultety, line 246: There's clearly something I'm missing here about those requests, so please clarify the following to me:
- we filter requests by a given commit ref
- we then filter those results to limit the set to only those which are related to `repo_name` 
- finally we order the set by highest request ID

So far so good. But why do we need to check  `get_repo_name(first_request.repo) == repo_name` again? Doesn't the SQL filter take care of that already?
Is it because of git protocol in the URI? Or is it because e.g. `"repo": "https://github.com/git/org/eggs"` and `"repo": "https://github.com/org/eggs"` are 2 different repos?

Also, how slow is loading those requests by pages of 10 results? I'm only asking since you're special-handling the first entry and then falling back to iterating over the rest of the set instead of just iterating it right away. What's the performance gain here on this API? Do we plan on running it that often instead only occasionally in a cron job?

cachito/web/api_v1.py
```diff
@@ -184,6 +184,70 @@ def get_request(request_id):
     return flask.jsonify(json)
 
 
+def get_latest_request() -> flask.Response:
+    """
+    Retrieve the latest request for a repo_name/ref.
```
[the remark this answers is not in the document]
eskultety: IIUC this one's just the public API entrypoint to the private "backend" handler which this one wraps with JSON.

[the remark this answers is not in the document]
ben-alkov: I guess my point was that we shouldn't have two different methods with identical docstrings.

cachito/web/api_v1.py
taylormadore: > Also, how slow is loading those requests by pages of 10 results? I'm only asking since you're special-handling the first entry and then falling back to iterating over the rest of the set instead of just iterating it right away. What's the performance gain here on this API? Do we plan on running it that often instead only occasionally in a cron job?

I expect that the initial query will cover nearly all requests and that the fall-back will just handle the odd edge-case. We'll probably just be running this as-needed and when it comes time to write the script, my plan was to rate-limit the requests anyway to avoid hammering the API for a non-critical maintenance task.

cachito/web/api_v1.py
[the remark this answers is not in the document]
taylormadore: I can update the docstring to make it more clear :+1:

cachito/web/api_v1.py
```diff
@@ -184,6 +184,70 @@ def get_request(request_id):
...
+    # mypy: Connexion ensures that these cannot be None
+    repo_name = cast(str, flask.request.args.get("repo_name"))
+    ref = cast(str, flask.request.args.get("ref"))
+
+    request = _get_latest_request_by_repo_name_and_ref(repo_name, ref)
+
+    if not request:
+        raise NotFound
+
+    return flask.jsonify(request.to_json())
```
brunoapimentel, line 206: If we don't set `request.to_json(verbose=False)`, it will return the fully resolved request (with packages and deps). I think we want the shorter version, right?

cachito/web/static/api_v1.yaml
```diff
@@ -341,6 +341,49 @@ paths:
...
+  "/requests/latest":
+    get:
+      operationId: cachito.web.api_v1.get_latest_request
+      summary: Get the latest request for a repo/ref
+      description: Return the latest request for a specified repo_name and ref
+      parameters:
+      - name: repo_name
+        required: true
+        in: query
+        description: The namespaced repository name (namespace/name)
```
brunoapimentel, line 353: I know it's described here, but :stupid: me took several minutes to figure out the right format for the `repo_name`. I ended up using the `cachito/common/utils.py::get_repo_name` function to figure it out.

Wdyt about adding a note here such as "The domain and protocol/scheme must not be present in this param (e.g. https://github.com/my-org/my repo translates to my-org/my-repo)" or something along these lines.

brunoapimentel (approved):
LGTM

ejegrova (approved):

eskultety (approved):

ben-alkov (approved):

PR #956, opened 2023-12-11
toil(reqs): pin connexion and pydantic to silence Dependabot...

ben-alkov:
...and regenerate requirements*.txt files to get out in front of DpB, at least a little.

This will also allow me to perma-close several Dependabot PRs (for instance, "greenlet", which DpB just opened a PR for)


eskultety:
If there's a benefit for the project itself then I think we can do it this way, although I'd extract most of the regenerated requirements to a separate patch to keep the changes more logically co-located. However, if this is solely about DpB annoyingly and persistently opening PRs for the affected deps maybe the better fix would be to mark those deps explicitly ignored by DpB which will also make it close them.

ben-alkov:
> If there's a benefit for the project itself then I think we can do it this way, although I'd extract most of the regenerated requirements to a separate patch to keep the changes more logically co-located. However, if this is solely about DpB annoyingly and persistently opening PRs for the affected deps maybe the better fix would be to mark those deps explicitly ignored by DpB which will also make it close them.

It's both, and they're intimately related

- I've pinned two things - connexion and pydantic - which **must not** be bumped without major code rework
- pydantic actually has an update (1.10.7 -> 1.10.13) which DpB is ignoring in favor of >2.0, so regenerating the reqs was *necessary*
- While regenerated the reqs, it seemed like a good idea to do it from scratch, updating everything (i.e. deleting all 'requirements*.txt' files and re-running e.g. `pip-compile --allow-unsafe --generate-hashes --output-file=requirements.txt requirements.in`), so that we're up-to-date on **all** *unpinned* reqs (and in fact DPB opened a PR for greenlet (which this PR updates) while I was working on this)

Also, I don't see the point of pinning reqs in one PR, and then actually regenerating the reqs in another - it makes no sense. Separate *commits*, maybe, but not separate PRs

eskultety:
> Also, I don't see the point of pinning reqs in one PR, and then actually regenerating the reqs in another - it makes no sense. Separate _commits_, maybe, but not separate PRs

That's exactly what I mean when I said 'patch' (i.e. not a separate PR):

> I'd extract most of the regenerated requirements to a  separate patch.

my bad for causing this bad wording confusion.

> I've pinned two things - connexion and pydantic - which must not be bumped without major code rework

I always understood pinning as a way of making things reproducible or to make sure we'd get **at least** <ver> because of some bug or CVE.

> While regenerated the reqs, it seemed like a good idea to do it from scratch, updating everything (i.e. deleting all 'requirements*.txt' files and re-running e.g. pip-compile --allow-unsafe --generate-hashes --output-file=requirements.txt requirements.in), so that we're up-to-date on all unpinned reqs (and in fact DPB opened a PR for greenlet (which this PR updates) while I was working on this)

Sure thing, the only point I was trying to make was - if this is solely about silencing DpB, then a dependabot-specific change is likely more appropriate rather than pinning, but thinking about this some more, if major code reword would be needed on our side, then you're right we should do it via req version pinning. However, like I said I'd still separate changes unrelated to connexion and pydantic to an individual commit and then pin those 2 along with any related regenerated req changes.

brunoapimentel:
> Sure thing, the only point I was trying to make was - if this is solely about silencing DpB, then a dependabot-specific change is likely more appropriate rather than pinning, but thinking about this some more, if major code reword would be needed on our side, then you're right we should do it via req version pinning. 

This was exactly my first though: let's just make DpB ignore the major versions for Connexion and Pydantic, instead of doing code changes. But on second thought, the project will stay in `<3` and `<2`, so maybe it's best to just declare it in the requirements file.

ben-alkov:
I'm more than happy to break it up into more commits; thanks for the reviews, guys.

slimreaper35:
I manually bumped Pydantic to V2 in _cachi2_. It was fun :harold:

taylormadore (approved):

PR #967, opened 2024-01-23
Add script for pruning stale source archives

taylormadore:
This PR adds a script that can list and (potentially) delete source archives older than a certain threshold. 

### DIFF
--- a/cachito/web/models.py
+++ b/cachito/web/models.py
@@ -406,12 +406,16 @@ class Request(db.Model):  # type: ignore[name-defined]
         if self.user:
             user = self.user.username
 
+        created = None
+        if self.created is not None:
+            created = self.created.isoformat(timespec="microseconds")
+
         env_vars_json = OrderedDict()
         for env_var in self.environment_variables:
             env_vars_json[env_var.name] = env_var.value
         rv = {
             "id": self.id,
-            "created": None if self.created is None else self.created.isoformat(),
+            "created": created,
             "repo": self.repo,
             "ref": self.ref,
             "pkg_managers": pkg_managers,
@@ -428,7 +432,7 @@ class Request(db.Model):  # type: ignore[name-defined]
             return {
                 "state": RequestStateMapping(state.state).name,
                 "state_reason": state.state_reason,
-                "updated": state.updated.isoformat(),
+                "updated": state.updated.isoformat(timespec="microseconds"),
             }
 
         def _error_to_json(error):
@@ -776,7 +780,7 @@ class RequestError(db.Model):  # type: ignore[name-defined]
             "origin": self.origin,
             "error_type": self.error_type,
             "message": self.message,
-            "occurred": self.occurred.isoformat(),
+            "occurred": self.occurred.isoformat(timespec="microseconds"),
         }
 
     @classmethod
--- a/cachito/workers/prune_archives.py
+++ b/cachito/workers/prune_archives.py
@@ -1,0 +1,233 @@
+import logging
+import re
+from dataclasses import dataclass
+from datetime import datetime, timedelta, timezone
+from itertools import islice
+from pathlib import Path
+from typing import Annotated, Any, Generator, NamedTuple, Optional
+
+import requests
+import typer
+from ratelimit import limits, sleep_and_retry
+
+from cachito.errors import NetworkError
+from cachito.workers.config import get_worker_config
+from cachito.workers.requests import get_requests_session
+
+app = typer.Typer()
+config = get_worker_config()
+log = logging.getLogger(__name__)
+session = get_requests_session()
+
+ARCHIVE_DIR = Path(config.cachito_sources_dir)
+ARCHIVE_PATTERN = re.compile(r"^[a-f0-9]{40}(-with-submodules)?\.tar\.gz$")
+DEFAULT_AGE_DATETIME = datetime.now(timezone.utc) - timedelta(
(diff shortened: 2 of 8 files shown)


cachito/workers/prune_archives.py
```diff
@@ -0,0 +1,217 @@
...
+            f"Archive {archive.path} is stale. The most recent request_id="
+            f"{archive.latest_request_id} at {archive.created}"
+        )
+        if execute:
+            log.error(f"Deleting {archive.path}")
+            archive.path.unlink()
+
+
+@app.command()
+def list(
```
tildebyte, line 192: You're shadowing builtin `list`.

cachito/workers/prune_archives.py
```diff
@@ -0,0 +1,217 @@
...
+    api_calls_per_second: Annotated[
+        int, typer.Option(min=1, max=5, help="The API requests-per-second limit.")
+    ] = 1,
+    limit: Annotated[
+        Optional[int], typer.Option(min=1, help="The maximum number of stale archives to process.")
+    ] = None,
+    execute: Annotated[bool, typer.Option(help="Actual deletion will only occur if True.")] = False,
+):
+    """
+    List and delete stale source archives.
```
tildebyte, line 177: Does it make more sense for `delete()` to call `list()`, rather than duplicating the code?

taylormadore: Reworked this a bit in the final. Let me know if the changes work for you :+1:

cachito/workers/prune_archives.py
taylormadore: Fixed

ben-alkov (approved):

cachito/workers/prune_archives.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+@app.command("delete")
+def delete_archives(
+    older_than: Annotated[
+        Optional[datetime],
+        typer.Option(
+            callback=_validate_older_than,
+            formats=["%Y-%m-%d"],
+            help="Deletes archives that are older than the specified date (UTC). YYYY-MM-DD",
```
eskultety, line 184: **nitpick:** 
> "Deletes archives that are older than the specified date (UTC)

1.  We should never ask end users to enter anything in UTC, because nobody wants to do the calculation in their head, we need to do the conversions automatically
2. mentioning UTC in context of plain dates might lead to the following question "Uhm, so I want to prune everything older than <date> in my timezone, but <date> is always `T00:00` which means that the UTC date would look differently, so which value do I put in there???" I mean this is pure bikeshedding and for the purpose of this script it doesn't matter at all, just something to be aware of.
3. using `replace()` on `datetime` instances doesn't adjust the time values underneath [1], it just forces a new field to the object meaning that you may have filtered out some values out of the evaluation, but here again, not a big deal in this particular case, it just brings questions to however reads the code.

For the sake of argument what I would suggest is to drop the UTC part from the help message, treat all inputs as local timezone to a particular system and then do
```
datetime.datetime.strptime("<timespec>", <format>).astimezone(datetime.timezone.utc)
```
in all related places to make sure we represent and base everything correctly.

[1] https://docs.python.org/3/library/datetime.html#datetime.datetime.replace

cachito/web/models.py
```diff
@@ -411,7 +411,9 @@ def to_json(self, verbose=True):
             env_vars_json[env_var.name] = env_var.value
         rv = {
             "id": self.id,
-            "created": None if self.created is None else self.created.isoformat(),
+            "created": None
+            if self.created is None
+            else self.created.isoformat(timespec="microseconds"),
```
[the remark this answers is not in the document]
taylormadore: Fixed

cachito/workers/prune_archives.py
taylormadore: Fixed :+1: 

User inputs should now be localtime and are converted to UTC

cachito/workers/prune_archives.py
```diff
@@ -0,0 +1,233 @@
...
+def _resolve_source_archive(parsed_archive: _ParsedArchive) -> Optional[_ResolvedArchive]:
+    """Return a _ResolvedArchive if a matching request is found via the API."""
+    latest_request = _get_latest_request(parsed_archive)
+    if latest_request is None:
+        log.debug("Archive %s could not be resolved via the API.", parsed_archive.path)
+        return None
+
+    return _ResolvedArchive(
+        parsed_archive.path,
+        datetime.strptime(latest_request["created"], "%Y-%m-%dT%H:%M:%S.%f").replace(
```
eskultety, line 122: Are the request timestamps already UTC adjusted or do we need to use the same treatment here as with user CLI input? I'm just curious since I don't see a 'Z' in the unit test timestamps and I'm too lazy to go and check the deployment for real data :).
It's still a nitpick though, so even if it weren't no big deal since I doubt the time shift would make a significant difference in terms of archive cleanup anyway.

eskultety (approved):

cachito/workers/prune_archives.py
taylormadore: Yeah, the request timestamps returned from the DB all should be UTC already. They're the values formatted by datetime.isoformat() in the first commit here

eskultety: @taylormadore can you specifically request review from me so that my approval counts?

eskultety (approved):

PR #978, opened 2024-02-12
Add preliminary Go 1.21 support

eskultety:
This PR just deals with the necessary bits to get projects building with 1.21. Proper 1.21 support including support for the toolchain will be proposed in the future.

This PR is based on https://github.com/containerbuildsystem/cachi2/pull/427 from which much of the code was ported with one notable exception -  as proposed cachito won't download the 1.20 toolchain locally like cachi2 does in case it's missing. The reason for that is that only the container environment use case is considered and hence the worker images will have 1.20 baked into them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -747,7 +747,7 @@ The table below shows the supported package managers and their support level in 
 
 Tool     | Version |
 ---      |---------|
-Go       | 1.20.7  |
+Go*      | 1.20.7, 1.21.0  |
 Npm      | 9.5.0   |
 Node     | 18.16.1 |
 Pip      | 22.3.1  |
@@ -761,6 +761,12 @@ Bundler* | 2.x     |
 * Cachito does not use the Ruby runtime (no ruby is interpreted from `Gemfile`s). 
   The processing of Gemfile.lock files is handled by
   [gemlock-parser](https://github.com/containerbuildsystem/gemlock-parser).
+* Starting with Go 1.21 Go changed the meaning of the `go` directive in `go.mod` file slightly and
+  made the constraint stricter in that the line now denotes the **minimum required** version of Go
+  instead of a suggested version of Go. If a project recommending an older version of Go is
+  processed with Go >=1.21 it might happen (based on other dependencies) that its own required
+  version of Go will be bumped to 1.21+, hence dirtying the git repo - to prevent this cachito
+  uses two releases of Go SDK concurrently.
 
 ### gomod
 
--- a/cachito/workers/pkg_managers/gomod.py
+++ b/cachito/workers/pkg_managers/gomod.py
@@ -312,6 +312,7 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
             "GOPROXY": f"{athens_url}|{athens_url}",
             "PATH": os.environ.get("PATH", ""),
             "GOMODCACHE": "{}/pkg/mod".format(temp_dir),
+            "GOTOOLCHAIN": "local",
         }
         if "cgo-disable" in request.get("flags", []):
             env["CGO_ENABLED"] = "0"
--- a/cachito/workers/pkg_managers/gomod.py
+++ b/cachito/workers/pkg_managers/gomod.py
@@ -10,7 +10,7 @@ import tempfile
 from datetime import datetime
 from itertools import chain
 from pathlib import Path, PureWindowsPath
-from typing import Any, Dict, Iterable, Iterator, List, Literal, Optional, Tuple, Union
+from typing import Any, Iterable, Iterator, List, Literal, Optional, Tuple, Union
 
 import backoff
 import git
@@ -44,7 +44,6 @@ __all__ = [
 
 log = logging.getLogger(__name__)
 tracer = trace.get_tracer(__name__)
-run_gomod_cmd = functools.partial(run_cmd, exc_msg="Processing gomod dependencies failed")
 
 MODULE_VERSION_RE = re.compile(r"/v\d+$")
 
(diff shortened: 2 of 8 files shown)


eskultety:
> All of the errors in the integration tests look related to the pathlib changes in python 3.12, which is included with fedora 39 🥲

Sigh. What if we install and fallback to an older version of Python in the container image for now and fix it properly in a standalone PR?

taylormadore:
This all LGTM :rocket: 

I've run a few tests locally and was able to build projects from the output with both go < 1.21 and 1.21. I think updating the README is the only thing left to do

taylormadore (approved):

eskultety:
> > All of the errors in the integration tests look related to the pathlib changes in python 3.12, which is included with fedora 39 🥲
> 
> Sigh. What if we install and fallback to an older version of Python in the container image for now and fix it properly in a standalone PR?

FWIW ^this turned out to cause more problems we'll have to figure out in the foreseeable future, but for the purposes of this PR decided to drop bumping to F39 just yet, so this comment can be considered resolved.

brunoapimentel:
Just finished my first read, approach and execution look amazing, Erik :rocket::rocket::rocket:

I will dive deeper and hopefully I can give you my final ack today.

eskultety:
@taylormadore @brunoapimentel do we need something specific about Go 1.21.0 to the README apart from adding it to the versions of tools? I mean the Go contents is already pretty generic and adding some 1.21 specific bits doesn't seem like it would fit.

taylormadore:
> do we need something specific about Go 1.21.0 to the README apart from adding it to the versions of tools?

I think that's fine :+1:

cachito/workers/pkg_managers/gomod.py
```diff
@@ -91,38 +94,118 @@ class GoPackage(_GolangModel):
...
+        self._release = release
+
+        if self._release:
+            self._bin = f"/usr/local/go/{release}/bin/go"
+
+        self._version: Optional[pkgver.Version] = None
+
+    @tracer.start_as_current_span("Go.__call__")
+    def __call__(self, cmd: list[str], params: Optional[dict] = None, retry: bool = False) -> str:
+        """Run a Go command using the underlying toolchain, same as running GoToolchain()().
```
brunoapimentel, line 126: Should this be "... same as running Go()()"?

cachito/workers/pkg_managers/gomod.py
```diff
@@ -228,12 +312,47 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
...
             env["CGO_ENABLED"] = "0"
 
         run_params = {"env": env, "cwd": app_source_path}
 
+        go = Go()
+        go1_21 = pkgver.Version("1.21")
+        go_base_version = go.version
+        go_mod_version_msg = "go.mod recommends/requires Go version: {}"
+
+        if (_ := _get_gomod_version(git_dir_path)) is None:
```
brunoapimentel, line 327: Very nitpickish comment, but I'd rather that this `_` variable had a name, since it's being used outside of the if clause context.

README.md
```diff
@@ -747,7 +747,7 @@ The table below shows the supported package managers and their support level in
 
 Tool     | Version |
 ---      |---------|
-Go       | 1.20.7  |
+Go       | 1.20.7, 1.21.0  |
```
brunoapimentel, line 750: Maybe this line also deserves a `*` explaining why we have two Go versions?

cachito/workers/pkg_managers/gomod.py
```diff
@@ -91,38 +94,118 @@ class GoPackage(_GolangModel):
...
+            self._version = pkgver.Version(self.release[2:])
+        return self._version
+
+    @property
+    def release(self) -> str:
+        """Release name of the Go Toolchain, e.g. go1.20 ."""
+        # lazy evaluation: defer running 'go'
+        if not self._release:
+            output = self(["version"])
+            log.info(f"Go release: {output}")
```
brunoapimentel, line 155: This will usually generate a log very similar to what this [line](https://github.com/containerbuildsystem/cachito/blob/master/cachito/workers/tasks/gomod.py#L133) does:

```
INFO Go version: go version go1.20.13 linux/amd64
(...)
INFO Running '['go', 'version']'
INFO Go release: go version go1.20.13 linux/amd64
INFO go.mod recommends/requires Go version: 1.21
(...)
```

I think it would be a little clearer if the logs were something such as:
```
INFO go.mod recommends/requires Go version: 1.21
INFO Processing request using go version go1.21.0 linux/amd64
```

Or just remove the original line in the tasks.gomod.

tests/test_workers/test_pkg_managers/test_gomod.py
```diff
@@ -239,6 +243,8 @@ def test_resolve_gomod_vendor_dependencies(
     mock_run.side_effect = run_side_effects
 
     mock_golang_version.return_value = "v0.1.0"
+    mock_go_release.return_value = "go0.1.0"
+    mock_get_gomod_version.return_value = "0.1.1"
```
brunoapimentel, line 247: If we parametrized these values, we could test the algorithm that chooses the go binary that will be used to process the request (or maybe split this into a separate test?). The coveralls report is broken, but see the following uncovered lines in pkg_managers.gomod: L332-333, L341 and L354.

brunoapimentel (approved):
LGTM! 

I left a few comments, but they're mostly nitpicks/improvement suggestions, so feel free to ignore them and merge the PR.

tests/test_workers/test_pkg_managers/test_gomod.py
eskultety: trying to adjust the existing `resolve_gomod` test by parametrizing this and figuring out whether we used the right version would be non-trivial, so instead I extracted the logic to a new helper and tested the helper instead as suggested.

eskultety:
@brunoapimentel do you agree with the changes?

brunoapimentel:
> @brunoapimentel do you agree with the changes?

Looks great! Thanks Erik.

PR #980, opened 2024-02-19
Fix go module not in root repo

ejegrova:
A bug was introduced in the 1.17 release. We were relying on the git repo path instead of the actual module path, and this caused an issue when the module is not in the repo's root.

 JIRA: STONEBLD-2191

integration test repo: https://github.com/cachito-testing/gomod-multiple-modules/tree/no_root_go_mod

- [n/a]] DB schema change has corresponding DB migration (if applicable)
- [n/a]] README updated (if worker configuration changed, or if applicable)
- [n/a]] Draft release notes are updated before merging

### DIFF
--- a/cachito/workers/pkg_managers/gomod.py
+++ b/cachito/workers/pkg_managers/gomod.py
@@ -285,7 +285,7 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
     """
     Resolve and fetch gomod dependencies for given app source archive.
 
-    :param str app_source_path: the full path to the application source code
+    :param Path app_source_path: the full path to the application source code
     :param dict request: the Cachito request this is for
     :param list dep_replacements: dependency replacements with the keys "name" and "version"; this
         results in a series of `go mod edit -replace` commands
@@ -319,7 +319,7 @@ def resolve_gomod(app_source_path, request, dep_replacements=None, git_dir_path=
 
         run_params = {"env": env, "cwd": app_source_path}
 
-        go = _select_go_toolchain(git_dir_path)
+        go = _select_go_toolchain(app_source_path)
 
         # Collect all the dependency names that are being replaced to later report which
         # dependencies were replaced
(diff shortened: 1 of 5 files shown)


eskultety:
We likely need this hunk too (I noticed we don't even check the generated env variables in our unit tests, otherwise this would have been caught already; the unit test update can IMO be a follow up)
```
diff --git a/cachito/workers/tasks/gomod.py b/cachito/workers/tasks/gomod.py
index 32b5f01..50036ab 100644
--- a/cachito/workers/tasks/gomod.py
+++ b/cachito/workers/tasks/gomod.py
@@ -166,6 +166,7 @@ def fetch_gomod_source(request_id, dep_replacements=None, package_configs=None):
         "GOCACHE": {"value": "deps/gomod", "kind": "path"},
         "GOPATH": {"value": "deps/gomod", "kind": "path"},
         "GOMODCACHE": {"value": "deps/gomod/pkg/mod", "kind": "path"},
+        "GOTOOLCHAIN": {"value": "local", "kind": "literal"},
     }
     env_vars.update(config.cachito_default_environment_variables.get("gomod", {}))
     update_request_env_vars(request_id, env_vars)
```

ejegrova:
> We likely need this hunk too (I noticed we don't even check the generated env variables in our unit tests, otherwise this would have been caught already; the unit test update can IMO be a follow up)
> 
> ```
> diff --git a/cachito/workers/tasks/gomod.py b/cachito/workers/tasks/gomod.py
> index 32b5f01..50036ab 100644
> --- a/cachito/workers/tasks/gomod.py
> +++ b/cachito/workers/tasks/gomod.py
> @@ -166,6 +166,7 @@ def fetch_gomod_source(request_id, dep_replacements=None, package_configs=None):
>          "GOCACHE": {"value": "deps/gomod", "kind": "path"},
>          "GOPATH": {"value": "deps/gomod", "kind": "path"},
>          "GOMODCACHE": {"value": "deps/gomod/pkg/mod", "kind": "path"},
> +        "GOTOOLCHAIN": {"value": "local", "kind": "literal"},
>      }
>      env_vars.update(config.cachito_default_environment_variables.get("gomod", {}))
>      update_request_env_vars(request_id, env_vars)
> ```

Just added this change

brunoapimentel (approved):

ben-alkov (approved):

PR #989, opened 2024-04-09
add pkg scheme to npm generic file PURLs

taylormadore:
The pkg scheme is required for each component according to the PURL specification.

- [n\a] New code has type annotations
- [n\a] OpenAPI schema is updated (if applicable)
- [n\a] DB schema change has corresponding DB migration (if applicable)
- [n\a] README updated (if worker configuration changed, or if applicable)

### DIFF
--- a/tests/integration/test_data/npm_packages.yaml
+++ b/tests/integration/test_data/npm_packages.yaml
@@ -782,22 +782,22 @@ workspaces:
   content_manifest:
   - purl: "pkg:github/cachito-testing/cachito-npm-workspaces@b4ec59868cb5667deb62930859762b107f23598c"
     dep_purls:
-      - "pkg:generic/bar?file%3Abar"
-      - "pkg:generic/eggs?file%3Aeggs-packages%2Feggs"
-      - "pkg:generic/foo?file%3Afoo"
-      - "pkg:generic/not-baz?file%3Abaz"
-      - "pkg:generic/spam?file%3Aspam-packages%2Fspam"
+      - "pkg:generic/bar?file=bar"
+      - "pkg:generic/eggs?file=eggs-packages%2Feggs"
+      - "pkg:generic/foo?file=foo"
+      - "pkg:generic/not-baz?file=baz"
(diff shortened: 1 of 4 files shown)


eskultety:
LGTM

tests/integration/test_data/npm_packages.yaml
```diff
@@ -782,22 +782,22 @@ workspaces:
...
-      - "generic/bar?file%3Abar"
-      - "generic/eggs?file%3Aeggs-packages%2Feggs"
-      - "generic/foo?file%3Afoo"
-      - "generic/not-baz?file%3Abaz"
-      - "generic/spam?file%3Aspam-packages%2Fspam"
+      - "pkg:generic/bar?file%3Abar"
+      - "pkg:generic/eggs?file%3Aeggs-packages%2Feggs"
+      - "pkg:generic/foo?file%3Afoo"
+      - "pkg:generic/not-baz?file%3Abaz"
+      - "pkg:generic/spam?file%3Aspam-packages%2Fspam"
```
chmeliik, line 789: I'm fairly certain these are still invalid purls (even syntactically) :disappointed: 

See qualifiers in https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#rules-for-each-purl-component. TLDR they must be key=value pairs

brunoapimentel: Just my 2¢. If it's easy to fix the qualifiers to be `key=value`, let's do it in this PR, but let's not try to make the PURLs perfect, as they're already broken now, and we're not making them worse.

We can file a follow up story for any other fixes that are needed, and we can work on it when we have time :man_shrugging:

taylormadore: I'll revisit this in the morning :+1: 
Just to confirm, we are OK with a PURL like this: `pkg:generic/%40jaegertracing/plexus?file=packages%2Fplexus`

brunoapimentel (approved):

PR #996, opened 2024-05-06
Full Go 1.21 support

eskultety:

- [ ] ~OpenAPI schema is updated (if applicable)~
- [ ] ~DB schema change has corresponding DB migration (if applicable)~
- [ ] ~README updated (if worker configuration changed, or if applicable)~

### DIFF
--- a/cachito/workers/config.py
+++ b/cachito/workers/config.py
@@ -31,7 +31,6 @@ class Config(object):
     cachito_default_environment_variables = {
         "gomod": {
             "GOSUMDB": {"value": "off", "kind": "literal"},
-            "GOTOOLCHAIN": {"value": "local", "kind": "literal"},
         },
         "npm": {
             "CHROMEDRIVER_SKIP_DOWNLOAD": {"value": "true", "kind": "literal"},
@@ -164,7 +163,6 @@ class TestingConfig(DevelopmentConfig):
         "gomod": {
             "GO111MODULE": {"value": "on", "kind": "literal"},
             "GOSUMDB": {"value": "off", "kind": "literal"},
-            "GOTOOLCHAIN": {"value": "local", "kind": "literal"},
         },
         "npm": {
             "CHROMEDRIVER_SKIP_DOWNLOAD": {"value": "true", "kind": "literal"},
--- a/cachito/workers/pkg_managers/gomod.py
+++ b/cachito/workers/pkg_managers/gomod.py
@@ -1042,7 +1042,7 @@ def get_golang_version(module_name, git_path, commit_sha, update_tags=False, sub
     )
 
 
-def _get_gomod_version(go_mod_file: Path) -> Optional[str]:
+def _get_gomod_version(go_mod_file: Path) -> Tuple[Optional[str], Optional[str]]:
     """Return the required/recommended version of Go from go.mod.
 
     We need to extract the desired version of Go ourselves as older versions of Go might fail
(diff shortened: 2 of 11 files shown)


cachito/workers/config.py
```diff
@@ -31,7 +31,7 @@ class Config(object):
     cachito_default_environment_variables = {
         "gomod": {
             "GOSUMDB": {"value": "off", "kind": "literal"},
-            "GOTOOLCHAIN": {"value": "local", "kind": "literal"},
+            "GOTOOLCHAIN": {"value": "auto", "kind": "literal"},
```
taylormadore, line 34: Did we want to override the user-side GOTOOLCHAIN with auto?

brunoapimentel:
Overall, LGTM! We just need to fix the failing test.

brunoapimentel:
> As far as I could tell, the tlog record is ever changing, so I don't see any other way to work around this besides somehow ignoring it during the tests. We can have some type of ignore list mechanism to handle such scenarios in `test_packages`, or simply making a special case for this specific test.

I went ahead and created a ignore list for that test: https://github.com/containerbuildsystem/cachito/compare/master...brunoapimentel:cachito:go-121. See if you think this is a good solution, the test is passing when pointing to the file on your fork of cachito-testing/test_files.

cachito/workers/pkg_managers/gomod.py
```diff
@@ -152,7 +152,7 @@ def release(self) -> str:
         # lazy evaluation: defer running 'go'
         if not self._release:
             output = self(["version"])
-            log.info(f"Go release: {output}")
+            log.debug(f"Go release: {output}")
```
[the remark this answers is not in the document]
eskultety: In which case I'll just drop the commit instead of dropping the log since it provides 2 different kinds of information: what the bundled version reports and what go.mod reports and transitively what we'll use.

cachito/workers/config.py
eskultety: Is this user-side? I thought this is for cachito itself, i.e. the default that will be used for the prefetch.

cachito/workers/pkg_managers/gomod.py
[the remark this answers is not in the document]
eskultety: Hmm, the commit was cherry-picked from cachi2 where I apparently mistakenly paired another log line in the commit which should have been placed in a standalone one. So what I'll do is I'll drop the log-level change and only keep the other log info on the actual go release used and reword the commit message.

eskultety:
> > As far as I could tell, the tlog record is ever changing, so I don't see any other way to work around this besides somehow ignoring it during the tests. We can have some type of ignore list mechanism to handle such scenarios in `test_packages`, or simply making a special case for this specific test.
> 
> I went ahead and created a ignore list for that test: [master...brunoapimentel:cachito:go-121](https://github.com/containerbuildsystem/cachito/compare/master...brunoapimentel:cachito:go-121). See if you think this is a good solution, the test is passing when pointing to the file on your fork of cachito-testing/test_files.

I applied both with one nitpick change and a slight adjustment to one of the commit messages, otherwise untouched. I confirm that with your commits the failing test finally passes. In the meantime I also pushed an update to the test data tarball, so hopefully the CI should finally be green.

brunoapimentel:
> I applied both with one nitpick change and a slight adjustment to one of the commit messages, otherwise untouched. I confirm that with your commits the failing test finally passes. In the meantime I also pushed an update to the test data tarball, so hopefully the CI should finally be green. Thank you for suggesting the commits!

I'm completely fine with the changes! The failure in the integration test now is something I had not noticed. It is probably because we're overwriting the ignore defaults:

```
>>> filecmp.DEFAULT_IGNORES
['RCS', 'CVS', 'tags', '.git', '.hg', '.bzr', '_darcs', '__pycache__']
```

So we'll probably have to join those with what we pass in `ignore_files`.

eskultety:
> > > I applied both with one nitpick change and a slight adjustment to one of the commit messages, otherwise untouched. I confirm that with your commits the failing test finally passes. In the meantime I also pushed an update to the test data tarball, so hopefully the CI should finally be green. Thank you for suggesting the commits!
> > 
> > 
> > I'm completely fine with the changes! The failure in the integration test now is something I had not noticed. It is probably because we're overwriting the ignore defaults:
> > ```
> > >>> filecmp.DEFAULT_IGNORES
> > ['RCS', 'CVS', 'tags', '.git', '.hg', '.bzr', '_darcs', '__pycache__']
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So we'll probably have to join those with what we pass in `ignore_files`.
> 
> Yep, I already have that particular change locally, but something else is happening as well after I had to change the order of arguments to `assert_expected_files` to silence CodeQL because the gomod test keeps failing the old way (I can reproduce locally as well).

<facepalm> a typo in the `ignore_files` directory - I forgot the trailing slash that the deps dir in `expected_files` has and since we use that as key to the dictionary, it failed...</facepalm>

Outside of scope of this PR, but the way integration tests are handled is way too fragile.

cachito/workers/config.py
taylormadore: Yes, these should get passed through to the request output: https://github.com/containerbuildsystem/cachito/blob/59967a1341ad4ee99c43f5492459c53e6e781180/cachito/workers/tasks/gomod.py#L171

tests/helper_utils/__init__.py
```diff
@@ -4,14 +4,16 @@
 from typing import Union
 
 
-def assert_directories_equal(dir_a, dir_b):
+def assert_directories_equal(dir_a, dir_b, ignore_files=[]):
```
brunoapimentel, line 7: I just remembered that setting an array as default for a function is a bad practice (my bad, I was the one that wrote this). Although we're not mutating the default array in this function, do we want to still fix that?

brunoapimentel (approved):
I just left a minor comment, but I'll go ahead and give my ack.

tests/helper_utils/__init__.py
eskultety: I noticed it, but kept it in solely for the reason you mention - we're not mutating the array. I'm in full agreement we should fix this, it is by no means critical given the current code base so I'll propose a minor fix in a follow-up.

taylormadore (approved):

PR #999, opened 2024-05-10
Add support for Go workspaces

brunoapimentel:
This commit changes the resolve_gomod function so that it can handle a json stream coming from `go list -m`, which happens when workspaces are used in a Go project.

It also removes the previous restrictions on using workspaces.

### DIFF
--- a/cachito/workers/pkg_managers/gomod.py
+++ b/cachito/workers/pkg_managers/gomod.py
@@ -6,6 +6,7 @@ import os.path
 import re
 import shutil
 import tempfile
+from dataclasses import dataclass
 from datetime import datetime
 from itertools import chain
 from pathlib import Path, PureWindowsPath
@@ -75,6 +76,7 @@ class GoModule(_GolangModel):
     """
 
     path: str
+    dir: Optional[Path] = None
     version: Optional[str] = None
     main: bool = False
     replace: Optional["GoModule"] = None
@@ -91,6 +93,43 @@ class GoPackage(_GolangModel):
     standard: bool = False
     module: Optional[GoModule]
     deps: list[str] = []
+
+
+@dataclass
+class LocalModules:
+    """The modules present in the current repository being processed by Cachito.
+
+    It contains a main module, which represents the current input package being processed, and all
+    existing worskpaces nested under the nearest directory containing a go.work file (if it exists).
+
+    Note that this class does not account for any local replacements present in the current go.mod
+    or go.work file.
+    """
+
+    main: GoModule
+    workspaces: list[GoModule]
+
+    @staticmethod
+    def from_json_stream(modules_json_stream: str, app_source_path: Path) -> "LocalModules":
+        main_module = None
+        workspaces = []
+
+        for item in load_json_stream(modules_json_stream):
+            module = GoModule.parse_obj(item)
+
+            if module.dir == app_source_path:
+                main_module = module
+            else:
+                workspaces.append(module)
+
+        # should never happen, since the main module will always be a part of the json stream
+        if not main_module:
+            raise RuntimeError('Failed to find the main module info in the "go list -m" output.')
+
+        return LocalModules(main_module, workspaces)
+
+    def all(self) -> list[GoModule]:
+        return [self.main] + self.workspaces
 
 
 class Go:
@@ -380,23 +419,12 @@ def resolve_gomod(
             # Make Go ignore the vendor dir even if there is one
             go_list.extend(["-mod", "readonly"])
 
-        main_module_name = go([*go_list, "-m"], run_params).rstrip()
-        main_module_version = get_golang_version(
-            main_module_name,
-            git_dir_path,
-            request["ref"],
-            update_tags=True,
-            subpath=(
-                None
-                if app_source_path == git_dir_path
-                else str(app_source_path).replace(f"{git_dir_path}/", "")
-            ),
+        local_modules = LocalModules.from_json_stream(
+            go([*go_list, "-m", "-json"], run_params).rstrip(),
+            app_source_path,
         )
-        main_module = {
-            "type": "gomod",
-            "name": main_module_name,
-            "version": main_module_version,
-        }
+
+        _set_local_modules_versions(local_modules, git_dir_path, request)
 
         def go_list_deps(pattern: Literal["./...", "all"]) -> Iterator[GoPackage]:
             """Run go list -deps -json and return the parsed list of packages.
@@ -413,7 +441,7 @@ def resolve_gomod(
             mod for pkg in go_list_deps("all") if (mod := pkg.module) and not mod.main
         )
         main_module_deps = _deduplicate_to_gomod_dicts(
-            chain(package_modules, downloaded_modules), deps_to_replace
+            chain(package_modules, downloaded_modules, local_modules.workspaces), deps_to_replace
         )
 
         log.info("Retrieving the list of packages")
@@ -446,7 +474,7 @@ def resolve_gomod(
                 elif not dep.module or dep.module.main:
                     # Standard=false, Module.Main=true
                     # Standard=false, Module=null   <- probably a to-be-generated package
-                    dep_version = main_module_version
+                    dep_version = local_modules.main.version
                 else:
                     _, dep_version = _get_name_and_version(dep.module)
 
@@ -455,7 +483,7 @@ def resolve_gomod(
             main_pkg = {
                 "type": "go-package",
                 "name": pkg.import_path,
-                "version": main_module_version,
+                "version": local_modules.main.version,
             }
             main_packages.append({"pkg": main_pkg, "pkg_deps": pkg_deps})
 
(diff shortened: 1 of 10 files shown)


cachito/workers/pkg_managers/gomod.py
```diff
@@ -472,6 +459,81 @@ def go_list_deps(pattern: Literal["./...", "all"]) -> Iterator[GoPackage]:
...
+        update_tags=True,
+        subpath=(
+            None
+            if app_source_path == git_dir_path
+            else str(app_source_path).replace(f"{git_dir_path}/", "")
+        ),
+    )
+
+    for module in workspace_modules:
+        module.version = main_module_version
```
brunoapimentel, line 498: As far as I could tell, Cachito is ignoring version tags applied to submodules. So we can just use the main module version for all workspaces, since we're pretty much keeping the current behavior.

We should probably fix this in a follow up PR, though.

brunoapimentel: I tested it with:
```
{
  "repo": "https://github.com/cachito-testing/gomod-pandemonium",
  "ref": "0c6890c3280a00271891f4bd04705a56151428f0",
  "pkg_managers": ["gomod"]
}
```

And I got:
```
{
    "name": "github.com/cachito-testing/gomod-pandemonium/terminaltor",
    "purl": "pkg:golang/github.com%2Fcachito-testing%2Fgomod-pandemonium%2Fterminaltor@v0.1.0",
    "type": "library",
    "version": "v0.1.0"
}
```

Although the version for terminaltor should've been [v1.0.0](https://github.com/cachito-testing/gomod-pandemonium/releases/tag/terminaltor%2Fv1.0.0).

brunoapimentel: (The SBOM also had an entry in which the version for terminaltor was `./terminaltor`, so the behavior is a little odd)

chmeliik: I think Cachito does support submodule versioning. But only if you tell it to process the submodule :smiling_face_with_tear:

brunoapimentel: So, I tried specifying all modules using the input you provided, and it still is correctly picking up for terminaltor:
```
  {
    "name": "github.com/cachito-testing/gomod-pandemonium/terminaltor",
    "purl": "pkg:golang/github.com%2Fcachito-testing%2Fgomod-pandemonium%2Fterminaltor@v1.0.0",
    "type": "library",
    "version": "v1.0.0"
  }
```

Which means we're still preserving the same behavior: submodule versions will only be picked up if the submodule is specified as a Cachito input package.

brunoapimentel: I didn't see any duplication in the resulting SBOM that happens as a side effect of specifying all the submodules/workspaces present in the repo. The terminaltor module has a duplication regardless of what input packages you specify, though:

```
 {
    "name": "github.com/cachito-testing/gomod-pandemonium/terminaltor",
    "purl": "pkg:golang/github.com%2Fcachito-testing%2Fgomod-pandemonium%2Fterminaltor@v1.0.0",
    "type": "library",
    "version": "./terminaltor"
  },
  {
    "name": "github.com/cachito-testing/gomod-pandemonium/terminaltor",
    "purl": "pkg:golang/github.com%2Fcachito-testing%2Fgomod-pandemonium%2Fterminaltor@v1.0.0",
    "type": "library",
    "version": "v1.0.0"
  }
```

This does not happen with `weird`... I'm not sure what is causing it yet. :thinking:

chmeliik: On the one hand, with workspaces it doesn't make any sense to require that the user specify all submodules

On the other hand, ¯\\\_(ツ)\_/¯

chmeliik (approved):

eskultety (approved):

taylormadore (approved):

cachito/workers/pkg_managers/gomod.py
brunoapimentel: Well, so I found out that the previous solution was indeed creating duplicate results in the SBOM. I rewrote it so:
- it picks the right version for each workspace
- the user does not need to specify multiple input packages anymore

brunoapimentel:
New push: rewrites the logic that parses the workspace data so that the versions applied to submodules are properly picked up.

cachito/workers/pkg_managers/gomod.py
```diff
@@ -465,13 +491,39 @@ def go_list_deps(pattern: Literal["./...", "all"]) -> Iterator[GoPackage]:
...
+
         return {
-            "module": main_module,
+            "module": main_module_dict,
             "module_deps": main_module_deps,
             "packages": main_packages,
         }
 
 
+def _set_local_modules_versions(
```
brunoapimentel, line 507: I don't like the idea of updating the object attributes in place, but I also didn't want to do a deep dive into refactoring `get_golang_version`. This function takes a lot of attributes, and anything I wrote in the LocalModules class would bring all this crap together :confused:

cachito/workers/pkg_managers/gomod.py
```diff
@@ -93,6 +94,42 @@ class GoPackage(_GolangModel):
     deps: list[str] = []
 
 
+class LocalModules(pydantic.BaseModel):
```
chmeliik, line 97: IMO there's no reason to make this a pydantic model, the class doesn't represent any external data structure

A dataclass would work just fine

cachito/workers/pkg_managers/gomod.py
```diff
@@ -446,7 +472,7 @@ def go_list_deps(pattern: Literal["./...", "all"]) -> Iterator[GoPackage]:
                 elif not dep.module or dep.module.main:
                     # Standard=false, Module.Main=true
                     # Standard=false, Module=null   <- probably a to-be-generated package
-                    dep_version = main_module_version
+                    dep_version = local_modules.main.version
```
chmeliik, line 475: The version will probably be wrong for packages coming from workspaces other than the one selected by the user. It seems to be handled correctly in cachi2 though, so I wouldn't mind pretending I never noticed this :smile:

chmeliik: Either that or maybe we don't report such packages at all, since as far as I can tell the `./...` pattern isn't really workspace-aware https://github.com/containerbuildsystem/cachi2/pull/553#discussion_r1654667805. Anyway, to-be-fixed in cachi2?

cachito/workers/pkg_managers/gomod.py
```diff
@@ -465,13 +491,39 @@ def go_list_deps(pattern: Literal["./...", "all"]) -> Iterator[GoPackage]:
...
+    for module in local_modules.all():
+        module.version = get_golang_version(
+            module.path,
+            git_dir_path,
+            request["ref"],
+            update_tags=True,
+            subpath=(
+                None
+                if module.dir == git_dir_path
+                else str(module.dir).replace(f"{git_dir_path}/", "")
```
chmeliik, line 522: Is the subpath handling copy-pasted from somewhere else for consistency? If not, we should do path operations properly `module_dir.relative_to(git_dir_path)`

brunoapimentel: This is how it was written before. I changed it, but it was just a small improvement. We still want to keep sending `None` or `str(subpath)` since that's how the called functions still expect it.

cachito/workers/pkg_managers/gomod.py
brunoapimentel: So I did a few tests targeting our [test repo](https://github.com/cachito-testing/cachi2-gomod/tree/go_workspaces), and here's what I found:

The workspace packages are only identified by `go list ./...` when the command is run inside the worskpace folder. That means that Cachito will only report them if the workspace was specified as an input package (in which case, the version will end up being correct, since it correspond to the actual parent module).

In case the workspace containing packages was not specified as an input package, the package won't be reported at all.

A quick test to check this:
```
$ curl -H "content-type: application/json" -d '{
  "repo": "https://github.com/cachito-testing/cachi2-gomod",
  "ref": "b97aca303b29dca71b89185da6935249b0deef68",
  "pkg_managers": ["gomod"],
  "packages": {
    "gomod": [{"path": "workspace_modules/echo"}, {"path": "workspace_modules/hi/hiii"}, {"path": "workspace_modules/hello"}]
  }
}' localhost:8080/api/v1/requests

$ curl -s localhost:8080/api/v1/requests/{request_id} | grep "echo/v4/middleware"
   "name": "github.com/labstack/echo/v4/middleware",
```
When we remove the echo workspace from the list of input packages, the `echo/v4/middleware` package is absent:
```
$ curl -H "content-type: application/json" -d '{
  "repo": "https://github.com/cachito-testing/cachi2-gomod",
  "ref": "b97aca303b29dca71b89185da6935249b0deef68",
  "pkg_managers": ["gomod"],
  "packages": {
    "gomod": [ {"path": "workspace_modules/hi/hiii"}, {"path": "workspace_modules/hello"}]
  }
}' localhost:8080/api/v1/requests

$ curl -s localhost:8080/api/v1/requests/{request_id} | grep "echo/v4/middleware"
  
```

brunoapimentel: I agree we should fix it in Cachi2, then backport the solution here.

chmeliik (approved):
LGTM

Just to make sure I understand the limitations
* all the workspaces will be reported with correct versions, doesn't matter if the user tell cachito to process just one workspace or all of them
* packages from a workspace will only be reported if the user tells cachito to process that workspace

More or less?

brunoapimentel:
>     * all the workspaces will be reported with correct versions, doesn't matter if the user tells cachito to process just one workspace or all of them

Yes, because `go list -m` will always list all workspaces, and we're running `_get_golang_version` individually for each one of them.

> * packages from a workspace will only be reported if the user tells cachito to process that workspace

Yes, because of the limitation of `go list ./...`.

PR #1003, opened 2024-05-29
Follow-up fixes for !996 

eskultety:
- follow good practice of not using a list as a default argument due to immutability in integration tests
- drop most explicit `GOSUMDB=off` assignments and only set it to its default value for `resolve_gomod`


ben-alkov (commented):
LGTM (but IANAE)

PR #1004, opened 2024-05-30
Preliminary support for Go 1.22

eskultety:
This PR basically just bumps the maximum supported version to 1.22 from 1.21. The changes don't include everything that 1.22 brings, most notably vendored workspaces (we don't even support base workspaces).

### DIFF
--- a/cachito/workers/pkg_managers/gomod.py
+++ b/cachito/workers/pkg_managers/gomod.py
@@ -1077,7 +1077,7 @@ def _get_gomod_version(go_mod_file: Path) -> Tuple[Optional[str], Optional[str]]
 def _select_go_toolchain(go_mod_file: Path) -> Go:
     go = Go()
     target_version = None
-    go_max_version = pkgver.Version("1.21")
+    go_max_version = pkgver.Version("1.22")
     go_121_version = pkgver.Version("1.21")
     go_base_version = go.version
     go_mod_version_msg = "go.mod reported versions: '{}'[go], '{}'[toolchain]"
--- a/cachito/workers/pkg_managers/gomod.py
+++ b/cachito/workers/pkg_managers/gomod.py
@@ -1078,6 +1078,7 @@ def _select_go_toolchain(go_mod_file: Path) -> Go:
     go = Go()
     target_version = None
     go_max_version = pkgver.Version("1.21")
+    go_121_version = pkgver.Version("1.21")
     go_base_version = go.version
     go_mod_version_msg = "go.mod reported versions: '{}'[go], '{}'[toolchain]"
 
@@ -1111,13 +1112,13 @@ def _select_go_toolchain(go_mod_file: Path) -> Go:
             f"Required/recommended Go toolchain version '{target_version}' is not supported yet.",
         )
 
-    if target_version >= go_max_version:
+    if target_version >= go_121_version:
         # Project makes use of Go >=1.21:
         # - always use the 'X.Y.0' toolchain to make sure GOTOOLCHAIN=auto fetches anything newer
         # - container environments need to have it pre-installed
         # - local environments will always install 1.21.0 SDK and then pull any newer toolchain
         go = Go(release="go1.21.0")
-    elif go_base_version >= go_max_version:
+    elif go_base_version >= go_121_version:
         # Starting with Go 1.21, Go doesn't try to be semantically backwards compatible in that
         # the 'go X.Y' line now denotes the minimum required version of Go, not a "suggested"
         # version. What it means in practice is that a Go toolchain >= 1.21 enforces the biggest
--- a/cachito/workers/tasks/gomod.py
+++ b/cachito/workers/tasks/gomod.py
@@ -4,13 +4,7 @@ import os
 from pathlib import Path
 
 from cachito.common.packages_data import PackagesData
-from cachito.errors import (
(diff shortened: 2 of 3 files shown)


taylormadore:
Not specifically related to this PR, but what do you think about having a log message with the version of the toolchain being executed? As I was testing these changes with a project that was using 1.22.0, I didn't see that toolchain version anywhere in the cachito request logs (just athens). It might be helpful for some future debugging to have that information available

taylormadore (approved):

eskultety:
> Not specifically related to this PR, but what do you think about having a log message with the version of the toolchain being executed? As I was testing these changes with a project that was using 1.22.0, I didn't see that toolchain version anywhere in the cachito request logs (just athens). It might be helpful for some future debugging to have that information available

That version isn't explicitly reported by Go anywhere though. The only information on the target toolchain we have is `go.mod` so IIUC we can't really report the toolchain being executed, because from our POV that's 1.21.0 and anything newer that ends up being used by Go is totally opaque to us. If anything, I can tweak the current `Using Go release: go1.21.0` message and expand it with a clause that any newer toolchain reported in go.mod will be fetched and used.

eskultety:
> > Not specifically related to this PR, but what do you think about having a log message with the version of the toolchain being executed? As I was testing these changes with a project that was using 1.22.0, I didn't see that toolchain version anywhere in the cachito request logs (just athens). It might be helpful for some future debugging to have that information available
> 
> That version isn't explicitly reported by Go anywhere though. The only information on the target toolchain we have is `go.mod` so IIUC we can't really report the toolchain being executed, because from our POV that's 1.21.0 and anything newer that ends up being used by Go is totally opaque to us. If anything, I can tweak the current `Using Go release: go1.21.0` message and expand it with a clause that any newer toolchain reported in go.mod will be fetched and used.

@taylormadore Here's my take on the log message - I added an extra [commit](https://github.com/containerbuildsystem/cachito/pull/1004/commits/0bc2e1ee06df2fde2bb242c66a3acdbaeeb85bb0) which tweaks the `Using Go release` message hinting at downloading another target toolchain automatically. I'm ambivalent on whether this really improves the status quo, so I'll leave it to you whether you think it helps at all, I'm fine dropping the commit again and leave it as is.

taylormadore:
> That version isn't explicitly reported by Go anywhere though

My initial thought was to report the output of the `go version` command (and making sure to use the env_vars so that the toolchain is used to report the version)

eskultety:
> > That version isn't explicitly reported by Go anywhere though
> 
> My initial thought was to report the output of the `go version` command (and making sure to use the env_vars so that the toolchain is used to report the version)

Oh, right. That means we have to stuff another go call in there somewhere before we commence with the module download.Okay, but that might require some bigger changes and actually not be that transparent from code perspective - because on one hand we're instantiating a specific toolchain as `Go('<toolchain_string>')` whereas where you want to use the bundled one (from which we report the version at the beginning) you just do it as `Go()`. Your suggestion falls kind of in between where it would become sort of a hybrid where you'd do `go=Go()`, but then instead of `go.version` you'd have to do `go('version', env={'GOTOOLCHAIN':'auto', 'GOSUMDB':'sum.golang.org'})` just to comply with the requirement here, correct? I'm not saying it's a wrong requirement, I'm only stating that it may lead to some bigger changes rather than a simple log string, I'm only clarifying.

eskultety:
> @brunoapimentel @taylormadore FYI several openshift teams are waiting on 1.22 support in cachito for the 4.16 OCP release :) Please review :)

This won't likely help any of them since kubernetes uses vendored workspaces which this doesn't add support for.

eskultety:
> > > That version isn't explicitly reported by Go anywhere though
> > 
> > 
> > My initial thought was to report the output of the `go version` command (and making sure to use the env_vars so that the toolchain is used to report the version)
> 
> Oh, right. That means we have to stuff another go call in there somewhere before we commence with the module download.Okay, but that might require some bigger changes and actually not be that transparent from code perspective - because on one hand we're instantiating a specific toolchain as `Go('<toolchain_string>')` whereas where you want to use the bundled one (from which we report the version at the beginning) you just do it as `Go()`. Your suggestion falls kind of in between where it would become sort of a hybrid where you'd do `go=Go()`, but then instead of `go.version` you'd have to do `go('version', env={'GOTOOLCHAIN':'auto', 'GOSUMDB':'sum.golang.org'})` just to comply with the requirement here, correct? I'm not saying it's a wrong requirement, I'm only stating that it may lead to some bigger changes rather than a simple log string, I'm only clarifying.

Unless we run `'go version` with the `GOTOOLCHAIN=auto` itself, which would comply with your request. I guess I'd be okay with that, it's just I specifically designed the toolchain download as a lazy evaluation, IOW defer downloading toolchains until the very last moment where we know we're going to download other modules rather than slow down other logic by downloading the toolchain upfront and then potentially fail anyway.

taylormadore:
I'd be ok with that approach and it's definitely not a blocker for this PR and we can resolve it in a follow-up issue.

eskultety:
> I'd be ok with that approach and it's definitely not a blocker for this PR and we can resolve it in a follow-up issue.

https://github.com/containerbuildsystem/cachito/issues/1005

ben-alkov (approved):
LGTM

PR #1013, opened 2024-07-03
Remove PATH deps allowlist for rubygems

taylormadore:
Remove the PATH deps allowlist for rubygems.

There is also an additional commit that allows PATH dependencies from anywhere in the request repository rather than just from the package directory. I can drop that commit if you disagree.


eskultety (approved):
the fewer obstacles the better :)

chmeliik (approved):
LGTM :+1:

PR #1019, opened 2024-07-12
Explicitly forbid workspace vendoring

brunoapimentel:
The ability to vendor dependencies in projects that make use of workspaces was[ introduced in Go 1.22](https://github.com/golang/go/issues/60056). Cachito currently does not support the use of "go work vendor", so this commit introduces an explicit check for that scenario and a clearer error message.

Note that by design, Go does not allow the use of "go mod vendor" if workspaces are present, so the check assumes that the existence of a vendor folder in a project with workspaces means that workspace vendoring is being used.

### DIFF
--- a/tests/test_workers/test_pkg_managers/test_gomod.py
+++ b/tests/test_workers/test_pkg_managers/test_gomod.py
@@ -211,7 +211,9 @@ def test_resolve_gomod(
 @mock.patch("cachito.workers.pkg_managers.gomod.GoCacheTemporaryDirectory")
 @mock.patch("subprocess.run")
 @mock.patch("cachito.workers.pkg_managers.gomod.RequestBundleDir")
+@mock.patch("cachito.workers.pkg_managers.gomod._vet_workspace_vendoring")
 def test_resolve_gomod_vendor_dependencies(
+    mock_vet_workspace_vendoring: mock.Mock,
     mock_bundle_dir: mock.Mock,
     mock_run: mock.Mock,
     mock_temp_dir: mock.Mock,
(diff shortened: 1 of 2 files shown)


eskultety:
> The ability to vendor dependencies in projects that make use of workspaces wa shttps://github.com/golang/go/issues/60056.

Linking the official release notes for completeness: https://go.dev/doc/go1.22#tools.

tests/test_workers/test_pkg_managers/test_gomod.py
```diff
@@ -881,6 +883,42 @@ def test_vet_local_file_dep_paths_outside_repo():
...
+    if vendor_folder_exists:
+        (tmp_path / "vendor").mkdir()
+
+    expect_error = "Workspace vendoring introduced in Go 1.22 is not supported"
+
+    if raise_error:
+        with pytest.raises(UnsupportedFeature, match=expect_error):
+            gomod._vet_workspace_vendoring(go, {})
+    else:
+        gomod._vet_workspace_vendoring(go, {})
```
eskultety, line 917: I believe the general idea in testing is to have 2 unit tests, one to test the positive behaviour, one for the negative to avoid these types of conditionals (and avoid arbitrary booleans in `parametrize` toggling the behaviour) at the expense of code repetition (to some extent).

eskultety (approved):
Mostly nitpicks. I have no doubts you can take care of them -> ACK.

taylormadore (approved):

tests/test_workers/test_pkg_managers/test_gomod.py
brunoapimentel: Well, to remove all the conditionals, I ended up with three tests. I think it's clearer now, but it is more verbose as well.

eskultety: Maybe I wasn't completely clear in my comment, but all I wanted to point out was the dichotomy between positive/negative test cases. In other words, having parametrized tests and booleans is okay within the same test domain, i.e. whether I'm testing a positive outcome or a negative one. The reason I pointed it out is because in general, one can have N positive combinations for a positive outcome and N negative ones (e.g. different errors/exceptions) and so the tests should be paired as such. It all boils down to something like this:

```
diff --git a/tests/test_workers/test_pkg_managers/test_gomod.py b/tests/test_workers/test_pkg_managers/test_gomod.py
index d06b94d..fc13e10 100644
--- a/tests/test_workers/test_pkg_managers/test_gomod.py
+++ b/tests/test_workers/test_pkg_managers/test_gomod.py
@@ -1549,3 +1549,37 @@ class TestGo:
         error_msg = f"Could not extract Go toolchain version from Go's output: '{go_output}'"
         with pytest.raises(GoModError, match=error_msg):
             gomod.Go(release=None).release
+
+
+@pytest.mark.parametrize(
+    "go_work_env",
+    [
+        pytest.param("go.work", id="go_work_set"),
+        pytest.param("off", id="go_work_off"),
+        pytest.param("", id="go_work_empty"),
+    ],
+)
+def test_vet_workspace_vendoring(
+    go_work_env: str,
+    tmp_path: Path,
+):
+    go = mock.Mock()
+
+    if go_work_env == "go.work":
+        go.return_value = str(tmp_path / "go.work")
+    else:
+        go.return_value = go_work_env
+
+    gomod._vet_workspace_vendoring(go, {})
+    go.assert_called_once_with(["env", "GOWORK"], {})
+
+
+def test_vet_workspace_vendoring_fail(tmp_path: Path):
+    go = mock.Mock()
+    go.return_value = str(tmp_path / "go.work")
+    (tmp_path / "vendor").mkdir()
+
+    expect_error = "Workspace vendoring introduced in Go 1.22 is not supported"
+
+    with pytest.raises(UnsupportedFeature, match=expect_error):
+        gomod._vet_workspace_vendoring(go, {})
```

(passed unit tests) which IMO is equal to the original proposal (sans additional booleans) and tests both positive outcomes as well as the negative one. Hopefully I haven't missed anything important but ^that was what I more or less expected and had in mind. We don't need the extra verbosity of the latest revision as I feel it hinders readability a bit. Let me know your thoughts, maybe I've overlooked some important bit which led you to that extra verbosity when splitting the tests cases.

brunoapimentel: Yeah, the only bit you missed was that the `test_vet_workspace_vendoring` will need to have a vendor folder for the go_work_off and go_work_empty cases (we're testing that vendoring is allowed in case there are no workspaces).

But I can now see a way to do it without the actual boolean test param that was present (vendor_folder_exist), so let me do it and you can leave a final review.

eskultety (approved):

PR #1033, opened 2024-09-05
Bump the supported Go version to 1.23

eskultety:
Go 1.23 was released last month, so bump the max supported version to reflect that. Most notable change from 1.23 regarding Go tooling is the introduction of telemetry data gathering which is opt-in by default. Patch 2 disables telemetry explicitly just in case Go would change stance on this in the future.
The lack of workspace vendoring support from 1.22 still remains true here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -747,7 +747,7 @@ The table below shows the supported package managers and their support level in 
 
 Tool     | Version |
 ---      |---------|
-Go*      | 1.20.7, 1.21.0  |
+Go*      | 1.20.7, 1.22.0 (no workspace vendoring support)  |
 Npm      | 9.5.0   |
 Node     | 18.16.1 |
 Pip      | 22.3.1  |
--- a/README.md
+++ b/README.md
@@ -747,7 +747,7 @@ The table below shows the supported package managers and their support level in 
 
 Tool     | Version |
 ---      |---------|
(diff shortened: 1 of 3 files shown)


taylormadore:
The changes LGTM, but we need something like https://github.com/containerbuildsystem/cachito/pull/1034 to fix the CI

README.md
```diff
@@ -747,7 +747,7 @@ The table below shows the supported package managers and their support level in
 
 Tool     | Version |
 ---      |---------|
-Go*      | 1.20.7, 1.21.0  |
+Go*      | 1.20.7, 1.22.0 (no workspace vendoring support), 1.23.0  |
```
brunoapimentel, line 750: Nitpick: I wonder if we need to specify 1.22. The second version in this table refers to the maximum supported version, so I'd just keep it as 1.23.
```suggestion
Go*      | 1.20.7, 1.23.0 (no workspace vendoring support)|
```

brunoapimentel (commented):
This LGTM. We just need to fix the CI.

I left a niptick comment that can be ignored

README.md
eskultety: Uhm, we don't, but I figured if we need to mention limitations better tie it with the version that they apply to, but it might be just me who'd be confused what particular feature that is and which version introduced it. If you don't think it's valuable I can drop it as you suggest.

taylormadore (approved):

PR #1034, opened 2024-09-09
Makefile: use docker compose v2

taylormadore:
Docker compose v1 has been deprecated and has been removed from Github actions runner images. Use v2 instead.

See https://github.com/actions/runner-images/issues/9692 and https://docs.docker.com/compose/migrate/ for more information.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-CACHITO_COMPOSE_ENGINE ?= docker-compose
+CACHITO_COMPOSE_ENGINE ?= docker compose
 PYTHON_VERSION_VENV ?= python3.11
 TOX_ENVLIST ?= python3.11
 TOX_ARGS ?=
--- a/README.md
+++ b/README.md
@@ -87,13 +87,13 @@ To verify that your code meets these standards, you may run `tox -e black,flake8
 
 ## Quick Start
 
-Run the application locally (requires [docker-compose](https://docs.docker.com/compose/)):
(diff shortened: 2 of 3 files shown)


brunoapimentel (approved):

Makefile
```diff
@@ -1,4 +1,11 @@
-CACHITO_COMPOSE_ENGINE ?= docker-compose
+DOCKER_COMPOSE_V2_AVAILABLE := $(shell docker compose version >/dev/null 2>&1 && echo true || echo false)
+
+ifeq ($(DOCKER_COMPOSE_V2_AVAILABLE), true)
+	CACHITO_COMPOSE_ENGINE ?= docker compose
+else
+	CACHITO_COMPOSE_ENGINE ?= docker-compose
+endif
+
```
eskultety, line 8: If v1 has been deprecated, why can't we just get rid of it completely? Why do we want to fall back to it?

eskultety: I mean any recentish distro should have v2 already, shouldn't it?

>  I've always used podman-compose, so I really don't care either way

[off-topic] good for you, podman-compose either isn't available or sucks big time on SoC Debian ARM spins :(

brunoapimentel (approved):

Makefile
taylormadore: Oops, I forgot to respond here that I switched it to v2

eskultety (approved):

